PR #17, opened 2020-04-16
Amount and signatures use CompressedRistretto

mfaulk:
Previously, Amount.commitment was a RistrettoPoint (type aliased as "Commitment'), while the rest of the transaction members are CompressedRistretto. Using RistrettoPoint in Amount causes unnecessary compress and decompress operations. For example, computing the hash, digest, serialization, or deserialization of a Tx involves compressing the Amount commitment for each ring element and transaction output.

This PR adds the newtypes Commitment and CompressedCommitment, and updates Amount to use CompressedCommitment. This should prevent most of the unnecessary compress/decompress operations related to Amount.

This also changes SignatureRctBulletproof's sign and verify methods to take one-time addresses and amount commitments in compressed form (CompressedRistrettoPublic and CompressedCommitment). I think this nicely encapsulates the compress/decompress logic close to where it is used.

### DIFF
--- a/transaction/core/src/amount.rs
+++ b/transaction/core/src/amount.rs
@@ -7,7 +7,10 @@
 
 #![cfg_attr(test, allow(clippy::unnecessary_operation))]
 
-use crate::ring_signature::{Blinding, Commitment, CurveScalar, GENERATORS};
+use crate::{
+    ring_signature::{Blinding, CurveScalar, GENERATORS},
+    CompressedCommitment,
+};
 use blake2::{Blake2b, Digest};
 use curve25519_dalek::scalar::Scalar;
 use digestible::Digestible;
@@ -30,7 +33,7 @@ pub enum AmountError {
 pub struct Amount {
     /// A Pedersen commitment `v*G + b*H` to a quantity `v` of MobileCoin, with blinding `b`,
     #[prost(message, required, tag = "1")]
-    pub commitment: Commitment,
+    pub commitment: CompressedCommitment,
 
     /// `masked_value = value + Blake2B(shared_secret)`
     #[prost(message, required, tag = "2")]
@@ -55,13 +58,12 @@ impl Amount {
         blinding: Blinding,
         shared_secret: &RistrettoPublic,
     ) -> Result<Amount, AmountError> {
-        let value: Scalar = Scalar::from(value);
-
         // Pedersen commitment `v*G + b*H`.
-        let commitment: Commitment = Commitment::from(GENERATORS.commit(value, blinding.into()));
+        let commitment = CompressedCommitment::new(value, blinding.into());
 
         // `v + Blake2B(shared_secret)`
         let masked_value: Scalar = {
+            let value: Scalar = Scalar::from(value);
             let mask = get_value_mask(&shared_secret);
             value + mask
         };
@@ -92,8 +94,7 @@ impl Amount {
         let value: u64 = self.unmask_value(shared_secret);
         let blinding = self.unmask_blinding(shared_secret);
 
-        let expected_commitment =
-            Commitment::from(GENERATORS.commit(Scalar::from(value), blinding.into()));
+        let expected_commitment = CompressedCommitment::new(value, blinding.into());
         if self.commitment != expected_commitment {
             // The commitment does not agree with the provided value and blinding.
             // This either means that the commitment does not correspond to the shared secret, or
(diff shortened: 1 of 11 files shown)


transaction/core/src/amount.rs
```diff
@@ -30,7 +33,7 @@ pub enum AmountError {
 pub struct Amount {
     /// A Pedersen commitment `v*G + b*H` to a quantity `v` of MobileCoin, with blinding `b`,
     #[prost(message, required, tag = "1")]
-    pub commitment: Commitment,
+    pub commitment: CompressedCommitment,
```
mfaulk, line 36: Previously, "Commitment" was a type alias for CurvePoint, which we used for any RistrettoPoint that we wanted to serialize.

This now uses the newtype CompressedCommitment which wraps CompressedRistretto. The dedicated newtype gives better type checking, and I think that storing this as a compressed point avoids ambiguity and is more performant (similar to how KeyImages are represented as compressed points).

transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -46,6 +48,7 @@ pub struct SignatureRctBulletproofs {
     pub ring_signatures: Vec<RingMLSAG>,
 
     /// Commitments of value equal to each real input.
+    /// TODO: Make this a CompressedCommitment.
```
mfaulk, line 51: This will be a follow-on PR.

transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -153,7 +156,7 @@ impl SignatureRctBulletproofs {
     /// * `output_values_and_blindings` - Value and blinding for each output amount commitment.
     pub fn sign<CSPRNG: RngCore + CryptoRng>(
         message: &[u8; 32],
-        rings: &[Vec<(Address, Commitment)>],
+        rings: &[Vec<(CompressedRistrettoPublic, CompressedCommitment)>],
```
mfaulk, line 159: This now takes compressed points, which can be read directly from the TxOuts in a transaction.

transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -180,8 +183,8 @@ impl SignatureRctBulletproofs {
     pub fn verify<CSPRNG: RngCore + CryptoRng>(
         &self,
         message: &[u8; 32],
-        rings: &[Vec<(Address, Commitment)>],
-        output_commitments: &[Commitment],
+        rings: &[Vec<(CompressedRistrettoPublic, CompressedCommitment)>],
```
mfaulk, line 186: This now takes compressed points, which can be read directly from the TxOuts in a transaction.

transaction/core/src/ring_signature/mlsag.rs
```diff
@@ -133,7 +135,7 @@ impl RingMLSAG {
     // * `rng` - Randomness.
     pub fn sign<CSPRNG: RngCore + CryptoRng>(
         message: &[u8; 32],
-        ring: &[(Address, Commitment)],
```
mfaulk, line 136: `Address` and `Commitment` were both type aliases for RistrettoPoint. The aliases gave some readability, but they didn't actually provide type safety (for example, you could pass an Address to a function that takes a Commitment). This PR removes them in favor of more specific newtypes.

transaction/core/src/ring_signature/mlsag.rs
```diff
@@ -124,7 +126,7 @@ impl RingMLSAG {
     //
     // # Arguments
     // * `message` - Message to be signed.
-    // * `ring` - A ring of input onetime addresses and amount commitments.
+    // * `ring` - A ring of input onetime addresses and uncompressed amount commitments.
```
mfaulk, line 129: Oops, a couple of comments are wrong...

transaction/core/src/validation/validate.rs
```diff
@@ -173,14 +173,13 @@ pub fn validate_transaction_signature<R: RngCore + CryptoRng>(
     let tx_prefix_hash = tx.prefix.hash();
     let message = tx_prefix_hash.as_bytes();
 
-    let mut rings: Vec<Vec<(Address, Commitment)>> = Vec::new();
-    for tx_in in &tx.prefix.inputs {
-        let mut ring: Vec<(Address, Commitment)> = Vec::new();
-        for tx_out in &tx_in.ring {
-            let address = RistrettoPublic::try_from(&tx_out.target_key)?;
```
mfaulk, line 180: This `try_from` performs a decompression.

transaction/core/src/ring_signature/mlsag.rs
```diff
@@ -347,11 +364,24 @@ impl RingMLSAG {
     }
 }
 
+fn decompress_ring(
```
mfaulk, line 367: This should now be the only place that input rings are decompressed on the consensus server.

PR #22, opened 2020-04-16
Unit tests that TransactionBuilder returns error if value is not conserved

mfaulk:
Just an extra unit test or two.

### DIFF
--- a/transaction/std/src/transaction_builder.rs
+++ b/transaction/std/src/transaction_builder.rs
@@ -242,43 +242,56 @@ pub mod transaction_builder_tests {
     use std::convert::TryFrom;
     use transaction::{
         account_keys::{AccountKey, DEFAULT_SUBADDRESS_INDEX},
+        get_tx_out_shared_secret,
         onetime_keys::*,
         tx::TxOutMembershipProof,
         validation::validate_transaction_signature,
     };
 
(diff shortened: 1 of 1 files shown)


transaction/std/src/transaction_builder.rs
```diff
@@ -326,16 +339,25 @@ pub mod transaction_builder_tests {
...
             assert!(view_key_matches_output(
                 &bob.view_key(),
-                &RistrettoPublic::try_from(&tx_out.target_key).unwrap(),
-                &RistrettoPublic::try_from(&tx_out.public_key).unwrap()
+                &RistrettoPublic::try_from(&output.target_key).unwrap(),
+                &RistrettoPublic::try_from(&output.public_key).unwrap()
             ));
         }
 
+        // The output should have the correct value.
```
mfaulk, line 353: Extra test.

transaction/std/src/transaction_builder.rs
```diff
@@ -346,4 +368,57 @@ pub mod transaction_builder_tests {
     fn test_inputs_with_different_ring_sizes() {
         unimplemented!()
     }
+
+    #[test]
+    // `build` should return an error if the sum of inputs does not equal the sum of outputs and the fee.
```
mfaulk, line 373: Extra test.

PR #23, opened 2020-04-17
RingMLSAG derives Message

mfaulk:
RingMLSAG derives Message instead of manually implementing it.

### DIFF
--- a/transaction/core/src/ring_signature/mlsag.rs
+++ b/transaction/core/src/ring_signature/mlsag.rs
@@ -26,7 +26,7 @@ use crate::{
     onetime_keys::compute_key_image,
     ring_signature::{
         encoding::{read_u8_32, write_u8_32},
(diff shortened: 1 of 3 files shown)


transaction/core/src/ring_signature/mlsag.rs
```diff
@@ -34,88 +34,21 @@ fn hash_to_point(address: &Address) -> RistrettoPoint {
 
 /// MLSAG for a ring of public keys and amount commitments.
 /// Note: Serialize and Deserialize appear to be cruft left over from sdk_json_interface.
-#[derive(Clone, Debug, Default, Digestible, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Clone, Digestible, PartialEq, Eq, Serialize, Deserialize, Message)]
```
mfaulk, line 37: The changes to this struct are the main point of this PR.

PR #24, opened 2020-04-17
MC-1117 8-byte masked_value

mfaulk:


consensus/api/proto/external.proto
```diff
@@ -75,10 +75,10 @@ message Amount {
     // A Pedersen commitment `v*G + s*H`
     CurvePoint commitment = 1;
 
-    // `masked_value = value + SHA3-512_scalar(shared_secret || n)`
-    CurveScalar masked_value = 2;
+    // `masked_value = value XOR_8 Blake2B("value_mask" || shared_secret)`
```
mfaulk, line 78: This is the main change.

PR #25, opened 2020-04-17
SignatureRctBulletproofs derives Message

mfaulk:


transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -42,109 +42,23 @@ use crate::{
 };
 
 /// An RCT_TYPE_BULLETPROOFS_2 signature.
-#[derive(Clone, Debug, Default, Digestible, Eq, PartialEq, Serialize, Deserialize)]
+#[derive(Clone, Digestible, Eq, PartialEq, Serialize, Deserialize, Message)]
```
mfaulk, line 45: The changes to this struct are the main point of this PR.

PR #30, opened 2020-04-17
Minor cleanup for Amount

mfaulk:
The main change here is removing the type alias "Blinding", and letting Amount::new take the blinding as a plain old Scalar. 

I think this removes an irritating layer of indirection ("Blinding is a type alias for CurveScalar which is a Prost-friendly wrapper for Scalar") in several places where a Prost-friendly type is not required, like in the non-public methods of Amount.

### DIFF
--- a/transaction/core/src/amount.rs
+++ b/transaction/core/src/amount.rs
@@ -8,7 +8,7 @@
 #![cfg_attr(test, allow(clippy::unnecessary_operation))]
 
 use crate::{
-    ring_signature::{Blinding, CurveScalar, GENERATORS},
+    ring_signature::{CurveScalar, GENERATORS},
     CompressedCommitment,
 };
 use blake2::{Blake2b, Digest};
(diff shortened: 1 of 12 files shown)


transaction/core/src/amount.rs
```diff
@@ -55,11 +55,11 @@ impl Amount {
     #[inline]
     pub fn new(
         value: u64,
-        blinding: Blinding,
+        blinding: Scalar,
```
mfaulk, line 58: This is the main change.

transaction/core/src/range_proofs/mod.rs
```diff
@@ -8,23 +8,25 @@ use merlin::Transcript;
...
 ///
 /// Creates a proof that each secret value is in the range [0,2^64).
 ///
 /// # Arguments
 /// `values` - Secret values that we want to prove are in [0,2^64).
-/// `serials` - Transaction output serial numbers.
+/// `blindings` - Pedersen commitment blinding for each value.
 ///
+/// # Returns
+/// The proof and the (padded) vector of Pedersen commitments from `values` and `blindings`.
```
xoloki, line 26: It might be good here to mention that the vector is padded to a poiwer of 2.

PR #32, opened 2020-04-18
Namespace and versions

sugargoat:
Fixes MC-1218.

This PR unifies the version number and namespace for our crates, following the Principle of Least Surprise. The idea is to reduce mental overhead (and frequency of grepping the codebase) to determine the origin of crates.

### NOTICE
This PR changes crate names and affects `cargo build/run -p` commands, so you may need to update  your workflows, as well as rebase any WIP PRs.

## Changes
* all crates are fully specified according to the directory structure
* all `[lib.name]` removed, and all references to crates now use the full path
    * this includes features
    * care was taken not to change generated proto file names
* some crates were moved to match logical associations or to deduplicate information in paths and crate names
* Updates documentation for `cargo build/run -p ...`


util/build/script/src/env.rs
```diff
@@ -320,9 +320,8 @@ impl Environment {
                     if value.is_empty() {
                         None
                     } else {
-                        Some(value.parse().map_err(|e| {
-                            EnvironmentError::ParseInt(ENV_CARGO_PKG_VERSION_PRE.to_owned(), e)
-                        })?)
+                        // FIXME: 0.1.3-pre1 cannot be parsed due to "not a digit"
```
sugargoat, line 323: @jcape - this is erroring currently - I'm ham-fisting it to a string, but I wonder if I'm missing something here.

Cargo.toml
```diff
@@ -17,51 +19,49 @@ members = [
     "consensus/scp",
     "consensus/scp/play",
     "consensus/service",
-    "crypto/ake/mcnoise",
+    "crypto/ake/noise",
```
jcape, line 22: Let's make this `mc-crypto-noise`?

Cargo.toml
```diff
@@ -17,51 +19,49 @@ members = [
...
-    "sgx/sgx_css",
-    "sgx/sgx_debug_edl",
-    "sgx/sgx_panic_edl",
-    "sgx/sgx_slog_edl",
+    "sgx/backtrace_edl",
+    "sgx/compat_edl",
+    "sgx/css",
+    "sgx/debug_edl",
+    "sgx/panic_edl",
+    "sgx/slog_edl",
```
jcape, line 44: Could you also remove the underscores from these names (e.g. `backtrace-edl`)

Cargo.toml
```diff
@@ -17,51 +19,49 @@ members = [
...
     "testnet-client",
     "transaction/core",
     "transaction/core/test-utils",
     "transaction/std",
     "util/b58-payloads",
+    "util/build/enclave",
+    "util/build/sgx",
+    "util/build/script",
     "util/build-grpc",
```
jcape, line 53: This should also become `build/grpc`

attest/ake/Cargo.toml
```diff
@@ -1,26 +1,29 @@
...
+[lib]
+name = "attest_ake"
+
 [dependencies]
-attest = { path = "../../attest/core", default-features = false }
-common = { path = "../../common", default-features = false }
-keys = { path = "../../crypto/keys", default-features = false }
-mcnoise = { path = "../../crypto/ake/mcnoise", default-features = false }
-mcserial = { path = "../../util/mcserial", default-features = false }
+mc-attest-core = { version = "0.1.3-alpha-1", path = "../../attest/core", default-features = false }
```
jcape, line 22: Actually looking at this, I think we should leave these version requirements off, what they're buying us is basically "if I introduce a breaking change and correctly bump semver of some crate in my repo, the build will fail", which is a pretty tame advantage to having 2400 lines of changes every time we want to bump our version, so I'd say just strip them entire and just use path like we were.

Sorry for sending you down the rabbit hole on that one.

sugargoat: Np - removed the versions. How do you feel about leaving the versions at the top of the Cargo.toml (currently at 0.1.4-alpha-1)?

jcape: In the `[package]` section? Those are required...

sugargoat: Yeah, my question is whether to leave the changes from this PR, that set them all to 0.14-alpha-1.

jcape: I think path dependencies should not include versions, though the package should. Additionally, we'll now need to make everything 0.1.5-alpha-1, because we've already tagged 0.1.4.

sugargoat: SGTM! Thanks, @jcape

garbageslam:
wow, this is a big cleanup :)

sugargoat:
@jcape - I can do that. The main reason I didn't was to reduce the footprint of this PR to just toml files. Removing those lib names means that every rust file that uses those libs has to change their `use` statements.

PR #36, opened 2020-04-19
Add enclave security linker flags to the enclave linker step

garbageslam:
And commentary about why they are needed / what they do

### DIFF
--- a/mcbuild/enclave/src/lib.rs
+++ b/mcbuild/enclave/src/lib.rs
@@ -547,6 +547,34 @@ impl Builder {
         static_archive.set_extension("a");
         self.build_enclave()?;
 
+        // Note: Some of the linker flags here are important for security [1]
+        //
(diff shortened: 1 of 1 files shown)


garbageslam:
I think these flags got lost in translation from the Makefile to the build.rs based system, but as I have read about enclave hardening, I think they might be important, especially the `noexecstack` one.

I think it might be a good idea to either
(1) integrate a tool like `checksec.sh` into the signing process: https://github.com/slimm609/checksec.sh but it's a little wierd, because if you read that tool, it sometimes wants to hit the internet.
(2) Look at the checks it is doing for PIE and NX, which use the `readelf` tool, and make the build.rs do those two checks after the linking step, to validate that turning on the securiy features in the linking step works. I think any build that is building the enclave is going to be an x86 linux build anyways, so adding build dep on readelf seems reasonable? Alternatively, if we can't find readelf, just print a warning, but make sure it is installed an available in the actual build docker container?

PR #47, opened 2020-04-21
MC-1330 Removes masked_blinding from Amount

mfaulk:
The Amount blinding is now derived from the output shared secret, which allows us to remove the `masked_blinding` field in Amount.

### DIFF
--- a/consensus/enclave/impl/src/lib.rs
+++ b/consensus/enclave/impl/src/lib.rs
@@ -28,7 +28,6 @@ use consensus_enclave_api::{
     WellFormedEncryptedTx, WellFormedTxContext,
 };
 use core::convert::{TryFrom, TryInto};
-use digest::Input;
 use digestible::Digestible;
 use identity::Ed25519Identity;
 use keys::{Ed25519Pair, Ed25519Public, RistrettoPrivate, RistrettoPublic, X25519Public};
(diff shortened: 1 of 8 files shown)


consensus/enclave/impl/src/lib.rs
```diff
@@ -429,6 +425,7 @@ impl ConsensusEnclave for SgxConsensusEnclave {
         let fee_tx_private_key = {
             let hash_value: [u8; 32] = {
                 let mut hasher = Blake2b256::new();
+                // TODO: domain separator.
```
mfaulk, line 428: Domain separators are also on my agenda, https://mobilecoin.atlassian.net/browse/MC-539

transaction/core/src/amount.rs
```diff
@@ -47,10 +47,6 @@ pub struct Amount {
     /// `masked_value = value XOR_8 Blake2B(value_mask | shared_secret)`
     #[prost(uint64, required, tag = "2")]
     pub masked_value: u64,
-
-    /// `masked_blinding = blinding + Blake2B(blinding_mask | shared_secret))
-    #[prost(message, required, tag = "3")]
-    pub masked_blinding: CurveScalar,
```
mfaulk, line 53: Removing this is the main change.

transaction/core/src/tx.rs
```diff
@@ -271,14 +271,14 @@ impl TxOut {
         recipient: &PublicAddress,
         tx_private_key: &RistrettoPrivate,
         hint: EncryptedFogHint,
-        rng: &mut RNG,
+        _rng: &mut RNG,
     ) -> Result<Self, AmountError> {
```
xoloki, line 275: Can we remove this rng from the API?

mfaulk: Good catch. I forgot about that. Will do.

PR #50, opened 2020-04-21
MC-1144 mc-sgx-core-types

jcape:
This is mostly copy-paste of rust types from the attest crate, with some slight modifications:

1. This is just local attestation and basic enclave types---IAS/EPID stuff will live in other crates.
1. The `impl_X` macros are public, for use in other crates---e.g. the IAS/EPID stuff should use it.
1. The `FfiWrapper` type is just a marker trait to ensure everything implements a common API.
1. This uses the `mc_encodings::{ToX64, FromX64}` traits, and drops the customized `SgxWrapperType` FFI stuff.
1. Attribute flags are now bitflags.
1. Serialization of complex types includes more validation by parsing into the rust type, then extracting the inner FFI types.
1. The `Into<Vec<u8>>` implementations removed (redundant to `ToX64`).
1. `ToX64` and `FromX64` have simplified method names.
1. Bindgen's `-sys` crates aren't implementing copy/clone, so we're just going to add `Clone` impl as a requirement. This is probably for the best, even if a 512B struct is memcpy-able, you probably don't want rust to do that for you automatically.

The longer-term goal here is to build up a more rust-like SGX codebase, rooted in auto-generated FFI structs with an eye towards migrating away from EPID/IAS alongside our existing "attest/baidu" stuff we have now, then cut over.

### DIFF
--- a/sgx/epid-types-sys/build.rs
+++ b/sgx/epid-types-sys/build.rs
@@ -2,8 +2,52 @@
 
 //! Build script to generate bindings for the Intel SGX SDK core FFI types
 
-use bindgen::Builder;
+use bindgen::{
+    callbacks::{IntKind, ParseCallbacks},
+    Builder,
+};
 use mcbuild_utils::Environment;
+
+#[derive(Debug)]
+struct Callbacks;
+
+impl ParseCallbacks for Callbacks {
+    fn int_macro(&self, name: &str, _value: i64) -> Option<IntKind> {
+        if name.ends_with("_SIZE") || name.ends_with("_BYTES") {
+            Some(IntKind::Custom {
+                name: "usize",
+                is_signed: false,
+            })
+        } else {
+            None
+        }
+    }
+
+    fn item_name(&self, name: &str) -> Option<String> {
+        if name.starts_with("_sgx_") {
+            Some(name[1..].to_owned())
+        } else if name.starts_with('_') {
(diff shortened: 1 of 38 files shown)


sgx/epid-types-sys/build.rs
```diff
@@ -2,16 +2,60 @@
...
+            retval.push_str(name);
+            // Cleanup the _quote_nonce, _platform_info, and _update_info_bit types.
+            //
+            // Yelling in the wilderness:
+            //
+            // Is there someone going around teaching students that "`_t` means it's a C type"
+            // without further explanation? The `_t` suffix exists for POSIX to namespace itself
+            // from the code of mere mortals, not for every damned fool to remind themselves what
+            // language they are writing code for.
+            //   -jmc
```
garbageslam, line 40: hahaha...

My recollection from C++ is that it is common that someone will put `_t` on a type, and no `_t` for variables of that type. Because in both cases they are typically snake case, so it can make it a little easier to humans to parse it at a glance. At least that was what I understood about it. I think it's common in C standard library also with like the `cstdint.h` header having `uint64_t` and `size_t` and `usize_t`.

jcape: Oh yeah, I've seen it all over, my complaint is that it conflicts with http://www.gnu.org/software/libc/manual/html_node/Reserved-Names.html ;-)

PR #51, opened 2020-04-21
MC-1313 Includes range proofs and pseudo output commitments in the signature

mfaulk:


transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -317,13 +305,53 @@ fn sign_with_balance_check<CSPRNG: RngCore + CryptoRng>(
...
+    let extended_message = extend_message(message, &pseudo_output_commitments, &range_proof_bytes)?;
+
+    // Prove that the signer is allowed to spend a public key in each ring, and that
+    // the input's value equals the value of the pseudo_output.
+    let mut ring_signatures: Vec<RingMLSAG> = Vec::new();
+    for i in 0..num_inputs {
+        let real_index = real_input_indices[i];
+        let (onetime_private_key, value, blinding) = input_secrets[i];
+        let ring_signature = RingMLSAG::sign(
+            &extended_message,
```
mfaulk, line 320: This is the main change. Here, the input `message` is the hash of the transaction prefix which does not contain the range proof or pseudo output commitments.

transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -317,13 +305,53 @@ fn sign_with_balance_check<CSPRNG: RngCore + CryptoRng>(
...
     })
 }
 
+/// Concatenates [message || pseudo_output_commitments || range_proof].
+fn extend_message(
+    message: &[u8],
+    pseudo_output_commitments: &[CompressedCommitment],
+    range_proof_bytes: &[u8],
+) -> Result<Vec<u8>, Error> {
+    let pseudo_output_bytes = serialize(&pseudo_output_commitments).unwrap(); // TODO
```
mfaulk, line 345: Will do

PR #57, opened 2020-04-22
MC-1337 form_block sorts outputs and key images

mfaulk:


consensus/enclave/impl/src/lib.rs
```diff
@@ -470,10 +470,13 @@ impl ConsensusEnclave for SgxConsensusEnclave {
         }
         outputs.push(fee_output);
 
-        // TODO: sort outputs and key images.
+        // Sort outputs and key images. This removes ordering information which could be used to
+        // infer the per-transaction relationships among outputs and/or key images.
```
garbageslam, line 474: you probably eventually want these to be like, side-channel resistant sorts

PR #70, opened 2020-04-22
MC-1342 Move FromRandom trait from keys to mc-utils-from-random

jcape:
This splits off the FromRandom trait into it's own crate that isn't specifically tied to keys. This is done to allow other, non-key-related stuff (e.g. random nonces) to use the same API without pulling in all of Dalek crypto.


garbageslam:
my only thought is, maybe keep re-exporting `FromRandom` out of keys?

idk, it's a judgment call

util/from-random/README.md
```diff
@@ -0,0 +1,24 @@
...
+impl FromRandom for MyStruct {
+    fn from_random<R: CryptoRng, RngCore>(csprng: &mut R) -> Self {
+        let mut bytes = [0u8; 32];
+        csprng.fill_bytes(&bytes);
+        MyStruct {
+            bytes,
+        }
+    }
+}
+```
```
garbageslam, line 24: this is nice, I guess I should put more actual example code in my readme's or something

sugargoat: Yeah this is great - speaking of, @jcape you could throw in the `#![doc(include = "../README.md")]` at the top of lib.rs.

util/from-random/src/lib.rs
```diff
@@ -0,0 +1,13 @@
+//! A trait which provides a common API for types which can be initialized
```
sugargoat, line 1: also drop the Copyright in here

sugargoat, line 1:
```suggestion
// Copyright (c) 2018-2020 MobileCoin Inc.

#![feature(external_doc)]
#![doc(include = "../README.md")]
//! A trait which provides a common API for types which can be initialized
```

PR #73, opened 2020-04-23
MC-1324 Merkle tree uses digest instead of mcserial::serialize

mfaulk:


transaction/core/src/membership_proofs/mod.rs
```diff
@@ -16,31 +16,39 @@ use mcserial::serialize;
...
 
-// Merkle tree Hash function for internal nodes.
-pub fn internal_hash_fn(bytes: &[u8]) -> [u8; 32] {
-    const NODE_PREFIX: u8 = 0x01;
-    hash_with_prefix(&[NODE_PREFIX], bytes)
+/// Merkle tree hash function for a leaf node.
+pub fn hash_leaf(tx_out: &TxOut) -> [u8; 32] {
+    let mut hasher = Blake2b256::new();
+    hasher.input(&TXOUT_MERKLE_LEAF);
+    tx_out.digest(&mut hasher);
```
mfaulk, line 35: Uses `digest`

ledger/db/src/tx_out_store.rs
```diff
@@ -1005,15 +996,11 @@ pub mod tx_out_store_tests {
                              tx_out_0                tx_out_1
         */
         let tx_out_one: &TxOut = tx_outs.get(1).unwrap();
-        let leaf_hash_one = leaf_hash_fn(&serialize(tx_out_one).unwrap());
+        let leaf_hash_one = hash_leaf(&tx_out_one);
```
mfaulk, line 999: This is the main change.

garbageslam:
this looks good, im really glad to see the `mcserial::serialize` go away
there's some parts of this function I don't really understand yet, like, why do we have to sort the ranges? we could instead require that they already be formatted in the correct order?

but it's not clear to me that that's actually needed for correctness. i'm going to try to drill down on the merkle proof validation routine later

PR #74, opened 2020-04-23
Make Ristretto implement Kex, make CryptoBox generic over Kex

garbageslam:
There's some potentially controversial stuff here, esp. with the
`GetBytes` trait in `keys` crate. We can look at redoing that
aspect or making it better.

I'm going to review the diff now and mark what lines were the "motivation" for the change. There's two distinct steps here.


crypto/box/src/hkdf_box.rs
```diff
@@ -0,0 +1,167 @@
...
+/// Represents an implementation of Ristretto-Box using Hkdf<Blake2b> and Aes128Gcm
+///
+/// This structure contains the actual cryptographic primitive details, and
+/// specifies part of the wire format of the "footer" where the ephemeral
+/// public key comes first, and the mac comes second.
+///
+/// Note: If when instantiating this, the IsLessOrEqual bounds fail,
+/// it may indicate that your Kex algorithm does not supply enough entropy to properly
+/// drive the Aead algorithm that you have chosen, and so the combination would be insecure.
+pub struct HkdfBox<KexAlgo, DigestAlgo, AeadAlgo>
```
garbageslam, line 26: Rewriting the CryptoBox implementation to be generic in this way (over Kex and Aead) has been a longstanding goal, that was blocked in part by lack of `Ristretto: Kex`.

crypto/box/src/hkdf_box.rs
```diff
@@ -0,0 +1,167 @@
...
+where
+    KexAlgo: Kex,
+    for<'privkey> <KexAlgo as Kex>::Public: From<&'privkey <KexAlgo as Kex>::EphemeralPrivate>,
+    DigestAlgo: Digest + Input + FixedOutput + Default + Clone + BlockInput + Reset,
+    AeadAlgo: Aead + NewAead,
+    AeadAlgo::NonceSize:
+        IsLessOrEqual<<KexAlgo::Secret as KexSecret>::EntropyLowerBound, Output = B1>,
+    AeadAlgo::KeySize:
+        IsLessOrEqual<<KexAlgo::Secret as KexSecret>::EntropyLowerBound, Output = B1>,
```
garbageslam, line 35: This bound is checking that the construction (the combination of Kex or Aead) is correct: The Aead key size does not require more entropy than could be obtained from the shared secret. You need this to satisfy the assumptions of the security proof of the Aead, in general.

This will prevent bad things like trying to drive Aes256Gcm using shared secret extracted from Ristretto -- you won't be able to achieve a 256 bit security level that way.

crypto/box/src/hkdf_box.rs
```diff
@@ -0,0 +1,167 @@
...
+    <<KexAlgo as Kex>::Public as PublicKey>::Size: Add<<AeadAlgo as aead::Aead>::TagSize>,
+    Sum<<KexAlgo::Public as PublicKey>::Size, AeadAlgo::TagSize>: ArrayLength<u8>,
+{
+    /// KDF part, factored out to avoid duplication
+    /// This part must produce the key and IV/nonce for aes-gcm
+    /// Blake2b produces 64 bytes of private key material which is more than we need,
+    /// so we don't do the HKDF-EXPAND step.
```
garbageslam, line 129: I have to update this comment. Until rust implements Specialization of generics, there's not an easy way to implement that optimization. So we now just always do the HKDF-EXPAND step.

crypto/ake/mcnoise/src/symmetric_state.rs
```diff
@@ -138,8 +138,8 @@ where
     ///
     /// Updates our hash to reflect a new hash of itself and the provided
     /// data.
-    pub fn mix_hash(&mut self, data: &[u8]) {
-        self.hash += data;
+    pub fn mix_hash<D: AsRef<[u8]>>(&mut self, data: D) {
```
garbageslam, line 141: this very small change allows to avoid writing `.as_ref()` in a lot of places

crypto/keys/src/ed25519.rs
```diff
@@ -192,11 +190,11 @@ impl Verifier<Ed25519Signature> for Ed25519Public {
 }
 
 impl TryFrom<&[u8]> for Ed25519Public {
-    type Error = SignatureError;
+    type Error = KeyError;
```
garbageslam, line 193: This change seems insubstantial, and it makes it much much easier to do generic code with `Kex` when you know the error type here will be `KeyError`

crypto/keys/src/traits.rs
```diff
@@ -106,10 +123,15 @@ pub trait PublicKey:
     + Ord
     + Serialize
     + Sized
-    + for<'bytes> TryFrom<&'bytes [u8]>
+    + for<'bytes> TryFrom<&'bytes [u8], Error = KeyError>
```
garbageslam, line 126: this change (adding `Error = KeyError` bound) makes it much easier to implement the generic cryptobox, otherwise I have to assume that any error type could fly out of this thing. the only thing not returning `KeyError` was `ed25519` which was returning a default `SignatureError`, so seems like not a big deal.

crypto/keys/src/traits.rs
```diff
@@ -43,6 +44,22 @@ impl From<LengthMismatch32> for KeyError {
...
+///
+/// This trait attempts to bridge this gap, using a zero-cost abstraction.
+/// An object like X25519 can have GetBytes::Result = &[u8],
+/// and this object already implements AsRef<[u8]>, which can be inlined to a no-op.
+/// An object like Ristretto can have GetBytes::Result = [u8; 32] or similar.
+pub trait GetBytes {
+    type Output: AsRef<[u8]>;
+
+    fn get_bytes(&self) -> Self::Output;
+}
```
garbageslam, line 61: this is possibly the most controversial part of the whole PR
I introduce this trait because in `trait Kex` we were requiring the `PublicKey` to implement `AsRef<[u8]>` which `RistrettoPublic` simply cannot do -- it must be compressed before it can be viewed as bytes.

This trait is another layer of indirection, so that it can return e.g. `Output = [u8;32]` which itself implements `AsRef<[u8]>`.

It was my hope that `X25519` could keep implementing this with `get_bytes(&self) -> &[u8]` since `&[u8]` implements `AsRef<[u8]>`, but I haven't been able to convince the compiler to build that yet.

crypto/keys/src/traits.rs
```diff
@@ -195,7 +231,7 @@ where
 pub trait Kex {
     type Public: KexPublic<KexEphemeralPrivate = Self::EphemeralPrivate>
         + for<'reusable> From<&'reusable Self::Private>
-        + AsRef<[u8]>;
+        + GetBytes;
```
garbageslam, line 234: The motivation for this change was so that `RistrettoPublic` can satisfy `Kex::Public`.

crypto/box/src/hkdf_box.rs
```diff
@@ -0,0 +1,169 @@
...
+        GenericArray<u8, AeadAlgo::KeySize>,
+        GenericArray<u8, AeadAlgo::NonceSize>,
+    ) {
+        let kdf = Hkdf::<DigestAlgo>::new(Some(b"dei-salty-box"), dh_secret.as_ref());
+        let mut key: GenericArray<u8, AeadAlgo::KeySize> = Default::default();
+        let mut nonce: GenericArray<u8, AeadAlgo::NonceSize> = Default::default();
+        kdf.expand(b"aead-key", key.as_mut_slice())
+            .expect("Aead::KeySize is too large compared to Digest output size");
+        kdf.expand(b"aead-nonce", nonce.as_mut_slice())
+            .expect("Aead::NonceSize is too large compared to Digest output size");
```
jcape, line 144: Not sure if you intended this, but nonce and key will not be successive expansions on an underlying state, it will run the same expansion twice, with different info parameters. If you want to derive multiple keys in the "usual" KDF way, you need to call expand with a buffer that is as long as both your key/nonce combined, and splice them.

Now that I'm thinking about it, we could probably use a [one-way Noise handshake](https://noiseprotocol.org/noise.html#one-way-handshake-patterns) to accomplish all this...

crypto/ake/mcnoise/src/symmetric_state.rs
garbageslam: It could in theory, but I think it doesn't as long as the thing you are passing is actually a ref. But at this specific revision, yes it is a copy, because the `get_bytes()` thing in this revision is basically the same as `to_bytes` in the `ReprBytes` PR.

I'm going to rebase this branch on the `ReprBytes` PR and get rid of the `GetBytes` thing so we can see better what the end state is actually going to be. It will make the PR larger but I think this PR is not that big in the first place. I guess I'm going to make that a third PR and then close this one and the other one?

crypto/box/src/hkdf_box.rs
garbageslam: yeah, you're right

I looked up Aes-256-GCM here: https://github.com/RustCrypto/AEADs/blob/master/aes-gcm/src/lib.rs#L142

I guess that this version has a 12 byte nonce, the same size as the Aes-128-GCM version?

It seems weird that the nonce is not as large as the key, and does not get larger as the security parameter increases. Maybe they are being combined somehow?

RistrettoSecret has very slightly less than 32 bytes of entropy, and AES-256-GCM needs 32 bytes of entropy for the key, but that should be fine.

The only way I think that this bound will fail is if the user picks really weird combinations, and as you said off-line, it should be obvious to them that the chain is only as strong as the weakest link. We can offer some guidance in the README

crypto/box/src/hkdf_box.rs
garbageslam: thanks, i didn't think of this, i was mostly just mimicking how we do the account key derivation. there was something I saw online that suggested using hkdf this way... i'm going to try to find a link

garbageslam: > we could probably use a one-way Noise handshake to accomplish all this...

that would be cool! i'm sure if you have embedded devices that already want to use noise for secure channels, they would appreciate a smaller footprint from not having hkdf?

i'm not sure if we need hkdf anywhere else in any of the enclaves?

garbageslam: also, bonus points for making a type called "noise box"

crypto/ake/mcnoise/src/symmetric_state.rs
garbageslam: i guess we should revert this in the next version, since you're right, it could be a performance leak

crypto/box/src/hkdf_box.rs
garbageslam: @jcape does the `mcnoise` crate have a dependency on `Vec` / `alloc`? Cause right now, `cryptobox` and `hkdf_box` do not. That would be a reason to keep hkdf box around

garbageslam: yah it looks to me like it does

if you want for like, a tiny off-line device to be able to construct transactions, including fog hints, then it might not want to have noise as a dependency?

attest/ake/src/initiator.rs
```diff
@@ -186,6 +186,7 @@ where
             HandshakeStatus::Complete(result) => {
                 let remote_report: VerificationReport =
                     deserialize(&output.payload).map_err(|_e| Error::ReportDeserialization)?;
+                #[allow(clippy::redundant_closure)]
```
garbageslam, line 189: clippy is apparently wrong, the code doesn't build without the closure

crypto/box/src/hkdf_box.rs
garbageslam: i think its better to have a simple proof of security than a complicated one... the one in the readme for hkdf box is like a few lines -- the statement being proved in the paper you linked is like vastly more complicated and proves something far more powerful. also it makes it clear that different properties are reached at different stages. i can't figure out what it guarantees for my scenario without reading the whole paper

crypto/box/src/hkdf_box.rs
```diff
@@ -0,0 +1,164 @@
...
+    Sum<<KexAlgo::Public as ReprBytes>::Size, AeadAlgo::TagSize>: ArrayLength<u8>,
+    GenericArray<u8, <<KexAlgo as Kex>::Public as ReprBytes>::Size>: Concat<
+        u8,
+        AeadAlgo::TagSize,
+        Rest = GenericArray<u8, AeadAlgo::TagSize>,
+        Output = GenericArray<
+            u8,
+            <<<KexAlgo as Kex>::Public as ReprBytes>::Size as Add<AeadAlgo::TagSize>>::Output,
+        >,
+    >,
```
jcape, line 60: This isn't a valid constraint for an Aead.

crypto/box/src/hkdf_box.rs
jcape: My thinking on this is simple: a published proof is easier to trust than an unpublished one, an existing scheme is easier to trust than a novel one, and audited code is easier to trust than un-audited code. If/when vec becomes a problem, we can replace it, `aead` has in-place APIs to handle things now.

crypto/box/src/hkdf_box.rs
garbageslam: The compiler tells me this is needed to make `curve_point_bytes.concat(mac)` compile,
I think all I'm telling it is that `KexAlgo::Public::Size + AeadAlgo::TagSize` is a number, and the size of a valid generic array.
I'm not sure though, why this is needed at this revision and not the previous revision.

jcape:
I'll review again when this is updated after #79.

garbageslam:
this is now rebased on the repr bytes PR

util/serial/src/lib.rs
```diff
@@ -7,9 +7,6 @@ use alloc::vec::Vec;
 
 pub extern crate prost;
 
-pub mod helpers;
-pub use helpers::{LengthMismatch32, ReprBytes32};
-
```
garbageslam, line 12: removing this is one of the main points of the PR

garbageslam:
rebased on master

crypto/keys/src/x25519.rs
```diff
@@ -187,13 +187,23 @@ impl AsRef<[u8]> for X25519Public {
...
-impl Into<Vec<u8>> for X25519Public {
-    /// Dump a public key into a new vector.
-    fn into(self) -> Vec<u8> {
-        Vec::from(&self.0.as_bytes()[..])
+impl TryFrom<&[u8]> for X25519Public {
+    type Error = KeyError;
+
+    /// Try to load the given byte slice as a public key.
+    fn try_from(src: &[u8]) -> Result<Self, <Self as TryFrom<&[u8]>>::Error> {
+        if src.len() != 32 {
```
jcape, line 195: As long as we're touching this, can we make "32" here a constant?

attest/ake/src/initiator.rs
```diff
@@ -196,14 +197,12 @@ where
...
-                            .as_ref()
-                            .ok_or(Error::MissingRemoteIdentity)?
-                            .as_ref(),
-                    )
-                    .map_err(|_e| Error::BadRemoteIdentity)?,
+                    &result
+                        .remote_identity
+                        .as_ref()
+                        .ok_or(Error::MissingRemoteIdentity)?
+                        .map_bytes(|bytes| ReportDataMask::try_from(bytes))
```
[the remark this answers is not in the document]
garbageslam: thank you!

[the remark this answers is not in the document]
garbageslam: this gave me some wierd errors:

```
error[E0631]: type mismatch in function arguments
   --> attest/ake/src/initiator.rs:203:36
    |
203 |                         .map_bytes(ReportDataMask::try_from)
    |                                    ^^^^^^^^^^^^^^^^^^^^^^^^
    |                                    |
    |                                    expected signature of `for<'r> fn(&'r [u8]) -> _`
    |                                    found signature of `fn(_) -> _`

error[E0271]: type mismatch resolving `for<'r> <fn(_) -> core::result::Result<mc_attest_core::types::report_data::ReportDataMask, <mc_attest_core::types::report_data::ReportDataMask as core::convert::TryFrom<_>>::Error> {<mc_attest_core::types::report_data::ReportDataMask as core::convert::TryFrom<_>>::try_from} as core::ops::FnOnce<(&'r [u8],)>>::Output == _`
   --> attest/ake/src/initiator.rs:203:26
    |
203 |                         .map_bytes(ReportDataMask::try_from)
    |                          ^^^^^^^^^ expected bound lifetime parameter, found concrete lifetime

error[E0631]: type mismatch in function arguments
   --> attest/ake/src/responder.rs:161:28
    |
161 |                 .map_bytes(ReportDataMask::try_from)
    |                            ^^^^^^^^^^^^^^^^^^^^^^^^
    |                            expected signature of `for<'r> fn(&'r [u8]) -> _`
    |                            found signature of `fn(_) -> _`

error[E0271]: type mismatch resolving `for<'r> <fn(_) -> core::result::Result<mc_attest_core::types::report_data::ReportDataMask, <mc_attest_core::types::report_data::ReportDataMask as core::convert::TryFrom<_>>::Error> {<mc_attest_core::types::report_data::ReportDataMask as core::convert::TryFrom<_>>::try_from} as core::ops::FnOnce<(&'r [u8],)>>::Output == _`
   --> attest/ake/src/responder.rs:161:18
    |
161 |                 .map_bytes(ReportDataMask::try_from)
    |                  ^^^^^^^^^ expected bound lifetime parameter, found concrete lifetime

error: aborting due to 4 previous errors
```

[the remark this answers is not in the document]
garbageslam: https://users.rust-lang.org/t/why-do-i-see-expected-bound-lifetime-parameter-found-concrete-lifetime-when-closure-does-not-specify-explicit-param-types/26175

PR #76, opened 2020-04-23
MC-539 Transaction domain separator tags

mfaulk:
Collects domain separator tags into a central place so that they can be standardized, documented, etc.

This PR moves the existing domain separators. Other PRs will add more where needed.

### DIFF
--- a/transaction/core/src/domain_separators.rs
+++ b/transaction/core/src/domain_separators.rs
@@ -2,7 +2,7 @@
 
 //! Domain separation tags for hash functions used in the MobileCoin transaction protocol.
 //!
-//! Domain separation allows multiple distinct hash functions to be derived a single base function:
+//! Domain separation allows multiple distinct hash functions to be derived from a single base function:
 //!   Hash_1(X) = Hash("Hash_1" || X),
 //!   Hash_2(X) = Hash("Hash_2" || X),
 //!   etc.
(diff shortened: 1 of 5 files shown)


transaction/core/src/ring_signature/mlsag.rs
```diff
@@ -16,11 +16,15 @@ use serde::{Deserialize, Serialize};
 use crate::{
     commitment::Commitment,
     compressed_commitment::CompressedCommitment,
+    domain_separators::RING_MLSAG_CHALLENGE_DOMAIN_TAG,
     ring_signature::{CurveScalar, Error, KeyImage, Scalar, GENERATORS},
 };
 
+// This needs to be the same "Hp" function used by the onetime keys.
```
mfaulk, line 23: This is probably next on my todo list.

transaction/core/src/domain_separators.rs
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) 2018-2020 MobileCoin Inc.
+
+//! Domain separation tags for hash functions used in the MobileCoin transaction protocol.
+//!
+//! Domain separation allows multiple distinct hash functions to be derived a single base function:
```
sugargoat, line 5:
```suggestion
//! Domain separation allows multiple distinct hash functions to be derived from a single base function:
```

PR #77, opened 2020-04-23
Removes unused imports in the `transaction` crate

mfaulk:


transaction/core/src/lib.rs
```diff
@@ -2,7 +2,6 @@
 
 #![no_std]
 // #![deny(missing_docs)]
-#![allow(unused_imports)] // During development...
```
mfaulk, line 5: My bad.

PR #78, opened 2020-04-23
KeyImage cleanup

mfaulk:
Replaces some "compute_key_image" functions with KeyImage::from. I think this makes it easier to figure out how to use KeyImage by reading key_image.rs .

### DIFF
--- a/transaction/core/src/ring_signature/key_image.rs
+++ b/transaction/core/src/ring_signature/key_image.rs
@@ -10,11 +10,13 @@ use serde::{Deserialize, Serialize};
 
 use super::Error;
 use crate::ring_signature::Scalar;
+use blake2::Blake2b;
 use core::convert::TryInto;
 use curve25519_dalek::constants::RISTRETTO_BASEPOINT_POINT;
+use keys::{RistrettoPrivate, RistrettoPublic};
 
 #[derive(Copy, Clone, Default, Eq, Serialize, Deserialize, Digestible)]
-/// The "image" of a private key `x`: I = x * H(x * G) = x * H(P).
+/// The "image" of a private key `x`: I = x * Hp(x * G) = x * Hp(P).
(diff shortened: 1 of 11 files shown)


transaction/core/src/ring_signature/key_image.rs
```diff
@@ -10,87 +10,97 @@ use serde::{Deserialize, Serialize};
...
+impl fmt::Debug for KeyImage {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "KeyImage({})", hex_fmt::HexFmt(self.as_bytes()))
     }
 }
 
-impl AsRef<[u8]> for KeyImage {
-    fn as_ref(&self) -> &[u8] {
-        &self.0.as_bytes()[..]
+impl From<&RistrettoPrivate> for KeyImage {
```
mfaulk, line 42: This is the main change.

transaction/core/src/ring_signature/key_image.rs
```diff
@@ -10,87 +10,97 @@ use serde::{Deserialize, Serialize};
...
 use core::convert::TryInto;
 use curve25519_dalek::constants::RISTRETTO_BASEPOINT_POINT;
+use keys::{RistrettoPrivate, RistrettoPublic};
 
 #[derive(Copy, Clone, Default, Eq, Serialize, Deserialize, Digestible)]
-/// The "image" of a private key `x`: I = x * H(x * G) = x * H(P).
-pub struct KeyImage(pub(crate) CompressedRistretto);
+/// The "image" of a private key `x`: I = x * Hp(x * G) = x * Hp(P).
+pub struct KeyImage {
+    pub point: CompressedRistretto,
```
mfaulk, line 21: I changed this from a tuple to a struct because I find `key_image.point` easier to think about than `key_image.0`.

PR #79, opened 2020-04-23
Move ReprBytes32 to its own crate and generalize it

garbageslam:
We currently have a lot of code that is like manually grabbing
`prost::encoding::*;` and using it to create custom serializations
with prost.

This is a hazard, because that is not part of the public-facing
interface of prost and it is not documented. It also changes
periodically as upstream refactors the library. The only reason
these symbols are publicly visible at all is that there is no way
that `prost-derive` can function if the proc-macro generated code
cannot access these symbols. When we grab `prost::encodings::*;`
it is creating a maintanence hazard that will make it very hard
ever to uprev prost again.

In the past, every cryptographic primitive that went into `prost`
or `serde` required custom bindings code, which was very long.
ReprBytes32 was a quick stopgap to prevent the code duplication
and take us down a route that would be more maintainable.

It turns out that we have some types that are not exactly 32
bytes, and then we have started grabbing the `prost::encodings::*`
again to support those.

Separately from this, we have had a problem in `keys` crate where
we need to be able to get the bytes of any `Kex` public key.
Previously, `Kex` was only implemented for `X25519` and was not
implemented for `Ristretto`, which prevented us from writing generic
(curve agnostic) code in a lot of places.

In https://github.com/mobilecoinofficial/mobilecoin/pull/74 we fixed
this, but we needed to introduce a trait `GetBytes` as a stopgap.

I propose that `GetBytes` can go away and be replaced with this
version of `ReprBytes`. I propose that `ReprBytes32` can also go
away and be replaced with this version of `ReprBytes`.

I believe that `ReprBytes` should potentially be moved out of
mobilecoin repo and treated as a "fundamental trait" in the RustCrypto
ecosystem. Then, crates like `prost` and `serde` would integrate with
`ReprBytes` trait, and e.g. `Dalek` crates would not have to know
about `prost` or `Dalek` would not have to know about `prost` or
`serde`, they would only have to know about `ReprBytes`.

Even if RustCrypto folks don't want to take ownership of this,
this will reduce a lot of nasty code duplication in our repo now,
and may help to clean up the keys crate, and make it able to achieve
its goals of abstracting both X25519 and Ristretto.

---

Alternatives: In `keys` crate, the main alternative is, try to make everything implement `AsRef<[u8]>` and `Into<Vec<u8>>`. Because these are the "standard library interfaces" for doing this.
However, traits like

```
/// A trait which all public key structures should implement.
pub trait PublicKey:
    Clone
    + Debug
    + DeserializeOwned
    + Digestible
    + Eq
    + Hash
    + Into<Vec<u8>>
    + PartialEq
    + PartialOrd
    + Ord
    + Serialize
    + Sized
    + for<'bytes> TryFrom<&'bytes [u8]>
{
    /// Retrieve the size of this public key's raw bytes.
    fn size() -> usize;
}
```

Are bad because they force a requirement on `extern crate alloc`. Because they have a *trait bound* on a core trait `Into<Vec<u8>>`, there is no way that a consumer of this API can implement the trait unless they also have access to `alloc` crate.

Suppose we wanted to take `keys` crate and add an `alloc` feature, so that it can be used in microcontroller environments without a standard allocator. All existing elliptic curve implementations do not require an allocator, and all of the code in RustCrypto is friendly to no-alloc development.

If we try to simply do `#[cfg(feature = "alloc")]` to make the `Into<Vec<u8>>` go away, then we are creating global feature coordination problems. Anything that implements `keys` traits will have to have an alloc feature, and that feature must be configured exactly the same way as the way that `keys` crate is configured. This is exactly the same problem that serde has, and the source of an enormous number of build problems, for us, for months. This would not be an "additive" use of features.

Cargo expects / requires that when a crate has features, turning them on is "additive", only adding things to the API, and not removing them. Otherwise, cargo's global feature unification is unsound and causes build breakage. Making a trait bound *stricter* when a feature is turned on is not additive, and the fact that serde itself does this, is a bug.

(My criticism of serde, in this way, with respect to non-additive `std::error::Error`, is the reason that they added this release of serde: https://github.com/serde-rs/serde/releases/tag/v1.0.100. I think the reason that it ended up this way in the first place is that serde is older than `no_std`, and when they created `no_std` they failed / neglected to move `std::error::Error` to core.)

So I think that we should expect to redesign the keys crate so that it does not have any dependence on `Into<Vec<u8>>` if we expect that it will eventually be taken up by RustCrypto folks. I think that a trait like `ReprBytes` can be an adequate replacement for it.

### DIFF
--- a/util/repr-bytes/Cargo.toml
+++ b/util/repr-bytes/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "mc-util-repr-bytes"
+version = "0.1.0"
+authors = ["MobileCoin"]
+edition = "2018"
+readme = "README.md"
+
+[dependencies]
+generic-array = "0.13"
+
+[dev-dependencies]
+prost = { version = "0.6.1", default-features = false, features = ["prost-derive"] }
+serde = { version = "1.0", default-features = false, features = ["alloc"] }
+serde_cbor = { version = "0.11.1", default-features = false, features = ["alloc"] }
--- a/util/repr-bytes/Cargo.toml
+++ b/util/repr-bytes/Cargo.toml
@@ -7,8 +7,12 @@ readme = "README.md"
 
 [dependencies]
 generic-array = "0.13"
+prost = { version = "0.6.1", optional = true, default-features = false, features = ["prost-derive"] }
+serde = { version = "1.0", optional = true, default-features = false, features = ["alloc"] }
+
+[features]
+default = ["alloc", "prost", "serde"]
+alloc = []
 
 [dev-dependencies]
-prost = { version = "0.6.1", default-features = false, features = ["prost-derive"] }
-serde = { version = "1.0", default-features = false, features = ["alloc"] }
 serde_cbor = { version = "0.11.1", default-features = false, features = ["alloc"] }
--- a/util/repr-bytes/src/lib.rs
+++ b/util/repr-bytes/src/lib.rs
@@ -1,0 +1,492 @@
+#![no_std]
+
+pub use generic_array::{typenum, ArrayLength, GenericArray};
+
+use core::fmt::Display;
+use typenum::Unsigned;
+
+/// ReprBytes represents a type that has a canonical representation as a fixed
+/// number of bytes. This interface is meant to support generic programming.
+///
+/// ReprBytes is meant to be general enough to support many forms of cryptographic
+/// primitives. Most cryptographic primitives implement AsRef<[u8]> and TryFrom<&[u8]>,
+/// but not all of them can -- RistrettoPoint requires an (expensive) compression
+/// step before the bytes of the canonical representation can be accessed.
+///
+/// ReprBytes provides an API very close to AsRef<[u8]> and TryFrom<&[u8]> which
+/// can be used in generic code that handles cryptographic primitives, and in
+/// glue code so that these primitives can be used easily with serialization
+/// libraries and frameworks.
+///
+/// The error types are constrained with Display so that both Prost and Serde can make effective use of them
+///
+/// To be useful, ReprBytes wants to provide many "blanket implementations" that
+/// connect it with core traits and traits from Prost and Serde.
+/// However, blanket implementations don't work very well in rust outside of stdlib.
+/// Instead, we provide macros so that these blanket implementations can be obtained
+/// on a per-type basis, and these macros are exported from this crate. We believe
+/// that this is consistent with current best practices around blanket implementations.
+pub trait ReprBytes: Sized {
+    /// A typenum representing the size, in bytes, of the canonical representation
+    type Size: ArrayLength<u8>;
+
+    /// The error type which may be returned by from_bytes.
+    type Error: Display;
+
+    /// Try to convert from canonical representation bytes to this type
+    fn from_bytes(src: &GenericArray<u8, Self::Size>) -> Result<Self, Self::Error>;
+
+    /// Convert to canonical representation bytes.
+    fn to_bytes(&self) -> GenericArray<u8, Self::Size>;
+
+    /// In-place visitation of the canonical bytes representation, using a closure
+    ///
+    /// Implementation note: The default implementation is not the best when your
+    /// type implements AsRef<[u8]>, it will make a needless copy in that case.
+    /// If your type implements AsRef<[u8]>, then you are strongly recommended to use
+    /// the macro `derive_repr_bytes_from_as_ref_try_from`.
+    /// Otherwise the default implementation is probably the best.
+    /// See also the suggested impl `derive_into_vec_from_repr_bytes`.
+    fn map_bytes<T, F: FnOnce(&[u8]) -> T>(&self, f: F) -> T {
+        f(self.to_bytes().as_slice())
+    }
+
+    /// Convenient helper: Get the representation size as a usize
+    fn size() -> usize {
+        Self::Size::USIZE
+    }
+}
+
+/// Error that indicates that we got a different number of bytes than expected
+#[derive(Debug)]
+pub struct LengthMismatch {
+    expected: usize,
+    found: usize,
+}
+
+impl Display for LengthMismatch {
+    fn fmt(&self, formatter: &mut core::fmt::Formatter) -> core::fmt::Result {
+        write!(
+            formatter,
+            "Expected exactly {} bytes, got {}",
+            self.expected, self.found
+        )
+    }
+}
+
+////
+// Suggested Implementations:
+// These macros provide instances of what can be thought of as "blanket implementations"
+// on a per-type basis.
+// These are macros because rust's coherence rules mean that blanket implementations
+// won't work out well for this use-case.
+//
+// There are two types of suggested impls:
+// - Impls of other traits in terms of ReprBytes
+// - Impls of ReprBytes in terms of other traits
+////
+
+/// Derive ReprBytes from AsRef<[u8]>, TryFrom<&[u8]>, and Size as a typenum.
+/// This is expected to be the right implementation for almost all cryptographic
+/// primitives, e.g. X25519, CompressedRistretto, etc.
+/// It can't work for e.g. RistrettoPoint, which doesn't have AsRef<[u8]>.
+///
+/// Arguments:
+///   - $mytype is the type you want to impl ReprBytes
+///   - $mysize is a typenum, representing the size of the canonical representation
+///
+/// Requirements:
+///   - <AsRef<[u8]> for $mytype>::as_ref().len() always equals $mysize::USIZE
+///   - <TryFrom<&[u8]> for $mytype>::Error implements core::fmt::Display
+///   - <TryFrom<&'a[u8]> for $mytype>::Error is the same for all values of 'a,
+///     OR they are all convertible to the value when 'a = 'static, via core::convert::From.
+#[macro_export]
+macro_rules! derive_repr_bytes_from_as_ref_and_try_from {
+    ($mytype:ty, $mysize:ty) => {
+        impl $crate::ReprBytes for $mytype {
+            type Size = $mysize;
+            type Error = <$mytype as TryFrom<&'static [u8]>>::Error;
+
+            fn from_bytes(src: &$crate::GenericArray<u8, Self::Size>) -> Result<Self, Self::Error> {
+                Ok(<Self as TryFrom<&[u8]>>::try_from(src.as_slice())?)
+            }
+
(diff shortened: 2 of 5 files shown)


garbageslam:
Please give me feedback on the general direction here, I want to make sure we agree this is a good idea before I start trying to nuke ReprBytes32.

Please also review https://github.com/mobilecoinofficial/mobilecoin/pull/74

eranrund:
I'm not sure about the whole keys/GetBytes thing (I think @jcape would have valuable input on that), but as a generic solution for the custom serializations we currently have in a few places, this seems like a good alternative to me if it works (I'd like to see how it looks like when replacing one or two of the custom ones).
I can see this living under `mcserial` if the only usage ends up being the prost/serde macros.

garbageslam:
@eranrund the ReprBytes32 lives under `mcserial` right now, but @jcape discussed in the past that that should not be there, because we want for the `keys` crate to have a nice future and `mcserial` is like an internal serialization API. The end goal for `keys` crate I think is that the stuff in `traits.rs` becomes a part of this repo: https://github.com/RustCrypto/traits just like the `aead` trait that James worked on.

garbageslam:
@eranrund I added tests that demonstrate this functionality, and I rebased the Ristretto Kex PR on this so we can see how it looks: https://github.com/mobilecoinofficial/mobilecoin/pull/74

util/repr-bytes/src/lib.rs
```diff
@@ -0,0 +1,492 @@
...
+                        return Err(::prost::DecodeError::new(
+                            LengthMismatch {
+                                expected: expected_size,
+                                found: len as usize,
+                            }
+                            .to_string(),
+                        ));
+                    }
+                    let result = <Self as $crate::ReprBytes>::from_bytes(
+                        (&buf.bytes()[0..expected_size]).try_into().unwrap(),
```
eranrund, line 217: What's try_into converting from/to here?

eranrund:
This looks good to me when it comes to solving the serde/prost ugly implementations. I am not sure on its effects with regards to Kex stuff. This has my okay to replace `ReprBytes32` is @jcape feels good about it.

util/repr-bytes/src/lib.rs
jcape: Tries to copy the slice into an array of the size required by from_bytes in this ccontext. As the slice may have the wrong length, this is fallible. The unwrap should be an expect, tho.

garbageslam:
@jcape This is in limbo for almost a month

We've had some off-line discussions about how this could be redone to be nicer in the future.

There is no clearly scoped path forwards on that.

We should not let perfect be the enemy of good, we have to get things done. This PR has been very carefully thought out, and if not merged, this will be blocking things very soon.

I tried to create a large window of time for discussion and requesting of changes.

Unless you can have concrete changes to request, please don't block this PR any longer.

jcape:
@garbageslam You promised to "circle back" on the in-place API, conducive to merging with `mc_util_encoding::{FromX64, ToX64}` multiple times, and never did. Eventually on May 1, I said I'd try to help out after I was done with the micro-architectural hardening, which turned out to be much tougher than anticipated, and continued until last Wednesday. I appreciate this PR has been sitting for a while, but I thought we had a direction for it.

I'm sorry I forgot to take this over last Wednesday, but I was reviewing your DB PR instead, because it seemed like that was a higher priority.

util/repr-bytes/src/lib.rs
```diff
@@ -0,0 +1,492 @@
+#![no_std]
```
jcape, line 1: Needs copyright and module description

util/repr-bytes/src/lib.rs
```diff
@@ -0,0 +1,492 @@
...
+            }
+        }
+    };
+}
+
+/// Derive Into<Vec<u8>> from a ReprBytes implementation
+#[macro_export]
+macro_rules! derive_into_vec_from_repr_bytes {
+    ($mytype:ty) => {
+        impl Into<::alloc::vec::Vec<u8>> for $mytype {
```
jcape, line 136: Same story here, in case the crate downstream of us isn't doing it's own `extern crate alloc;` directly.

util/repr-bytes/src/lib.rs
```diff
@@ -0,0 +1,492 @@
...
+///   - <AsRef<[u8]> for $mytype>::as_ref().len() always equals $mysize::USIZE
+///   - <TryFrom<&[u8]> for $mytype>::Error implements core::fmt::Display
+///   - <TryFrom<&'a[u8]> for $mytype>::Error is the same for all values of 'a,
+///     OR they are all convertible to the value when 'a = 'static, via core::convert::From.
+#[macro_export]
+macro_rules! derive_repr_bytes_from_as_ref_and_try_from {
+    ($mytype:ty, $mysize:ty) => {
+        impl $crate::ReprBytes for $mytype {
+            type Size = $mysize;
+            type Error = <$mytype as TryFrom<&'static [u8]>>::Error;
```
jcape, line 108: You should make `TryFrom` a re-export from this crate, and then reference it as `$crate::_exports::TryFrom` (or whatever).

util/repr-bytes/src/lib.rs
```diff
@@ -0,0 +1,492 @@
...
+    };
+}
+
+/// Derive prost::Message from a ReprBytes implementation
+/// The corresponding protobuf has exactly one member, of type `bytes`.
+#[macro_export]
+macro_rules! derive_prost_message_from_repr_bytes {
+    ($mytype:ty) => {
+        impl ::prost::Message for $mytype {
+            #[inline]
```
jcape, line 173: These inlines should probably get removed in favor of `lto = true` in the `Cargo.toml` file.

util/repr-bytes/src/lib.rs
jcape: As noted, the `unwrap()` should be an `expect()`.

util/repr-bytes/src/lib.rs
```diff
@@ -0,0 +1,492 @@
...
+    };
+}
+
+/// Derive serde::{Deserialize, Serialize} from a ReprBytes implementation
+/// This is represented within serde as a bytes primitive. During deserialization,
+/// a sequence of individual bytes also works, which helps serde_json.
+#[macro_export]
+macro_rules! derive_serde_from_repr_bytes {
+    ($mytype:ty) => {
+        impl ::serde::ser::Serialize for $mytype {
```
jcape, line 248: Same, should be re-export.

garbageslam:
@jcape i wanted to avoid making prost and serde explicit dependencies of repr bytes, because that creates versioning problems. also they are opt in, the user of repr bytes need not depend on prost and serde if they don't want or don't need, they only get those if they use the macros.

lmk if you think they should become hard dependencies of the lib

garbageslam:
if there is a use-case for having core-only targets and targets which use prost built from the same workspace, then we should not make prost even an optional dependency of this lib

garbageslam:
i guess if we have such a use-case, we can just split this crate into two crates, one with the traits, and one with the macros.

garbageslam:
if we think this trait will eventually go away anyways then i'm not going to worry about that for now.

util/repr-bytes/src/lib.rs
garbageslam: the user cannot choose not to import core, so i just went with `::core::convert::TryFrom`

garbageslam:
@jcape "You promised to "circle back" on the in-place API, conducive to merging with mc_util_encoding::{FromX64, ToX64} multiple times, and never did."

Yeah, sorry, I got over-burdened. I will help to try to merge with mc_util_encoding but I need to have a faster path to stabilize things based on generic kex traits so that this work does not block stabilizing those things

util/repr-bytes/src/lib.rs
```diff
@@ -0,0 +1,515 @@
...
+                    <Self as $crate::ReprBytes>::Size,
+                >>::from_slice(src))
+            }
+        }
+    };
+}
+
+/// Derive prost::Message from a ReprBytes implementation
+/// The corresponding protobuf has exactly one member, of type `bytes`.
+#[macro_export]
```
jcape, line 190: This should get gated on `cfg(feature = "prost")`, which I think would make this the only time we've been able to actually use a cargo feature as it was intended :-)

util/repr-bytes/src/lib.rs
```diff
@@ -0,0 +1,515 @@
...
+                F: FnOnce(&[u8]) -> T,
+            {
+                f(<Self as AsRef<[u8]>>::as_ref(self))
+            }
+        }
+    };
+}
+
+/// Derive Into<Vec<u8>> from a ReprBytes implementation
+#[macro_export]
```
jcape, line 154: This should get feature-gated on `alloc`.

util/repr-bytes/src/lib.rs
```diff
@@ -0,0 +1,515 @@
...
+            }
+        }
+    };
+}
+
+/// Derive serde::{Deserialize, Serialize} from a ReprBytes implementation
+/// This is represented within serde as a bytes primitive. During deserialization,
+/// a sequence of individual bytes also works, which helps serde_json.
+#[macro_export]
+macro_rules! derive_serde_from_repr_bytes {
```
jcape, line 267: Same (feature gating)

util/repr-bytes/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+readme = "README.md"
+
+[dependencies]
+generic-array = "0.13"
+prost = { version = "0.6.1", optional = true, default-features = false, features = ["prost-derive"] }
+serde = { version = "1.0", optional = true, default-features = false, features = ["alloc"] }
+
+[features]
+default = ["alloc", "prost", "serde"]
+alloc = []
```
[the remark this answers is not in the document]
garbageslam: i think that doesn't work -- it will mean that turning on alloc at all forces serde to come on
instead what i tried to do is, put the alloc feature in the `serde = ...` part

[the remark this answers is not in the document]
garbageslam: because cargo doesn't let you do stuff like, make some feature contingent on two other features coming on. and serde here is itself optional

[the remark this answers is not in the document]
jcape: Ugh, yeah, OK.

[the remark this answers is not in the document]
garbageslam: maybe serde and prost should fill in alloc? but i don't know how I can express that, cargo tells that a feature and a library cannot have the same name.

PR #87, opened 2020-04-23
MC-1344 Merkle proofs omit nil hashes

mfaulk:
Allows Merkle proofs in transactions to omit hash values that can be inferred by the consensus enclave.

Many of the hashes in each Merkle proof are likely to be "Nil" hashes, which correspond to "missing" leaves compared to a complete binary tree. These Nils can be inferred server-side. Omitting them may be a significant savings in transaction size.

### DIFF
--- a/ledger/db/src/tx_out_store.rs
+++ b/ledger/db/src/tx_out_store.rs
@@ -294,15 +294,11 @@ impl TxOutStore {
         let mut range_to_hash: HashMap<Range, [u8; 32]> = HashMap::default();
         for &(low, high) in ranges.iter().chain(other_ranges.iter()) {
             let range = Range::new(low, high)?;
-            let hash = if low >= num_tx_outs {
-                // Supply the nil hash if the range contains no data.
-                // Note: Nil hashes could probably be omitted as an optimization if validation
-                // knows that it must supply them for any range where `low >= num_tx_outs`.
-                *NIL_HASH
-            } else {
-                self.get_merkle_hash(&range, db_transaction)?
-            };
-            range_to_hash.insert(range, hash);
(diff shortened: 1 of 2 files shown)


ledger/db/src/tx_out_store.rs
```diff
@@ -294,15 +294,11 @@ impl TxOutStore {
         let mut range_to_hash: HashMap<Range, [u8; 32]> = HashMap::default();
         for &(low, high) in ranges.iter().chain(other_ranges.iter()) {
             let range = Range::new(low, high)?;
-            let hash = if low >= num_tx_outs {
-                // Supply the nil hash if the range contains no data.
-                // Note: Nil hashes could probably be omitted as an optimization if validation
-                // knows that it must supply them for any range where `low >= num_tx_outs`.
-                *NIL_HASH
```
mfaulk, line 301: Omitting this NIL_HASH when constructing a proof of membership is the main change.

ledger/db/src/tx_out_store.rs
```diff
@@ -1127,23 +1123,24 @@ pub mod tx_out_store_tests {
...
+                                                   ___H(4,5)__          H(6,7) = H(Nil) Omitted
                                                    |         |
                                                  H(4,4)   H(5,5)
                                                    |        |
                    0       1       2       3       4       [5]       Nil       Nil
 
             */
 
-            assert_eq!(7, proof.elements.len());
+            assert_eq!(6, proof.elements.len());
```
mfaulk, line 1134: The point of omitting Nil hashes is to reduce the size of the Merkle proof. Here, the proof size is reduced by about 14%.

PR #105, opened 2020-04-27
AccountKey cleanup

mfaulk:
- Updates PublicAddress docs to refer to subaddresses
- Unit tests that private subaddress keys agree with public subaddress keys.
- Removes redundant math in AccountKey.subaddress()

### DIFF
--- a/transaction/core/src/onetime_keys.rs
+++ b/transaction/core/src/onetime_keys.rs
@@ -18,6 +18,17 @@ use rand_core::{CryptoRng, RngCore};
 
 const G: RistrettoPoint = RISTRETTO_BASEPOINT_POINT;
 
+/// Generate a tx pubkey for a subaddress transaction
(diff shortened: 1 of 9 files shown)


transaction/core/src/onetime_keys.rs
```diff
@@ -18,6 +18,17 @@ use rand_core::{CryptoRng, RngCore};
 
 const G: RistrettoPoint = RISTRETTO_BASEPOINT_POINT;
 
+/// Generate a tx pubkey for a subaddress transaction
+pub fn compute_tx_pubkey(
```
mfaulk, line 22: This just moved up in the file. I'm trying to organize the file in the order that the protocol uses the functions (the sender creates keys for the output, the recipient identifies an owned output, the recipient recovers the onetime private key for the output)

PR #109, opened 2020-04-28
MC-1388 Fix build script to output libconsensus-enclave.signed.so as before.

jcape:
This will need to get merged into master, and then to the release branch to fix the deployment.


joekottke:
Once this is merged, I will cherry-pick it into the release branch

PR #114, opened 2020-04-28
MC-1378 Subaddress hash functions use domain separator

mfaulk:
Adds a domain separator tag to the subaddress hashes, and updates the Fee account keys.

This changes all subaddresses, which breaks compatibility with the existing ledger.

### DIFF
--- a/transaction/core/src/account_keys.rs
+++ b/transaction/core/src/account_keys.rs
@@ -12,7 +12,7 @@
 
 #![allow(non_snake_case)]
 
-use crate::view_key::ViewKey;
+use crate::{domain_separators::SUBADDRESS_DOMAIN_TAG, view_key::ViewKey};
 use alloc::string::{String, ToString};
 use blake2::{Blake2b, Digest};
 use core::{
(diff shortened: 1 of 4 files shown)


transaction/core/src/tx.rs
```diff
@@ -449,28 +446,4 @@ mod tests {
         let recovered_tx: Tx = Tx::decode(&buf[..]).unwrap();
         assert_eq!(tx, recovered_tx);
     }
-
-    #[test]
-    fn generate_fee_view_key() {
```
mfaulk, line 454: This moved to constants.rs

transaction/core/src/account_keys.rs
```diff
@@ -284,6 +284,7 @@ impl AccountKey {
         let Hs: Scalar = {
             let n = Scalar::from(index);
             let mut digest = Blake2b::new();
+            digest.input(SUBADDRESS_DOMAIN_TAG);
```
mfaulk, line 287: This is the main change.

PR #117, opened 2020-04-28
MC-511 TransactionBuilder sorts outputs by public key

mfaulk:
TransactionBuilder should be misuse-resistant. Here, TransactionBuilder sorts the transaction outputs so that patterns in client code, like always adding a change output last, do not leak into submitted transactions.

Also updates mobilecoind unit tests which assumed that the change output was the last output in the transaction.

### DIFF
--- a/transaction/std/src/transaction_builder.rs
+++ b/transaction/std/src/transaction_builder.rs
@@ -25,8 +25,7 @@ use std::collections::HashSet;
 #[derive(Debug)]
 pub struct TransactionBuilder {
     input_credentials: Vec<InputCredentials>,
-    outputs: Vec<TxOut>,
-    output_shared_secrets: Vec<RistrettoPublic>,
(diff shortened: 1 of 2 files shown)


transaction/std/src/transaction_builder.rs
```diff
@@ -117,7 +115,26 @@ impl TransactionBuilder {
             })
             .collect();
 
-        let tx_prefix = TxPrefix::new(inputs, self.outputs.clone(), self.fee, self.tombstone_block);
+        // Sort outputs by public key.
+        self.outputs_and_shared_secrets
+            .sort_by(|(a, _), (b, _)| a.public_key.cmp(&b.public_key));
```
mfaulk, line 120: This is the main change.

rjwalters: we could sort by any field to solve the ordering problem... is there a performance reason to prefer `public_key`?

PR #147, opened 2020-05-08
Add watcher node and incorporate signatures into block explorer

sugargoat:
Adds a watcher node, with its own main, as well as an option to mobilecoind.

The watcher node pulls signatures from each of the `--tx-src-urls`, and provides them from the watcher_db to mobilecoind clients, including the block_explorer.

Screen Shot 2020-05-07 at 8 34 49 PM
Screen Shot 2020-05-07 at 6 32 39 PM

### TODO:
- [x] Make watcher DB file creation/opening easier

### Future Work:
* Make signatures in block explorer associated with signers, and show continuity, as per:

```
********************************************************************************
MobileCoin Blockchain Explorer
********************************************************************************
There are 456 blocks and 9876 transactions in the ledger.
Index  Size        Hash                          Enclave Signatures
(#)    (txos/~kB)  (bytes)                       0 1 2 3 4 5 6 7 8 9 A B C D E F
*****  **********  ***************************   *******************************
456    3   / 0.08  a1379d86f532u98709273878...                     | | | | |      
455    3   / 0.08  a1379d86f532u98709273878...                     | | | | |     
454    5   / 0.18  a1379d86f532u98709273878...   | | | | | |     
453    3   / 0.08  a1379d86f532u98709273878...   | | | | | |     
452    3   / 0.08  a1379d86f532u98709273878...   | | | | | |     
451    4   / 0.12  a1379d86f532u98709273878...   | | | | | | | | | 
450    3   / 0.08  a1379d86f532u98709273878...   | | | | | | | | | 
449    3   / 0.08  a1379d86f532u98709273878...   | | | | | | | | | 
448    3   / 0.08  a1379d86f532u98709273878...   | | | | | | | | | 
447    3   / 0.08  a1379d86f532u98709273878...   | | | | | | | | | 
446    3   / 0.08  a1379d86f532u98709273878...   | | | | | | | | | 
```


sugargoat:
@jcape - good question. The downside of watching is that it slows down the sync, because it has to fetch from all of the watched peers. Mobilecoind's default sync behavior is to round-robin from the nodes it is watching.

mobilecoind/src/service.rs
```diff
@@ -812,6 +819,20 @@ impl<T: BlockchainConnection + UserTxConnection + 'static> ServiceApi<T> {
                 .push(mc_consensus_api::external::TxOut::from(&output));
         }
 
+        if let Some(watcher_db) = self.watcher_db.clone() {
```
eranrund, line 822: I don't think you need to clone it here. Does `as_ref()` work?

watcher/src/watcher.rs
```diff
@@ -0,0 +1,231 @@
...
+                let url = src_url.join(&filename)?;
+
+                // Try and get the block.
+                log::debug!(
+                    self.logger,
+                    "Attempting to fetch block {} from {}",
+                    block_index,
+                    url
+                );
+                match self.transactions_fetcher.block_from_url(&url) {
```
eranrund, line 84: What happens if this fails for one of the URLs? Not all nodes would upload at the exact same time.
Additionally, for example, if one node is down for 24 hours, syncing would halt for all other nodes.

watcher/src/watcher_db.rs
```diff
@@ -0,0 +1,252 @@
...
+    pub fn add_block_signatures(
+        &self,
+        block_index: u64,
+        signatures: &[BlockSignature],
+    ) -> Result<(), WatcherDBError> {
+        // NOTE: num_blocks is actually "block height" - some blocks may be missing.
+        // This allows us to skip the origin block, and assumes we are adding signatures
+        // to monotonically increasing block indices.
+        let mut db_txn = self.env.begin_rw_txn()?;
+        let num_blocks = self.num_blocks()?;
```
eranrund, line 94: This is not using transactions correctly. Instead of using `db_txn` created above, it creates another transaction.

watcher/src/watcher_db.rs
```diff
@@ -0,0 +1,252 @@
...
+        }
+        db_txn.commit()?;
+        Ok(())
+    }
+
+    /// Get the signatures for a block.
+    pub fn get_block_signatures(
+        &self,
+        block_index: u64,
+    ) -> Result<Vec<BlockSignature>, WatcherDBError> {
```
eranrund, line 130: Commenting here since this is where the thought occurred to me - it might be nice to store the block URL the signature was obtained from? That could allow us to see how nodes change signatures over time. I'd wrap `BlockSignature` in a struct that contains it + the base URL + the filename.

sugargoat:
@jcape re: where the signatures are being verified, it is in the [ReqwestTransactionsFetcher](https://github.com/mobilecoinofficial/mobilecoin/blob/master/ledger/sync/src/reqwest_transactions_fetcher.rs#L164), when the block is first fetched. Since the ReqwestTransactionsFetcher is also used by ledger_sync, it means the blocks pulled from the archive always have their signatures validated. I think we should likely move this more prominently in the TransactionsFetcher trait, to make it a requirement of any transaction fetching implementation that signatures are verified.

watcher/src/watcher_db.rs
sugargoat: Yep - this is the plan for addressing the next phase of the block explorer, that tracks the change of the signature over time per enclave. I'm planning to do that as part of the next PR.

watcher/src/watcher.rs
sugargoat: Good point - since the watcher's signature store is based entirely from archive blocks in S3, I'll store a map of "last_synced_archive_block" per URL in the watcher DB. 

Re: consensus validator node being down, to walk through what that will look like in the signature data, any new blocks will not have signatures from that node, because that node will get those from catchup, and did not participate in consensus. 

It is possible that the node may go down before publishing signed blocks to S3, so in that case, having the "last_synced_archive_block" will catch those.

Re: the watcher's view of network block height when using WatcherSync, it uses the local ledger_db, which was synced according to ledger_sync, based on the `is_blocking_and_quorum` evaluation of block height, so it's a decent indicator of network block height, without having to duplicate the logic in ledger sync, or make any requests to nodes for block height.

One other case to plan for is if the user restarts the watcher with a new consensus validator, then it will need to include that new node with a `last_synced_archive_block` of 0, and sync its signatures from S3.

watcher/src/watcher_db.rs
sugargoat: Took care of the plumbing for this in the current PR - will use the data in the next PR.

watcher/src/watcher.rs
sugargoat: This is addressed!

sugargoat:
I've updated the display for signatures to match the spec. 

The block explorer now displays the signers associated with each block, and in the block display, it shows which url those signers are associated with.

Here are examples for what this looks like when watching node1 and node2:

Screen Shot 2020-05-10 at 11 12 46 AM
Screen Shot 2020-05-10 at 11 13 06 AM

mobilecoind/clients/python/blockchain_explorer.py
```diff
@@ -8,65 +8,79 @@
...
+                           signers=signers)
+
 
 @app.route('/block/<block_num>')
 def block(block_num):
     block = client.get_block(int(block_num))
     size_of_block = getsizeof(block)
+
+    for signature in block.signatures:
+        signature.src_url = signature.src_url.split('/')[-2]
```
eranrund, line 70: This might come and bite us in the future, for partners who don't have the node name be a part of the URL.

watcher/src/watcher.rs
```diff
@@ -0,0 +1,284 @@
...
+            .iter()
+            .map(|url| (url.as_str().to_string(), false))
+            .collect();
+
+        loop {
+            let mut last_synced: HashMap<String, u64> = self.watcher_db.last_synced_blocks()?;
+            log::debug!(self.logger, "Last synced: {:?}", last_synced);
+            log::debug!(self.logger, "max blocks: {:?}", max_blocks);
+            if let Some(max_blocks) = max_blocks {
+                last_synced.retain(|_url, block_index| *block_index < max_blocks);
```
eranrund, line 127: Shouldn't this be `start + max_blocks`?

watcher/src/bin/main.rs
```diff
@@ -0,0 +1,33 @@
...
+            .expect("Failed creating ReqwestTransactionsFetcher");
+
+    let watcher_db = create_or_open_watcher_db(
+        config.watcher_db,
+        &transactions_fetcher.source_urls,
+        logger.clone(),
+    )
+    .expect("Could not create or open watcher db");
+    let watcher = Watcher::new(watcher_db, transactions_fetcher, logger);
+    // For now, ignore origin block, as it does not have a signature.
```
eranrund, line 29: Why do we need to ignore it? It should be okay for blocks to not have signatures. Does anything break if it starts from 0?

watcher/src/config.rs
```diff
@@ -0,0 +1,36 @@
...
+    #[structopt(long = "tx-source-url", required = true, min_values = 1)]
+    pub tx_source_urls: Vec<String>,
+
+    /// (Optional) Number of blocks to sync
+    #[structopt(long)]
+    pub max_blocks: Option<u64>,
+
+    /// Poll Interval
+    #[structopt(long, default_value = "5", parse(try_from_str=parse_duration_in_seconds))]
+    pub poll_interval: Duration,
```
eranrund, line 31: I don't see this being used in watcher/src/bin/main.rs

watcher/src/bin/main.rs
```diff
@@ -0,0 +1,33 @@
...
+    watcher
```
eranrund, line 30: Shouldn't this be in a loop?

mobilecoind/clients/python/blockchain_explorer.py
sugargoat: Agreed - for now they are matching this convention, and we can always address this with different heuristics as we need.

watcher/src/bin/main.rs
sugargoat: Nothing breaks if it starts from 0. The mobilecoind usage of the watcher starts from 0, for example. In the watcher main, I started from 1 since the purpose of the watcher main is just to sync signatures to the watcher_db as a test, and in our test environments, the first block is often very large.

watcher/src/bin/main.rs
sugargoat: This main just syncs and stops. The loop management is done by the WatcherSyncThread, which is currently only used in mobilecoind. If we decide to start shipping a watcher product, we can adjust this main to use the WatcherSync, but for now this is a test utility to test the watcher in isolation from mobilecoind.

I'll update the documentation to reflect that this tool syncs signatures until it finds no more blocks and then stops.

watcher/src/watcher.rs
sugargoat: The start + max_blocks calculation is currently done in the sync thread, [here](https://github.com/mobilecoinofficial/mobilecoin/blob/520397707ec388e4c604c953960ff52dce20584c/watcher/src/watcher.rs#L263), so the `max_blocks` means `max_block_height`. I will update the docs and change name to disambiguate.

PR #148, opened 2020-05-08
apply a patch to dialoguer and console until next release

rjwalters:
In order to avoid printing escape control sequence in `dialoguer::input::interact()`, I offered some patches to the libraries we are using. The maintainer accepted my PR but it will be a few weeks before the updated `interact_text()` functions are available in a release.

This adds the patched libraries to the workspace Cargo.toml and changes our test-client to use the new functions. I've also adjusted some prompts to help users understand that entering an empty value can be used to cancel an action.


Cargo.toml
```diff
@@ -96,3 +96,5 @@ rpath = true
 [patch.crates-io]
 prost = { git = "https://github.com/mobilecoinofficial/prost", rev = "4e1905329369ca7a1cac3eda978ee9379167ee95" }
 prost-derive = { git = "https://github.com/mobilecoinofficial/prost", rev = "4e1905329369ca7a1cac3eda978ee9379167ee95" }
+dialoguer = { git = "https://github.com/mitsuhiko/dialoguer" }
+console = { git = "https://github.com/mitsuhiko/console" }
```
jcape, line 100: This should tag the specific rev that you're looking for, in case they introduce other changes before they release 0.6.

PR #152, opened 2020-05-11
Add a pre-push hook, prevents accidentally pushing branches to origin

garbageslam:
When we set up mobilecoin repo, some engineers wanted that we prevent
new branches from being created for pull requests in the main repo
itself, they want everyone to use forks, and this makes some tools
work better.

Initially we tried to make github itself reject such pushes but
it was found that there is no way to do it.

Instead, this pre-push hook can be installed locally to prevent
accidentally pushing to origin, and suggest pushing to fork and
opening a PR from your fork instead.

### DIFF
--- a/hooks/pre-push
+++ b/hooks/pre-push
@@ -2,15 +2,19 @@
 # Use this file to prevent pushes to origin
 # Based on https://gist.githubusercontent.com/stefansundin/d465f1e331fc5c632088/raw/pre-push-2
 #
-# This is not installed by install_git_settings.sh, to turn this on, use a similar command
+# This is not installed by install_git_settings.sh.
+# To turn this on, use a similar command
 # as the one there that creates a symlink in .git/hooks
+#
+# It may require manual configuration depending on how you have configured your remotes
+# https://git-scm.com/docs/githooks#_pre_push
 
 ORIGIN="$1"
 REMOTE_URL="$2"
 
-if [[ "$REMOTE_URL" == *"origin/"* ]]; then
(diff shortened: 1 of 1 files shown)


hooks/pre-push
```diff
@@ -0,0 +1,19 @@
...
+# Use this file to prevent pushes to origin
+# Based on https://gist.githubusercontent.com/stefansundin/d465f1e331fc5c632088/raw/pre-push-2
+#
+# This is not installed by install_git_settings.sh, to turn this on, use a similar command
+# as the one there that creates a symlink in .git/hooks
+
+ORIGIN="$1"
+REMOTE_URL="$2"
+
+if [[ "$REMOTE_URL" == *"origin/"* ]]; then
```
jcape, line 11: I think this needs to specifically mention mobilecoinofficial, my local checkout names `jcape/mobilecoin.git` asd `origin`, and `mobilecoinofficial/mobilecoin.git` as `upstream`

garbageslam: okay I'll play with it

i'm not sure but I think there might be no way to make it work for every person, it may depend on how you created your local checkout / how you configured remotes

https://git-scm.com/docs/githooks#_pre_push

garbageslam:
the version james' suggested works for me:

```
chris:~/mobilecoinofficial/mobilecoin$ git push origin pre-push-hook 

Prevented pushing to mobilecoinofficial/mobilecoin. Please push to your fork instead.
If you really want to do this, use --no-verify to bypass this pre-push hook.

error: failed to push some refs to 'git@github.com:mobilecoinofficial/mobilecoin'
```

PR #155, opened 2020-05-11
984 sort transactions by fee

mfaulk:
Previously, SCP Ballot values were sorted by transaction hash. Now, Ballot values are sorted by the ordering defined by WellformedTxContext, which sorts by fee, and then by transaction hash.

Many of the changes here relate to replacing BTreeSet with HashSet. In the past, a lot of SCP code used sorted sets of TxHash because that matched the ballot ordering. That's no longer the case, so this PR prefers the faster and simpler HashSet data structure.

This was slam tested and performance seems unchanged.

### DIFF
--- a/consensus/scp/src/slot.rs
+++ b/consensus/scp/src/slot.rs
@@ -512,8 +512,9 @@ impl<V: Value, ValidationError: Display> Slot<V, ValidationError> {
         self.update_YZ();
 
         if !self.Z.is_empty() && self.B.is_zero() {
-            let values: Vec<_> = self.Z.iter().cloned().collect();
-            self.B = Ballot::new(1, &values);
+            let values = (self.combine_fn)(self.Z.clone());
+            let values_as_vec: Vec<V> = values.into_iter().collect();
+            self.B = Ballot::new(1, &values_as_vec);
         }
     }
 
@@ -526,11 +527,13 @@ impl<V: Value, ValidationError: Display> Slot<V, ValidationError> {
         // Invariant: X and Y are disjoint.
         assert!(self.X.is_disjoint(&self.Y));
 
-        let mut new_Z = self.additional_values_confirmed_nominated();
-        if !new_Z.is_empty() {
-            new_Z.append(&mut self.Z);
-            self.Z = (self.combine_fn)(new_Z);
-        }
+        self.Z
+            .extend(self.additional_values_confirmed_nominated().into_iter());
+        // let mut new_Z = self.additional_values_confirmed_nominated();
+        // if !new_Z.is_empty() {
+        //     new_Z.append(&mut self.Z);
+        //     self.Z = (self.combine_fn)(new_Z);
+        // }
     }
 
     fn do_ballot_protocol(&mut self) {
@@ -1110,7 +1113,8 @@ impl<V: Value, ValidationError: Display> Slot<V, ValidationError> {
         // then "ballot.value" is taken as the output of the deterministic combining function
         // applied to all confirmed nominated values."
         if !self.Z.is_empty() {
-            return Some(self.Z.iter().cloned().collect());
+            let values: Vec<V> = (self.combine_fn)(self.Z.clone()).into_iter().collect();
+            return Some(values);
         }
 
         // "Otherwise, if no ballot is confirmed prepared and no value is confirmed nominated,


consensus/scp/src/predicates.rs
```diff
@@ -126,33 +126,6 @@ impl<V: Value> Predicate<V> for ValueSetPredicate<V> {
     }
 }
 
-impl<V: Value> ValueSetPredicate<V> {
-    /// Given a list of results, each containg a set of values, find the "biggest" set of values.
-    /// Sets of values are sorted by their length, and if the lenght matches then by their values.
-    pub fn filter_to_max_values(
```
mfaulk, line 132: This was unused.

consensus/enclave/api/src/lib.rs
```diff
@@ -42,12 +42,12 @@ pub struct WellFormedEncryptedTx(pub Vec<u8>);
 /// Tx data we wish to expose to untrusted from well-formed Txs
 #[derive(Clone, Debug, Default, Deserialize, Eq, Hash, Ord, PartialEq, PartialOrd, Serialize)]
 pub struct WellFormedTxContext {
```
mfaulk, line 44: This struct is now ordered by fee, then by tx_hash.

consensus/scp/src/slot.rs
```diff
@@ -526,11 +526,8 @@ impl<V: Value, ValidationError: Display> Slot<V, ValidationError> {
         // Invariant: X and Y are disjoint.
         assert!(self.X.is_disjoint(&self.Y));
 
-        let mut new_Z = self.additional_values_confirmed_nominated();
-        if !new_Z.is_empty() {
-            new_Z.append(&mut self.Z);
-            self.Z = (self.combine_fn)(new_Z);
```
mfaulk, line 532: Previously, Z always held a set of values that had been through the combine function. This had the advantage that Z could be copied into a Ballot at any time, but had the downside that a node might "forget" that a value had been confirmed nominated if that value were excluded by the combine function. I think this situation is benign, but might mean repeatedly checking if a value has been accepted nominated by a quorum.

In this PR, Z contains all values that have been confirmed nominated, and the combine function is only applied when the values in Z are moved into a Ballot. I think this is logically clearer, and might be slightly more efficient.

consensus/scp/src/slot.rs
```diff
@@ -58,16 +58,16 @@ pub struct Slot<V: Value, ValidationError: Display> {
     M: HashMap<NodeID, Msg<V>>,
 
     /// Set of values that have been proposed, but not yet voted for.
-    W: BTreeSet<V>,
+    W: HashSet<V>,
```
mfaulk, line 61: These are no longer ordered sets. The ordering of values in a Ballot is entirely determined by the combine function, which may be a function of data that is not present in V -- specifically, when V is TxHash, but the ordering depends on the contents of WellFormedTxContext.

consensus/scp/src/msg.rs
```diff
@@ -30,6 +30,22 @@ pub struct NominatePayload<V: Value> {
...
 
+impl<V: Value> NominatePayload<V> {
+    /// Create a NominatePayload
+    ///
+    /// # Arguments
+    /// * `x` - Values voted nominated.
+    /// * `y` - Values accepted nominated.
+    ///
+    /// `x` and `y` must be disjoint sets.
+    pub fn new(x: &HashSet<V>, y: &HashSet<V>) -> Self {
```
mfaulk, line 41: Syntactic sugar.

Under the hood, NominatePayload still uses BTreeSets for X and Y so that Nominate messages have a unique encoding. However, the actual ordering of X and Y is meaningless.

consensus/scp/src/slot.rs
```diff
@@ -1394,18 +1386,25 @@ impl<V: Value, ValidationError: Display> Slot<V, ValidationError> {
...
+        accepted_nominated.extend(accepted_from_blocking_set.into_iter());
+        accepted_nominated.extend(accepted_from_quorum.into_iter());
 
         accepted_nominated
     }
 
     /// "Confirmed Nominated" values that are not yet in self.Z.
-    fn additional_values_confirmed_nominated(&self) -> BTreeSet<V> {
+    fn additional_values_confirmed_nominated(&self) -> HashSet<V> {
+        let mut initial_values: BTreeSet<V> = Default::default();
```
mfaulk, line 1398: These predicates still use BTreeSets. I don't see any reason why they need to be BTreeSets (instead of hashSets), but changing them to HashSet breaks some of the crate's integration tests. MC-1444 is about investigating this.

consensus/scp/src/slot.rs
```diff
@@ -1332,9 +1324,9 @@ impl<V: Value, ValidationError: Display> Slot<V, ValidationError> {
     }
 
     /// "Accepted Nominated" values that are not yet in self.Y.
-    fn additional_values_accepted_nominated(&self) -> BTreeSet<V> {
+    fn additional_values_accepted_nominated(&self) -> HashSet<V> {
         // 1) Find values that can be accepted because a blocking set has issued accept nominate.
-        let mut accepted_from_blocking_set: BTreeSet<V> = {
+        let accepted_from_blocking_set: BTreeSet<V> = {
```
mfaulk, line 1329: These predicates still use BTreeSet (instead of HashSet). I don't see why BTreeSet is necessary, but changing it to HashSet breaks some of the crate integration tests. MC-1444 is about investigating this.

eranrund:
Has this been successfully slammed?

consensus/enclave/api/src/lib.rs
eranrund: Having a comment mentioning that the ordering of the fields in this struct have an effect is probably a good idea.

eranrund:
Do the `scp` crate tests pass? Asking explicitly, because https://github.com/mobilecoinofficial/mobilecoin/blob/master/.circleci/config.yml#L40 skips them

consensus/service/src/test_utils/mod.rs
```diff
@@ -25,17 +24,12 @@ impl UntrustedInterfaces for TrivialTxManagerUntrustedInterfaces {
...
-    ) -> BTreeSet<TxHash> {
-        BTreeSet::from_iter(
-            tx_contexts
-                .iter()
-                .take(max_elements)
-                .map(|tx_context| tx_context.tx_hash())
-                .cloned(),
-        )
+    fn combine(&self, tx_contexts: &[&WellFormedTxContext], max_elements: usize) -> Vec<TxHash> {
+        tx_contexts
```
eranrund, line 28: The previous implementation ensured the returned set of values is sorted (by TxHash).
This implementation does not sort values, which I believe is not a correct implementation of `combine`.

consensus/service/src/validators.rs
```diff
@@ -101,47 +100,54 @@ impl<L: Ledger> TxManagerUntrustedInterfaces for DefaultTxManagerUntrustedInterf
...
-
-        for tx_context in tx_contexts.iter() {
-            if allowed_hashes.len() >= max_elements {
-                // Enforce maximum size.
-                break;
-            }
+    fn combine(&self, tx_contexts: &[&WellFormedTxContext], max_elements: usize) -> Vec<TxHash> {
+        // Dedup by TxHash
+        let mut candidates: Vec<&WellFormedTxContext> = {
+            let hash_to_candidate: HashMap<&TxHash, &WellFormedTxContext> = tx_contexts
```
eranrund, line 115: Why is this building a HashMap and then discarding the keys and just grabbing the values?

eranrund:
Updating I ran the `scp` tests locally on both debug and release and they passed. Weirdly, debug noticeably slower compared to `master`, but release was faster:
```
This branch, debug:
real	31m19.990s
user	162m4.250s
sys	0m4.185s

master, debug:
real	25m5.928s
user	128m5.607s
sys	0m12.480s

This branch, release:
real	3m14.477s
user	19m48.508s
sys	0m7.478s

master, release:
real	8m19.475s
user	57m0.452s
sys	1m45.538s
```

Probably need to run it a bunch more times to rule out variances, but I don't think that matters too much. I'm more curious how `slam` performs.
```

**Comment by mfaulk** [comment]
@eranrund Yes, this was slam tested a few days ago. The first pulse of transactions is with the master branch; the second pulse is with this branch.

Screenshot from 2020-06-02 17-16-32

File: consensus/service/src/test_utils/mod.rs (hunk 8 above)
Comment (mfaulk) [reply consensus/service/src/test_utils/mod.rs:28]: Good catch!

File: consensus/service/src/validators.rs (hunk 9 above)
Comment (mfaulk) [reply consensus/service/src/validators.rs:115]: The intent is to remove any duplicates contained in the collection `tx_contexts`. Building a HashMap and then taking the values has roughly the same "dedup-ing" effect as building a HashSet and then taking its values. I think I took the HashMap route instead of the HashSet route because a HashSet would allow two instances of WellFormedTxContext that contain the same TxHash but have other differing fields, while the HashMap keyed by TxHash would not. 

In retrospect, that's not really an important difference. I'll switch it to a HashSet, which should be more readable.

File: consensus/service/src/test_utils/mod.rs [hunk 10]
```diff
@@ -25,7 +26,16 @@ impl UntrustedInterfaces for TrivialTxManagerUntrustedInterfaces {
     }
 
     fn combine(&self, tx_contexts: &[&WellFormedTxContext], max_elements: usize) -> Vec<TxHash> {
-        tx_contexts
+        // Remove duplicates.
+        let uniques: HashSet<_> = tx_contexts.iter().cloned().collect();
+        let mut as_vec: Vec<_> = uniques.into_iter().collect();
+
+        // Sort, descending.
+        as_vec.sort();
```
eranrund, line 34: You could skip the HashSet and use https://doc.rust-lang.org/std/vec/struct.Vec.html#method.dedup

mfaulk: Will do

PR #158, opened 2020-05-12
extend_message should not use serialize

mfaulk:
Removes a use of `serialize`, when the intent is to get canonical bytes for a compressed Ristretto point.

Also removes a redundant error type for Ristretto points that do not decompress.


transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -348,14 +348,16 @@ fn extend_message(
     message: &[u8],
     pseudo_output_commitments: &[CompressedCommitment],
     range_proof_bytes: &[u8],
-) -> Result<Vec<u8>, Error> {
-    let pseudo_output_bytes = serialize(&pseudo_output_commitments)?;
```
mfaulk, line 352: Removing this use of serialize is the main change.

transaction/core/src/ring_signature/error.rs
```diff
@@ -42,10 +41,6 @@ pub enum Error {
     #[fail(display = "There was an opaque error returned by another crate or library")]
     InternalError,
 
-    /// Public keys must be valid Ristretto points.
-    #[fail(display = "KeyError")]
-    KeyError,
```
mfaulk, line 47: This enum contains another error variant for RistrettoPoints that do not compress/decompress (i.e. InvalidCurvePoint).

PR #162, opened 2020-05-13
interact_text -> interact

sugargoat:


sugargoat:
Yeah - I will close this PR. I believe the issue is related to the interaction of tmux with the interactive_text() command. We may see some issues if people are running the testnet-client from within tmux, and I can submit an issue upstream.

PR #177, opened 2020-05-22
MC-1339 mc-sgx-epid #in-review

jcape:
This implements the rust wrapper around those bits of the SGX API implemented by `libsgx_epid`. It accomplishes this by monkey-patching additional functionality onto the `TargetInfo`, `Quote`, `PlatformInfo`, and `EpidGroupId`.

This will (eventually) supersede the analogous EPID functionality in `mc-attest-trusted`.


sgx/core-types/src/error.rs
```diff
@@ -3,7 +3,7 @@
 //! SGX Error types
 
 use core::{convert::TryFrom, result::Result as StdResult};
-use failure::Fail;
+use displaydoc::Display;
```
mfaulk, line 6: Cool

sgx/epid/src/lib.rs
```diff
@@ -0,0 +1,188 @@
+// Copyright (c) 2018-2020 MobileCoin Inc.
+
+//! Traits and support for EPID-based remote attestation.
+
+#![deny(missing_docs)]
```
mfaulk, line 5: Thanks

sgx/epid/src/lib.rs
```diff
@@ -0,0 +1,188 @@
...
+                quote_size,
+            )
+        }
+        .into_result((quote, qe_report))
+    }
+}
+
+bitflags! {
+    /// A set of bitflags describing what kind of reprovisioning to perform, if any.
+    pub struct EpidUpdateConfig: u32 {
```
mfaulk, line 117: I notice that EpidUpdateStatus starts with 1, while this starts with a shift (1 << 1). Just checking if that's intentional.

jcape: Yeah, bit zero is reserved in the config u32 in the Intel API.

PR #192, opened 2020-06-01
[MC-1470] Ring elements must be sorted

mfaulk:
Soundtrack of this PR: [April 26, 1992](https://www.youtube.com/watch?v=qi8KJ0boov8&t=91s)

### Motivation

Sorting ring elements prevents the ordering from leaking any information about the true input. 

### In this PR
* InputCredentials sorts the ring elements (instead of randomly shuffling them)
* validate.rs checks that each ring is sorted.

### DIFF
--- a/transaction/std/src/input_credentials.rs
+++ b/transaction/std/src/input_credentials.rs
@@ -36,7 +36,6 @@ impl InputCredentials {
     /// * `real_index` - Index in `ring` of the output being spent.
     /// * `onetime_private_key` - Private key for the output being spent.
     /// * `view_private_key` - The view private key belonging to the owner of the real output.
-    /// * `rng` - Randomness.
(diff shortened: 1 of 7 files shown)


transaction/std/src/input_credentials.rs
```diff
@@ -39,45 +37,48 @@ impl InputCredentials {
     /// * `onetime_private_key` - Private key for the output being spent.
     /// * `view_private_key` - The view private key belonging to the owner of the real output.
     /// * `rng` - Randomness.
```
rjwalters, line 39:
```suggestion
```

rjwalters:
> Implementation seems reasonable, I can't speak to the privacy/security aspects.

the sorting happens client side, so I don't think side channel concerns are in scope?

PR #201, opened 2020-06-04
Dont panic on ristretto decompression errors

garbageslam:
Soundtrack of this PR: https://www.youtube.com/watch?v=sVReXN2sFck

### Motivation

Before this commit, a b58 payload with invalid ristretto curvepoints
will cause the application to panic. This is wrong, we should not
panic when deserialization fails.

### In this PR

This commit changes the b58 payloads so that they don't implement
From, which assumes the conversion cannot fail. Instead they
implement TryFrom, and makes sites that were using this
use either TryFrom or TryInto.

### DIFF
--- a/util/b58-payloads/src/payloads.rs
+++ b/util/b58-payloads/src/payloads.rs
@@ -3,7 +3,7 @@
 use crate::error::Error;
 use core::{convert::TryFrom, fmt};
 use crc::crc32;
-use mc_crypto_keys::RistrettoPublic;
+use mc_crypto_keys::{KeyError, RistrettoPublic};
 use mc_transaction_core::account_keys::{AccountKey, PublicAddress};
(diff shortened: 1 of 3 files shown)


util/b58-payloads/src/payloads.rs
```diff
@@ -250,16 +250,17 @@ impl RequestPayload {
 }
 
 /// Decodes a RequestPayload to an account_keys::PublicAddress
-impl From<&RequestPayload> for PublicAddress {
-    fn from(src: &RequestPayload) -> Self {
-        let spend_key = RistrettoPublic::try_from(&src.spend_public_key).unwrap();
-        let view_key = RistrettoPublic::try_from(&src.view_public_key).unwrap();
```
garbageslam, line 256: getting rid of these unwraps is the point of this PR

PR #209, opened 2020-06-11
add resize to lru cache

rjwalters:
Soundtrack of this PR: [dropical](https://soundcloud.com/dropical/no-games-feat-deja-solis)

### Motivation

We use the LRU cache in scp to store received messages on nodes. In crate tests, we don't have the catch-up process to fall back on if a node falls far behind its peers. It is convenient to increase the size of the lru cache during difficult crate tests so that a node that does fall behind can still collect the old messages it needs from peers and successfully externalize the test data. 

[This fn was available in the library we recently replaced.](https://github.com/mobilecoinofficial/mobilecoin/blob/708b0e944fc802629283a985ca7c0eb9247d45a6/common/src/lru.rs#L581)

### In this PR

Adds a resize fn to our lru cache.
Adds tests from [previous library](https://raw.githubusercontent.com/jeromefroe/lru-rs/master/src/lib.rs)

### DIFF
--- a/common/src/lru.rs
+++ b/common/src/lru.rs
@@ -48,6 +48,40 @@ impl<K: PartialEq + Eq + Hash, V> LruCache<K, V> {
         self.used_indexes.len()
     }
 
+    /// Resizes the cache.
+    pub fn resize(&mut self, new_capacity: usize) {
+        // return early if capacity won't change
+        if new_capacity == self.used_indexes.capacity() {
+            return;
+        }
+        // re-initialize storage
+        self.free_indexes = VecDeque::with_capacity(new_capacity);
+        self.key_to_entry_index.clear();
+        let mut new_entries: Vec<Option<(Arc<K>, V)>> = Vec::with_capacity(new_capacity);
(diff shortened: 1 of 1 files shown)


eranrund:
Can’t we just initialize the LRU in the tests to the maximal value needed?
Memory usage for tests is not a concern and it’s a much simpler solution.
I’d prefer to not have to maintain this code just for the sake of a part of
Scp tests.

rjwalters:
> Can’t we just initialize the LRU in the tests to the maximal value needed? Memory usage for tests is not a concern and it’s a much simpler solution. I’d prefer to not have to maintain this code just for the sake of a part of Scp tests.

I'm not sure; the lru cache size is set using a default constant in scp/node. There are a bunch of functions in the original LRU cache library that we are dropping. I assumed we would add in those functions over time as we needed them... but maybe it's better to find a different way to run crates tests with 10x larger lru caches?

rjwalters:
@eranrund would you prefer to close this?

eranrund:
Yes

PR #219, opened 2020-06-16
Encoding for address requests, right now just a URL to POST to

tsegaran:
Soundtrack of this PR: [Mr Postman](https://www.youtube.com/watch?v=425GpjTSlS4)

### Motivation

For some potential workflows we'd like a way to specify a URL for a user to submit their public address 

### In this PR
* Defines a payload for a b58 encoding of an "address request" which is just URL
* Tests round-trip encoding and decoding

### Future Work
* Integrate into APIs

### DIFF
--- a/util/b58-payloads/src/payloads.rs
+++ b/util/b58-payloads/src/payloads.rs
@@ -14,6 +14,7 @@ enum PayloadType {
     Transfer = 1,
     // Wallet = 2,
     // Envelope = 3,
+    AddressRequest = 5,
 }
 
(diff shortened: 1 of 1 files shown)


util/b58-payloads/src/payloads.rs
```diff
@@ -389,6 +390,59 @@ impl From<&TransferPayload> for AccountKey {
     }
 }
 
+/// AddressRequestPayload encodes a URL to which a user to send a public address
```
rjwalters, line 393:
```suggestion
/// AddressRequestPayload encodes a URL to which a client can POST a public address
```

PR #220, opened 2020-06-17
Add JSON Admin Status to Info Endpoint

sugargoat:
Soundtrack of this PR: [Paper Planes](https://www.youtube.com/watch?v=ewRjZoRtu0Y)

### Motivation

We need to expose some more information about the consensus validator's view of its status, as well as the status of its peers, to make monitoring easier.

### In this PR
* Adds the following info to the consensus validator "info" admin endpoint:
    * version
    * broadcast_peer_count
    * known_peer_count
    * block_height
    * sync_status
    * blocks_behind
    * latest_block_hash
    * latest_block_timestamp

#done [MC-1555](https://mobilecoin.atlassian.net/browse/MC-1555)

### DIFF
--- a/consensus/service/src/consensus_service.rs
+++ b/consensus/service/src/consensus_service.rs
@@ -711,8 +711,8 @@ impl<E: ConsensusEnclaveProxy, R: RaClient + Send + Sync + 'static> ConsensusSer
         let ledger_db = self.ledger_db.clone();
         let byzantine_ledger = self.byzantine_ledger.clone();
         let config = self.config.clone();
+        let logger = self.logger.clone();
         Arc::new(move || {
-            let block_height = ledger_db.num_blocks().unwrap_or(0);
             let mut sync_status = "synced";
             let mut peer_block_height: u64 = 0;
             let byzantine_ledger = byzantine_ledger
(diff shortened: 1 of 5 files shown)


consensus/service/src/consensus_service.rs
```diff
@@ -704,24 +708,45 @@ impl<E: ConsensusEnclaveProxy, R: RaClient + Send + Sync + 'static> ConsensusSer
...
+                peer_block_height = byzantine_ledger.highest_peer_block();
+            }
             Ok(json!({
                 "config": {
                     "public_key": config.node_id().public_key,
                     "peer_responder_id": config.peer_responder_id,
                     "client_responder_id": config.client_responder_id,
                     "message_pubkey": config.msg_signer_key.public_key(),
                     "network": config.network_path,
-                    "ias_api_key": config.ias_api_key,
```
sugargoat, line 716: These shouldn't actually be on the info endpoint.

consensus/service/src/consensus_service.rs
```diff
@@ -704,24 +708,45 @@ impl<E: ConsensusEnclaveProxy, R: RaClient + Send + Sync + 'static> ConsensusSer
...
                 },
                 "network": config.network(),
+                "status": {
+                    "block_height": block_height,
+                    "version": VERSION,
+                    "broadcast_peer_count": config.network().broadcast_peers.len(),
+                    "known_peer_count": config.network().known_peers.map(|x| x.len()),
+                    "sync_status": sync_status,
+                    "blocks_behind": std::cmp::min(peer_block_height - block_height, 0),
+                    "latest_block_hash": ledger_db.get_block(block_height - 1).map(|x| x.id).ok(),
```
eranrund, line 748: How does serializing of an Option looks like in JSON?

consensus/service/src/consensus_service.rs
```diff
@@ -704,24 +708,45 @@ impl<E: ConsensusEnclaveProxy, R: RaClient + Send + Sync + 'static> ConsensusSer
 
     /// Helper method for creating the get config json function needed by the GRPC admin service.
     fn create_get_config_json_fn(&self) -> GetConfigJsonFn {
+        let ledger_db = self.ledger_db.clone();
+        let byzantine_ledger = self.byzantine_ledger.clone();
         let config = self.config.clone();
         Arc::new(move || {
+            let block_height = ledger_db.num_blocks().unwrap_or(0);
```
eranrund, line 715: I think this should panic, or somehow be changed to report the error back. Otherwise, if this failed, executing `ledger_db.get_block(block_height - 1)` below is weird behavior.

consensus/service/src/consensus_service.rs
sugargoat: If it's None, it's a json "null" - my intent here was to provide the hash unless it errored, in which cause display "null."

consensus/service/src/consensus_service.rs
sugargoat: Reporting the error back makes sense, and can then display "null" for the block hash.

sugargoat:
Tested on internal deployed network:
Screen Shot 2020-06-25 at 1 37 59 PM

consensus/service/src/consensus_service.rs
```diff
@@ -704,24 +708,81 @@ impl<E: ConsensusEnclaveProxy, R: RaClient + Send + Sync + 'static> ConsensusSer
...
+            }
+            let block_height;
+            let latest_block_hash;
+            let latest_block_timestamp;
+            let blocks_behind;
+            // If we do not get a num_blocks, several status points will be null
+            match ledger_db.num_blocks() {
+                Ok(b) => {
+                    block_height = Some(b);
+                    latest_block_hash = ledger_db
```
eranrund, line 735: I wouldn't call this last block hash, it should be last block id

eranrund:
> Tested on internal deployed network:
> Screen Shot 2020-06-25 at 1 37 59 PM

It's weird that `known_peer_count` is `null` and not `0`.

sugargoat:
@eranrund - I thought so too - I just added a `map_or(0, ...)` so that it will be 0 of there are no known_peers.

consensus/service/src/consensus_service.rs
sugargoat: Discussed offline - we're going to go with latest_block_hash to match the request for the endpoint and reduce cognitive load, as the block_id is our name for the block hash, as it functions as the identifier for the block.

PR #226, opened 2020-06-19
Enable mc-common/std for mc-common/loggers

kylefleming:
Soundtrack of this PR: [Sahalé - Nlreb Mra Alrrih (Playing With The Wind)](https://soundcloud.com/sahaleprods/sahale-nlreb-mra-alrrih-playing-with-the-wind)

### Motivation

`mc-util-serial/std` must be enabled whenever `serde/std` is enabled, even if `serde/std` is enabled by a third-party dependency. Currently `mc-common/loggers` enables multiple dependencies (`sentry`, `slog-gelf`, and `slog-json`) that all enable `serde/std`. This causes a compile error for any crate that depends on `mc-common/loggers` without something in the dependency tree also enabling `mc-util-serial/std`. Since `mc-common/loggers` is the crate that introduces these dependencies, it should also enable `mc-util-serial/std` for correctness.

### In this PR
* Enables `mc-util-serial/std` for `mc-common/loggers`
* Removes redundant `serde/std` from `mc-common/std`

### DIFF
--- a/common/Cargo.toml
+++ b/common/Cargo.toml
@@ -9,7 +9,6 @@ default = []
 std = [
     "failure/std",
     "mc-util-serial/std",
-    "serde/std",
 ]
 log = [
     "backtrace",
@@ -19,7 +18,11 @@ log = [
     "slog-scope",
 ]
 loggers = [
+    # Features
     "log",
+    "std",
+
+    # Dependencies
     "hostname",
     "lazy_static",
     "mc-util-build-info",


common/Cargo.toml
```diff
@@ -9,7 +9,6 @@ default = []
 std = [
     "failure/std",
     "mc-util-serial/std",
-    "serde/std",
```
garbageslam, line 12: are you sure this is redundant?

kylefleming: `mc-util-serial/std` enables it.

common/Cargo.toml
```diff
@@ -23,6 +22,8 @@ loggers = [
     "hostname",
     "lazy_static",
     "mc-util-build-info",
+    # `mc-util-serial/std` must be enabled since `sentry`, `slog-gelf`, and `slog-json` enable `serde/std`
+    "mc-util-serial/std",
```
garbageslam, line 26: i would suggest just saying `loggers => std` in this crate

kylefleming: Because `failure/std` should also be enabled? Or for a different reason?

garbageslam: i just spent a lot of time trying to make these features work in the past, debugging very annoying enclave-not-building problems, and i have even less ability to debug it when these core crates live in a different repo. i have a few rules of thumb, usually, the simplest implication that is correct is what you should use. If any build that uses loggers must have std available (which is true), the loggers should imply the std feature.

garbageslam: the more like, piecemeal additions like this people just add arbitrarily to the cargo tomls, the less maintainable all of this becomes.

common/Cargo.toml
garbageslam: i think at this revision, when loggers => std, you still need `serde/std` to be implied by the std feature

garbageslam: oh i see

yeah do whatever you like here.

PR #229, opened 2020-06-20
[MC-984] Sort transactions by fee, again.

mfaulk:
I must have made a mistake in merging or rebasing the first PR, because some of the intended commits didn't get merged. 

This has been slam tested, and passes CI in `internal`.

### Motivation

The combine function sorts transactions by fee, then lexicographically over the other fields of WellformedTxContext. In several places, data structures containing TxHashes are changed from BTreeSet to HashSet (when uniqueness is required, but ordering is not) or Vec (when ordering is required), which reflects the fact that transactions in a block do not need to follow the same ordering as their hashes.

### Future Work
* Emit statistics about the transaction fees in a block.


mfaulk:
Slam tested, looks good.
Screenshot from 2020-06-23 16-28-28

consensus/scp/src/predicates.rs
```diff
@@ -153,60 +153,6 @@ impl<V: Value> ValueSetPredicate<V> {
     }
 }
 
-/// A predicate for narrowing down (min, max) ranges. Works in conjunction with
-/// `Msg.accepts_commits()` and `Msg.votes_or_accepts_commits()`.
-#[derive(Clone)]
-pub struct MinMaxPredicate<V: Value> {
```
mfaulk, line 159: This was unused.

consensus/service/src/lib.rs
```diff
@@ -21,6 +21,3 @@ mod counters;
 mod grpc_error;
 mod peer_api_service;
 mod peer_keepalive;
-
-#[cfg(test)]
-mod test_utils;
```
mfaulk, line 26: This was unused.

consensus/service/src/validators.rs
```diff
@@ -1193,4 +1175,42 @@ mod combine_tests {
         assert_eq!(combined_transactions.len(), 2);
         assert!(combined_transactions.contains(third_client_tx.tx_hash()));
     }
+
+    #[test]
+    // `combine` should return hashes in the order defined by WellformedTxContext.
```
mfaulk, line 1180: This functionality is the main point of the PR.

consensus/enclave/api/src/lib.rs
```diff
@@ -13,6 +13,7 @@ pub use crate::{error::Error, messages::EnclaveCall};
 
 use alloc::vec::Vec;
 use core::{hash::Hash, result::Result as StdResult};
+use failure::_core::cmp::Ordering;
```
eranrund, line 16: `core::cmp::Ordering`

consensus/enclave/api/src/lib.rs
```diff
@@ -39,26 +40,26 @@ pub struct LocallyEncryptedTx(pub Vec<u8>);
...
     /// Fee included in the tx.
-    fee: u64,
+    pub fee: u64,
+
+    /// Tx hash.
+    pub tx_hash: TxHash,
 
     /// Tombstone block.
-    tombstone_block: u64,
+    pub tombstone_block: u64,
```
eranrund, line 53: Why were all of these made `pub`? They were originally not `pub` intentionally to prevent them getting changed by mistake somewhere. There are accessors for all of them below.

consensus/enclave/api/src/lib.rs
```diff
@@ -100,6 +101,83 @@ impl From<&Tx> for WellFormedTxContext {
     }
 }
 
+/// Defines a sort order for transactions in a block.
+/// Transactions are sorted by fee (high to low), then by transaction hash and any other fields.
+impl Ord for WellFormedTxContext {
+    fn cmp(&self, other: &Self) -> Ordering {
+        if self.fee != other.fee {
+            // Sort by fee, descending.
+            other.fee.cmp(&self.fee)
```
eranrund, line 110: Isn't this already the behavior when `fee` is the first field in the struct?

consensus/service/src/validators.rs
```diff
@@ -125,61 +122,46 @@ impl<L: Ledger> TxManagerUntrustedInterfaces for DefaultTxManagerUntrustedInterf
...
-        max_elements: usize,
-    ) -> BTreeSet<TxHash> {
-        // Allow transactions that do not introduce key image double-spends.
-        let mut allowed_hashes = BTreeSet::new();
-        let mut used_key_images = HashSet::default();
-        let mut seen_output_public_keys = HashSet::default();
-
-        for tx_context in tx_contexts.iter() {
+    fn combine(&self, tx_contexts: &[&WellFormedTxContext], max_elements: usize) -> Vec<TxHash> {
+        // WellformedTxContext defines the sort oder of transactions within a block.
```
eranrund, line 134: Typo `oder`

consensus/scp/play/src/main.rs
```diff
@@ -63,8 +57,12 @@ fn trivial_validity_fn(_value: &TxHash) -> Result<(), TransactionValidationError
     Ok(())
 }
 
-fn trivial_combine_fn(values: BTreeSet<TxHash>) -> BTreeSet<TxHash> {
-    BTreeSet::from_iter(values.into_iter().take(MAX_TRANSACTIONS_PER_BLOCK))
+fn trivial_combine_fn(values: &[TxHash]) -> Vec<TxHash> {
```
rjwalters, line 60: Does this need to be different from the `trivial_combine_fn` in [src/test_utils#L26](https://github.com/mobilecoinofficial/mobilecoin/blob/88f850c80ac70a7e1f9e8b49cd511ddb099eb085/consensus/scp/src/test_utils.rs#L26)?

consensus/enclave/api/src/lib.rs
mfaulk: They were made `pub` so that instances of WellformedTxContext can be directly initialized, which simplifies testing. Previously, we did things like WellFormedTxContext::from(tx), which involves the more complicated task of creating a Tx.

I will replace the `pub`s with a constructor, which will also work to keep the unit tests clean.

consensus/enclave/api/src/lib.rs
mfaulk: Not quite. Deriving `Ord` with `fee` as the first field means that WellFormedTxContext is sorted by fee in ascending order, which is the opposite of what we want.

mfaulk:
@eranrund I don't think it makes sense to add a test to form_block. That function is called after consensus is complete and accepts whatever list of transactions was agreed upon. In fact, form_block ought to be indifferent to transaction ordering because the resulting block contents (key images and outputs) are themselves sorted.

eranrund:
> @eranrund I don't think it makes sense to add a test to form_block. That function is called after consensus is complete and accepts whatever list of transactions was agreed upon. In fact, form_block ought to be indifferent to transaction ordering because the resulting block contents (key images and outputs) are themselves sorted.

Good points.

PR #244, opened 2020-06-28
Remove external.RistrettoPublic, replace with CompressedRistretto

garbageslam:
Soundtrack of this PR: https://www.youtube.com/watch?v=q_7l0ozbHVA

### Motivation

Generally, we should not use RistrettoPublic in serialized types,
we should try to use CompressedRistretto, because decompressing
a ristretto point requires elliptic curve operations and is pretty
slow. In early performance profiling, consensus spent most of its
time doing this, and that is why we created the CompressedRistretto
type.

The protobuf type external.RistrettoPublic is actually identical
to external.CompressedRistretto, we don't need both, and we should
discourage people from using RistrettoPublic in their serialized
structs.

### In this PR

Remove `external.RistrettoPublic`
Fixup mobilecoind_api.
Turns out we have enough conversions on external.CompressedRistretto
now that no code changes are required.

Fog-125

### DIFF
--- a/api/proto/external.proto
+++ b/api/proto/external.proto
@@ -13,13 +13,6 @@ option java_outer_classname = "MobileCoinAPI";
 // `keys` crate
 ///////////////////////////////////////////////////////////////////////////////
 
-/// Note: This is used only in mobilecoind. It is not recommended to use
-/// uncompressed RistrettoPublic in protobuf types because the decompressions
-/// may be unnecessary and can cause significant performance hit.
-message RistrettoPublic {
-    bytes data = 1;
-}
-
 /// A Ristretto private key.
 message RistrettoPrivate {
     bytes data = 1;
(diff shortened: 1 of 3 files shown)


api/proto/external.proto
```diff
@@ -31,9 +24,12 @@ message CompressedRistretto {
...
 /// A 32-byte scalar associated to the ristretto group.
-/// This is the same as RistrettoPrivate, but they are used in different places,
-/// the transaction crate generally uses CurveScalar and higher-level crates use
-/// RistrettoPrivate.
+/// This is the same as RistrettoPrivate, but they are used in different places.
+/// Many crates only need keys to perform key exchange operations, and strictly
+/// use the RistrettoPublic and RistrettoPrivate types.
+/// The transaction crate often uses CurveScalar and sometimes must perform
+/// raw elliptic curve operations etc.
+/// This type can possibly be factored out.
```
kylefleming, line 32: I think we should leave this out since this file is mainly meant to be used as an interface file. Suggestions about improvements should probably take the form of a comment on the PR or a dedicated Jira ticket.

It's also not clear to me whether this is a confirmed statement or just a thought that needs more investigation.

garbageslam: created MC-1605 to track this and removed commentary from the proto comment

PR #246, opened 2020-06-29
Initial fog authority key signature with schnorrkel

sugargoat:
Soundtrack of this PR: [Sandstorm](https://www.youtube.com/watch?v=y6120QOlsfU)

### Motivation

Currently we have placed the plumbing for the fog authority signature, but not the fog authority signature itself.

### In this PR
* Uses [schnorrkel](https://crates.parity.io/schnorrkel/index.html) to construct a signature of the subaddress_view_private_key over the fog authority fingerprint for the public address
* Adds test for round-trip sign and verify

First part of [FOG-106](https://mobilecoin.atlassian.net/browse/FOG-106)

### Future Work
* Incorporate changes into internal client
* This PR required some [changes to upstream schnorrkel](https://github.com/w3f/schnorrkel/compare/master...sugargoat:no-std?expand=1) in order to build in no_std and produce deterministic signatures. These changes will need to go through PR review and be merged upstream.


transaction/core/src/account_keys.rs
```diff
@@ -362,8 +363,24 @@ impl AccountKey {
 
         // Compute fog_authority_sig as a signature over self.fog_authority_key_fingerprint
         if !self.fog_report_url.is_empty() {
-            // FIXME: FOG-106 fog_authority_sig should be a Schnorrkel sig using subaddress_view_private
-            result.fog_authority_sig.extend(&[9u8, 9u8, 9u8, 9u8]);
+            // Construct a MiniSecretKey from the bytes directly, then expand to SecretKey
+            let scalar_bytes = self.subaddress_view_private(index).to_bytes();
+            let mut bytes_slice = [0u8; 32];
+            bytes_slice[..32].copy_from_slice(&scalar_bytes);
+            let mini_secret = MiniSecretKey::from_bytes(&bytes_slice)
```
mfaulk, line 370: I think MiniSecretKey is a "precursor" that is used to generate a new scalar (i.e. our RistrettoPrivate). See:

https://github.com/w3f/schnorrkel/blob/master/src/keys.rs#L51-L56

and 

See: https://tools.ietf.org/html/rfc8032#section-5.1.5

Here, however, the scalar -- the `subaddress_view_private` key -- already exists. I think you can use that directly to create a SecretKey via `SecretKey::from_bytes(bytes)`, where bytes is a [u8; 64] containing `subaddress_view_private.to_bytes()` and a random 32-byte nonce.

jcape: I haven't read through the code/comments, but I'm pretty confident that we cannot use our `PrivateKey` as the `SecretKey` directly, we should use it as a `MiniSecretKey`

sugargoat: @jcape - we need to follow up on this - created [MC-1611](https://mobilecoin.atlassian.net/browse/MC-1611) to capture changes in our keys crate that need to be made to support using MiniSecretKey

crypto/keys/src/ristretto.rs
```diff
@@ -233,6 +237,10 @@ impl RistrettoPublic {
     pub fn to_bytes(&self) -> [u8; 32] {
         self.0.compress().to_bytes()
     }
+
+    pub fn point(&self) -> RistrettoPoint {
+        self.0
+    }
```
jcape, line 243: We already have an `AsRef<RistrettoPoint>`, please don't add redundant/non-generic API to this, or use it in production code.

transaction/core/src/account_keys.rs
```diff
@@ -360,10 +363,27 @@ impl AccountKey {
...
+            let seed: [u8; 32] = Sha256::digest(&scalar.to_bytes()).into();
+            let mut cspring: FixedRng = SeedableRng::from_seed(seed);
+            let mut nonce = [0u8; 32];
+            cspring.fill_bytes(&mut nonce);
+
+            let mut secret_bytes = [0u8; 64];
+            secret_bytes[0..32].copy_from_slice(&scalar.to_bytes());
+            secret_bytes[32..64].copy_from_slice(&nonce);
+
+            let secret_key = SecretKey::from_bytes(&secret_bytes).unwrap();
```
jcape, line 381: Why are we not using the mini secret key, and letting them do the nonce generation?

sugargoat: The MiniSecretKey expansion creates a new KeyPair that does not verify with the RistrettoPublic associated with the scalar. That's what [MC-1611](https://mobilecoin.atlassian.net/browse/MC-1611) should address, so that our keys are doing the same thing that Schnorrkel keys are doing.

crypto/keys/src/ristretto.rs
sugargoat: Fixed!

crypto/keys/src/ristretto.rs
```diff
@@ -75,6 +75,10 @@ impl RistrettoPrivate {
     pub fn to_bytes(&self) -> [u8; 32] {
         *self.0.as_bytes()
     }
+
+    pub fn scalar(&self) -> Scalar {
```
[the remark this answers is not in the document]
sugargoat: Fixed!

transaction/core/src/account_keys.rs
```diff
@@ -360,10 +363,28 @@ impl AccountKey {
...
+        // NOTE: The fog_authority_sig is deterministic due to using the private key hash as the rng seed
         if !self.fog_report_url.is_empty() {
-            // FIXME: FOG-106 fog_authority_sig should be a Schnorrkel sig using subaddress_view_private
-            result.fog_authority_sig.extend(&[9u8, 9u8, 9u8, 9u8]);
+            // Construct the fog authority signature over the fingerprint using the view privkey
+            let view_private = self.subaddress_view_private(index);
+            let scalar: &Scalar = view_private.as_ref();
+
+            // Create rng seed from hash of private key in order to create nonce and sign
+            let seed: [u8; 32] = Sha256::digest(&scalar.to_bytes()).into();
```
garbageslam, line 373: at this place, I would want to append `self.fog_authority_key_fingerprint` to the scalar bytes, so that the digest depends on both, because that ensures that this seed is not reused for two different messages that we are signing, even with the same private key.

and it would be my preference to use sha3 instead of sha256, because of djb's comments in this paper: http://ed25519.cr.yp.to/ed25519-20110926.pdf

> Pseudorandom generation of `r`
>
> ...
> Standard  PRF  hypotheses  imply  that  this  pseudorandom  session  key `r` is indistinguishable from a truly random string generated independently for each M, so there is no loss of security. Well-known length-extension properties prevent secret-prefix SHA-512 from being a PRF, but also do not threaten the security of Ed25519-SHA-512, since `r` is not visible to the attacker. All remaining SHA-3  candidates  are  explicitly  designed  to  be  PRFs,  and  we  will  not  hesitate  to recommend Ed25519-SHA-3 after SHA-3 is standardized. It would of course alsobe safe to generaterwith a cipher such as AES, combined with standard PRF-stretching mechanisms to support a long input; but we prefer to reuse H to save area in hardware implementations.

sugargoat: Re: appending the message - that is done - will look at switching to sha3 now!

sugargoat: sha3 done!

transaction/core/src/account_keys.rs
```diff
@@ -360,10 +363,32 @@ impl AccountKey {
...
+            let view_private = self.subaddress_view_private(index);
+            let scalar: &Scalar = view_private.as_ref();
+
+            // Nonce is hash( private_key || message )
+            let nonce_digest: Vec<u8> = [
+                &scalar.to_bytes(),
+                self.fog_authority_key_fingerprint.as_slice(),
+            ]
+            .concat();
+            let nonce: [u8; 32] = Sha3_256::digest(&nonce_digest).into();
```
mfaulk, line 378: I have a slight preference for using Blake2B here instead of adding a new dependency on sha3 ("mo dependencies mo problems", as Biggie would say), but this is fine.

transaction/core/src/account_keys.rs
```diff
@@ -360,10 +363,32 @@ impl AccountKey {
...
+
+            let mut secret_bytes = [0u8; 64];
+            secret_bytes[0..32].copy_from_slice(&scalar.to_bytes());
+            secret_bytes[32..64].copy_from_slice(&nonce);
+            let secret_key = SecretKey::from_bytes(&secret_bytes).unwrap();
+            let keypair = secret_key.to_keypair();
+
+            // Context provides domain separation for signature
+            let ctx = signing_context(b"Fog authority signature");
```
mfaulk, line 387: This string should probably be a public constant.

garbageslam:
this looks good to me

if you want to make a security statement about the sha3 part, what i would write is something like:

Schnorr signatures are proved to be secure when the nonce is truly uniformly random, however in practice it can happen that there is a defect in a random number generator that causes them not to be random, causing some high profile breaks. This observation informed RFC 6979, ed25519 and much subsequent work.

In this code, the nonce for the schnorr signature is derived as `SHA3(ristretto_private, message)`. This has the advantage of being deterministic -- for a given signing key and message there is only one nonce that will ever be used. And the nonce the we choose for a given message, is the result of secret-prefix SHA3. Under the cryptographic assumption that secret-prefix SHA3 is indistinguishable from a PRF, this is as good as choosing perfectly uniformly random nonces.

transaction/core/src/account_keys.rs
```diff
@@ -360,10 +363,32 @@ impl AccountKey {
...
+            let view_private = self.subaddress_view_private(index);
+            let scalar: &Scalar = view_private.as_ref();
+
+            // Nonce is hash( private_key || message )
+            let nonce_digest: Vec<u8> = [
+                &scalar.to_bytes(),
+                self.fog_authority_key_fingerprint.as_slice(),
+            ]
+            .concat();
+            let nonce: [u8; 32] = Blake2b256::digest(&nonce_digest).into();
```
jcape, line 378: Just create a new blake2b instance and feed the two slices into it separately.

transaction/core/src/account_keys.rs
sugargoat: In off-line discussions, we agreed that using the SecretKey with the Hash(ristretto_private | message) as the nonce is likely sufficient, and we plan to do a careful review and analysis of this scheme wrt the Scalar also being used and derived through CryptoNote math.

transaction/core/src/account_keys.rs
sugargoat: Similar discussion here: https://github.com/mobilecoinofficial/mobilecoin/pull/246#discussion_r448661514

garbageslam:
if we go with blake2 we could also say "blake was one of the sha3 finalists and was also explicitly designed to model a PRF"

transaction/core/src/account_keys.rs
sugargoat: Done

garbageslam:
sorry i had so many examples, that is going to be a PITA. i think its okay to disable the comparison tests for now if you want

sugargoat:
@cbeck - no worries! Happy to put the examples through!

sugargoat:
@garbageslam @mfaulk @jcape - thanks for the reviews!

Re: the cryptographer review, the current plan is to move forward with this implementation of the Schnorrkel signature in its current state in order to meet requirements for internal fog usage. The signature will not be created or consumed in production on TestNet until we have had an audit of the current scheme.

PR #247, opened 2020-06-30
[MC-1601] Node maintains a single current slot

mfaulk:
### Motivation

Currently, our nodes only handle messages for their current slot. The desired behavior is that a node always processes messages for its current slot, and also handles messages for some number of previous slots that it has externalized in order to help other nodes in the network reach consensus. In general, we want nodes to finish consensus instead of relying on catch-up, because the node does not validate transactions and sign blocks during catch-up. Also, the catch-up mode is stricter than the normal quorum rules (again, because it can't validate the transactions), and so is less likely to succeed. I think I've also found a few differences between how we handle slots in production vs how we run SCP tests, so there are a few interrelated things that need to get straightened out.

### In this PR
- Node maintains a single `current` slot, and a bounded queue of previous externalized slots.
- ByzantineLedger passes all "recent" messages into Node, which passes them to the appropriate slot. This makes use of the Externalize phase of the SCP protocol, instead of stopping at the first Externalize message.
- Node and Slot batch-process incoming messages.


consensus/scp/src/node.rs
```diff
@@ -63,12 +62,22 @@ impl<V: Value, ValidationError: Display> Node<V, ValidationError> {
         Q: QuorumSet,
         validity_fn: ValidityFn<V, ValidationError>,
         combine_fn: CombineFn<V>,
+        current_slot_index: SlotIndex,
```
rjwalters, line 65: should this always be zero?

consensus/scp/src/node.rs
```diff
@@ -32,8 +29,8 @@ pub struct Node<V: Value, ValidationError: Display> {
     /// Local node quorum set.
     pub Q: QuorumSet,
 
-    /// Map of last few slot indexes -> slots.
-    pub pending: LruCache<SlotIndex, Slot<V, ValidationError>>,
```
mfaulk, line 36: Replacing this with a single Slot instance is one of the main changes of this PR. It doesn't make sense to run more than one non-externalized slot at a time. Using a cache also has some negative side effects: if a node handled a bunch of (malicious) messages for different slots, the cache could "flush" the slot it's supposed to be working on, which would break the protocol by resetting the node's state for that slot.

consensus/scp/src/node.rs
mfaulk: My intent is that Node gets instantiated to work on whatever the next slot is. For example, if the consensus service restarts it should start running consensus at the current slot index for is local ledger, instead of starting over at slot zero.

consensus/scp/src/node.rs
```diff
@@ -138,12 +133,8 @@ pub trait ScpNode<V: Value>: Send {
     /// Get local node quorum set.
     fn quorum_set(&self) -> QuorumSet;
 
-    /// Submit a list of values of nomination.
-    fn nominate(
-        &mut self,
-        slot_index: SlotIndex,
```
mfaulk, line 144: The `slot_index` argument was removed, because it only makes sense to nominate values for the current slot.

consensus/service/src/byzantine_ledger.rs
```diff
@@ -519,7 +521,7 @@ impl<
                 log::info!(self.logger, "sync_service reports we are no longer behind!");
 
                 // Reset scp state.
-                self.scp.clear_pending_slots();
+                self.scp.reset_slot_index(self.cur_slot);
```
eranrund, line 524: Have you tested catchup? I believe this is wrong since `self.cur_slot` points to the old current slot, before catchup has completed.

mfaulk: Good catch. I changed this to reset the scp state after cur_slot has been updated, and so far catch-up seems to work.

eranrund:
Any updates on this?

consensus/scp/src/node.rs
```diff
@@ -223,26 +207,30 @@ impl<V: Value, ValidationError: Display> ScpNode<V> for Node<V, ValidationError>
...
-                    return Ok(None);
-                }
-            }
+        // Ignore messages for future slots.
+        if msg.slot_index > self.current_slot.get_index() {
+            // TODO: return an error?
+            return Ok(None);
         }
 
+        // // Log an error if another node Externalizes different values.
```
mfaulk, line 216: Will re-enable this...

consensus/scp/src/node.rs
```diff
@@ -138,34 +132,29 @@ pub trait ScpNode<V: Value>: Send {
...
 
     /// Process pending timeouts.
     fn process_timeouts(&mut self) -> Vec<Msg<V>>;
 
-    /// Get metrics for a specific slot.
-    fn get_slot_metrics(&mut self, slot_index: SlotIndex) -> Option<SlotMetrics>;
+    /// Get the current slot's index.
+    fn current_slot_index(&self) -> SlotIndex;
+
+    /// Get metrics for the current slot.
```
rjwalters, line 150: This seems like it might be problematic since node auto-advances... could we add the slot_index and keep a vector of metrics around, similar to externalized slots?

consensus/scp/src/node.rs
```diff
@@ -15,28 +15,25 @@ use mc_crypto_digestible::Digestible;
...
 
-    /// Map of last few slot indexes -> slots.
-    pub pending: LruCache<SlotIndex, Slot<V, ValidationError>>,
+    /// The current slot that this node is attempting to reach consensus on.
+    pub current_slot: Slot<V, ValidationError>,
 
-    /// Map of last few slot indexes -> externalized slots.
-    pub externalized: LruCache<SlotIndex, ExternalizePayload<V>>,
+    /// Previous, externalized slots, ordered by increasing slot index.
+    pub externalized_slots: Vec<Slot<V, ValidationError>>,
```
rjwalters, line 36: I think I agree that there's no advantage to using lrucache here, since nodes must process all slots in sequence if they fall behind. Does this mean we can drop it as a requirement in Cargo.toml for the crate?

consensus/service/src/byzantine_ledger.rs
```diff
@@ -709,113 +707,115 @@ impl<
...
-                block.index,
-                signature,
-                ext_vals,
-            );
+                // Write to ledger.
+                {
+                    let (block, block_contents, signature) = self
+                        .tx_manager
+                        .tx_hashes_to_block(&ext_vals)
+                        .unwrap_or_else(|e| {
```
rjwalters, line 739: My understanding is that we're striving to avoid this pattern in favor of `expect("Failed to build block")`, wtih the error enum including the details in it's Display implementation. Then `expect()` will panic with the output "Failed to build block: {err}" @jcape

consensus/service/src/byzantine_ledger/worker.rs
```diff
@@ -418,183 +404,209 @@ impl<
...
+                consensus_msg.scp_msg().to_display(),
+                consensus_msg.prev_block_id(),
+            );
+        }
 
-                if !self.fetch_missing_txs(scp_msg, &from_responder_id) {
-                    continue;
-                }
+        // Process compatible messages in batches.
+        let batch_size = 5;
```
rjwalters, line 444: should this be a defined const?

consensus/scp/src/node.rs
```diff
@@ -9,31 +9,35 @@ use crate::{
...
+use std::{
+    collections::{BTreeSet, HashMap},
+    fmt::Display,
+    time::Duration,
+};
 
-/// Max number of externalized slots to store.
-const MAX_EXTERNALIZED_SLOTS: usize = 10;
+/// Default limit on number of externalized slots to store.
+const MAX_EXTERNALIZED_SLOTS: usize = 1;
```
rjwalters, line 23: is there a disadvantage to keeping more slots around?

consensus/scp/src/node.rs
rjwalters: I am still concerned about this

rjwalters (approved):

consensus/service/src/byzantine_ledger/worker.rs
```diff
@@ -366,19 +362,12 @@ impl<
                 // SCP Statement
                 TaskMessage::ConsensusMsg(consensus_msg, from_responder_id) => {
                     // Only look at messages that are not for past slots.
-                    if consensus_msg.scp_msg().slot_index >= self.cur_slot {
-                        // Feed network state. The sync service needs this
-                        // to be able to tell if we fell behind based on the slot values.
-                        // Block ID checking is skipped here, since if we fell behind
-                        // we are not going to have blocks to compare to.
+                    if consensus_msg.scp_msg().slot_index >= self.current_slot_index {
```
eranrund, line 365: Wasn't part of the idea of the work done here to allow for nodes to get back to nodes that are a bit behind by properly handling the externalize phase? This would ignore any nodes that are currently behind.

consensus/service/src/byzantine_ledger/worker.rs
```diff
@@ -418,183 +404,209 @@ impl<
...
-                if !self.fetch_missing_txs(scp_msg, &from_responder_id) {
-                    continue;
-                }
+        // Process compatible messages in batches.
+        let batch_size = 5;
+        for chunk in compatible_msgs.chunks(batch_size) {
+            // Omit a message if it references a transaction that cannot be obtained.
+            let (resolved, failed): (Vec<_>, Vec<_>) =
+                chunk
+                    .iter()
```
eranrund, line 449: I wonder how this would behave if we used a parallel iterator (https://docs.rs/rayon/0.6.0/rayon/par_iter/index.html)
(out of scope for this PR but would be interesting to try it out)

consensus/service/src/byzantine_ledger/worker.rs
```diff
@@ -418,183 +404,209 @@ impl<
...
         }
     }
 
     fn attempt_complete_cur_slot(&mut self) {
         // See if we have externalized values for the current slot.
-        let ext_vals = self.scp.get_externalized_values(self.cur_slot);
-        if ext_vals.is_empty() {
-            return;
-        }
+        match self.scp.get_externalized_values(self.current_slot_index) {
```
eranrund, line 490: `if let Some(ext_vals) = self.scp.get_externalized_values(self.current_slot_index) {` is more elegant, imo (and is nested one level less)

consensus/scp/src/node.rs
mfaulk: I'm not sure yet, but I think the cost of each additional externalized slot is pretty low.

I have a hunch that we should rate-limit the production of new blocks (say, no more than one per second), then size the queue of externalized slots so that each node responds to messages for X blocks (i.e. X seconds) of recent data. That just makes it easy to reason about data availability, when to perform catch-up, etc.

consensus/scp/src/node.rs
```diff
@@ -129,36 +156,40 @@ pub trait ScpNode<V: Value>: Send {
...
     /// Process pending timeouts.
     fn process_timeouts(&mut self) -> Vec<Msg<V>>;
 
-    /// Get metrics for a specific slot.
-    fn get_slot_metrics(&mut self, slot_index: SlotIndex) -> Option<SlotMetrics>;
+    /// Get the current slot's index.
+    fn current_slot_index(&self) -> SlotIndex;
+
+    /// Get metrics for the current slot.
+    fn get_slot_metrics(&mut self) -> SlotMetrics;
```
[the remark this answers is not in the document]
mfaulk: I'm going to go with making this explicitly about the current slot's metrics. I'm happy to support metrics for other slots later -- I think that will involve several other changes, so I don't want to make it part of this PR.

consensus/scp/src/node.rs
```diff
@@ -168,206 +199,667 @@ impl<V: Value, ValidationError: Display + 'static> ScpNode<V> for Node<V, Valida
...
+        assert_eq!(node.propose_values(values), Ok(Some(msg)));
+
+        // The `slot_index` slot should now be extnalized, and current_slot should be at `slot_index + 1`.
+        assert_eq!(node.current_slot.get_index(), slot_index + 1);
+        assert_eq!(node.externalized_slots.len(), 1);
+        assert_eq!(node.externalized_slots[0].get_index(), slot_index)
+    }
+
+    // Messages for the "handle_messages" tests.
+    #[allow(unused)]
```
mfaulk, line 481: Still unused. Will remove.

consensus/service/src/byzantine_ledger/worker.rs
mfaulk: Good catch! Yes, the intent is that it should process messages for the current slot and however many recently externalized slots that Node is keeping around. I will update this and run it through another round of testing.

eranrund:
Unfortunately this branch is misbehaving locally :( (both before and after the recent changes):
image

eranrund (approved):
While this might be displaying some weird issues, I believe getting this in is a net positive.

PR #254, opened 2020-07-01
RistrettoPublic(schnorrkel::PublicKey)

sugargoat:
Soundtrack of this PR: [Quarantine Love Song](https://www.youtube.com/watch?v=pE49WK-oNjU)

### Motivation

Currently, our keys implementation does not take advantage of the schnorrkel API for constructing keypairs, which handles proper expansion from secret key material.

### In this PR
* Changes RistrettoPublic to wrap a schnorrkel::PublicKey
* Updates resulting fee constants

[MC-1611](https://mobilecoin.atlassian.net/browse/MC-1611)

### Future Work
* Update schnorrkel signature verification in #246

### DIFF
--- a/crypto/keys/src/ristretto.rs
+++ b/crypto/keys/src/ristretto.rs
@@ -11,7 +11,6 @@ use core::{
     hash::{Hash, Hasher},
 };
 use curve25519_dalek::{
-    constants::RISTRETTO_BASEPOINT_POINT,
     ristretto::{CompressedRistretto, RistrettoPoint},
     scalar::Scalar,
 };
@@ -26,6 +25,7 @@ use mc_util_repr_bytes::{
 };
 use rand_core::{CryptoRng, RngCore};
 use serde::{Deserialize, Serialize};
+use schnorrkel::{PublicKey, MiniSecretKey, ExpansionMode};
 use zeroize::Zeroize;
 
 /// A Ristretto-format private scalar
@@ -114,7 +114,7 @@ impl KexEphemeralPrivate for RistrettoPrivate {
         self,
         their_public: &<Self as PrivateKey>::Public,
     ) -> <Self as KexPrivate>::Secret {
-        RistrettoSecret((self.0 * their_public.0).compress().to_bytes())
+        RistrettoSecret((self.0 * their_public.0.as_point()).compress().to_bytes())
     }
 }
 
@@ -123,7 +123,7 @@ impl KexReusablePrivate for RistrettoPrivate {
         &self,
         their_public: &<Self as PrivateKey>::Public,
     ) -> <Self as KexPrivate>::Secret {
-        RistrettoSecret((self.0 * their_public.0).compress().to_bytes())
+        RistrettoSecret((self.0 * their_public.0.as_point()).compress().to_bytes())
     }
 }
 
@@ -154,7 +154,7 @@ impl KexEphemeralPrivate for RistrettoEphemeralPrivate {
         self,
         their_public: &<Self as PrivateKey>::Public,
     ) -> <Self as KexPrivate>::Secret {
-        RistrettoSecret((self.0 * their_public.0).compress().to_bytes())
+        RistrettoSecret((self.0 * their_public.0.as_point()).compress().to_bytes())
     }
 }
 
@@ -174,25 +174,26 @@ impl Debug for RistrettoEphemeralPrivate {
     }
 }
 
-/// A Ristretto-format curve point for use as a public key
-#[derive(Clone, Copy, Default, Digestible)]
-pub struct RistrettoPublic(pub(crate) RistrettoPoint);
+/// A schnorrkel PublicKey derived from the RistrettoPrivate Scalar
+#[derive(Clone, Copy, Default)]
+pub struct RistrettoPublic(pub(crate) PublicKey);
 
 impl AsRef<RistrettoPoint> for RistrettoPublic {
     fn as_ref(&self) -> &RistrettoPoint {
-        &self.0
+        &self.0.as_point()
     }
 }
 
 impl FromRandom for RistrettoPublic {
     fn from_random<R: CryptoRng + RngCore>(csprng: &mut R) -> RistrettoPublic {
-        Self(RistrettoPoint::random(csprng))
+        let mini_secret = MiniSecretKey::generate_with(csprng);
+        Self(mini_secret.expand_to_keypair(ExpansionMode::Uniform).public)
     }
 }
 
 impl From<RistrettoPoint> for RistrettoPublic {
     fn from(point: RistrettoPoint) -> Self {
-        Self(point)
+        Self(PublicKey::from_point(point))
     }
 }
 
@@ -201,15 +202,11 @@ impl ReprBytes for RistrettoPublic {
     type Error = KeyError;
 
     fn to_bytes(&self) -> GenericArray<u8, U32> {
-        self.0.compress().to_bytes().into()
+        self.0.as_compressed().to_bytes().into()
     }
 
     fn from_bytes(src: &GenericArray<u8, U32>) -> Result<Self, KeyError> {
-        Ok(Self(
-            CompressedRistretto::from_slice(src.as_slice())
-                .decompress()
-                .ok_or(KeyError::InvalidPublicKey)?,
-        ))
+        Ok(Self(PublicKey::from_compressed(CompressedRistretto::from_slice(src.as_slice()))?))
     }
 }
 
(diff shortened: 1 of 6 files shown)


crypto/keys/src/ristretto.rs
```diff
@@ -105,7 +105,19 @@ impl PrivateKey for RistrettoPrivate {
...
-        Self(Scalar::random(csprng))
+        let mini_secret = MiniSecretKey::generate_with(csprng);
+        let secret = mini_secret
+            .expand_to_keypair(ExpansionMode::Uniform)
+            .secret
+            .clone();
+        // The scalar is the first 32 bytes in canonical representation
+        let mut scalar_bytes: [u8; 32] = [0u8; 32];
+        scalar_bytes[0..32].copy_from_slice(&secret.to_bytes()[0..32]);
+        // FIXME: error handling
```
garbageslam, line 116: fwiw I think the code as written is correct, there should not be any error case for generating a ristretto private from an rng, and if we want to get rid of these .expect then we should fix schnorrkel crate and not attempt to bubble these errors up.

garbageslam: generating a random private key cannot logically fail. if you feel tempted to bubble errors out of that then something is wrong

sugargoat: Ok great - I will leave as `expect()`

garbageslam:
I think this is going to break the wire representation of `AccountKey` as protobuf, which might not actually matter depending on how people are using it, maybe everyone is really using root_entropy representation  instead of that.
but it's something to consider, if we are trying to stabilize this.

my two cents is that we should not change `RistrettoPublic`, we should derive the nonce from the secret key, just as ed25519. i think schnorrkel's mini secret key thing is not a good idea.

jcape:
The mini secret thing is what Ed does, tho

garbageslam:
i'm not going to try to block this PR though, I think it's important that we start to integrate the signature code, and we should be able to refactor this without too much difficulty. it's really nice that we have used the keys abstraction layer in our project so that we can make big changes like this without breaking everything :)

garbageslam:
@jcape the schnorrkel secret key structure is larger than a dalek scalar, it holds the nonce on the side, next to the private key scalar:

https://github.com/w3f/schnorrkel/blob/master/src/keys.rs#L357

that seems different from ed25519

garbageslam:
@jcape I guess this is also how isis implemented ed25519-dalek: https://github.com/dalek-cryptography/ed25519-dalek/blob/master/src/secret.rs#L287

jcape:
```
    /// Expand this `MiniSecretKey` into a `SecretKey` using
    /// ed25519-style bit clamping.
    ///
    /// At present, there is no exposed mapping from Ristretto
    /// to the underlying Edwards curve because Ristretto invovles
    /// an inverse square root, and thus two such mappings exist.
    /// Ristretto could be made usable with Ed25519 keys by choosing
    /// one mapping as standard, but doing so makes the standard more
    /// complex, and possibly harder to implement.  If anyone does
    /// standardize the mapping to the curve then this method permits
    /// compatable schnorrkel and ed25519 keys.
    ///
    /// # Examples
    ///
    /// ```compile_fail
    /// # fn main() {
    /// use rand::{Rng, rngs::OsRng};
    /// use schnorrkel::{MiniSecretKey, SecretKey};
    ///
    /// let mini_secret_key: MiniSecretKey = MiniSecretKey::generate_with(OsRng);
    /// let secret_key: SecretKey = mini_secret_key.expand_ed25519();
    /// # }
    /// ```
    fn expand_ed25519(&self) -> SecretKey {
        use sha2::{Sha512, digest::{Input,FixedOutput}};

        let mut h = Sha512::default();
        h.input(self.as_bytes());
        let r = h.fixed_result();

        // We need not clamp in a Schnorr group like Ristretto, but here
        // we do so to improve Ed25519 comparability.
        let mut key = [0u8; 32];
        key.copy_from_slice(&r.as_slice()[0..32]);
        key[0]  &= 248;
        key[31] &=  63;
        key[31] |=  64;
        // We then devide by the cofactor to internally keep a clean
        // representation mod l.
        scalars::divide_scalar_bytes_by_cofactor(&mut key);
        let key = Scalar::from_bits(key);

        let mut nonce = [0u8; 32];
        nonce.copy_from_slice(&r.as_slice()[32..64]);

        SecretKey{ key, nonce }
    }

````

vs:

```
impl<'a> From<&'a SecretKey> for ExpandedSecretKey {
    /// Construct an `ExpandedSecretKey` from a `SecretKey`.
    ///
    /// # Examples
    ///
    /// ```
    /// # extern crate rand;
    /// # extern crate sha2;
    /// # extern crate ed25519_dalek;
    /// #
    /// # fn main() {
    /// #
    /// use rand::rngs::OsRng;
    /// use ed25519_dalek::{SecretKey, ExpandedSecretKey};
    ///
    /// let mut csprng = OsRng{};
    /// let secret_key: SecretKey = SecretKey::generate(&mut csprng);
    /// let expanded_secret_key: ExpandedSecretKey = ExpandedSecretKey::from(&secret_key);
    /// # }
    /// ```
    fn from(secret_key: &'a SecretKey) -> ExpandedSecretKey {
        let mut h: Sha512 = Sha512::default();
        let mut hash:  [u8; 64] = [0u8; 64];
        let mut lower: [u8; 32] = [0u8; 32];
        let mut upper: [u8; 32] = [0u8; 32];

        h.input(secret_key.as_bytes());
        hash.copy_from_slice(h.result().as_slice());

        lower.copy_from_slice(&hash[00..32]);
        upper.copy_from_slice(&hash[32..64]);

        lower[0]  &= 248;
        lower[31] &=  63;
        lower[31] |=  64;

        ExpandedSecretKey{ key: Scalar::from_bits(lower), nonce: upper, }
    }
}
```
```
https://github.com/dalek-cryptography/ed25519-dalek/blob/3ffa9ff530750f590645e3886c40c2184c04a963/src/secret.rs#L267

jcape:
Also:
```
// ² It is the author's view (specifically, isis agora lovecruft, in the event
// you'd like to complain about me, again) that this is "ill-designed" because
// this doesn't actually provide true hash domain separation, in that in many
// real-world applications a user wishes to have one key which is used in
// several contexts (such as within tor, which does domain separation
// manually by pre-concatenating static strings to messages to achieve more
// robust domain separation).  In other real-world applications, such as
// bitcoind, a user might wish to have one master keypair from which others are
// derived (à la BIP32) and different domain separators between keys derived at
// different levels (and similarly for tree-based key derivation constructions,
// such as hash-based signatures).  Leaving the domain separation to
// application designers, who thus far have produced incompatible,
// slightly-differing, ad hoc domain separation (at least those application
// designers who knew enough cryptographic theory to do so!), is therefore a
// bad design choice on the part of the cryptographer designing primitives
// which should be simple and as foolproof as possible to use for
// non-cryptographers.  Further, later in the ed25519 signature scheme, as
// specified in RFC8032, the public key is added into *another* hash digest
// (along with the message, again); it is unclear to this author why there's
// not only one but two poorly-thought-out attempts at domain separation in the
// same signature scheme, and which both fail in exactly the same way.  For a
// better-designed, Schnorr-based signature scheme, see Trevor Perrin's work on
// "generalised EdDSA" and "VXEdDSA".
```

garbageslam:
Here's from djb:

> EdDSA avoids these issues by generating r = H(hb, . . . , h2b−1, M), so that
different messages will lead to different, hard-to-predict values of r. No permessage randomness is consumed. This idea of generating random signatures
in a secretly deterministic way, in particular obtaining pseudorandomness by
hashing a long-term secret key together with the input message, was proposed
by Barwood in [9];

So it's actually a bit funny -- he's mentioning that you could get the nonce as `hash(scalar, message)` and that people tried to patent that, but earlier in the same sentence he's actually writing `r = H(hb, . . . , h2b−1, M)`, and I `hb ... h2b` are actually the hash of the minisecret key. I didn't realize this until now.

garbageslam:
thanks, sorry for confusion!

crypto/keys/src/ristretto.rs
```diff
@@ -105,7 +105,18 @@ impl PrivateKey for RistrettoPrivate {
...
+            .expand_to_keypair(ExpansionMode::Uniform)
+            .secret
+            .clone();
+        // The scalar is the first 32 bytes in canonical representation
+        let mut scalar_bytes: [u8; 32] = [0u8; 32];
+        scalar_bytes[0..32].copy_from_slice(&secret.to_bytes()[0..32]);
+        let scalar = Scalar::from_canonical_bytes(scalar_bytes)
+            .ok_or(KeyError::InvalidPrivateKey)
+            .expect("Could not get scalar from RistrettoPrivate");
+        Self(scalar)
```
mfaulk, line 119: How does the result of this implementation differ from `Self(Scalar::random(csprng))`? How is it better?

I tried to walk through this, and it looks like it generates a MiniSecreyKey containing a random [u8; 32], then "expands" it by creating a Merlin transcript containing the MiniSecretKey, then extracting another [u8; 32] from the transcript (i.e. the hash of the transcript), which becomes the final scalar. 

In total, that seems like we've basically replaces Scalar::random, which is implemented as:

```
let mut scalar_bytes = [0u8; 64];
        rng.fill_bytes(&mut scalar_bytes);
        Scalar::from_bytes_mod_order_wide(&scalar_bytes)
```
with something that ultimately calls Scalar::from_bytes_mod_order_wide on the 64-byte result of hashing a 32-byte random value (i.e. the MiniSecretKey).

garbageslam: afaict the scalar part is not different from `Scalar::random(csprng)`,

the only thing it is doing is creating the Scalar and the nonce in a way that has a relationship similar to that defined in ed25519 spec (but they have a variation called "uniform"). the Scalar we end up with has exactly the same distribution as `Scalar::random(csprng)`. the nonce is only used when creating signatures.

sugargoat:
Closing for now

PR #256, opened 2020-07-01
new fog fields require a new payload version for backwards compatibilty

rjwalters:
### Motivation

Recent changes to the b58 code scheme were not backwards compatible. This breaking change needs to be coordinated with our release schedule.

Even if we decide not to use the b58 codes in the SDK work, there's no reason to break backwards compatibility in the test client.

### In this PR

* move new fields to version 4

### Future Work

* actually verify the test vectors by comparison to calculation in javascript
* do this right! Separate serialization from encoding and use protobuf

### DIFF
--- a/util/b58-payloads/src/payloads.rs
+++ b/util/b58-payloads/src/payloads.rs
@@ -629,7 +629,7 @@ mod testing {
         );
         let alice_request_payload = RequestPayload::try_from(&alice_public).unwrap();
         let alice_b58_str = alice_request_payload.encode();
-        assert_eq!(alice_b58_str, "fixme");
+        assert_eq!(alice_b58_str, "5F24DBwfEGBN18LevLEt3wZHknBk1tSC32QfMzSoiPcMrJyErBqBiNQpfFWboY1DHAjwjkBf2qfUbGvjvcdJ6Uhp6qcZ1NChFutR2MpiyTkrKm7NhCrT6KjtDkQBxr7rw74oFzvh7rkrK");
         let alice_payload = RequestPayload::decode(&alice_b58_str).unwrap();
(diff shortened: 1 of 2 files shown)


garbageslam:
the testnet client doesn't use fog, so this seems low priority

garbageslam:
it seems wierd though that we won't be able to encode a fog address with a signature.
like, this makes v1 completely useless

garbageslam:
I think you should consider deleting v1

rjwalters:
> I think you should consider deleting v1

yeah I think it's unlikely that we will want to go back and support "old style fog" with root certs. I'll see if I can mark this deprecated rather than delete it since it's used in `new_v2()`. 

Adding new unanticipated "mandatory" fields has revealed a disadvantage of having a scheme with a fixed parameter order. Amount and an empty memo only add about 10 bytes but it's still an annoyance. We're also just swimming upstream by trying to avoid concepts like "Public Address" in an effort to keep things simple for users. Given the inertia with exchanges it seems impossible to avoid surfacing pub addresses as their own thing, which is unnecessarily overloading the PaymentRequest deeplink concept.

util/b58-payloads/src/payloads.rs
```diff
@@ -206,26 +207,23 @@ impl RequestPayload {
...
-            fog_authority_sig: Default::default(),
             value: 0,
             memo: "".to_owned(),
+            fog_report_id: Default::default(),
+            fog_authority_sig: Default::default(),
         })
     }
 
-    /// Create a version 1 RequestPayload
+    /// Create a version 1 RequestPayload - this is deprecated becuause fog now requires a signature
```
kylefleming, line 217:
```suggestion
    /// Create a version 1 RequestPayload - this is deprecated because fog now requires a signature
```

PR #272, opened 2020-07-07
Fix watcher build

sugargoat:
Soundtrack of this PR: [My Favorite Mistake](https://www.youtube.com/watch?v=AmIlUKo4dQc)

### Motivation

Ended up in a situation where the previous PR #271 passed CI, built and tested locally, but then when merged to master, failed CI and deployment.

### In this PR
* Fixes build error and cleans up lingering resolved FIXME

### DIFF
--- a/watcher/src/watcher_db.rs
+++ b/watcher/src/watcher_db.rs
@@ -202,12 +202,16 @@ impl WatcherDB {
         let db_txn = self.env.begin_ro_txn()?;
         let mut cursor = db_txn.open_ro_cursor(self.last_synced)?;
 
-        // FIXME: do this in one db transaction
         let all_urls: Vec<Url> = cursor
             .iter()
-            .map(|(url_bytes, _block_index_bytes)| {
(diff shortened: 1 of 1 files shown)


sugargoat:
I think the unwraps should be ok because the URL bytes were constructed from calls to to_str, as_bytes, and this is in an anonymous function in the filter_map, so syntax is less convenient. There are several ways to peel this potato, though, so I’m happy to make changes if something else is preferred.

jcape:
> I think the unwraps should be ok because the URL bytes were constructed from calls to to_str, as_bytes, and this is in an anonymous function in the filter_map, so syntax is less convenient. There are several ways to peel this potato, though, so I’m happy to make changes if something else is preferred.

I think anywhere we do the unwrap-as-assert pattern, we should just use expect in the same way we'd add a string description to an assert. i.e. we shouldn't use unwrap anywhere.

sugargoat:
@jcape 👍 Fixed

PR #292, opened 2020-07-10
Mc sig

garbageslam:
Soundtrack of this PR: https://www.youtube.com/watch?v=tzR5NtB0xlY&t=5s

### Motivation

In https://github.com/mobilecoinofficial/mobilecoin/pull/246 we merged a custom adaptation
of the Schnorrkel crate, to support deterministic digital signatures and avoid creating extra
storage requirements for nonces in account key. This code is still being reviewed.

This PR moves the new scheme to a new crate, instead of the transaction core crate.
This PR is purely code movement and should not be a behavior change.

- It also exposes a simpler API so that we can more easily reuse this scheme instead of
copy pasting code around. Other crates can use these signatures without bringing in dependencies on
complex stuff like bulletproofs.
- This is also expected to make it easier to review, since it is small and well-scoped. The transaction core crate, by contrast, is enormous and has many dependencies. Its scope has grown to contain any cryptography directly related to mobilecoin transactions, and to the account key and public address structs, and serialization concerns related to this, so it contains things as diverse as bulletproofs, ring signatures, and merkle proofs, and now, Schnorr signatures.
- Normally it is a goal that cryptography crates have as few dependencies as possible, and especially, build in a no_std environment. It is not good if the cryptography cannot be separated from things like prost and serde. This particular crate works even without `extern crate alloc;`, just like Schnorrkel.

### In this PR

Purely code movement, re-org, and documentation improvements, buliding on
https://github.com/mobilecoinofficial/mobilecoin/pull/246.

- Move blake2b-256 module to a crate separate from transaction core crate. I discussed this
with Matt a few months ago in slack but we never actually merged a change. Because the
signatures need this hash function, if we want to use it in the signature crate, it needs to move
out of transaction crate as well.

- Move API for sign and verify logic for our schnorrkel sigs to its own crate, and add tests

### Future Work

I think it might be a good idea to move account key and public address to their own crate.
This would allow to e.g. build fog without building bulletproofs.


crypto/sig/src/lib.rs
```diff
@@ -0,0 +1,125 @@
+#![no_std]
+
+//! Simplified API for using Schnorrkel in a deterministic manner, with simple
+//! ristretto public and ristretto private key-pairs.
```
sugargoat, line 4: For audit purposes, it may be worth mentioning here that RistrettoPrivate is a wrapper of a Ristretto Scalar, and RistrettoPublic is a Ristretto Point

mfaulk:
Thanks, Chris. I feel like this is moving things in the right direction. I generally like the idea of keeping `transction_core` minimal, and I think this PR helps. Moving AccountKey and PublicAddress also seems like a good separation of concerns.

crypto/sig/README.md
```diff
@@ -0,0 +1,85 @@
...
+Putting this low-level code in its own module makes it easier to reuse
+and to audit, and hides sensitive cryptographic details from the users of the API.
+
+The main differences are:
+
+- The signature is completely deterministic, like RFC6979 and ed25519, and `sign_message`
+  does not take an RNG as it does in Schnorrkel. Instead, in order to use Schnorrkel without
+  patching it, we create a seeded RNG using a seed derived in a manner similar to the nonce
+  in RFC6979 and ed25519.
+  -  This is very important for mobilecoin because we want for the public address to be
```
sugargoat, line 26:
```suggestion
  -  This is very important for MobileCoin because we want for the public address to be
```

crypto/sig/README.md
```diff
@@ -0,0 +1,85 @@
...
+     completely deterministic from the private account key, so that we can detect malicious
+     tampering easily.
+- We do not use the "minisecret key" or the "minisecret key expansion" from Schnorrkel.
+  - This will create serious problems for mobilecoin, because we want to be able to use
```
sugargoat, line 30:
```suggestion
  - This will create serious problems for MobileCoin, because we want to be able to use
```

crypto/sig/README.md
```diff
@@ -0,0 +1,85 @@
...
+sign_message(signing_context: &[u8], private_key: &RistrettoPrivate, message: &[u8]) -> Signature
+verify_signature(signing_context: &[u8], public_key: &RistrettoPublic, message: &[u8], sig: &Signature) -> bool
+```
+
+Putting this low-level code in its own module makes it easier to reuse
+and to audit, and hides sensitive cryptographic details from the users of the API.
+
+The main differences are:
+
+- The signature is completely deterministic, like RFC6979 and ed25519, and `sign_message`
```
sugargoat, line 22: Can you provide links inline to RFC6979 and ed25519 documentation?

crypto/sig/README.md
```diff
@@ -0,0 +1,85 @@
...
+> but we prefer to reuse `H` to save area in hardware implementations.
+
+We point out that Blake was one of the SHA-3 finalists, and was also explicitly designed to model a PRF.
+We prefer Blake2b here because it reduces the number of different hash functions in our system overall.
+
+Assuming the Blake2b has the secret-prefix-PRF property, we can say that the signatures created this way
+are hard to distinguish from signatures created where the nonce is truly uniformly random, even if the messages
+that are signed are adversarially chosen. So, if Schnorrkel is secure when the nonces are truly random and
+the RNG is the OS-RNG, or, when the nonce is created using the mini-secret-key expansion, then this should also be secure.
+
```
sugargoat, line 79: I don't know if it's worth explicitly mentioning that we will only sign one message with this nonce, since the nonce depends on the message.

crypto/sig/README.md
```diff
@@ -0,0 +1,85 @@
...
+Rng required by Schnorrkel
+--------------------------
+
+Schnorrkel API deviates from RFC6979 and ed25519 in requiring the signer to provide a CSPRNG, which they
```
sugargoat, line 83: I think the Schnorrkel API allows us to bypass the OS rng via `attach_rng`, but I haven't been able to devote time to trying this out yet.

crypto/sig/README.md
garbageslam: :+1:

crypto/sig/README.md
garbageslam: yeah but my point is, RFC6979 and ed25519 don't require an RNG at all. like, that's what deterministic signatures means. i think Schnorrkel has departed from earlier work in asking us for an RNG at all.

crypto/sig/README.md
garbageslam: I'll do another editing pass

garbageslam:
@sugargoat lmk if you think the last edits make this better
I also added a FIXME comment in the code -- rather than change the behavior in this commit,
because the PR comments when I opened it say that this is just code movement. lmk what you think

PR #294, opened 2020-07-14
[MC-1557] Add building on macOS to CI

kylefleming:
### Motivation

The TestNet-client is built for macOS as part of the release process. This means it should be under test to avoid unintentional regressions.

### In this PR
* Adds a CircleCI job to build the project on macOS

fixes #MC-1557


kylefleming:
This PR has now been rebased on master.

PR #297, opened 2020-07-15
Export needed constant from account-keys lib

garbageslam:
This is needed by libmobilecoin

Fixes up previous commit

### DIFF
--- a/account-keys/src/domain_separators.rs
+++ b/account-keys/src/domain_separators.rs
@@ -1,4 +1,7 @@
 // Copyright (c) 2018-2020 MobileCoin Inc.
 
+/// The tag for the signature context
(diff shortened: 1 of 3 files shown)


account-keys/src/domain_separators.rs
```diff
@@ -1,4 +1,7 @@
 // Copyright (c) 2018-2020 MobileCoin Inc.
 
+/// The tag for the signature context
+pub const FOG_AUTHORITY_SIGNATURE_TAG: &[u8; 23] = b"Fog authority signature";
```
mfaulk, line 4: Just curious: Is there a particular reason to use `&[u8; 23]` instead of `&str`?

garbageslam: Not really, this is just what we merged before

PR #302, opened 2020-07-17
root entropy changes

garbageslam:
This PR is mostly code movement to put the `RootIdentity` struct in with `mc-account-key` but there is some new code around, adding support for all the fog data in the `RootIdentity`, and around, checking `root_entropy` values.

### Motivation

The "root entropy" / "root identity" concept is useful for making accounts that are still cryptographically secure (if the KDF is good) but can be memorized / printed as QR codes much more easily, since they are seeded from only 32 bytes of entropy instead of 64.

This concept has been used extensively in demos but we would like to try to stabilize a serialization format for these, using protobuf like everything else, and stabilize the derivation methods, and a method of "blacklisting" bad random values to the extent possible.

I would also like to move this to the `mc-account-keys` crate because IMO it doesn't really belong in `mc-transaction-std`.

### In this PR

- Move to `mc-account-keys` and remove many unnecessary dependencies on `mc-transaction-std`.
- Support prost serialization, and test that this works with the .proto file
- Remove protobuf serialization
- Support all account key fields in the root identity structure
- Remove the `.json` files from keyfile and make it use protobuf instead for root entropies.
- Move ViewKey over to use protobuf serialization, add stuff to external.proto for it

### Future work

???

In review comments, we decided to remove the root entropy smoke tests thing


account-keys/src/identity.rs
```diff
@@ -0,0 +1,198 @@
+// Copyright (c) 2018-2020 MobileCoin Inc.
+
+//! In cryptography, several private keys can be derived from a single source of
+//! entropy using a strong KDF (key derivation function).
+//! This is sound, so long as the input-key-material to the KDF itself has
+//! at least enough entropy as the length of any one of the derived keys.
+//!
+//! The RootIdentity object contains 32 bytes of "root entropy", used with HKDF
+//! to produce the other mobilecoin private keys. This is useful an AccountKey
```
rjwalters, line 9:
```suggestion
//! to produce the other MobileCoin private keys. This is useful because an AccountKey
```

account-keys/src/identity.rs
```diff
@@ -0,0 +1,198 @@
...
+    #[prost(bytes, tag = 4)]
+    pub fog_authority_fingerprint: Vec<u8>,
+}
+
+impl RootIdentity {
+    /// Generate a random root identity with a specific fog url configured
+    pub fn random<T: RngCore + CryptoRng>(rng: &mut T, fog_url: Option<&str>) -> Self {
+        let mut root_entropy = [0u8; 32];
+        // Filter out bad root entropies
+        // TODO: Break after 100 tries and fail?
```
garbageslam, line 54: i feel like, maybe a better way is, use a known good rng with fixed seed, and confirm in a unit test that it succeeds within at least 100 tries. if a developer is using this code and their thing is in an infinite loop, it implies that something is wrong with their Rng i guess.

account-keys/src/entropy_check.rs
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) 2018-2020 MobileCoin Inc.
+
+use displaydoc::Display;
+
+/// Check an entropy string for "obvious" statistical problems or blacklisted values.
+///
+/// Note: It would be good to try to make the checks constant-time, since the root
+/// entropy is generally a secret value.
+pub fn check_root_entropy(root_entropy: &[u8]) -> Result<(), RootEntropyProblem> {
```
jcape, line 9: I'd recommend factoring this into an implementation of the [NIST tests](https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-22r1a.pdf) against an RNG generally, and running it on startup---that is, on startup, grab an RNG and run the tester on it, and then simply refuse to start if it fails. This could happen on node/enclave startup as an environmental sanity check, and would not preclude the use of test vectors.

account-keys/src/identity.rs
jcape: As noted above, I think it's better to just run some tests on the RNG at startup, then just use it normally everywhere else. Though TBH, `CryptoRng` is supposed to be that already...

garbageslam:
@rjwalters in response to review and discussion with james, did the following:

- The .json files containing root entropies are gone, it is called ".root" and the format is protobuf rather than serde json, like the .pub files. No one has to use these for anything, and the known accounts is fine. but it seems reasonable to have a file format in case you want to transport them in files. AFAIK no one actually is using them, only the .pub part.
- the RootEntropy struct is created and represents exactly 32 bytes and implements prost::Message
- serde is gone entirely from account keys crate
- the tests for "bad root entropy" are deleted

api/proto/external.proto
```diff
@@ -95,10 +87,57 @@ message PublicAddress {
...
+message RootEntropy {
+    bytes data = 1;
+}
+
+/// A ViewKey is a reduced AccountKey -- it contains the private key necessary to
+/// view your transactions and see the amounts, but not to send new transactions.
+/// This concept is part of Cryptonote.
+/// In Mobilecoin, all public addresses correspond to subaddresses, and often
+/// the "default subaddress" is used.
+/// The ViewKey similarly corresponds to a particular subaddress.
```
garbageslam, line 115: @xoloki lmk if you think this is a good description

api/tests/prosty.rs
```diff
@@ -0,0 +1,81 @@
+//! Tests that prosty versions of things in external.proto roundtrip with protobuf versions
```
kylefleming, line 1: Opinion:

I find the word `prosty` (as in `prosty.rs`) to be confusing since it doesn't actually correspond to anything. Would the reader assume it's related to `prost` or would they google it assuming it might be something entirely different? I think we should stick to using the word `prost`.

(Although, while still a bit weird, using phrasing like `prost-y` in comments is at least not as confusing in the same way as `prosty` since it's more clear that it's referring to `prost`.)

garbageslam: yeah if you don't like this I can change it to `prost.rs`

in casual conversation some engineers have referred to "prosty versions" of structs as opposed to the protobuf lib generated ones, that's where this came from

garbageslam: done

garbageslam:
@sugargoat i realized this is going to break the mobilecoind strategies, i'm thinking about either trying to make the mobilecoind strategies use protobuf to load the account keys files, or just leave the json format the way it was, lmk what you think plz

garbageslam:
going to come up with a more creative way of preserving the json format

garbageslam:
this passed deploy, going to merge

PR #303, opened 2020-07-17
Run unit tests in parallel in CI

kylefleming:
### Motivation

This PR optimizes the CircleCI checks.

This PR started out incorporating what has now been split out as PR #306. This PR now only includes the change that converts the unit tests to run in parallel in CI.

### In this PR

* Runs unit tests in parallel (currently set to 4 concurrent jobs)
* Lowers typical time for `all-tests` check from 10-12 mins to 7-8 mins

### Future Work

* This PR splits up the tests but only naively. In the future, it would do this based on timing data. There are currently issues with outputting test results with timing data from Cargo in a way that CircleCI can use. These issues will need to be worked out before CircleCI's test split feature can be used. However, at the moment the naive approach should provide sufficient performance.
* Add macOS to CI. (MC-1557)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -211,6 +211,30 @@ commands:
               | tee /tmp/test-results.json
             cat /tmp/test-results.json | cargo2junit > /tmp/test-results.xml || true
 
+  run-parallel-tests:
+    parameters:
+      test_command:
+        type: string
+        default: cargo test --frozen
+    steps:
+      - run:
+          name: Run parallel unit tests
+          command: |
+            # Generate list of test packages
+            all_packages="$(<< parameters.test_command >> --message-format json --no-run \
+              | jq -r 'select(.profile.test).package_id | split(" ")[0]' | sort -u)"
+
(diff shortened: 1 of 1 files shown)


eranrund:
What are the cost implications of this?
Running 4 containers for ~8 min instead of one container for ~10 min is almost a x4 in our CI costs. We vary between $500-1500 a month as it is. Doubling it x4 seems not worth it to me to shave a few minutes:
image

kylefleming:
> What are the cost implications of this?

For the month of June, this change would have increased our CircleCI usage by $125-$150.

$0.0006 per credit, 40 credits per xl linux docker minute, 258 workflows run in June (excluding canceled workflows), 20-24 extra minutes per workflow (10 mins - ~4*8 mins).

sugargoat:
I'm also concerned re: the points that Eran raised. I'm not sure that the cost-benefit is sufficient to tip in favor of optimizing in this way.

kylefleming:
Closing for now.

PR #308, opened 2020-07-21
Fixup code comments in keyfile json_format module per discussion

garbageslam:
earlier PR was #302

### DIFF
--- a/util/keyfile/src/json_format.rs
+++ b/util/keyfile/src/json_format.rs
@@ -1,7 +1,7 @@
 //! We mostly try to serialize our types using protobuf,
 //! but early in the project we were using a json-based representation of
-//! account key root entropy. There is now some mobilecoind strategies
-//! code based on this format, so we maintain it here.
+//! account key root entropy. As part of our CD, we maintain the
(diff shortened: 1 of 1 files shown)


util/keyfile/src/json_format.rs
```diff
@@ -1,9 +1,10 @@
-//! We mostly try to seriliaze our types using protobuf,
+//! We mostly try to serialize our types using protobuf,
 //! but early in the project we were using a json-based representation of
-//! account key root entropy. There is now a lot of mobilecoind strategies
+//! account key root entropy. There is now some mobilecoind strategies
```
sugargoat, line 3:
```suggestion
//! account key root entropy. As part of our CD, we maintain strategies/test_client 
```

util/keyfile/src/json_format.rs
```diff
@@ -1,9 +1,10 @@
...
 //! code based on this format, so we maintain it here.
```
sugargoat, line 4:
```suggestion
//! account key root entropy. As part of our CD, we maintain the
//! strategies/test_client code based on this format.
```

PR #311, opened 2020-07-21
Remove redundant filtering and validation from Node

mfaulk:
Soundtrack of this PR: [Paid](https://www.youtube.com/watch?v=Df4vDfYxTRk)

### Motivation

Previously, Node and Slot both validated nominated values and filtered out stale or redundant incoming messages, which is confusing and maybe inefficient. This PR makes Slot primarily responsible for "doing the right thing" with proposed values and incoming messages.

This has been slam tested, and seems to have identical performance.

### DIFF
--- a/consensus/scp/src/node.rs
+++ b/consensus/scp/src/node.rs
@@ -185,21 +185,8 @@ impl<V: Value, ValidationError: Display> ScpNode<V> for Node<V, ValidationError>
             return Ok(None);
         }
 
-        let valid_values: BTreeSet<V> = values
-            .into_iter()
-            .filter(|value| (self.validity_fn)(&value).is_ok())
-            .collect();
-
-        if valid_values.is_empty() {
-            log::error!(
-                self.logger,
-                "nominate() called with only invalid values. Ignoring."
-            );
-            return Ok(None);
-        }
-
         let slot = self.get_or_create_pending_slot(slot_index);
-        let outbound = slot.propose_values(&valid_values)?;
+        let outbound = slot.propose_values(&values)?;
 
         match &outbound {
             None => Ok(None),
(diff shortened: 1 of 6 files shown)


consensus/scp/src/node.rs
```diff
@@ -44,10 +39,6 @@ pub struct Node<V: Value, ValidationError: Display> {
     /// Application-specific function for combining multiple values. Must be deterministic.
     combine_fn: CombineFn<V>,
 
-    /// Hashes of messages we've already processed.
-    /// (We store hashes instead of message content to reduce memory footprint.)
-    pub seen_msg_hashes: LruCache<Hash, ()>,
```
mfaulk, line 49: Slot ignores duplicate or lesser messages on a per-node basis, which I think is simpler and more effective than maintaining a cache of messages here.

consensus/scp/src/node.rs
```diff
@@ -138,8 +128,8 @@ pub trait ScpNode<V: Value>: Send {
     /// Get local node quorum set.
     fn quorum_set(&self) -> QuorumSet;
 
-    /// Submit a list of values of nomination.
-    fn nominate(
```
mfaulk, line 142: This was renamed to more closely match Slot.propose_values, and to more clearly indicate that a proposed value does not necessarily get nominated immediately.

consensus/scp/src/slot.rs
```diff
@@ -303,12 +303,14 @@ impl<V: Value, ValidationError: Display> Slot<V, ValidationError> {
...
         }
 
-        // Reject invalid values.
-        for value in values {
-            self.is_valid(value)?;
-        }
+        // Omit any invalid values.
+        let valid_values: Vec<V> = values
+            .iter()
+            .filter(|value| self.is_valid(value).is_ok())
```
mfaulk, line 309: Here, slot validates proposed values.

mfaulk:
@eranrund Yes, crate tests and slam both look good.

PR #334, opened 2020-08-03
BASE_FEE -> MINIMUM_FEE, and increase minimum to 10mMOB

sugargoat:
Soundtrack of this PR: [She Blinded Me with Science](https://www.youtube.com/watch?v=V83JR2IoI8k)

### Motivation

More accurate naming and increase minimum fee from 10 pMOB to 10 mMOB. 

### In this PR
* Change BASE_FEE -> MINIMUM_FEE
* Set minimum to 10mMOB

[MC-1733](https://mobilecoin.atlassian.net/browse/MC-1733)

### DIFF
--- a/transaction/core/src/constants.rs
+++ b/transaction/core/src/constants.rs
@@ -25,7 +25,7 @@ pub const TOTAL_MOB: u64 = 250_000_000;
 /// one milliMOB = 1e9 picoMOB
 const MILLIMOB = 1_000_000_000;
 
-/// Minimum allowed fee, denominated in milliMOB (Note, base unit is picoMOB).
+/// Minimum allowed fee, denominated in picoMOB.
 pub const MINIMUM_FEE: u64 = 10 * MILLIMOB;
 
 lazy_static! {
--- a/transaction/core/src/constants.rs
+++ b/transaction/core/src/constants.rs
@@ -23,10 +23,10 @@ pub const MAX_TOMBSTONE_BLOCKS: u64 = 100;
 pub const TOTAL_MOB: u64 = 250_000_000;
 
 /// one milliMOB = 1e9 picoMOB
-pub const MILLIMOB: u64 = 1_000_000_000;
+pub const MILLIMOB_TO_PICOMOB: u64 = 1_000_000_000;
 
 /// Minimum allowed fee, denominated in picoMOB.
-pub const MINIMUM_FEE: u64 = 10 * MILLIMOB;
+pub const MINIMUM_FEE: u64 = 10 * MILLIMOB_TO_PICOMOB;
 
 lazy_static! {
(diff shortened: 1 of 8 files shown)


transaction/core/src/constants.rs
```diff
@@ -22,8 +22,8 @@ pub const MAX_TOMBSTONE_BLOCKS: u64 = 100;
 /// The MobileCoin network will contain a fixed supply of 250 million mobilecoins (MOB).
 pub const TOTAL_MOB: u64 = 250_000_000;
 
-/// Minimum allowed fee, denominated in picoMOB.
-pub const BASE_FEE: u64 = 10;
+/// Minimum allowed fee, denominated in milliMOB (Note, base unit is picoMOB).
```
kylefleming, line 25: Wouldn't the denomination of `MINIMUM_FEE` still be picoMOB? The `10` is denominated in `milliMOB` but the doc comment is meant to document the `MINIMUM_FEE` variable itself.

If someone was looking at autogenerated docs they might get the impression that `MINIMUM_FEE` itself is denominated in milliMOB and thus requires multiplying by `1_000_000_000` in order to be denominated in picoMOB.

rjwalters, line 25:
```suggestion
/// one milliMOB = 1e9 picoMOB
const MILLIMOB = 1_000_000_000;

/// Minimum allowed fee, denominated in milliMOB (Note, base unit is picoMOB).
```

transaction/core/src/constants.rs
```diff
@@ -22,8 +22,8 @@ pub const MAX_TOMBSTONE_BLOCKS: u64 = 100;
...
+pub const MINIMUM_FEE: u64 = 10 * 1_000_000_000;
```
rjwalters, line 26:
```suggestion
pub const MINIMUM_FEE: u64 = 10 * MILLIMOB;
```

transaction/core/src/constants.rs
sugargoat: Ah thank you - I originally was going to do the conversion elsewhere. Will fix up the comment.

transaction/core/src/constants.rs
```diff
@@ -22,8 +22,11 @@ pub const MAX_TOMBSTONE_BLOCKS: u64 = 100;
 /// The MobileCoin network will contain a fixed supply of 250 million mobilecoins (MOB).
 pub const TOTAL_MOB: u64 = 250_000_000;
 
+/// one milliMOB = 1e9 picoMOB
+pub const MILLIMOB: u64 = 1_000_000_000;
```
kylefleming, line 26: This should probably be named something more like `MILLIMOB_FROM_PICOMOB` (or `PICOMOB_TO_MILLIMOB`).

PR #341, opened 2020-08-04
fixup ed25519-dalek version

garbageslam:
This moves off of our patched ed25519-dalek

In developing this PR I learned the following things:

- `digest` trait is used as an API among an enormous number of our crates,
  and mixing `digest 0.8` and `digest 0.9` generally breaks the build
- Many RustCrypto crates are already on `digest 0.9` but `curve25519-dalek` ecosystem is still on `digest 0.8`
- We will likely have to pin things to versions that use `digest 0.8` until `curve25519-dalek` uprevs their version of digest

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -945,10 +945,12 @@ dependencies = [
 
 [[package]]
 name = "ed25519-dalek"
-version = "1.0.0-pre.3"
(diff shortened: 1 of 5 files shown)


garbageslam:
See also: https://github.com/dalek-cryptography/curve25519-dalek/pull/327

PR #350, opened 2020-08-05
Add a basic constant time FogHint::ct_decrypt function

xoloki:
Soundtrack of this PR: [Walk Like an Egyptian](https://www.youtube.com/watch?v=Cv6tuzHUuuk)

### Motivation

In order to make the ingest loop constant time, we need a constant time decrypt function for the fog hint.  We can't really do this now, due to limitations in the aes-gcm library (https://github.com/RustCrypto/AEADs/issues/184).  I have a PR against aes-gcm that will allow us to do so (https://github.com/RustCrypto/AEADs/pull/188), but there is no guarantee that it will be merged.  We may have to carry a fork of aes-gcm, but that can wait for the future.  This PR will allow us to move forwards, which is good enough for now.

### In this PR
* Adds a FogHint::ct_decrypt function, which returns a (FogHint, bool)

This will help make progress on FOG-96 and FOG-97.

### Future Work
* Use this code in the ingest enclave impl
* Make a constant time aes-gcm decrypt and update this function to use it

### DIFF
--- a/transaction/core/src/fog_hint.rs
+++ b/transaction/core/src/fog_hint.rs
@@ -160,10 +160,7 @@ impl FogHint {
             }
         }
 
-        let default_fog_hint = FogHint::from_slice(
-            &default_plaintext.as_ref()[0..RISTRETTO_PUBLIC_LEN],
-        )
-        .expect("default plaintext should always contain a valid RistrettoPublic at the beginning");
+        let default_fog_hint = FogHint::new(default_pubkey);
 
         match FogHint::from_slice(&plaintext.as_ref()[0..RISTRETTO_PUBLIC_LEN]) {
             Ok(fog_hint) => (fog_hint, success),
--- a/transaction/core/src/fog_hint.rs
+++ b/transaction/core/src/fog_hint.rs
@@ -15,6 +15,8 @@ use mc_crypto_box::{
 use mc_crypto_keys::{
     CompressedRistrettoPublic, ReprBytes, Ristretto, RistrettoPrivate, RistrettoPublic,
 };
+use mc_crypto_rand::McRng;
+use mc_util_from_random::FromRandom;
 use rand_core::{CryptoRng, RngCore};
 
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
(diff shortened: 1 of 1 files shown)


transaction/core/src/fog_hint.rs
```diff
@@ -117,6 +119,57 @@ impl FogHint {
...
         FogHint::from_slice(&plaintext.as_ref()[0..RISTRETTO_PUBLIC_LEN])
     }
+
+    /// ct_decrypt
+    ///
+    /// Try to decrypt an encrypted payload onto this FogHint object in constant time.
+    /// Fails if decryption fails, or the magic number is wrong.
+    ///
+    /// # Arguments
+    /// * acct_server_private_key
```
sugargoat, line 129:
```suggestion
    /// * ingest_server_private_key
```

transaction/core/src/fog_hint.rs
```diff
@@ -117,6 +119,57 @@ impl FogHint {
...
+        for byte in &plaintext[RISTRETTO_PUBLIC_LEN..] {
+            if *byte != MAGIC_NUMBER {
+                success = false;
+            }
+        }
+
+        let default_fog_hint = FogHint::from_slice(
+            &default_plaintext.as_ref()[0..RISTRETTO_PUBLIC_LEN],
+        )
+        .expect("default plaintext should always contain a valid RistrettoPublic at the beginning");
```
sugargoat, line 166: Should this return `(default_plaintext, false)`? Or it's ok to have the panic?

transaction/core/src/fog_hint.rs
```diff
@@ -117,6 +119,57 @@ impl FogHint {
...
+    ///
+    /// Try to decrypt an encrypted payload onto this FogHint object in constant time.
+    /// Fails if decryption fails, or the magic number is wrong.
+    ///
+    /// # Arguments
+    /// * acct_server_private_key
+    /// * 128 byte encrypted payload
+    ///
+    /// # Returns
+    /// * Fog hint on success, cryptobox error otherwise
```
sugargoat, line 133:
```suggestion
    /// * (Fog hint, true) on success, (Default fog hint, false) otherwise
```

transaction/core/src/fog_hint.rs
xoloki: Fixed.

transaction/core/src/fog_hint.rs
xoloki: I changed it to make the default fog hint directly from the default_pubkey, so it can't fail.

transaction/core/src/fog_hint.rs
xoloki: Fixed.

garbageslam:
i'd suggest the following, but this is a larger change which we could hold off on:

1. Get rid of `FogHint` type, we only need to have `EncryptedFogHint`. It can encrypt from `CompressedRistrettoPublic`? And it can `ct_decrypt` to either `GenericArray<u8, U32>` or `[u8; 32]` whatever you prefer. The only reason i would not want to decrypt to `CompressedRistrettoPublic` is that i don't think that type supports constant time operations (like `subtle::ConstantTimeSelectable`). So we could skip `CompressedRistrettoPublic` on encryption side also?
2. `ct_decrypt` could take `&mut [u8;32]` as an output parameter and set the output only if decryption succeeded, and return a bool indicating if decryption succeeded. Conventionally in rust we use result but lacking a `CtResult` IMO its fine to use output parameter, it's usually gonna save a copy. If you use an output parameter then you don't need to worry about passing a default in.

garbageslam:
We could also just make `FogHint` a type alias for `GenericArray<u8, U32>` or something

PR #356, opened 2020-08-07
[MC-1766] Use mobilecoinofficial/serde_cbor fork

kylefleming:
### Motivation

`serde_cbor` currently uses serde in such a way that it fails to compile if the `serde_cbor/std` feature is not enabled but `serde/std` is. The error is:
```
error[E0277]: the trait bound `error::Error: std::error::Error` is not satisfied
   --> /root/.cargo/registry/src/github.com-1ecc6299db9ec823/serde_cbor-0.11.1/src/de.rs:781:18
    |
781 |     type Error = Error;
    |                  ^^^^^ the trait `std::error::Error` is not implemented for `error::Error`
```

This issue has been fixed in https://github.com/pyfisch/cbor/pull/198, but the maintainer of serde_cbor is no longer merging PRs (see https://github.com/pyfisch/cbor/issues/179#issuecomment-654071549).

A fork has been created that fixes https://github.com/pyfisch/cbor/pull/198, located here: https://github.com/mobilecoinofficial/cbor/tree/fix-serde-std

### In this PR
* Switches to using the https://github.com/mobilecoinofficial/cbor fork of https://github.com/pyfisch/cbor.

### Future Work
* Removing `serde_cbor` as a dependency entirely since it's no longer being maintained and the maintainer is no longer accepting PRs. (MC-1767)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4860,8 +4860,7 @@ dependencies = [
 [[package]]
 name = "serde_cbor"
 version = "0.11.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1e18acfa2f90e8b735b2836ab8d538de304cbb6729a7360729ea5a895d15a622"
+source = "git+https://github.com/mobilecoinofficial/cbor?rev=4c886a7c1d523aae1ec4aa7386f402cb2f4341b5#4c886a7c1d523aae1ec4aa7386f402cb2f4341b5"
 dependencies = [
  "half",
  "serde",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -116,3 +116,8 @@ console = { git = "https://github.com/mitsuhiko/console", rev = "1307823" }
 
 # Override lmdb-rkv for a necessary bugfix (see https://github.com/mozilla/lmdb-rs/pull/80)
 lmdb-rkv = { git = "https://github.com/mozilla/lmdb-rs", rev = "df1c2f5" }
+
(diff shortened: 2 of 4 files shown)


garbageslam:
does this have the patch to serde_cbor? https://github.com/mobilecoinofficial/cbor/commits/master
I couldn't find the commit

kylefleming:
> does this have the patch to serde_cbor? [mobilecoinofficial/cbor@`master` (commits)](https://github.com/mobilecoinofficial/cbor/commits/master)
> I couldn't find the commit

Yea, in the https://github.com/mobilecoinofficial/cbor/tree/fix-serde-std branch. I'll push it to master as well though.

garbageslam:
do you want to turn on branch protection on master also? (or i can do it)

kylefleming:
> do you want to turn on branch protection on master also? (or i can do it)

If you want, go for it. Nothing is referencing master though, so I think it's less of an issue than for a repo with more active development.

garbageslam:
i did it -- its just that philosophically, if someone force pushed this repo, that would break historical mobilecoin builds, so i would like to prevent that, to preserve the ability to checkout old commits and build

kylefleming:
> i did it -- its just that philosophically, if someone force pushed this repo, that would break historical mobilecoin builds, so i would like to prevent that, to preserve the ability to checkout old commits and build

Not unless they also delete the `fix-serde-std` branch and the `v0.11.1-rev1` tag, but I take your point.

PR #358, opened 2020-08-09
Make slog actually work inside of sgx enclave/impl crates

garbageslam:
When we planned out the way that we build enclaves, one of the main
questions was "how do we unit test the mission critical code that
lives in enclaves". We split this logic into `_impl` crates that
can compile as std, without having any sgx-only bits inside them.

Currently, none of these crates support logging, because the
`mc-common::logger` module is std-only. We built an `mc-sgx-slog`
interface which sends slog structures logs over an OCALL to untrusted
to be logged on the global logger. But because this code was
enclave-specific, there was no way to actually use this in the
enclave/impl crates without breaking the tests. And so `mc-sgx-slog`
has been basically unused. None of the `_impl` crates use it, and none
of the tests even use the `[test_with_logger]` functionality, because
they can't actually use the logger in the code under test.

After this revision we make the following changes:

- `mc-common` depends unconditionally on `slog`, and configures it
  in a `no_std` mode.
- `mc-common` unconditionally re-exports `mc-common::logger`, giving
  the `mc-common::logger::log` and `mc-common::logger::slog` parts,
  because these are portable to the enclaves.
- Everything that depends on `slog` now gets it from the `mc-common`
  re-export instead of duplicating the dependency in its cargo toml.
- `slog` is patched to fix minor problems in its no_std build when
  `dynamic_keys` feature is turned on.
- Functionality in `mc-common::logger` that depends on `slog_scope`
  is gated on `log` feature as before, and this requires `std`.
- Functionality in `mc-common::logger` that depends on `loggers`
  feature is still gated on `loggers` as before.
- The `mc-sgx-slog::untrusted` module is moved to the `mc-sgx-urts`
  crate with the other logging related stuff. This also means that
  some times that a server needs to depend `mc-sgx-slog` "only for linkage"
  are removed, because it gets the needed code from `mc-sgx-urts` now.
- `mc-sgx-urts` now sends `eprintln` and `panic` messages to
  `global_log` instead of `eprintln`.
- `consensus-enclave-impl` owns a `Logger` object where it sends log
   messages
- Tests in `consensus-enclave-impl` use `test_with_logger` and build
  the enclave using the test logger, to demonstrate that the functionality
  works.

The biggest drawback in this commit is that it makes the `mc-common::logger`
module messy, but this module was already messy and was re-exporting
a lot of complicated and variously portable stuff from `slog-*` crates. The alternative
here is that everything starts depending on my patched slog, and if they don't all
get the right version then the build breaks. In this commit, we have tried to keep
that module as nice as possible without breaking the API that it defined.

We propose to consider changes to the API of the `mc-common::logger` module
in another commit, because that will be onerous and will require touching
almost the entire project, anything that uses the logger. By scoping this
commit so that we don't change the API of `mc-common` we make it possible
to land it and unblock other work that needs logging in enclaves.

It seems that the logger module needs to be redesigned with the idea that
there are three different components that different consumers need:
- `slog` itself, which is totally portable
- `slog_scope`, which is not portable to enclave, depends on `std` and a
  quite complex `ArcSwap` crate. This is used to manage the lifetime
  of the global logger instance. It also depends on thread local vectors of
  logger pointers, to manage the scope loggers. This stuff is no_std
  compatible only in nightly compiler because it relies on thread_local.
  But we already have several pieces of code that are nightly only.
  IMO We should consider whether
  to patch this crate to make it `no_std` compatible, which would
  simplify stuff somewhat. We would have to get rid of ArcSwap but
  that might be for the best anyways, it's difficult for me to see
  why we need all that, we could require that `set_global_logger`
  is only called once and not allow swapping out global logger on
  the fly. I have never seen a legitimate use-case for that in any project.
  But it's not for sure that we even need
  the `global_log` API in enclaves, we can probably get by without it.
- `sentry` etc., which depends on `std` and an http stack, which we
  cannot compile with `libmobilecoin` or for the phones.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -111,6 +111,7 @@ rpath = true
 [patch.crates-io]
 prost = { git = "https://github.com/danburkert/prost", rev = "6113789f70b69709820becba4242824b4fb3ffec" }
 prost-derive = { git = "https://github.com/danburkert/prost", rev = "6113789f70b69709820becba4242824b4fb3ffec" }
+slog = { git = "https://github.com/garbageslam/slog", rev = "cc1562258b81353426d17a55b7447256fa6eabbf" }
 dialoguer = { git = "https://github.com/mitsuhiko/dialoguer", rev = "a0c6c1e" }
 console = { git = "https://github.com/mitsuhiko/console", rev = "1307823" }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -109,8 +109,13 @@ overflow-checks = false
 rpath = true
 
 [patch.crates-io]
+# prost is patched with no_std support (https://github.com/danburkert/prost/pull/319)
+# current revision is from jun 13 2020, waiting for a new prost release
+# https://github.com/danburkert/prost/issues/329
 prost = { git = "https://github.com/danburkert/prost", rev = "6113789f70b69709820becba4242824b4fb3ffec" }
 prost-derive = { git = "https://github.com/danburkert/prost", rev = "6113789f70b69709820becba4242824b4fb3ffec" }
+# slog is patched in order to build in no_std config with dynamic_keys feature on
+# https://github.com/slog-rs/slog/pull/268
 slog = { git = "https://github.com/garbageslam/slog", rev = "cc1562258b81353426d17a55b7447256fa6eabbf" }
 dialoguer = { git = "https://github.com/mitsuhiko/dialoguer", rev = "a0c6c1e" }
 console = { git = "https://github.com/mitsuhiko/console", rev = "1307823" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -116,7 +116,7 @@ prost = { git = "https://github.com/danburkert/prost", rev = "6113789f70b6970982
 prost-derive = { git = "https://github.com/danburkert/prost", rev = "6113789f70b69709820becba4242824b4fb3ffec" }
 # slog is patched in order to build in no_std config with dynamic_keys feature on
 # https://github.com/slog-rs/slog/pull/268
-slog = { git = "https://github.com/garbageslam/slog", rev = "cc1562258b81353426d17a55b7447256fa6eabbf" }
+slog = { git = "https://github.com/mobilecoinofficial/slog", rev = "cc1562258b81353426d17a55b7447256fa6eabbf" }
 dialoguer = { git = "https://github.com/mitsuhiko/dialoguer", rev = "a0c6c1e" }
 console = { git = "https://github.com/mitsuhiko/console", rev = "1307823" }
 
--- a/common/Cargo.toml
+++ b/common/Cargo.toml
@@ -14,7 +14,6 @@ log = [
     "backtrace",
     "chrono",
     "mc-util-logger-macros",
-    "slog",
     "slog-scope",
 ]
 loggers = [
@@ -58,7 +57,7 @@ siphasher = "0.3"
 backtrace = { version = "0.3", optional = true }
 chrono = { version = "0.4", optional = true }
 mc-util-logger-macros = { path = "../util/logger-macros", optional = true }
-slog = { version = "2.5", features = ["dynamic-keys", "max_level_trace", "release_max_level_trace"], optional = true }
+slog = { version = "2.6", default-features = false, features = ["dynamic-keys", "max_level_trace", "release_max_level_trace"] }
 slog-scope = { version = "4.1.2", optional = true }
 
 # loggers-only dependencies
--- a/common/README.md
+++ b/common/README.md
@@ -6,5 +6,14 @@ This crate contains several things right now:
   Putting the structs in `common` instead of in a larger crate allows to break
   dependencies, and minimize the amount of code that compiles as `std` and `no_std`.
 - Common error types, used at API boundaries. The rationale is similar.
-- A shared hashmap object based on hashbrown that works in and out of the enclave.
-- Logging functionality
+- A hashmap object based on hashbrown that works in and out of the enclave.
+- A simple LRU cache implementation
+- Logging functionality. Some of this is enclave compatible, some isn't,
+  this is controlled by `log` and `loggers` features.
+
+Note (enclave logging)
+----------------------
+
+The `mc-sgx-slog` crate provides an `slog::Logger` appropriate for the enclave.
+It is recommended to refer to `mc_common::logger::Logger` in the portable `enclave-impl` crates,
+and put the calls to `mc-sgx-slog` in the `enclave-trusted` crates.
--- a/common/README.md
+++ b/common/README.md
@@ -6,9 +6,5 @@ This crate contains several things right now:
   Putting the structs in `common` instead of in a larger crate allows to break
   dependencies, and minimize the amount of code that compiles as `std` and `no_std`.
 - Common error types, used at API boundaries. The rationale is similar.
-- Forward declarations of functionality that is "common" to both inside and outside
-  the enclave. For instance, `mchash` and `mc-util-serial` types are forwarded from the
-  `common` crate. This avoids needing to list these dependencies explicitly in
-  all of the high level targets that use them.
-- Constants associated to attestation. These appear in `ias_settings` module.
-- Utility functions such as `for_each_set_bit`, which appears in `bits` module.
+- A shared hashmap object based on hashbrown that works in and out of the enclave.
+- Logging functionality
--- a/sgx/urts/src/eprintln.rs
+++ b/sgx/urts/src/eprintln.rs
@@ -3,7 +3,14 @@
 // Mobilenode-side support for sgx_debug crate impl
 //
 // Allows dev-build debugging messages similar to eprintln!
+// from within the enclave.
+//
+// We print them to slog because that is our logging infrastructure
+//
+// Note: This is deprecated and the eprintln should go away, the logger
+// API is what you should use.
 
+use mc_common::logger::global_log;
 use std::str;
 
 #[no_mangle]
@@ -11,10 +18,11 @@ pub extern "C" fn eprintln_message(msg: *const u8, msg_len: usize) {
     let msg_bytes = unsafe { std::slice::from_raw_parts(msg, msg_len) };
 
     match str::from_utf8(msg_bytes) {
-        Ok(v) => eprintln!("Enclave log: {}", v),
-        Err(e) => eprintln!(
-            "Enclave log message contained invalid utf8:\n{}\n{:?}",
-            e, msg_bytes
+        Ok(v) => global_log::info!("Enclave eprintln: {}", v),
+        Err(e) => global_log::info!(
+            "Enclave eprintln message contained invalid utf8:\n{}\n{:?}",
+            e,
+            msg_bytes
         ),
     }
 }
--- a/sgx/urts/src/panic.rs
+++ b/sgx/urts/src/panic.rs
@@ -2,9 +2,10 @@
 
 // Mobilenode-side support for sgx_panic crate impl
 //
-// If the enclave panics, it will pass us a message, which we write to stderr
+// If the enclave panics, it will pass us a message, which we log as critical.
 // The enclave is expected to call rsgx_abort itself after this
 
+use mc_common::logger::global_log;
 use std::str;
 
 #[no_mangle]
@@ -12,10 +13,11 @@ pub extern "C" fn report_panic_message(msg: *const u8, msg_len: usize) {
     let panic_msg_bytes = unsafe { std::slice::from_raw_parts(msg, msg_len) };
 
     match str::from_utf8(panic_msg_bytes) {
-        Ok(v) => eprintln!("Enclave panic:\n{}\n", v),
-        Err(e) => eprintln!(
+        Ok(v) => global_log::crit!("Enclave panic:\n{}\n", v),
+        Err(e) => global_log::crit!(
             "Enclave panic message contained invalid utf8:\n{}\n{:?}",
-            e, panic_msg_bytes
+            e,
+            panic_msg_bytes
         ),
     }
 }
--- a/sgx/urts/src/slog.rs
+++ b/sgx/urts/src/slog.rs
@@ -1,8 +1,15 @@
 // Copyright (c) 2018-2020 MobileCoin Inc.
 
-use crate::common::EnclaveLogMessage;
+//! Backend which forwards EnclaveLogMessages to slog with appropriate source
+//! information
+
+use mc_common::logger::{
+    global_log, slog,
+    slog::{Level, Record, RecordLocation, RecordStatic},
+    slog_scope,
+};
+use mc_sgx_slog::EnclaveLogMessage;
 use prost::Message;
-use slog::{Level, Record, RecordLocation, RecordStatic};
 use std::slice;
 
 static ENCLAVE_SLOG_LOCATION: RecordLocation = RecordLocation {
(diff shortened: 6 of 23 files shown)


garbageslam:
Here is PR in slog: https://github.com/slog-rs/slog/pull/268

eranrund:
@garbageslam it is important to keep the logger swapping capability since that is the only way to change the log level, which is a thing we need and use when things start misbehaving. See https://github.com/mobilecoinofficial/mobilecoin/blob/master/util/grpc/src/admin_service.rs#L120

An alternative to that would be to figure out a different way for getting that change to apply, but would likely require forking slog or the crate that has the filtering code in it.

garbageslam:
idk why they dont just stick the log level in an atomic variable so that it can be mutated without locks, that is what i would have done

eranrund:
> idk why they dont just stick the log level in an atomic variable so that it can be mutated without locks, that is what i would have done

Because you don't want to parse that string every time you emit a lot message. They parse it into a bunch of structs if I recall correctly that make rule matching much faster.

garbageslam:
it seems okay to just keep using `slog_scope` if it is working fine for us now, i dont think we need it in the enclave -- there is not all that much global_log in our code base

garbageslam:
> Because you don't want to parse that string every time you emit a log message. They parse it into a bunch of structs if I recall correctly that make rule matching much faster.

which string though? in glog the log level is ultimately a number, and that u32 or whatever can be atomic

Cargo.toml
```diff
@@ -111,6 +111,7 @@ rpath = true
 [patch.crates-io]
 prost = { git = "https://github.com/danburkert/prost", rev = "6113789f70b69709820becba4242824b4fb3ffec" }
 prost-derive = { git = "https://github.com/danburkert/prost", rev = "6113789f70b69709820becba4242824b4fb3ffec" }
+slog = { git = "https://github.com/garbageslam/slog", rev = "cc1562258b81353426d17a55b7447256fa6eabbf" }
```
eranrund, line 114: Please include a comment explaining why we are patching this. I'd like us to move to a pattern where it is more obvious just by looking at this file, versus having to go and decipher what mandated that from the commit log.

common/README.md
```diff
@@ -6,9 +6,5 @@ This crate contains several things right now:
   Putting the structs in `common` instead of in a larger crate allows to break
   dependencies, and minimize the amount of code that compiles as `std` and `no_std`.
 - Common error types, used at API boundaries. The rationale is similar.
-- Forward declarations of functionality that is "common" to both inside and outside
-  the enclave. For instance, `mchash` and `mc-util-serial` types are forwarded from the
-  `common` crate. This avoids needing to list these dependencies explicitly in
-  all of the high level targets that use them.
-- Constants associated to attestation. These appear in `ias_settings` module.
-- Utility functions such as `for_each_set_bit`, which appears in `bits` module.
+- A shared hashmap object based on hashbrown that works in and out of the enclave.
```
eranrund, line 9: shared hashmap?

sgx/urts/src/panic.rs
```diff
@@ -2,20 +2,22 @@
...
 use std::str;
 
 #[no_mangle]
 pub extern "C" fn report_panic_message(msg: *const u8, msg_len: usize) {
     let panic_msg_bytes = unsafe { std::slice::from_raw_parts(msg, msg_len) };
 
     match str::from_utf8(panic_msg_bytes) {
-        Ok(v) => eprintln!("Enclave panic:\n{}\n", v),
-        Err(e) => eprintln!(
+        Ok(v) => global_log::crit!("Enclave panic:\n{}\n", v),
```
eranrund, line 16: How does this work? global_log uses slog_scope, which I believe you say cannot be linked into an enclave?

sgx/urts/src/slog.rs
```diff
@@ -1,9 +1,17 @@
...
+use mc_common::logger::{
+    global_log, slog,
+    slog::{Level, Record, RecordLocation, RecordStatic},
+    slog_scope,
+};
+use mc_sgx_slog::EnclaveLogMessage;
 use prost::Message;
-use slog::{Level, Record, RecordLocation, RecordStatic};
 use std::slice;
+//use slog::{Level, Record, RecordLocation, RecordStatic};
```
eranrund, line 14: Leftover

sgx/urts/src/eprintln.rs
```diff
@@ -3,18 +3,26 @@
...
 #[no_mangle]
 pub extern "C" fn eprintln_message(msg: *const u8, msg_len: usize) {
     let msg_bytes = unsafe { std::slice::from_raw_parts(msg, msg_len) };
 
     match str::from_utf8(msg_bytes) {
-        Ok(v) => eprintln!("Enclave log: {}", v),
-        Err(e) => eprintln!(
-            "Enclave log message contained invalid utf8:\n{}\n{:?}",
-            e, msg_bytes
+        Ok(v) => global_log::info!("Enclave eprintln: {}", v),
```
eranrund, line 21: How does global log works inside an enclave?

eranrund:
> > Because you don't want to parse that string every time you emit a log message. They parse it into a bunch of structs if I recall correctly that make rule matching much faster.
> 
> which string though? in glog the log level is ultimately a number, and that u32 or whatever can be atomic
> 
> i agree you dont want to parse env every log message

Its this stuff https://github.com/slog-rs/envlogger
The log level is not only global, its per module. We can have INFO globally, but then override scp, for example, to have DEBUG.

common/README.md
garbageslam: yeah this is bad phrasing, i was trying to say that it is shared between enclave and non-enclave code

Cargo.toml
garbageslam: yeah okay

garbageslam:
> The log level is not only global, its per module. We can have INFO globally, but then override scp, for example, to have DEBUG.

i see, i didn't realize that. does that overriding happen within the global logger object? i guess i can go read the code and investigate

sgx/urts/src/panic.rs
eranrund: I see, yeah, I got confused.

eranrund:
> > The log level is not only global, its per module. We can have INFO globally, but then override scp, for example, to have DEBUG.
> 
> i see, i didn't realize that. does that overriding happen within the global logger object? i guess i can go read the code and investigate

Sorry I don't understand the question. The issue is that there's no way to get envlogger to reparse the env variable and the solution we came up with is to recreate the logger chain, which creates a new instance of envlogger after the env var is changed. This doesn't affect only the global logger but also the specific Logger instances, as they all end up logging to the same sink.

garbageslam:
Here's slog-scope at current master `4.3.0`

https://github.com/slog-rs/scope/blob/7c93a607ebf14567840331228da0a7f19d1fcfa5/lib.rs#L64

- there is no `no_std`, and most stuff is grabbed from `std` namespace, but stuff like `std::sync::Arc` could be grabbed from `alloc` so that's not a big deal
- `std::thread_local!` macro is used. there is no core-only version of that thing. there is an unstable `[thread_local]` attribute which we are using in `mcrand` crate. so we could probably port this to be like that
- for the global logger, arcswap crate is used. that crate is pretty complicated

Here's arcswap crate at current master:

https://github.com/vorner/arc-swap/blob/3ea27a0580df026c97a6f37e9ce6b0be01729c64/src/lib.rs#L1070

I don't know what the enclave analogue of `std::thread::yield_now()` is, i don't think there is one, there's also some references to `std::process`. So it seems that they have made liberal use of `std`-only synchronization primitives, and they end up with like thousands of lines of code. It seems to me like this will be a lot harder to port to a core / alloc only config.

I guess there is discussion about design of arcswap here: https://github.com/crossbeam-rs/crossbeam/issues/160

eranrund:
I don't think we need slog-scope in the enclave. We should avoid global logger usage there.

Cargo.toml
garbageslam: done

sgx/urts/src/eprintln.rs
kylefleming, line 21: What's the impetus for using `global_log` instead of `println`? Doesn't `println` already get redirected to slog anyways / isn't `println` basically already a global log of sorts? (I don't think moving to `global_log` is wrong, I'm just wondering what the motivation is)

common/Cargo.toml
```diff
@@ -58,7 +57,7 @@ siphasher = "0.3"
 backtrace = { version = "0.3", optional = true }
 chrono = { version = "0.4", optional = true }
 mc-util-logger-macros = { path = "../util/logger-macros", optional = true }
-slog = { version = "2.5", features = ["dynamic-keys", "max_level_trace", "release_max_level_trace"], optional = true }
+slog = { version = "2.6", default-features = false, features = ["dynamic-keys", "max_level_trace", "release_max_level_trace"] }
```
kylefleming, line 60: Do all of these features (`dynamic-keys`, `max_level_trace`, `release_max_level_trace`) need to be enabled by default or should they be enabled in the `log` feature?

kylefleming, line 60: We should probably rename the `log` feature (either now or later) since it no longer enables log anymore (since `slog` is now enabled regardless).

sgx/urts/src/slog.rs
```diff
@@ -13,12 +20,10 @@ static ENCLAVE_SLOG_LOCATION: RecordLocation = RecordLocation {
     module: "<enclave>",
 };
 
-/// # Safety
-///
-/// This function is marked unsafe due to receiving a raw pointer from the caller.
+// This function is unsafe, if the caller misuses the API it can cause undefined behavior
```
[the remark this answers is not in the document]
garbageslam: 1. when you mark a whole function `unsafe`, then there are no `unsafe` blocks in its impl. by making the function not explicitly unsafe, it allows us to use `unsafe` scopes inside what is mostly safe code, which seems good
2. it feels redundant to say `extern "C" unsafe`
3. All our other ocalls follow this pattern

sgx/urts/src/eprintln.rs
garbageslam: the impetus is that right now, when an enclave in the discovery container does `eprintln`, the logs for the container dont show me what server it even was, while `global_log` does. so from practical point of view it makes it less confusing

additionally I think we should treat those logs as part of the server log and make sure they follow the same path as everything else.

fwiw i think we should kill the eprintln thing now that we have `log::info!` working in enclave

common/Cargo.toml
garbageslam: i think it may be confusing if `slog` is configured one way in the enclave and a different way in the logging sync that handles the serialized slog logs outside the enclave.

i suspect that the build will break if `dynamic-keys` is set differently, and i can't easily predict what will happen if `max_level_trace` or `max_release_level_trace` is different in different places

garbageslam: i mean, it ensures the same functionality is available that it did before, including an `std` dependency

idk i would like to not break the api of the crate right now if possible

sgx/urts/src/slog.rs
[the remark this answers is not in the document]
garbageslam: this rfc is also relevant -- they want to make it so that `unsafe fn` (function wtih unsafe prototype) still requires `unsafe { ` blocks in its implementation to highlight which parts of its implementations are safe and unsafe

http://rust-lang.github.io/rfcs/2585-unsafe-block-in-unsafe-fn.html

in current versions of the language you can't mark the prototype unsafe without marking the entire implementation unsafe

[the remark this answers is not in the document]
kylefleming: > 1. when you mark a whole function `unsafe`, then there are no `unsafe` blocks in its impl. by making the function not explicitly unsafe, it allows us to use `unsafe` scopes inside what is mostly safe code, which seems good, since less code overall is permitted to do unsafe things, which may ease review

Marking a function `unsafe` affects both the using of unsafe code within the function, as well as whether safe code can call it or not. I agree that the function is mostly safe and by marking the whole function unsafe, it's making it hard to see at a glace which parts of the function are the unsafe bits. However, if we want to separate out the unsafe bits from the safe bits, we should split it up into multiple functions.

The _most important_ reason for marking a function as `unsafe` is so that the compiler emits an error if safe code tries to call it. By marking the function as safe, we'd effectively be opening up a hole in Rust's safety paradigm. Safe code could call this function and could pass it arguments that cause the function to perform unsafe memory access.

As I understand it, marking this function as safe would be a violation of Rust's safety model.

> 2. it feels redundant to say `extern "C" unsafe`

It's not. `extern "C"` refers to symbol mangling and how the symbol is exported. `extern "C"` functions are often also unsafe because they need to access C-compatible arguments, which often means unsafe memory access, but in terms of the Rust language, `extern "C"` does not imply `unsafe`.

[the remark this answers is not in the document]
garbageslam: > The most important reason for marking a function as unsafe is so that the compiler emits an error if safe code tries to call it.

yes but this won't happen when it is extern "C", it is going to be called by C code.

> As I understand it, marking this function as safe would be a violation of Rust's safety model.

Rust's safety model already has a lot of holes, i don't think what we are doing here is a serious risk. If you think it is, then you should open a ticket so that we can review all of our OCALL signatures, i'm just following our established patterns. It's out of scope for this PR to find and change all the rust-side OCALL signatures.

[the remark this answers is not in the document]
garbageslam: > The most important reason for marking a function as unsafe is so that the compiler emits an error if safe code tries to call it. 

I disagree with this, because of the issue mentioned in the RFC. It is easier to review this code when less of it is inside of an unsafe block. It is standard best practice to make unsafe scopes as small as possible, for obvious reasons. What I did in this PR is apply that guideline.

common/Cargo.toml
kylefleming: Yea, that makes sense. I'm mainly just pointing out that the `log` feature no longer enables logging since it will now have already been enabled. Doesn't need to be addressed right now though.

sgx/urts/src/slog.rs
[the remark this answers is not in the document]
garbageslam: basically, in this code, because of the constraints of the language, we have to choose between either
(1) the compiler forcing the rust caller to use "unsafe" appropriately, or
(2) the compiler forcing us to use `unsafe` appropriately inside the function body.

because there is no rust caller, the second is clearly preferable.

I hope they implement the RFC soon.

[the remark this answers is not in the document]
jcape: I think @kylefleming is right here: the function should be unsafe, it's an FFI boundary, and it's manipulating raw pointers.

You can have a safe function this chains to which does all the real bits using the slice, but the function here should be unsafe.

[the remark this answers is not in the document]
garbageslam: Yeah, okay.

Do you remember if there are any other ocalls in the system besides in this crate? It was too long ago that I worked on that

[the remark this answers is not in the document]
garbageslam: I guess i'm adjusting my stance, i'm going to try to fix it as suggested in all instances in this crate in this PR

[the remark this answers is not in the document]
kylefleming: > I guess i'm adjusting my stance, i'm going to try to fix it as suggested in all instances in this crate in this PR

I think it would easier to code review if it was a seperate PR.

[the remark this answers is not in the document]
garbageslam: ok, i will do it in a separate PR, thanks

PR #361, opened 2020-08-10
Trait and Mock for Slot

mfaulk:
### Motivation

Currently, it has been difficult to test Node because some of its behavior requires carefully-constructed SCP messages which it passes to Slot. It would be easier to test if Slot could be mocked.

### In this PR
* Creates a Slot trait, and uses automock to generate mocks for testing.
* Adds some simple unit tests to Node that show mocking in action.

### Future Work
This PR leaves several parts of Node untested because those parts are likely to change soon.

### DIFF
--- a/consensus/scp/src/node.rs
+++ b/consensus/scp/src/node.rs
@@ -147,8 +147,8 @@ pub trait ScpNode<V: Value>: Send {
     /// Get metrics for a specific slot.
     fn get_slot_metrics(&mut self, slot_index: SlotIndex) -> Option<SlotMetrics>;
 
-    /// Get the slot internal state (for debug purposes).
-    fn get_slot_state(&mut self, slot_index: SlotIndex) -> Option<String>;
+    /// Additional debug info, e.g. a JSON representation of the Slot's state.
+    fn get_slot_debug_snapshot(&mut self, slot_index: SlotIndex) -> Option<String>;
 
     /// Clear the list of pending slots. This is useful if the user of this object realizes they
     /// have fallen behind their peers, and as such they want to abort processing of current slots.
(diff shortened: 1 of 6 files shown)


consensus/scp/src/node.rs
```diff
@@ -29,7 +28,7 @@ pub struct Node<V: Value, ValidationError: Display> {
     pub Q: QuorumSet,
 
     /// Map of last few slot indexes -> slots.
-    pub pending: LruCache<SlotIndex, Slot<V, ValidationError>>,
+    pub pending: LruCache<SlotIndex, Box<dyn ScpSlot<V>>>,
```
mfaulk, line 31: This is one of the main changes.

consensus/scp/src/node.rs
```diff
@@ -152,14 +148,14 @@ pub trait ScpNode<V: Value>: Send {
     fn get_slot_metrics(&mut self, slot_index: SlotIndex) -> Option<SlotMetrics>;
 
     /// Get the slot internal state (for debug purposes).
-    fn get_slot_state(&mut self, slot_index: SlotIndex) -> Option<SlotState<V>>;
+    fn get_slot_state(&mut self, slot_index: SlotIndex) -> Option<String>;
```
mfaulk, line 151: Previously, this returned a SlotState, which was serialized to JSON and logged. That created an implicit dependency between Node and Slot. Here, the conversion to JSON has been pushed down so that Node is not aware of SlotSate.

consensus/scp/src/slot.rs
```diff
@@ -43,6 +47,25 @@ pub enum Phase {
     Externalize,
 }
 
+/// A Single slot of the SCP protocol.
+#[cfg_attr(test, automock)]
+pub trait ScpSlot<V: Value>: Send {
```
mfaulk, line 52: This is the new trait for Slot.

consensus/scp/src/node.rs
rjwalters: maybe `get_slot_debug_snapshot()`?

PR #362, opened 2020-08-11
Split byzantine ledger

mfaulk:
### Motivation

The byzantine_ledger.rs file is large and complex, and it is difficult to see which parts have test coverage.

### In this PR
*  In preparation for adding more tests, this PR splits byzantine_ledger.rs into several smaller files. I think that makes it easier to grok the big picture, navigate the code, and see which parts do/don't have tests.

### Future Work
* Writing more tests


rjwalters:
It's a bit hard to see what changed... is it possible to refactor the PR as `consensus/service/src/byzantine_ledger.rs → mod.rs` with a new file for the `worker`?

consensus/service/src/byzantine_ledger/mod.rs
```diff
@@ -0,0 +1,617 @@
...
+impl ByzantineLedger {
+    pub fn new<
+        PC: BlockchainConnection + ConsensusConnection + 'static,
+        L: Ledger + Sync + 'static,
+    >(
+        node_id: NodeID,
+        quorum_set: QuorumSet,
+        peer_manager: ConnectionManager<PC>,
+        ledger: L,
+        tx_manager: Arc<Mutex<Box<dyn TxManager>>>,
```
garbageslam, line 60: So, I'm surprised to still see the `Arc<Mutex>` being used on the `TxManager` given that that wasn't necessary before this revision and it caused performance problems before.

Normally to accomplish dependency injection, you only need `Box<dyn TxManager>`.
If we end up with extra mutexes that we didn't need before, that seems like a bad thing that will make it harder to reason about the system and its performance. And most of the function on the TxManager are thread-safe, so this is a significant pessimization.

You could try something with RwLock if you really need this?

Can you speak to why `Arc<Mutex>` is needed here?

mfaulk: @garbageslam This file has conflicts with master, and the Arc<Mutex> should go away when that is resolved. It may actually be less work to close this PR and open a new one that starts from the current master, so I might do that instead.

PR #368, opened 2020-08-13
Schnorrkel uprev

sugargoat:
Soundtrack of this PR: [DJ Fauci](https://soundcloud.com/dj_fauci/ephem-all)

### Motivation

We we were previously carrying a patch of Schnorrkel in order to pass in our own rng, which enabled us to build in `no_std`. However, as mentioned in https://github.com/w3f/schnorrkel/pull/55, we can use attach_rng instead. This requires us to use the Merlin Transcript directly, as opposed to the signing_context, which is a wrapper around a Transcript. This is fine, as the signing context functionality is preserved in this revision. 

### In this PR
* Remove patch to schnorrkel and uprev to "0.9"
* Add context_tag to digest as well, addressing remaining FIXME

[MC-1619](https://mobilecoin.atlassian.net/browse/MC-1619), [MC-1761](https://mobilecoin.atlassian.net/browse/MC-1761)


crypto/sig/src/lib.rs
```diff
@@ -25,13 +26,9 @@ pub use schnorrkel::{Signature, SignatureError, SIGNATURE_LENGTH};
...
 pub fn sign(context_tag: &[u8], private_key: &RistrettoPrivate, message: &[u8]) -> Signature {
-    // Nonce is hash( private_key || message )
-    // FIXME: We should probably hash the context_tag in as well.
-    // Or just use something like merlin instead of Blake2b256.
-    // In that case we make the assumption that Keccak, which underlies STROBE,
-    // is a strong Pseudorandom permutation, and that consequently Merlin with
-    // a partially secret input is a PRF.
+    // Nonce is hash( context_tag || private_key || message )
     let mut hasher = Blake2b256::new();
+    hasher.input(context_tag);
```
garbageslam, line 31: I think it would be a good idea to add as a prefix, the length of the `context_tag` as well, since it is variable length
that prevents situations where `"", private_key1 "privatekey1 foo"` has the same nonce as a signature from another context`"privatekey1", privatekey1, "foo")`

garbageslam: (i dont think we need framing around the message because it is the last thing in the sequence)

garbageslam: i think this kind of framing is what merlin does: https://merlin.cool/transcript/ops.html#appending-messages

garbageslam:
LGTM otherwise!

crypto/sig/src/lib.rs
isislovecruft: I agree that this is a good idea. For simplicity's sake, you probably just want to use merlin again for the nonce generation here, using `Transcript.build_rng()`, so that you're getting identical behaviour w.r.t. domain separation as you are in the Schnorrkel signature (or, at least, no worse, since Schnorrkel is also using merlin).

crypto/sig/README.md
```diff
@@ -77,10 +75,13 @@ The [ed25519 manuscript]((http://ed25519.cr.yp.to/ed25519-20110926.pdf)) from 20
...
 > but we prefer to reuse `H` to save area in hardware implementations.
 
-We point out that Blake was one of the SHA-3 finalists, and was also explicitly designed to model a PRF.
-We prefer Blake2b here because it reduces the number of different hash functions in our system overall.
+We point out that Merlin Transcripts implement a subset of the STROBE protocol, which states in [STROBE Protocol Framework](https://strobe.sourceforge.io/),
 
-Assuming the Blake2b has the secret-prefix-PRF property, we can say that the signatures created this way
+>  Strobe is based on SHA-3, or rather Keccak-f and cSHAKE (draft NIST SP 800-185).
+
+We prefer using Merlin Transcripts here for simplicity, as we are constructing two transcripts in the course of constructing the signature, and because it reduces the number of different hash functions in our system overall.
```
garbageslam, line 82: suggestion:

"We prefer using Merlin transcripts to produce the nonce here, instead of a cryptographic hash function. Schnorrkel already relies on Merlin -- it has several benefits, like automatic domain separation and framing. By using this here instead of a hash function like SHA3, we can reduce the total number of cryptographic assumptions underpinning this signature scheme."

crypto/sig/README.md
```diff
@@ -77,10 +75,13 @@ The [ed25519 manuscript]((http://ed25519.cr.yp.to/ed25519-20110926.pdf)) from 20
...
+
+Assuming the Merlin `challenge_bytes` has the secret-prefix-PRF property, we can say that the signatures created this way
```
garbageslam, line 84: When using Merlin for this, instead of assuming that a hash function has the secret-prefix-PRF property, we assume that the following pseudo-rust function does, for any particular merlin transcript, where the private_key argument is identified with the secret prefix:

```
fn produce_nonce(transcript, private_key, message) -> [u8; 32]
    transcript.append_message("private", private_key);
    transcript.append_message("message", message);
    let mut nonce = [0u8; 32];
    transcript.challenge_bytes("nonce", &mut nonce);
    nonce
}
```

This assumption can be justified if we believe that the STROBE "PRF" operation functions as a PRF once STROBE has been keyed. We refer the reader to https://strobe.sourceforge.io/papers/strobe-20170130.pdf for a discussion of the usage and security properties of STROBE. For more discussion, see also the documentation around Merlin: https://merlin.cool/transcript/ops.html.

With this assumption in hand, we can say that signatures created this way are hard to distinguish from signatures ...

crypto/sig/src/lib.rs
```diff
@@ -25,16 +24,12 @@ pub use schnorrkel::{Signature, SignatureError, SIGNATURE_LENGTH};
...
-    let mut hasher = Blake2b256::new();
-    hasher.input(private_key.to_bytes());
-    hasher.input(message);
-    let nonce = hasher.result();
+    let mut transcript = Transcript::new(b"SigningNonce");
+    transcript.append_message(b"context", &context_tag);
+    transcript.append_message(b"private", &private_key.to_bytes());
+    transcript.append_message(b"message", &message);
+    let mut nonce = [0u8; 32];
+    transcript.challenge_bytes(b"nonce", &mut nonce);
```
garbageslam, line 32: it might be a little easier to read if we scoped this transcript, like

```
let nonce = {
    let mut transcript = Transcript::new(b"SigningNonce");
    ...
    let mut nonce = [0u8;32];
    transcript.challenge_bytes(b"nonce", &mut nonce);
};
```
just so that there's no confusion between this transcript and the later transcript?

crypto/sig/README.md
sugargoat: Excellent - ty!

crypto/sig/src/lib.rs
sugargoat: 👍

PR #372, opened 2020-08-18
Remove Ledger from TxManager

mfaulk:
Soundtrack of this PR: [Tycho](https://www.youtube.com/watch?v=dKh1yvmZpgU)

### Motivation

Components are easier to test if they have fewer sub-components. Previously, TxManager held a Ledger instance, but only used it in one place to query for the previous block.

### In this PR
* Removes the Ledger instance from TxManager
* TxManager::tx_hashes_to_block now takes a `parent_block` instead of querying the ledger for it.

### Future Work
* This is incremental progress towards simplifying and testing TxManager and the things that use TxManager.

### DIFF
--- a/consensus/service/src/tx_manager.rs
+++ b/consensus/service/src/tx_manager.rs
@@ -14,7 +14,7 @@ use mc_consensus_enclave::{
     WellFormedTxContext,
 };
 use mc_crypto_keys::CompressedRistrettoPublic;
-use mc_ledger_db::{Error as LedgerDbError, Ledger};
+use mc_ledger_db::Error as LedgerDbError;
 use mc_transaction_core::{
     constants::MAX_TRANSACTIONS_PER_BLOCK,
     ring_signature::KeyImage,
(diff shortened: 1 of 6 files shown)


consensus/service/src/tx_manager.rs
```diff
@@ -114,17 +114,10 @@ pub trait UntrustedInterfaces: Clone {
...
-    E: ConsensusEnclaveProxy,
-    L: Ledger,
-    UI: UntrustedInterfaces = crate::validators::DefaultTxManagerUntrustedInterfaces<L>,
-> {
+pub struct TxManager<E: ConsensusEnclaveProxy, UI: UntrustedInterfaces> {
     /// Enclave.
     enclave: E,
 
-    /// Ledger.
-    ledger: L,
```
mfaulk, line 126: Removing this is the main change.

consensus/service/src/tx_manager.rs
```diff
@@ -297,9 +287,9 @@ impl<E: ConsensusEnclaveProxy, L: Ledger, UI: UntrustedInterfaces> TxManager<E,
     pub fn tx_hashes_to_block(
         &self,
         tx_hashes: &[TxHash],
+        parent_block: &Block,
```
mfaulk, line 290: This is the other main change.

PR #373, opened 2020-08-18
[MC-1806] Automocks ScpNode

mfaulk:
Soundtrack of this PR: [more Tycho](https://www.youtube.com/watch?v=PRqKCgT0sQA)

### Motivation

It would be nice to test things like ByzantineLedgerWorker by mocking its dependencies, such as ScpNode.

### In this PR
* Generates a mock ScpNode duriing testing via mockall's `automock`.

### Future Work
* Mock ByzantineLedgerWorker's other dependencies.

### DIFF
--- a/consensus/scp/src/node.rs
+++ b/consensus/scp/src/node.rs
@@ -11,6 +11,8 @@ use mc_common::{
     logger::{log, Logger},
     LruCache, NodeID,
 };
+#[cfg(test)]
+use mockall::*;
 use std::{collections::BTreeSet, fmt::Display, time::Duration};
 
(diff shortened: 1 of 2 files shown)


consensus/scp/src/node.rs
```diff
@@ -118,6 +120,7 @@ impl<V: Value, ValidationError: Display + 'static> Node<V, ValidationError> {
 }
 
 /// A node capable of participating in SCP.
+#[cfg_attr(test, automock)]
```
mfaulk, line 123: This is the main change.

PR #382, opened 2020-08-20
MC-1814 Enclave target features inheritance. #in-review

jcape:
### Motivation

SHA2 instructions are not available on the platforms we currently have Kubernetes running, and there's no way to remove that other than by hard-code changes. This removes most of the hard-coding (it does keep LVI mitigations), and simply copies target features enabled for the untrusted build into the enclave build.

### In this PR
* Remove SHA target feature where it exists (not supported on our platform yet)
* Copy target features from untrusted into enclave build's RUSTFLAGS.
* Remove explicit skylake from individual enclave build.

### DIFF
--- a/util/build/enclave/src/lib.rs
+++ b/util/build/enclave/src/lib.rs
@@ -218,9 +218,27 @@ impl Builder {
             .exec()?;
 
         let mut cargo_builder = CargoBuilder::new(&env, staticlib_dir, false);
+
+        // copy our target features to the enclave's build
+        let features = env.target_features();
+        let mut feature_buf = String::with_capacity(features.len() * 32);
+        feature_buf.push_str("target-feature=+lvi-cfi,+lvi-load-hardening");
+        for feature in features {
+            feature_buf.push(',');
+            feature_buf.push('+');
+            // Cleanup cargo's nonsense.
+            match feature.as_str() {
+                "cmpxchg16b" => feature_buf.push_str("cx16"),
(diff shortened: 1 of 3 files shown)


util/build/enclave/src/lib.rs
```diff
@@ -218,9 +218,27 @@ impl Builder {
...
         let mut cargo_builder = CargoBuilder::new(&env, staticlib_dir, false);
+
+        // copy our target features to the enclave's build
+        let features = env.target_features();
+        let mut feature_buf = String::with_capacity(features.len() * 32);
+        feature_buf.push_str("target-feature=+lvi-cfi,+lvi-load-hardening");
+        for feature in features {
+            feature_buf.push(',');
+            feature_buf.push('+');
+            // Cleanup cargo's nonsense.
```
sugargoat, line 229: lol

util/build/enclave/src/lib.rs
```diff
@@ -218,9 +218,27 @@ impl Builder {
...
+            match feature.as_str() {
+                "cmpxchg16b" => feature_buf.push_str("cx16"),
+                "pclmulqdq" => feature_buf.push_str("pclmul"),
+                "rdrand" => feature_buf.push_str("rdrnd"),
```
garbageslam, line 233: omg... this one was infuriating when i was doing the `mc-crypto-rand` feature config stuff, especially because, referring to one of these cryptic features that happens to be mistyped is not a bug

jcape: IKR!?

PR #384, opened 2020-08-20
improve behavior of local debugging with enclaves in sw mode

cbeck88:
### Motivation

These are some changes that I applied locally when trying to debug stuff at the enclave boundary. on the trusted side
via printf debugging.
Because there is no global logger in the enclave, `eprintln` is pretty much the only logging option for this.
And because the slog async logger can't be flushed, it is useful to ensure that critical messages get out over an additional channel besides the slog logs, otherwise all you get is "enclave SIGILL".

### In this PR

Make enclave panic messages go out on STDERR as well as slog, in debug builds.
Add improved comments about the enclave eprintln functionality.

### Future Work

Improve the slog backend

### DIFF
--- a/sgx/urts/src/eprintln.rs
+++ b/sgx/urts/src/eprintln.rs
@@ -7,8 +7,10 @@
 //
 // We print them to slog because that is our logging infrastructure
 //
-// Note: This is deprecated and the eprintln should go away, the logger
-// API is what you should use.
+// Note: This should really only be used for local debugging of patches.
+// There is no global logger infrastructure in the enclave, so this eprintln
+// is all you get, especially when printf debugging something near the FFI layer
+// where you are unlikely to have a Logger object available.
 
 use mc_common::logger::global_log;
 use std::str;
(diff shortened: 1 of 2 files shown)


garbageslam:
> Why does eprintln survives a crash and the slog thing doesn't (with the addition of a little sleep for flushing)?

Im not sure why, all I know is that it didn't

> Wasn't there a plan to replace eprintln with global_log?

I think in the enclave, we didn't come up with a plan to do global_log, at least not based on slog_scope, because that is not portable to the enclave.

Idk, the eprintln is very effective when I just need to do a little println debugging

There is probably some way to make a shim that would work in the sgx compat crate, so that it works more or less like eprintln but sends a slog payload, but it's not completely straightforward, mainly because of how feature configuration works. If you want we can talk about it offline

remoun:
This still seems useful. Do we want to try and merge it?

jcape:
We can revisit if we want later.

PR #389, opened 2020-08-23
MCC-697 Digestible rework to use merlin

garbageslam:
# Motivation

Currently, hashes of transaction and the blockchain are derived using the `mc-crypto-digestible` crate. This is a trait, plus a proc-macro crate, which allows to bring arbitrary cryptographic hashes ("digest" implementors) to a digestible object and hash it. This essentially combines serialization and hashing into one step. The goals of this were (1) provide stability guarantees that aren't normally present in serialization, (2) provide security guarantees that the hash is not malleable.

When this crate was merged we did not all agree that this was the best way to do this -- there was an argument that we should use merlin instead of basing it on a hash function, because merlin is a framework and offers a higher-level API, and is well integrated with the rest of the dalek ecosystem, particularly the Schnorrkel digital signature crate.

In this PR we pivot to using merlin. One benefit of this switch is that I think it is possible to prove a slightly stronger security property after this revision than before -- we can say that no two distinct ASTs will have the same hash, not just no two instances of the same structure.

But the most interesting thing in this PR is likely that we are here building in a form of support for *schema evolution* which was not present in the initial code.

(3) *Schema evolution* here means that just as in protobuf, we can add new optional fields to structures without breaking compatibility with the old structures. In this case, compatibility means ledger-compatibility. We can add optional fields to thing without changing the hashes of old blocks.

With this PR we will be able to add new Optional members to e.g. TxOut type or BlockContents type without having to figure out how to migrate the ledger, because the hashes of the old blocks won't change as a result of the change. So, in some cases we can add features without having to figure out how to migrate the ledger, just as we normally would when using protobuf.

There will be minor ancillary benefits like, less low-level things will depend on the rust `digest` crate which has seen recent breaking changes. (although we seem to be past the breaking changes now, courtesy of Eran :) )

## In this PR

- Diff is +3000 lines, but the vast majority of that is tests (which become many lines after rustfmt) and readme
- Add substantial test coverage for the portions not related to digestible-derive, the portions related to digestible-derive, and the schema-evolution properties
- Add test-vectors for stability of hashes of origin blocks, tx-outs, test-utils generated block chain

In the tests for this PR, the `inspect_ast` test functionality can be used to show exactly what structure digestible produced for the user-provided type. The AST can be mapped to json and pretty-printed.

The AST is not actually computed in production when we need to digest something -- it is only computed explicitly in debugging tools. However, the AST completely captures the resulting hash -- if we capture the AST for a structure and then compute the hash from the AST, we always get the same hash. (There is quite a lot of test coverage backing up that claim.)

The AST can be displayed as json: the json faithfully represents the entire AST. So, looking at the json can give the reviewer a good sense of what is being captured by the hash.

Here's the AST for a random TxOut:

```
  "elems": [
    {
      "amount": "aggregate",
      "elems": [
        {
          "commitment": "primitive",
          "data": "[242, 152, 40, 146, 146, 42, 240, 192, 35, 99, 79, 54, 220, 236, 108, 181, 213, 143, 214, 94, 153, 71, 189, 181, 22, 77, 172, 211, 204, 73, 63, 3]",
          "type_name": "ristretto"
        },
        {
          "data": "[187, 146, 125, 76, 38, 34, 179, 187]",
          "masked_value": "primitive",
          "type_name": "uint"
        }
      ],
      "name": "Amount"
    },
    {
      "data": "[126, 132, 117, 253, 124, 40, 56, 37, 230, 94, 15, 206, 138, 168, 124, 53, 160, 123, 163, 167, 130, 15, 16, 157, 171, 117, 12, 8, 214, 240, 105, 113]",
      "target_key": "primitive",
      "type_name": "ristretto"
    },
    {
      "data": "[238, 35, 140, 56, 133, 221, 101, 105, 2, 101, 249, 201, 51, 21, 57, 222, 41, 51, 60, 205, 41, 220, 180, 61, 236, 162, 72, 97, 136, 209, 89, 88]",
      "public_key": "primitive",
      "type_name": "ristretto"
    },
    {
      "data": "[165, 234, 17, 98, 158, 70, 82, 80, 174, 54, 135, 135, 107, 141, 215, 81, 56, 69, 162, 101, 225, 237, 184, 140, 10, 219, 123, 17, 207, 187, 210, 165, 66, 245, 105, 183, 136, 31, 101, 233, 86, 191, 35, 6, 42, 81, 44, 59, 246, 77, 18, 141, 90, 67, 169, 243, 2, 56, 101, 10, 2, 55, 212, 240, 173, 69, 76, 136, 133, 105, 181, 252, 11, 22, 207, 187, 106, 149, 108, 150, 246, 214, 15, 106, 92, 189, 148, 227, 237, 135, 44, 121, 190, 52, 189, 0, 67, 68, 68, 248, 208, 231, 145, 194, 4, 77, 47, 49, 238, 63, 73, 195, 126, 92, 230, 175, 136, 76, 47, 237, 65, 50, 107, 139, 225, 12, 0, 0]",
      "e_account_hint": "primitive",
      "type_name": "bytes"
    }
  ],
  "name": "TxOut",
  "test": "aggregate"
}
```
Here's the AST for a random origin block:

```
{
  "elems": [
    {
      "data": "[138, 161, 54, 74, 202, 243, 220, 127, 18, 223, 79, 47, 175, 248, 56, 118, 93, 203, 45, 96, 32, 111, 2, 177, 99, 85, 125, 216, 81, 32, 223, 120]",
      "id": "primitive",
      "type_name": "bytes"
    },
    {
      "data": "[0, 0, 0, 0]",
      "type_name": "uint",
      "version": "primitive"
    },
    {
      "data": "[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]",
      "parent_id": "primitive",
      "type_name": "bytes"
    },
    {
      "data": "[0, 0, 0, 0, 0, 0, 0, 0]",
      "index": "primitive",
      "type_name": "uint"
    },
    {
      "cumulative_txo_count": "primitive",
      "data": "[5, 0, 0, 0, 0, 0, 0, 0]",
      "type_name": "uint"
    },
    {
      "elems": [
        {
          "elems": [
            {
              "data": "[0, 0, 0, 0, 0, 0, 0, 0]",
              "from": "primitive",
              "type_name": "uint"
            },
            {
              "data": "[0, 0, 0, 0, 0, 0, 0, 0]",
              "to": "primitive",
              "type_name": "uint"
            }
          ],
          "name": "Range",
          "range": "aggregate"
        },
        {
          "data": "[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]",
          "hash": "primitive",
          "type_name": "bytes"
        }
      ],
      "name": "TxOutMembershipElement",
      "root_element": "aggregate"
    },
    {
      "contents_hash": "primitive",
      "data": "[142, 234, 222, 1, 66, 175, 228, 166, 18, 94, 58, 74, 199, 168, 43, 52, 175, 1, 126, 90, 222, 153, 120, 59, 217, 187, 215, 47, 43, 213, 93, 67]",
      "type_name": "bytes"
    }
  ],
  "name": "Block",
  "test": "aggregate"
}
```


transaction/core/src/blockchain/block_signature.rs
```diff
@@ -51,10 +49,8 @@ impl BlockSignature {
         block: &Block,
         keypair: &Ed25519Pair,
     ) -> Result<Self, Ed25519SignatureError> {
-        // SHA512 is used for compatibility with Ed25519ph.
```
garbageslam, line 54: @jcape @mfaulk is there some reason that we need this compatibility? we could use SHA512 here via the pseudo-merlin thing if its important

transaction/core/src/membership_proofs/mod.rs
```diff
@@ -30,7 +29,7 @@ lazy_static! {
 pub fn hash_leaf(tx_out: &TxOut) -> [u8; 32] {
     let mut hasher = Blake2b256::new();
     hasher.update(&TXOUT_MERKLE_LEAF_DOMAIN_TAG);
-    tx_out.digest(&mut hasher);
+    hasher.update(&tx_out.hash());
```
garbageslam, line 32: i didn't want to mess around with this for now, we try to make this use merlin instead of blake2b though
i think its fine either way

transaction/core/src/blockchain/block_signature.rs
jcape: > For Ed25519ph, phflag=1 and PH is SHA512 instead.  That is, the input is hashed using SHA-512 before signing with Ed25519.

https://tools.ietf.org/html/rfc8032

jcape:
Won't this break the use of ed25519 as the leaf cert in the fog signing chain?

crypto/keys/src/ed25519.rs
```diff
@@ -469,13 +473,15 @@ mod ed25519_tests {
...
+        let sig = pair
+            .try_sign_digest(hasher.inner)
+            .expect("Failed to sign digest");
 
-        let mut hasher = Sha512::default();
-        data.digest(&mut hasher);
-        pair.verify_digest(hasher, &sig)
+        let mut hasher = PseudoMerlin(Sha512::default());
+        data.append_to_transcript(b"test", &mut hasher);
+        pair.verify_digest(hasher.inner, &sig)
```
garbageslam, line 484: @jcape here's how do you ed25519ph signature after this revision

transaction/core/src/blockchain/block_signature.rs
garbageslam: this doesn't answer my question -- there's nothing wrong with computing a merlin hash, and then signing that using ed25519. there might be a SHA512 computed on the merlin hash value -- that's fine, efficiency is not a concern

jcape: You want to set the phflag=1 on Ed25519 in order to distinguish between a signature over X bytes, and a signature over the hash of X bytes. In Ed25519ph, the requirement is that that pre-hash uses SHA512.

garbageslam: i see, thanks

crypto/digestible/README.md
```diff
@@ -1,175 +1,484 @@
...
+  Instead, as many fields as needed are appended, and then there is a `closer` that
+  is appended to the transcript.
+
+Together this means that:
+- New `Option` fields may be added to existing structures without breaking ledger compatibility.
+- Old fields that were not `Option` may be made optional without breaking ledger compatibility.
+
+Note that fields may not be re-ordered or renamed.
+
+However, in the future, we may support a proc-macro attribute in digestible-derive that allows
```
mfaulk, line 458: Cool, that would be useful.

Tangentially, the Jackson JSON processor in Java might be a good source of inspiration. That supports a variety of ways to annotate a class in order to customize how it gets serialized, e.g. https://www.baeldung.com/jackson-annotations

crypto/digestible/README.md
```diff
@@ -1,175 +1,484 @@
 mc-crypto-digestible
```
mfaulk, line 1: Excellent docs! Thank you!

crypto/digestible/derive/src/lib.rs
```diff
@@ -40,6 +149,7 @@ fn try_digestible_struct(
     };
 
     // call is a Vec<TokenStream> (I think)
```
mfaulk, line 151: Might be worth being explicit about the type, i.e. `let call: Vec<TokenStream> = ` instead of commenting about its type.

crypto/digestible/derive/src/lib.rs
```diff
@@ -48,16 +158,14 @@ fn try_digestible_struct(
                 // this is a regular struct
                 Some(field_ident) => {
                     quote! {
-                        hasher.update(stringify!(#field_ident).as_bytes());
-                        self.#field_ident.digest(hasher);
+                        self.#field_ident.append_to_transcript_allow_omit(stringify!(#field_ident).as_bytes(), transcript);
                     }
                 }
                 // this is a tuple struct, and the member doesn't have an identifier
```
mfaulk, line 164: This comment seems to use "member" and "field" interchangeably. Might be clearer as:
"This field is a tuple struct, and the field doesn't have an identifier.", or "This field is a tuple struct without an identifier."

crypto/digestible/derive/test/tests/behavior.rs
```diff
@@ -67,38 +42,81 @@ struct GenericFoo<X: Digestible> {
 #[test]
 fn foo1() {
     let arg = Foo { a: 0, b: 1, c: 2 };
```
mfaulk, line 44: It looks like a lot of the test data in this file is "benignly ordered", that is, the field names are often in alphabetical order and/or their values are in alphabetical order. It also seems like the field names are mostly/all single-letter names (i.e. don't contain underscores).

Seems like it would be good to diversify the test cases with more arbitrary field names and out-of-order fields and data.

crypto/digestible/derive/test/tests/behavior.rs
```diff
@@ -67,38 +42,81 @@ struct GenericFoo<X: Digestible> {
 #[test]
 fn foo1() {
```
mfaulk, line 43: It would be nice to have a one-line comment about what property this test should check. It seems to be something along the lines of, "A struct that only contains primitive fields should yield an AST containing an aggregate node with one primitive node for each field; and the instance should digest to some known bytes."

Writing that out, it feels like this is testing several different-but-related pieces of functionality, so maybe tests like this should be separated into testing the AST, and testing test vectors of digests. Just food for thought.

crypto/digestible/derive/test/tests/behavior.rs
```diff
@@ -67,38 +42,81 @@ struct GenericFoo<X: Digestible> {
...
+        arg.digest32::<MerlinTranscript>(b"foo1"),
+        [
+            44, 174, 200, 201, 207, 72, 151, 193, 63, 97, 49, 112, 135, 50, 252, 61, 211, 208, 41,
+            4, 241, 249, 90, 114, 181, 197, 177, 151, 39, 165, 76, 197
+        ]
+    );
 }
 
 #[test]
 fn blob1() {
```
mfaulk, line 78: A comment here like "A tuple struct should be treated as an Aggregate" would help make it clear how this test and the following test about TransBlob should differ.

crypto/digestible/derive/test/tests/behavior.rs
```diff
@@ -67,38 +42,81 @@ struct GenericFoo<X: Digestible> {
...
+        arg.digest32::<MerlinTranscript>(b"blob1"),
+        [
+            59, 63, 205, 99, 4, 221, 221, 230, 90, 1, 135, 226, 106, 52, 210, 105, 88, 37, 182, 26,
+            208, 240, 152, 4, 226, 0, 204, 11, 10, 187, 14, 48
+        ]
+    );
+}
+
+#[test]
+fn blob2() {
```
mfaulk, line 101: Again, it would help if the function name and/or comment describe what's happening here, e.g. "A tuple struct with the `transparent` attribute is not treated as an aggregate. This is intended to work with the "new type" pattern of wrapping an existing data type."

Also,  does #[digestible(transparent)] work for tuple structs with more than one field? Or structs with named fields? Or unit (fieldless) structs?

crypto/digestible/derive/test/tests/behavior.rs
```diff
@@ -108,32 +126,114 @@ fn bar1() {
...
+        arg.digest32::<MerlinTranscript>(b"bar1"),
+        [
+            237, 231, 204, 138, 55, 249, 219, 0, 154, 213, 236, 77, 123, 104, 185, 68, 165, 117,
+            179, 15, 85, 65, 13, 134, 163, 16, 206, 60, 249, 184, 194, 81
+        ]
+    );
+}
+
+#[test]
+fn bar2() {
```
mfaulk, line 184: As a general comment, I'm finding all the `foos` and `bars` to be unhelpful identifiers. For example, I think a Bar is a "normal, C-style struct" and a TBar is a "normal C-syle struct containing a transparent field". That could all be clarified with comments, but the tests might be a lot more self-documenting if the type names were meaningful, i.e. Bar -> PlainStruct, TBar -> StructWithTransField, etc.

crypto/digestible/derive/test/tests/behavior.rs
```diff
@@ -166,108 +291,772 @@ fn generic_foo2() {
...
+        [
+            73, 146, 142, 113, 190, 93, 124, 188, 118, 21, 72, 51, 19, 101, 141, 43, 101, 207, 234,
+            146, 65, 49, 191, 0, 40, 215, 129, 168, 55, 80, 189, 247
+        ]
+    );
+}
+
+// An example structure
+#[derive(Digestible)]
+struct Thing {
```
mfaulk, line 337: It looks like these types and the following tests are all about schema evolution. If they are self-contained, it might help to move them into a separate file, like tests/schema_evolution.rs

crypto/digestible/derive/test/tests/behavior.rs
```diff
@@ -166,108 +291,772 @@ fn generic_foo2() {
...
+        [
+            10, 34, 82, 129, 5, 30, 197, 99, 66, 246, 191, 25, 96, 23, 84, 249, 228, 156, 252, 247,
+            30, 194, 152, 147, 221, 244, 220, 46, 23, 236, 213, 203
+        ]
+    );
+}
+
+// Tests for struct_schema evolution
+#[test]
+fn struct_schema_evolution() {
```
mfaulk, line 398: These are really interesting

crypto/digestible/derive/test/tests/behavior.rs
mfaulk: For nastier test cases, strings like "prim", "seq", or "agg" would make good field names or values.

crypto/digestible/derive/test/tests/behavior.rs
```diff
@@ -166,108 +291,772 @@ fn generic_foo2() {
...
+struct ThingV4 {
+    a: Option<u64>,
+    b: Option<u64>,
+    c: Vec<bool>,
 }
 
-// Test digesting an enum.
 #[test]
-fn test_digest_enum() {
+fn thing_struct() {
```
mfaulk, line 367: I'm not sure what this is supposed to be showing me / proving.

crypto/digestible/src/lib.rs
```diff
@@ -3,28 +3,195 @@
...
+///
+/// For small objects it likely calls "append_bytes" once.
+/// For aggregates it likely calls "append_to_transcript" recursively on the members,
+/// passing member names as context. (See digestible-derive crate for specifics.)
+///
+/// In merlin documentation, having the transcript protocol call itself recursively
+/// is called "protocol composition".
+/// (We refer the reader to merlin docu for more discussion.)
+///
+/// Implementations of this trait should be not produce results that depends on
```
mfaulk, line 25: "should be not produce" --> "should not produce"

crypto/digestible/derive/test/tests/behavior.rs
mfaulk: A field containing an empty Vec would be a good test case, too.

crypto/digestible/test-utils/src/inspect_ast.rs
```diff
@@ -0,0 +1,518 @@
+use core::fmt;
+use mc_crypto_digestible::{DigestTranscript, Digestible, MerlinTranscript};
+use serde_json::json;
+use std::{ops::Deref, vec::Vec};
+
+// Digestible AST node types
+
+/// Represents a node in the AST
+#[derive(Clone, Eq, PartialEq)]
+pub enum ASTNode {
```
mfaulk, line 10: Could give each enum variant a short doc comment. For context, by the time i got here, I had forgotten what a `Variant` is, and when I re-read the readme, noticed that None was talked about separately from the other four:

- A primitive is a "simple" type (as opposed to a compound type that has a canonical representation as bytes.
- A sequence is a variable length sequence of values of some other type. A sequence has a length known at runtime.
- An aggregate is a fixed-length sequence of values ("fields"), of different types. Each field has a field name.
- A variant is a single value which may be one of several different types. Each possibility has an associated name, in the context of this variant. In the sequel we call this the "variant possibility name".

crypto/digestible/test-utils/README.md
```diff
@@ -0,0 +1,23 @@
...
+`mc-crypto-digest-test-utils` includes two implementations of `DigestTranscript`
+that can be used to verify that `digestible-derive` is generating the right code.
+
+`MockMerlin` can be used to capture the sequence of `append_bytes` calls when
+a given object is appended to the transcript.
+
+There is also a higher-level debugging tool.
+The interaction of an object with the `DigestTranscript` protocol naturally produces
+an AST, which is normally only implicit, but can be made explicit using the `InspectAST`
+visitor.
```
mfaulk, line 13: Nice

crypto/digestible/test-utils/src/mock_merlin.rs
```diff
@@ -0,0 +1,23 @@
+use mc_crypto_digestible::DigestTranscript;
+use std::vec::Vec;
+
+// This is a mock of merlin, which doesn't hash anything, but captures
+// the inputs to calls to "append_bytes", so that we can inspect them,
+// write tests against them, etc.
+#[derive(Default, Clone, Debug)]
+pub struct MockMerlin {
```
mfaulk, line 8: I have really been enjoying mockall for things like this, https://docs.rs/mockall/0.8.0/mockall/ , although what you've got here looks good, too.

transaction/core/src/blockchain/block_contents.rs
```diff
@@ -33,57 +31,35 @@ impl BlockContents {
         }
     }
 
-    /// The Blake2B256 digest of `self`.
+    /// The Merlin digest of `self`.
```
mfaulk, line 34: Just thinking out loud: At one point, we moved from SHA3 to Blake2B because Blake2B was faster. Switching to Merlin (specifically, STROBE) roughly brings us back to SHA3. I don't think performance will be an issue here, and it feels like the correctness hygiene we get from Merlin transcripts is much more important.

crypto/digestible/derive/test/tests/behavior.rs
garbageslam: the transparent thing only works when it has exactly one field, it's meant for like:

you have a type that's in the ledger somehow, and you need to create a newtype wrapper around it, but you don't want that wrapper to change the hash

i added `transparent` to a bunch of wrappers in the transaction crate in this commit. i'm not sure if that was a great idea -- it will mean that we can't add fields to those types. but mainly they are like "TxOutMembershipHash" and its hard to imagine that we would add fields to them. the main benefit of marking them transparent was that it made the AST things more readable for important structures like TxOut. but possibly we shouldn't actually make those transparent.

I also made `RistrettoPublic` transparent, because I don't think we want to add fields to it, and that made the AST more readable. but at the end of the day, it doesn't matter that much if the AST is readable, and its concievable, though unlikely that we might want to add fields to that thing. so that's kind of a difficult judgment call

crypto/digestible/derive/test/tests/behavior.rs
garbageslam: yeah this is a good criticism, thanks i will fix

garbageslam:
this is now rebased on master

thanks Matt for all the detailed comments! i will work to address them all after 5pm

crypto/digestible/derive/test/tests/behavior.rs
mfaulk: That makes sense, and that is a tricky judgement call.

I like that `transparent` lets us refactor a field to be a newtype (or a different newtype with the same inner data) without changing the digest. That might be handy if we decided that, say, RistrettoPublic wasn't specific enough, and we wanted different newtypes to denote different categories of Ristretto public keys.

crypto/digestible/derive/src/lib.rs
jcape: CLion with Rust does this:

Screenshot from 2020-09-02 13-42-44

crypto/digestible/src/lib.rs
```diff
@@ -3,28 +3,195 @@
...
+    /// If the primitive is the only thing being digested, the context string
+    /// comes from the call to digest32 / digest64.
+    ///
+    /// The typename is an arbitrary string representing the type of the data,
+    /// for example, `uint`, `bytes`, `string`, `ristretto`
+    ///
+    /// The data is the canonical bytes representing the primitive.
+    /// If the primitive does not have a canonical representation as bytes then
+    /// it isn't appropriate to treat it as a primitive in this hashing scheme.
+    #[inline]
```
jcape, line 103: These should either be `inline(always)`, or left to LTO.

crypto/digestible/src/lib.rs
```diff
@@ -49,128 +216,258 @@ pub trait Digestible {
...
+impl DigestibleAsBytes for [u8; 23] {}
+impl DigestibleAsBytes for [u8; 24] {}
+impl DigestibleAsBytes for [u8; 25] {}
+impl DigestibleAsBytes for [u8; 26] {}
+impl DigestibleAsBytes for [u8; 27] {}
+impl DigestibleAsBytes for [u8; 28] {}
+impl DigestibleAsBytes for [u8; 29] {}
+impl DigestibleAsBytes for [u8; 30] {}
+impl DigestibleAsBytes for [u8; 31] {}
+impl DigestibleAsBytes for [u8; 32] {}
```
jcape, line 394: Not 100% this will work, buuut:

```suggestion
impl<const N: usize> DigestibleAsBytes for [u8; N] {}
```

jcape (approved):
The macro processing seems reasonable, and it looks like my main concerns from the last PR are addressed (e.g. it's not possible to have `vec![u16; 2]` and `vec![u32; 1]` with identical bytes map to identical representations), but there are a couple nitpicks (e.g. I *think* we can use const generics now, since [`LengthAtMost32` was removed from rust](https://github.com/rust-lang/rust/pull/74060/files)).

crypto/digestible/src/lib.rs
garbageslam: I don't think this is right,
(1) my experience has been that benchmarks got significantly better in ORAM when I used `#[inline]`, instead of doing what you are suggesting, for functions in `aligned`, `aligned-cmov` and other crates
(2) I think the convention in many essential rust libraries is to use `#[inline]` when we want to permit the optimizer to inline this across crate boundaries, which is usually the case for tiny, low-level functions like this. For example, here's a related discussion in the `hashbrown` crate which is very informative: https://github.com/rust-lang/hashbrown/pull/119

I think it's better to permit the compiler to inline things in cases like this, if llvm decides its a good idea. Thin-LTO is a late stage optimization pass, with less information and less time to make optimization decisions, and it shouldn't be relied on or used as a replacement for things that would normally be done at -O2 or -O3 optimizations.
If we omit `#[inline]` then the optimizer is not permitted to inline this function across crate boundaries, even at `-O2` and `-O3` levels.

The main reason to remove inline annotations is to sacrifice runtime performance for better compile times. We shouldn't do that unless the compile times for this are actually slow. I see no evidence that that is the case.

crypto/digestible/src/lib.rs
garbageslam: will investigate, thanks :)

crypto/digestible/README.md
```diff
@@ -1,175 +1,484 @@
...
+
+Primitives:
+-----------
+
+For this discussion, a *primitive* is a type which *has a canonical, portable representation as bytes*.
+
+`Digestible` is implemented in this crate for *built-in integer types*, *byte slices and arrays*, and *strings*.
+
+For built-in integer types, we specify the use of little-endian byte encoding, as merlin uses internally
+for the encoding of buffer lengths. The type signifier is `"uint"`.
```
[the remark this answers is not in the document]
garbageslam: i just have a different string literal which is `int`

crypto/digestible/src/lib.rs
garbageslam: I guess we could have an `inline-more` feature like hashbrown does: https://github.com/rust-lang/hashbrown/blob/master/Cargo.toml#L53

It just seems kind of silly

jcape: Were you building with lto?

crypto/digestible/README.md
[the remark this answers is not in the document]
garbageslam: fixed

transaction/core/src/blockchain/block_contents.rs
garbageslam: yeah -- STROBE is lower-level than SHA3, so it might be somewhat faster than SHA3. I don't really know -- we can do some profiling

crypto/digestible/src/lib.rs
garbageslam: it gave me this:

```
error[E0658]: const generics are unstable
   --> crypto/digestible/src/lib.rs:372:12
    |
372 | impl<const N: usize> DigestibleAsBytes for [u8; N] {}
    |            ^
    |
    = note: see issue #44580 <https://github.com/rust-lang/rust/issues/44580> for more information
    = help: add `#![feature(const_generics)]` to the crate attributes to enable

error[E0277]: arrays only have std trait implementations for lengths 0..=32
   --> crypto/digestible/src/lib.rs:372:22
    |
358 | pub trait DigestibleAsBytes: AsRef<[u8]> + Sized {}
    |                              ----------- required by this bound in `DigestibleAsBytes`
...
372 | impl<const N: usize> DigestibleAsBytes for [u8; N] {}
    |                      ^^^^^^^^^^^^^^^^^ the trait `core::array::LengthAtMost32` is not implemented for `[u8; N]`
    |
    = note: required because of the requirements on the impl of `core::convert::AsRef<[u8]>` for `[u8; N]`

error: aborting due to 2 previous errors
```

garbageslam: i could add it, but right now, nothing is tying this crate to the nightly compiler, so i'd rather wait until that feature is stable

tsegaran (approved):
I'm happy with this pending CI

crypto/digestible/derive/test/tests/behavior.rs
garbageslam: done

PR #392, opened 2020-08-24
[MCC-483] TxManager trait

mfaulk:
Soundtrack of this PR: [All About That Trait](https://www.youtube.com/watch?v=7PCkvCPvDXk)

### Motivation

Creates a trait for TxManager so that it can be mocked with mockall.

### In this PR
* Splits TxManager into a few smaller files. I find that easier to navigate.
* Creates a trait and mock. Changes a method to avoid a higher-ranked trait bound, which mockall doesn't support.

### Future Work
* Update ByzantineLedger, ConsensusService, ClientApiService, PeerApiService, etc. to depend on the trait instead of the struct, so that they can be tested using the mock.
* Will probably rename TxManager and TxManagerTrait to TxManagerImpl and TxManager.


consensus/service/src/byzantine_ledger/worker.rs
```diff
@@ -610,33 +610,16 @@ impl<
...
-            log::debug!(
-                self.logger,
-                "attempting to resolve missing tx hashes ({}, left with {}): {:?}",
-                missing_hashes.len(),
-                all_missing_hashes.len(),
-                missing_hashes,
-            );
-
-            match conn.fetch_txs(&missing_hashes, Fibonacci::from_millis(100).take(10)) {
+        for chunk in missing_hashes[..].chunks(100) {
```
mfaulk, line 613: `chunks` makes this easier.

consensus/service/src/tx_manager/mod.rs
```diff
@@ -223,25 +158,18 @@ impl<E: ConsensusEnclave, UI: UntrustedInterfaces> TxManager<E, UI> {
         expired.keys().cloned().collect()
     }
 
-    /// Returns elements of `tx_hashes` that are not inside the cache.
-    pub fn missing_hashes<T>(&self, tx_hashes: &T) -> Vec<TxHash>
-    where
-        for<'a> &'a T: IntoIterator<Item = &'a TxHash>,
```
mfaulk, line 229: mockall doesn't like this higher-ranked trait bound (it lets you pass a reference to any iterable collection of TxHashes as a function argument). I'm guessing that the simpler `contains` will work OK, but I will also check if the extra locking causes performance problems.

transaction/core/src/onetime_keys.rs
```diff
@@ -121,7 +121,7 @@ pub fn view_key_matches_output(
 /// * `tx_public_key` - The transaction public key `R`.
 /// * `view_private_key` - A private view key `a`.
 /// * `subaddress_spend_private_key` - A private spend key `d = Hs(a || i) + b`.
-///
+///69
```
mfaulk, line 124: I blame my cat. Will fix.

consensus/service/src/tx_manager/mod.rs
```diff
@@ -7,72 +7,29 @@
 //! may operate on.
 
 use crate::counters;
-use failure::Fail;
+
```
eranrund, line 10: We generally don't split imports so this empty line isn't necessary.

consensus/service/src/tx_manager/mod.rs
eranrund: Yup, would be interesting to see what effect this has. I think this is going to move us from one lock per SCP statement, so N locks per SCP statement, where N is the number of tx hashes the statement refers to.

transaction/core/src/onetime_keys.rs
eranrund: 🐈 🐈 🐈 🐈

consensus/service/src/tx_manager/error.rs
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) 2018-2020 MobileCoin Inc.
+
+use failure::Fail;
```
rjwalters, line 3: deprecated?

consensus/service/src/tx_manager/mod.rs
rjwalters: https://stackoverflow.com/questions/3652056/how-efficient-is-locking-an-unlocked-mutex-what-is-the-cost-of-a-mutex

should we place bets? :)

consensus/service/src/tx_manager/error.rs
mfaulk: I'll pick that up on the next PR.

mfaulk:
Performance seems unchanged:
Screenshot from 2020-08-24 20-47-23

PR #393, opened 2020-08-25
Interpret public addresses as request codes

tsegaran:
### Motivation

We'd like clients to be able to enter either a payment request code or a public address. This PR interprets either case and determines the best thing to do.

### In this PR
* Allows `read_request_code` to take a plain public address, which is interpreted as a payment request with no amount and no memo

### DIFF
--- a/mobilecoind/src/service.rs
+++ b/mobilecoind/src/service.rs
@@ -358,19 +358,28 @@ impl<T: BlockchainConnection + UserTxConnection + 'static> ServiceApi<T> {
             request.get_b58_code().to_string(),
         )
         .map_err(|err| rpc_internal_error("PrintableWrapper_b58_decode", err, &self.logger))?;
-        if !request_wrapper.has_payment_request() {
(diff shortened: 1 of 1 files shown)


mobilecoind/src/service.rs
```diff
@@ -358,19 +358,28 @@ impl<T: BlockchainConnection + UserTxConnection + 'static> ServiceApi<T> {
...
+            let public_address = request_wrapper.get_public_address();
+            let mut response = mc_mobilecoind_api::ReadRequestCodeResponse::new();
+            response.set_receiver(public_address.clone());
+            response.set_value(0);
+            response.set_memo(String::new());
+            Ok(response)
+        } else {
+            Err(RpcStatus::new(
                 RpcStatusCode::INVALID_ARGUMENT,
                 Some("has_payment_request".to_string()),
```
sugargoat, line 380:
```suggestion
                Some("Neither payment request or public address".to_string()),
```

PR #394, opened 2020-08-25
Implement HSM cold storage

xoloki:

### Motivation

Some partners require cold storage using an HSM in order to satisfy compliance. We will use the HSM to provide ECDSA signatures, allowing us to store the cold storage transactions on the chain but preventing them from being spent without the private keys on the HSM.

### In this PR
* Add HsmParams and use in transaction::tx::Tx and external::Tx
* Update consensus enclave to do ECDSA checks and alter TxOuts for HSM transactions

Fixes #MCC-1193

### Future Work

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2383,6 +2383,7 @@ dependencies = [
 name = "mc-consensus-enclave-impl"
 version = "0.6.0"
 dependencies = [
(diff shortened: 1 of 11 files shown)


sugargoat:
Closing for now - will re-approach at some later date.

PR #396, opened 2020-08-25
[MCC-1694] ByzantineLedger uses TxManager trait

mfaulk:
Soundtrack of this PR: [Royals](https://www.youtube.com/watch?v=nlcIKh6sBtc)

### Motivation

ByzantineLedger currently has a single, large unit test that traces the "happy path" of a node reaching consensus on a single slot. We don't yet have unit test coverage for failure modes, catch-up, or broadcast.

### In this PR
* ByzantineLedger depend on the TxManager trait, instead of a concrete implementation. This will help with testing failure modes.
* Creates a simple unit test that uses the autogenerated mock for TxManager. This is mostly a proof-of-concept that sets up the pattern for further tests.
* Extracts some common setup code that will probably get reused in multiple unit tests.

### Future Work
* https://mobilecoin.atlassian.net/browse/MCC-480
* https://mobilecoin.atlassian.net/browse/MCC-486
* https://mobilecoin.atlassian.net/browse/MCC-498

### DIFF
--- a/consensus/service/src/byzantine_ledger/mod.rs
+++ b/consensus/service/src/byzantine_ledger/mod.rs
@@ -11,7 +11,7 @@ mod worker;
 use crate::{
     byzantine_ledger::{task_message::TaskMessage, worker::ByzantineLedgerWorker},
     counters,
-    tx_manager::TxManagerTrait,
+    tx_manager::TxManager,
 };
 use mc_common::{logger::Logger, NodeID, ResponderId};
 use mc_connection::{BlockchainConnection, ConnectionManager};
@@ -53,7 +53,7 @@ impl ByzantineLedger {
     pub fn new<
         PC: BlockchainConnection + ConsensusConnection + 'static,
         L: Ledger + Sync + 'static,
-        TXM: TxManagerTrait + Send + Sync + 'static,
+        TXM: TxManager + Send + Sync + 'static,
     >(
         node_id: NodeID,
         quorum_set: QuorumSet,
@@ -233,7 +233,7 @@ impl Drop for ByzantineLedger {
 mod tests {
     use super::*;
     use crate::{
-        tx_manager::{MockTxManagerTrait, TxManager},
+        tx_manager::{MockTxManager, TxManagerImpl},
         validators::DefaultTxManagerUntrustedInterfaces,
     };
     use hex;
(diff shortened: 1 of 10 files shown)


peers/Cargo.toml
```diff
@@ -27,6 +27,7 @@ crossbeam-channel = "0.3"
 ed25519 = { version = "1.0.1", default-features = false, features = ["serde"] }
 failure = "0.1.5"
 grpcio = "0.6.0"
+mockall = "0.7.2"
```
mfaulk, line 30: This moved from dev_dependencies to dependencies so that other crates can use mocks defined here in their tests. I think this is a limitation of cargo: if crate A depends on crate B, building A with the `test` attribute does not build B with the `test` attribute. 

I think this is OK, but would be happy to do something else if you've got an idea.

consensus/service/src/byzantine_ledger/mod.rs
```diff
@@ -52,16 +51,15 @@ pub struct ByzantineLedger {
...
         L: Ledger + Sync + 'static,
-        UI: UntrustedInterfaces + Send + Sync + 'static,
+        TXM: TxManager + Send + Sync + 'static,
     >(
         node_id: NodeID,
         quorum_set: QuorumSet,
         peer_manager: ConnectionManager<PC>,
         ledger: L,
-        tx_manager: Arc<TxManager<E, UI>>,
+        tx_manager: Arc<TXM>,
```
mfaulk, line 62: This is the main change that allows for convenient mocking.

consensus/service/src/byzantine_ledger/mod.rs
```diff
@@ -274,57 +274,152 @@ mod tests {
...
+            logger.clone(),
         );
 
-        let node_c_signer_key = Ed25519Pair::from_random(&mut seeded_rng);
-        let node_c_uri = test_peer_uri(44, hex::encode(&node_c_signer_key.public_key()));
-        let _node_c = (
-            test_node_id(node_c_uri.clone(), &node_c_signer_key),
-            trivial_quorum_set,
+        // Mock tx_manager
+        let tx_manager = Arc::new(MockTxManager::new());
```
mfaulk, line 370: This is the auto-generated mock for the TxManager trait. It doesn't really do anything here because this test is very simple, but in general it would make it easy to test how ByzantineLedger reacts to a transaction that is not well-formed, not valid, not in the cache, etc.

PR #398, opened 2020-08-26
[MCC-1700] Mocks Ledger trait

mfaulk:
### Motivation

The Ledger trait / LedgerDB are used by a lot of things, which makes those things tedious to test.

### In this PR
* A few tweaks to the Ledger trait so that it is compatible with automock.

### Future Work
* Use MockLedger in unit tests.

### DIFF
--- a/ledger/db/src/ledger_trait.rs
+++ b/ledger/db/src/ledger_trait.rs
@@ -8,16 +8,16 @@ use mc_transaction_core::{
     tx::{TxOut, TxOutMembershipProof},
     Block, BlockContents, BlockSignature,
 };
-// use mockall::*;
+use mockall::*;
 
-// #[automock]
+#[automock]
 pub trait Ledger: Send {
     /// Appends a block along with transactions.
     fn append_block(
         &mut self,
         block: &Block,
         block_contents: &BlockContents,
-        signature: Option<&BlockSignature>,
+        signature: Option<BlockSignature>,
     ) -> Result<(), Error>;
 
     /// Get the total number of blocks in the ledger.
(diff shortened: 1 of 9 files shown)


ledger/db/src/ledger_trait.rs
```diff
@@ -8,14 +8,16 @@ use mc_transaction_core::{
     tx::{TxOut, TxOutMembershipProof},
     Block, BlockContents, BlockSignature,
 };
+use mockall::*;
 
-pub trait Ledger: Clone + Send {
```
mfaulk, line 12: automock cannot mock traits that inherit from Clone.

ledger/db/src/ledger_trait.rs
```diff
@@ -8,14 +8,16 @@ use mc_transaction_core::{
...
+#[automock]
+pub trait Ledger: Send {
     /// Appends a block along with transactions.
     fn append_block(
         &mut self,
         block: &Block,
         block_contents: &BlockContents,
-        signature: Option<&BlockSignature>,
```
mfaulk, line 18: automock does not like this.

ledger/db/src/ledger_trait.rs
rjwalters: so the pattern here is to switch to using an Arc to wrap things instead of cloning? this seems like it will always be an acceptable tradeoff for us to have mocking? or are there conditions in which we would worry about performance?

mfaulk: I'm still feeling out the patterns that work well, but here are some observations:

The `automock` macro currently can't mock traits with trait bounds, so it doesn't work with Clone. I think it is fine with Send because that is a marker trait that doesn't declare any methods. The more flexible `mock!` macro does support trait bounds, but it requires you to copy-paste each trait into the body of the macro, which is ugly and awkward to maintain.

Removing Clone from the trait doesn't seem like a problem to me, because code that demands a clone-able ledger can always declare additional trait bounds on the types it uses, i.e. `L: Ledger + Clone`.

mfaulk:
Testing this with `internal` before merging...

PR #402, opened 2020-08-27
MCC-1705: Fix subtraction overflow for blocks behind

sugargoat:
### Motivation

Fix bug which encounters overflow if peer block height is behind current block height when calculating "blocks behind."

### In this PR
* Fixes bug described above

[MCC-1705](https://mobilecoin.atlassian.net/browse/MCC-1705)

### DIFF
--- a/consensus/service/src/consensus_service.rs
+++ b/consensus/service/src/consensus_service.rs
@@ -622,11 +622,8 @@ impl<E: ConsensusEnclaveProxy, R: RaClient + Send + Sync + 'static> ConsensusSer
                             )
                         })
                         .ok();
-                    blocks_behind = if peer_block_height > b {
-                        Some(peer_block_height - b)
-                    } else {
-                        Some(0)
-                    };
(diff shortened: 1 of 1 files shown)


consensus/service/src/consensus_service.rs
```diff
@@ -622,7 +622,11 @@ impl<E: ConsensusEnclaveProxy, R: RaClient + Send + Sync + 'static> ConsensusSer
                             )
                         })
                         .ok();
-                    blocks_behind = Some(std::cmp::min(peer_block_height - b, 0));
+                    blocks_behind = if peer_block_height > b {
```
mfaulk, line 625: Good catch!

You can probably make this a one-liner with `peer_block_height.saturating_sub(b)`, but what you've got here might be more readable. Up to you.

sugargoat: Oh neat - I like that! Will switch to `saturating_sub`

PR #404, opened 2020-08-27
MCC-959 side-channel resistant merkle proof validation

garbageslam:
The previous merkle proof has the following structure:

- index (indicates which number leaf this TxOut is in the merkle tree)
- highest_index (indicates how many TxOuts are in the merkle tree at time of proof)
- several TxOutMembershipElements, which include a hash corresponding to
  an internal node in the tree, and a range explaining which leafs
  are covered by this hash

The previous merkle proof validation routine assumes no particular structure
on the membership elements, it sorts them, throws them into hash maps,
etc. and as a result it is complicated and leaks information about the
path that the merkle proof is following over side channels. Because the
path determines the entire merkle proof, this means that we leak the
whole thing, in the SGX security model.

This PR does the following:

- Introduce a `compose_adjacent_membership_elements` function
  which takes two membership elements, checks if their ranges are
  adjacent in the tree, and computes the hash of their parent in
  the tree if so. This uses subtle constant time equality tests
  and conditional assignments to ensure that there are no branches
  at all, except leading to erroring code paths.
- Merkle proof validation then works strictly from the bottom up,
  using `try_fold` to fold successive membership elements from
  the membership proof together, and deducing the implied root
  element hash as a result.
  Finally we check the implied root element hash against the known
  root element hash.
- This imposes constraints on what protobufs are valid merkle proofs.
  The new routine won't accept proofs in arbitrary order, they must
  be in the order that makes validation easy. It is okay to push
  work from the server to the clients in this way. (IMO)
- This also changes the merkle proof generation code to follow this
  format. It turns out that generating the proofs in this order is
  quite easy as well -- the only reason they could end up in a wierd
  order is because of the hashmap or because of sorting ambiguities,
  for example, every proof should actually contain two leaves.

Not all the tests have been fixed yet, but this draft shows the basic idea.

One thing that we could do that might make this simpler still is,
in the TxOutMembershipProof, add a separate "leaf hash" field corresponding
to the hash of the leaf at `proof.index`, and don't make a separate membership
element for this leaf. This will make some types of invalid merkle proofs
unrepresentable, and so reduce the number of error conditions of the
validation routine, essentially pushing some of the error handling
into the protobuf layer. I tried to avoid changing any proto types
in this commit though.

---

One possibly significant motivation for this PR is that, after this PR, we no longer have the "internal nodes" in the merkle proof elements set, we only have the "other nodes" (see get_merkle_proof routine for variables with these names).
Basically, in the previous implementation, not only does the memebership proof contain the inputs to the hash function that are needed to verify the proof, it also contains all the intermediate hash values, which would be implied by the other data. This was needed to make the top-down validation approach work. But after the bottom-up validation approach is used, we don't need those hashes at all. This makes the proofs half the size, so if the tree has height 30, instead of 1800 bytes worth of merkle proof data, we only have 900 bytes worth of merkle proof data. This also allows to delete some unnecessary code when implementing the routine that "derives merkle proof at index"

### DIFF
--- a/ledger/db/src/tx_out_store.rs
+++ b/ledger/db/src/tx_out_store.rs
@@ -345,7 +345,12 @@ impl TxOutStore {
             });
         }
 
-        Ok(TxOutMembershipProof::new(index, num_tx_outs - 1, elements))
+        let result = TxOutMembershipProof::new(index, num_tx_outs - 1, elements);
+        debug_assert!(
+            mc_transaction_core::membership_proofs::compute_implied_merkle_root(&result).is_ok(),
+            "Freshly created membership proof was invalid"
+        );
+        Ok(result)
     }
 }
 
@@ -512,9 +517,8 @@ mod membership_proof_tests {
             // Tamper with proof after it is constructed. This bypasses checks in TxOutMembershipProof::new().
             proof.index = 6;
             assert_eq!(
-                false,
+                Err(MembershipProofError::HighestIndexMismatch),
                 is_membership_proof_valid(&tx_outs.get(5).unwrap(), &proof, &known_root_hash)
-                    .unwrap()
             );
         }
     }
@@ -554,9 +558,8 @@ mod membership_proof_tests {
             // `num_tx_outs` is less than the index of the `TxOut` referenced by the proof.
             proof.highest_index = 2;
             assert_eq!(
-                false,
+                Err(MembershipProofError::HighestIndexMismatch),
                 is_membership_proof_valid(&tx_outs.get(5).unwrap(), &proof, &known_root_hash)
-                    .unwrap()
             );
         }
 
@@ -570,9 +573,8 @@ mod membership_proof_tests {
             // ranges [8,15] and [0,15].
             proof.highest_index = 8;
             assert_eq!(
-                false,
+                Err(MembershipProofError::HighestIndexMismatch),
                 is_membership_proof_valid(&tx_outs.get(5).unwrap(), &proof, &known_root_hash)
-                    .unwrap()
             );
         }
     }
@@ -606,9 +608,8 @@ mod membership_proof_tests {
                 hash: TxOutMembershipHash::from([7u8; 32]),
             };
             assert_eq!(
-                false,
+                Err(MembershipProofError::UnexpectedMembershipElement(3)),
                 is_membership_proof_valid(&tx_outs.get(5).unwrap(), &proof, &known_root_hash)
-                    .unwrap()
             );
         }
     }
@@ -651,17 +652,17 @@ mod membership_proof_tests {
         let mut proof1 = proof.clone();
         proof1.elements.remove(0);
 
-        assert!(
-            !is_membership_proof_valid(&tx_outs.get(5).unwrap(), &proof1, &known_root_hash)
-                .unwrap()
+        assert_eq!(
+            Err(MembershipProofError::MissingLeafHash(5)),
+            is_membership_proof_valid(&tx_outs.get(5).unwrap(), &proof1, &known_root_hash)
         );
 
         let mut proof2 = proof.clone();
         proof2.elements.remove(1);
 
-        assert!(
-            !is_membership_proof_valid(&tx_outs.get(5).unwrap(), &proof2, &known_root_hash)
-                .unwrap()
+        assert_eq!(
+            Err(MembershipProofError::UnexpectedMembershipElement(2)),
+            is_membership_proof_valid(&tx_outs.get(5).unwrap(), &proof2, &known_root_hash)
         );
     }
 
@@ -1240,17 +1241,14 @@ pub mod tx_out_store_tests {
 
             */
 
-            assert_eq!(7, proof.elements.len());
+            assert_eq!(4, proof.elements.len());
 
             let ranges: Vec<Range> = proof.elements.iter().map(|e| e.range).collect();
 
             assert!(ranges.contains(&Range::new(3, 3).unwrap()));
             assert!(ranges.contains(&Range::new(2, 2).unwrap()));
-            assert!(ranges.contains(&Range::new(2, 3).unwrap()));
             assert!(ranges.contains(&Range::new(0, 1).unwrap()));
-            assert!(ranges.contains(&Range::new(0, 3).unwrap()));
             assert!(ranges.contains(&Range::new(4, 7).unwrap()));
-            assert!(ranges.contains(&Range::new(0, 7).unwrap()));
 
             for element in proof.elements {
                 let expected_hash = tx_out_store
@@ -1284,18 +1282,15 @@ pub mod tx_out_store_tests {
 
             */
 
-            assert_eq!(7, proof.elements.len());
+            assert_eq!(4, proof.elements.len());
 
             let ranges: Vec<Range> = proof.elements.iter().map(|e| e.range).collect();
 
             println!("{:?}", ranges);
             assert!(ranges.contains(&Range::new(4, 4).unwrap()));
             assert!(ranges.contains(&Range::new(5, 5).unwrap()));
-            assert!(ranges.contains(&Range::new(4, 5).unwrap()));
             assert!(ranges.contains(&Range::new(6, 7).unwrap()));
-            assert!(ranges.contains(&Range::new(4, 7).unwrap()));
             assert!(ranges.contains(&Range::new(0, 3).unwrap()));
-            assert!(ranges.contains(&Range::new(0, 7).unwrap()));
 
             for element in proof.elements {
                 let expected_hash = if element.range.from >= num_tx_outs as u64 {
--- a/ledger/db/src/tx_out_store.rs
+++ b/ledger/db/src/tx_out_store.rs
@@ -1229,11 +1229,11 @@ pub mod tx_out_store_tests {
             /*
                 The proof-of-membership for TxOut 3 should include the following ranges/hashes:
 
-                               _____________H(0,7)_________
+                               _______________|____________
                                |                          |
-                         ____H(0,3)_____                H(4,7)
+                         ______|________               H(4,7)
                         |              |
-                     H(0,1)          H(2,3)
+                     H(0,1)          __.____
                                     |      |
                                  H(2,2)  H(3,3)
                                    |       |
@@ -1245,10 +1245,10 @@ pub mod tx_out_store_tests {
 
             let ranges: Vec<Range> = proof.elements.iter().map(|e| e.range).collect();
 
-            assert!(ranges.contains(&Range::new(3, 3).unwrap()));
-            assert!(ranges.contains(&Range::new(2, 2).unwrap()));
-            assert!(ranges.contains(&Range::new(0, 1).unwrap()));
-            assert!(ranges.contains(&Range::new(4, 7).unwrap()));
+            assert_eq!(ranges[0], Range::new(3, 3).unwrap());
+            assert_eq!(ranges[1], Range::new(2, 2).unwrap());
+            assert_eq!(ranges[2], Range::new(0, 1).unwrap());
+            assert_eq!(ranges[3], Range::new(4, 7).unwrap());
 
             for element in proof.elements {
                 let expected_hash = tx_out_store
@@ -1270,11 +1270,11 @@ pub mod tx_out_store_tests {
             /*
                 The proof-of-membership for TxOut 5 should include the following ranges/hashes:
 
-                               __________H(0,7)___________________
+                               _____________|_____________________
                                |                                  |
-                            H(0,3)                       ______ H(4,7)_____
+                            H(0,3)                       _________|________
                                                          |                |
-                                                   ___H(4,5)__          H(6,7) = H(Nil)
+                                                   ______|____          H(6,7) = H(Nil)
                                                    |         |
                                                  H(4,4)   H(5,5)
                                                    |        |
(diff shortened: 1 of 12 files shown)


transaction/core/src/membership_proofs/errors.rs
```diff
@@ -1,29 +1,31 @@
 // Copyright (c) 2018-2020 MobileCoin Inc.
 
 use crate::range;
-use failure::Fail;
+use displaydoc::Display;
 
 /// Reasons why a creating or validating a proof of membership might fail.
-#[derive(Debug, Fail, PartialEq)]
+#[derive(Debug, Display, PartialEq)]
```
garbageslam, line 7: changed this to displaydoc, because as a policy we are trying to migrate to displaydoc, and so anyhting using failure that we touch should change to displaydoc

transaction/core/src/membership_proofs/errors.rs
```diff
@@ -1,29 +1,31 @@
...
+    HighestIndexMismatch,
+
     /// Failed to serialize a TxOut.
-    #[fail(display = "TxOutSerializationError")]
     TxOutSerializationError,
 
-    #[fail(display = "CapacityExceeded")]
-    CapacityExceeded,
+    /// Numeric limits exceeded
+    NumericLimitsExceeded,
```
garbageslam, line 28: i thought this was a better name for an error condition when u64 was not large enough to give a result to our math

transaction/core/src/membership_proofs/mod.rs
```diff
@@ -50,6 +50,139 @@ fn hash_nil() -> [u8; 32] {
...
+/// * b: TxOutMembershipElement to be combined with a, and which should be its left or right sibling
+///
+/// Returns:
+/// * A TxOutMembershipElement for the parent of a and b, with the implied hash value.
+/// * An error if a and b were not adjacent, which implies that the merkle proof was badly structured.
+///   This error can be mapped to e.g. Error::UnexpectedMembershipElement and the caller can provide context.
+pub fn compose_adjacent_membership_elements(
+    a: &TxOutMembershipElement,
+    b: &TxOutMembershipElement,
+) -> Result<TxOutMembershipElement, ()> {
```
garbageslam, line 77: this is the main ingredient that was missing before to get side-channel resistance

transaction/core/src/membership_proofs/mod.rs
```diff
@@ -66,96 +199,29 @@ pub fn is_membership_proof_valid(
...
+    if first.range.from != proof.index || first.range.to != proof.index {
+        return Err(Error::MissingLeafHash(proof.index));
     }
-
-    if proof.highest_index > root_range.to {
-        return Ok(false);
+    // The first element hash must match the tx_out
+    if !bool::from(first.hash.as_ref().ct_eq(&hash_leaf(tx_out))) {
+        return Err(Error::IncorrectLeafHash(first.range.from));
     }
```
garbageslam, line 216: this test got lost in the refactor, i will put it back in the next commit

transaction/core/src/membership_proofs/mod.rs
```diff
@@ -66,96 +199,29 @@ pub fn is_membership_proof_valid(
...
 
-    // * All internal node's hashes between leaf and root must be recomputable from their children's hashes.
```
garbageslam, line 115: after this change, the proof doesn't carry all the hashes of the internal nodes, those values are implied by the other values, so this check goes away, and we only check the final hash versus the known root hash

transaction/core/src/membership_proofs/mod.rs
```diff
@@ -66,96 +199,29 @@ pub fn is_membership_proof_valid(
...
-        .get(root_range)
-        .expect("`root_range` should be a key.");
-
-    if *root_hash != *known_root_hash {
-        // Incorrect root hash.
-        return Ok(false);
+        .first()
+        .ok_or_else(|| Error::MissingLeafHash(proof.index))?;
+    if first.range.from != proof.index || first.range.to != proof.index {
+        return Err(Error::MissingLeafHash(proof.index));
```
garbageslam, line 211: as explained in pull request, i think it would nice to make a separate field in `TxOutMembershipProof` that contains the hash of the first leaf, and then these two checks would go away

transaction/core/src/membership_proofs/mod.rs
```diff
@@ -50,6 +50,139 @@ fn hash_nil() -> [u8; 32] {
...
+
+    // The first element should correspond to proof.index in a well-formed proof
+    let first = proof
+        .elements
+        .first()
+        .ok_or_else(|| Error::MissingLeafHash(proof.index))?
+        .clone();
+    if first.range.from != proof.index || first.range.to != proof.index {
+        return Err(Error::MissingLeafHash(proof.index));
+    }
```
garbageslam, line 164: (these checks would also go away if we just add this to `proof.first_hash` or something and make it not part of `proof.elements`

transaction/core/src/membership_proofs/mod.rs
```diff
@@ -50,6 +50,139 @@ fn hash_nil() -> [u8; 32] {
...
+
+    // Try to fold subsequent elements together with the first element,
+    // combining them using `combine_adjacent_membership_elements`.
+    // idx is used only to provide context for the error message
+    let implied_root = proof.elements[1..].iter().enumerate().try_fold(
```
garbageslam, line 169: this slice would then also go away

transaction/core/src/membership_proofs/mod.rs
```diff
@@ -178,19 +244,18 @@ pub fn derive_proof_at_index(
     let derived_root_range: Range = {
         let num_leaves_full_tree_opt = (index + 1).checked_next_power_of_two();
         if num_leaves_full_tree_opt.is_none() {
-            return Err(Error::CapacityExceeded);
+            return Err(Error::NumericLimitsExceeded);
```
garbageslam, line 247: i just changed the name of this error, this is the only place that it appears

transaction/core/src/membership_proofs/mod.rs
```diff
@@ -207,28 +272,13 @@ pub fn derive_proof_at_index(
...
-            let right_child_hash = {
-                let right_child_range = Range::new(mid + 1, element.range.to)?;
-                *derived_elements
-                    .get(&right_child_range)
-                    .expect("Child range should already exist.")
-            };
-
-            TxOutMembershipHash::from(hash_nodes(&left_child_hash, &right_child_hash))
+            // This is unexpected, none of the proof elements should cover index.
+            return Err(Error::UnexpectedMembershipElement(element_idx));
```
garbageslam, line 276: this code goes away because the proof structure no longer contains intermediate values, their values are implied by the others, and so they dont need to be transmitted, nor updated

transaction/core/src/range.rs
```diff
@@ -9,6 +9,12 @@ use serde::{Deserialize, Serialize};
 #[derive(Debug, Eq, PartialEq, Copy, Clone)]
 pub struct RangeError {}
 
+impl core::fmt::Display for RangeError {
+    fn fmt(&self, fmt: &mut core::fmt::Formatter) -> core::fmt::Result {
+        write!(fmt, "Invalid range")
+    }
+}
```
garbageslam, line 16: i could not derive Display on range error so i had to implement it

transaction/core/src/tx.rs
```diff
@@ -316,21 +318,13 @@ impl TxOutMembershipProof {
...
-            .map(|(r, h)| TxOutMembershipElement {
-                range: r,
-                hash: TxOutMembershipHash(h),
-            })
-            .collect();
-        hashes.sort();
-
+    /// * `elements` - The tx out membership elements, containing ranges referring to subtrees in the tree, and hashes.
+    ///                These must be provided in the order in which they should be combined to validate the proof.
+    pub fn new(index: u64, highest_index: u64, elements: Vec<TxOutMembershipElement>) -> Self {
```
garbageslam, line 323: the most significant API change is that new takes the sequence of elements in the right order, and doesn't throw them into a hashmap or force the caller to do that. we want to keep them in the right order, which is the same order that they are generated in

mfaulk:
This is looking great! Thank you, Chris!

transaction/core/src/tx.rs
mfaulk: This is great.

garbageslam:
This has passed CD

transaction/core/src/validation/validate.rs
```diff
@@ -302,16 +302,16 @@ fn validate_membership_proofs(
                 return Err(TransactionValidationError::InvalidLedgerContext);
             }
             Ok(derived_proof) => {
-                match derived_proof.elements.last() {
-                    None => {
+                match crate::membership_proofs::compute_implied_merkle_root(&derived_proof) {
```
garbageslam, line 305: this had to change because the previous code assumes that `elements.last()` is the merkle root value given by the proof, but after this revision, the last element is actually the last thing we fold in, we have to call `compute_implied_merkle_root` to get the merkle root element

transaction/core/src/membership_proofs/mod.rs
```diff
@@ -50,6 +50,141 @@ fn hash_nil() -> [u8; 32] {
...
+    // If both are to the left according to this test, then one of the ranges must be reversed,
+    // contrary to the precondition.
+    if !bool::from(a_is_left ^ b_is_left) {
+        return Err(());
+    }
+
+    // Initialize the result as if a is the left and b is right,
+    // then use conditional_assign to fix it if not.
+    // The point of this is to get conditional logic without branching,
+    // subtle::conditional_assign is assumed to be side-channel resistant
```
sugargoat, line 97: Nit - but would maybe reword to indicate we're not just assuming, but subtle's docs [claim to be constant time](https://docs.rs/subtle/0.3.0/subtle/trait.ConditionallyAssignable.html#tymethod.conditional_assign), e.g.

```suggestion
    // subtle::conditional_assign is implemented to be side-channel resistant
```

transaction/core/src/membership_proofs/mod.rs
```diff
@@ -50,6 +50,141 @@ fn hash_nil() -> [u8; 32] {
...
+    result.hash.0 = hash_nodes(&left, &right);
+    Ok(result)
+}
+
+// Helper: Conditionally assign to [u8; 32]
+//
+// It seems subtle doesn't implement ConditionallySelectable on [u8; 32], not sure why,
+// so we can't just use the conditional_assign function from that API.
+//
+// If this is slow then we can later used a faster implementation, but it probably won't be that slow.
```
sugargoat, line 127:
```suggestion
// If this is slow then we can later use a faster implementation, but it probably won't be that slow.
```

transaction/core/src/membership_proofs/mod.rs
```diff
@@ -50,6 +50,141 @@ fn hash_nil() -> [u8; 32] {
...
+    let implied_root = proof.elements[1..].iter().enumerate().try_fold(
+        first,
+        |prev, (idx, next)| -> Result<TxOutMembershipElement, Error> {
+            compose_adjacent_membership_elements(&prev, next)
+                .map_err(|_| Error::UnexpectedMembershipElement(idx + 1))
+        },
+    )?;
+
+    // At this point, we could check that implied_root.range.to = proof.highest_index, or that -1, or something,
+    // but I'm not sure at this moment what exactly the right test is, and I don't think we really need this,
```
sugargoat, line 180: What would you need to be able to be sure what the right test is? Or how could we resolve the uncertainty around whether we need to test any other qualities?

transaction/core/src/membership_proofs/mod.rs
sugargoat: Is this test back? Or am I just missing it?

transaction/core/src/membership_proofs/mod.rs
```diff
@@ -66,96 +201,38 @@ pub fn is_membership_proof_valid(
...
-
-            let expected_hash = hash_nodes(left_child_hash, right_child_hash);
-            if *hash != expected_hash {
-                // Proof contains an incorrect hash value.
-                return Ok(false);
-            }
-        }
+    // This test is carried over from earlier revisions
+    // TODO: Should this be a test for equality?
+    // TODO: Should we also test that implied_root.range.from == 0?
```
sugargoat, line 226: What's needed to resolve these TODOs?

transaction/core/src/membership_proofs/mod.rs
garbageslam: i put it back, i'll send a link

transaction/core/src/membership_proofs/mod.rs
garbageslam: (this line is where that test came back)

garbageslam: to resolve these TODO's, I need more context on all the places this is used I guess -- is there some way that a valid merkle proof is generated that doesn't have implied_root.range.from == 0 ?
can using the derived merkle proof thing cause that?
we can just try it and see if it passes tests
we can ask matt if he sees a problem with that or a reason not to do it?
we can try to read more of the code again to figure this out

probably all of the above :)

transaction/core/src/membership_proofs/mod.rs
garbageslam: that's true -- they also say this: https://github.com/dalek-cryptography/subtle#warning

my take is, code comments here should not make assertions about the correctness of code in another library, because that code can change. code comments should instead state what assumptions this code makes for its own correctness.

so its like a maintenance note, rather than like, marketing.

the marketing pitch is certainly like, subtle is made by world class cryptographers and probably not broken, and i dont want to detract from that.

i just fundamentally don't like to write code comments in one software component that makes assertions about the implementation of a completely different software component, i'd rather just document APIs, contracts, assumptions

sugargoat: Got it - that makes sense to me!

ledger/db/src/tx_out_store.rs
```diff
@@ -1222,17 +1241,14 @@ pub mod tx_out_store_tests {
 
             */
 
-            assert_eq!(7, proof.elements.len());
+            assert_eq!(4, proof.elements.len());
```
mfaulk, line 1244: This might be a good place to test that the elements are in the expected order -- I don't think I've seen that tested specifically in this PR. Something like

```
let expected_ranges = vec![Range::new(3, 3).unwrap(), Range::new(2, 2).unwrap(), Range::new(0, 1).unwrap(), Range::new(4, 7).unwrap()];
assert_eq!(ranges, expected_ranges);
```

Bonus points for updating the ASCII art right above this to explain or illustrate that the hashes of internal nodes are not included because they are implied.

ledger/db/src/tx_out_store.rs
```diff
@@ -1266,18 +1282,15 @@ pub mod tx_out_store_tests {
...
 
             let ranges: Vec<Range> = proof.elements.iter().map(|e| e.range).collect();
```
mfaulk, line 1287: Also a good place to test for the expected order.

transaction/core/src/membership_proofs/mod.rs
garbageslam: ok, i think the highest_index is not supposed to be the same as merkle_root.to, it's supposed to be equal to index i think

i think this is resolved now

garbageslam:
this passed through CD successfully

PR #405, opened 2020-08-28
[MCC-695] Unit tests for onetime_keys

mfaulk:
Soundtrack of this PR: [Raio](https://www.youtube.com/watch?v=i3Z4isDQaiw)

### Motivation

onetime_keys.rs defines several functions used by a sender to create a transaction output, and used by the receiver to determine that they own the output and spend it. The code here has gone through a few a few stages, and so some names and doc comments are out of date or inconsistent. It is also thin on unit tests.

### In this PR
* Adds a module-level doc comment explaining how account keys, subaddresses, and output public keys are related.
* Removes an unused function.
* Moves a function that is not used when working with onetime keys.
* Lots of tweaks for clarity, brevity, and consistency.
* Adds unit tests

### Future Work
* Use proptest to increase testing coverage.

### DIFF
--- a/transaction/core/src/onetime_keys.rs
+++ b/transaction/core/src/onetime_keys.rs
@@ -180,20 +180,20 @@ pub fn recover_onetime_private_key(
     RistrettoPrivate::from(x)
 }
 
-/// Computes the shared secret `aB` from a private key `a` and a public key `B`.
+/// Computes the shared secret `xY` from a private key `x` and a public key `Y`.
 ///
 /// # Arguments
-/// * `public_key` - A public key `B`.
-/// * `private_key` - A private key `a`
+/// * `public_key` - A public key `Y`.
+/// * `private_key` - A private key `x`
 pub fn compute_shared_secret(
     public_key: &RistrettoPublic,
     private_key: &RistrettoPrivate,
 ) -> RistrettoPublic {
-    let a = private_key.as_ref();
-    let B = public_key.as_ref();
-    let aB = a * B;
+    let x = private_key.as_ref();
+    let Y = public_key.as_ref();
+    // let xY = x * Y;
 
-    RistrettoPublic::from(aB)
+    RistrettoPublic::from(x * Y)
 }
 
 /// Generate a tx keypair for a subaddress transaction
--- a/transaction/core/src/onetime_keys.rs
+++ b/transaction/core/src/onetime_keys.rs
@@ -369,8 +369,6 @@ mod tests {
         ));
     }
 
-    // TODO: test recover_onetime_private_key
-
     #[test]
     // Returns the private key corresponding to `onetime_public_key`.
     fn test_recover_onetime_private_key_valid_keypair() {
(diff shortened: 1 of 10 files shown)


transaction/core/src/onetime_keys.rs
```diff
@@ -1,124 +1,174 @@
...
 };
 use mc_account_keys::{PublicAddress, ViewKey};
 use mc_crypto_keys::{RistrettoPrivate, RistrettoPublic};
-use mc_util_from_random::FromRandom;
-use rand_core::{CryptoRng, RngCore};
 
 const G: RistrettoPoint = RISTRETTO_BASEPOINT_POINT;
 
-/// Applies a hash function and returns a RistrettoPoint.
-pub fn hash_to_point(ristretto_public: &RistrettoPublic) -> RistrettoPoint {
```
mfaulk, line 23: This moved to the ring_signature module because it is only used when working with key images.

transaction/core/src/onetime_keys.rs
```diff
@@ -127,48 +177,30 @@ pub fn recover_onetime_private_key(
...
-
-    RistrettoPublic::from(aB)
-}
-
-/// Generate a tx keypair for a subaddress transaction
-///
-/// # Arguments
-/// * `rng` - A RNG`.
-/// * `recipient_spend_key` - A recipient's public spend key `D`
-pub fn generate_tx_keypair<T: CryptoRng + RngCore>(
```
mfaulk, line 164: This was unused.

transaction/core/src/onetime_keys.rs
```diff
@@ -178,126 +213,243 @@ mod tests {
...
+        // (view, spend)
+        let (c, d) = (
+            account.subaddress_view_private(index),
+            account.subaddress_spend_private(index),
+        );
+
+        // (View, Spend)
+        let (C, D) = (RistrettoPublic::from(&c), RistrettoPublic::from(&d));
+        // Look out! The argument ordering here ie weird.
+        let subaddress = PublicAddress::new(&D, &C);
```
mfaulk, line 240: It's a bummer that PublicAddress::new takes the spend key before the view key, which leads to this counter-intuitive order. C and D have the same type, so I'm afraid to just swap the arguments.

rjwalters: There are only a few (3?) places in the code where we explicitly use `::new()` but swapping the order would render all the testnet utxos unspendable... maybe we should fix this for main net as a break change and add a stronger type wrapper on `spend` and `new` at the same time...

transaction/core/src/onetime_keys.rs
```diff
@@ -1,124 +1,177 @@
...
+//!
+//! When creating a transaction, the sender computes a onetime public key for each output in
+//! such a way that only the sender and the recipient know who the recipient is, and only the
+//! recipient is able to compute the corresponding onetime private key that is required to
+//! spend the output.
+//!
+//! To further protect recipient privacy, an output's onetime key is computed for a `subaddress`
+//! that the recipient generated from their CryptoNote-style address. This makes it easy for a
+//! recipient to use different subaddresses for different purposes and keep track of how much
+//! MobileCoin was sent to each subaddress.
```
rjwalters, line 13:
```suggestion
//! mobilecoins was sent to each subaddress.
```

rjwalters, line 13: ... how many mobilecoins were sent ...

transaction/core/src/onetime_keys.rs
```diff
@@ -178,126 +213,243 @@ mod tests {
...
+
+        let onetime_private_key =
+            recover_onetime_private_key(&tx_public_key, account.view_private_key(), &d);
         assert_eq!(
             onetime_public_key,
             RistrettoPublic::from(&onetime_private_key)
         );
     }
 
-    //    #[bench]
```
rjwalters, line 237: should we make a ticket to restore benchmarks someday?

mfaulk:
Will merge after companion PR in internal. This renames a few things, etc. so there are a few updates needed

PR #406, opened 2020-08-28
MCC-613 Reduce encrypted fog hint size to what is actually needed

garbageslam:
Previously it was 128 bytes, because it needed to accomodate two
elliptic curve points in the payload in very early drafts.

In this commit we reduce it to 90 bytes, with explanation:
- 50 bytes for mc-crypto-box header
- 32 bytes for curve point
- 8 bytes of magic / extra padding for future use, filled with
  magic number for now

This is a breaking change for the clients, but it doesn't break the software API I think, they will just have to recompile the SDK to get the new size constant

### DIFF
--- a/api/proto/external.proto
+++ b/api/proto/external.proto
@@ -194,7 +194,7 @@ message TxOut {
     // Encrypted fog hint payload.
     // This is an mc-crypto-box cryptogram for the fog ingest server,
     // or a random cryptogram indistinguishable from a real one.
-    EncryptedFogHint e_account_hint = 4;
+    EncryptedFogHint e_fog_hint = 4;
 }
 
 message TxIn {
--- a/api/proto/external.proto
+++ b/api/proto/external.proto
@@ -191,7 +191,9 @@ message TxOut {
     // Public key.
     CompressedRistretto public_key = 3;
 
-    // 128 byte encrypted fog hint
+    // Encrypted fog hint payload.
+    // This is an mc-crypto-box cryptogram for the fog ingest server,
+    // or a random cryptogram indistinguishable from a real one.
     EncryptedFogHint e_account_hint = 4;
 }
 
--- a/transaction/core/src/encrypted_fog_hint.rs
+++ b/transaction/core/src/encrypted_fog_hint.rs
@@ -10,7 +10,7 @@
 use alloc::{vec, vec::Vec};
 use core::{convert::TryFrom, fmt};
 use generic_array::{
-    typenum::{Diff, Unsigned, U128},
+    typenum::{Diff, Unsigned, U90},
     GenericArray,
 };
 use mc_crypto_box::{CryptoBox, VersionedCryptoBox};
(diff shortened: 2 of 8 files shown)


transaction/core/src/encrypted_fog_hint.rs
```diff
@@ -26,9 +26,9 @@ use rand_core::{CryptoRng, RngCore};
 use serde::{Deserialize, Serialize};
 
 // The length of the encrypted fog hint field in the ledger.
-// Must be at least as large as McRistrettoBox::FooterSize, or it can't hold a
-// Ristretto-Box encryption.
-pub type EncryptedFogHintSize = U128;
+// Must be at least as large as mc_crypto_box::VersionedCryptoBox::FooterSize.
+// Footersize = 50, + 32 for one curve point, + 8 bytes of magic / padding space for future needs
```
rjwalters, line 30: I suggest we drop the padding. I don't see what it is useful for and we can version a new transaction type later if we need more space... so `U82`?

garbageslam: The padding would allow us to make a change where, based on the value of the padding, fog treats something differently, in constant time

Maybe just 2 bytes? U84?

garbageslam: idk, maybe it's pointless, i'm not sure what it would do differently

api/proto/external.proto
```diff
@@ -191,7 +191,9 @@ message TxOut {
     // Public key.
     CompressedRistretto public_key = 3;
 
-    // 128 byte encrypted fog hint
+    // Encrypted fog hint payload.
+    // This is an mc-crypto-box cryptogram for the fog ingest server,
+    // or a random cryptogram indistinguishable from a real one.
     EncryptedFogHint e_account_hint = 4;
```
garbageslam, line 197: @rjwalters @kylefleming @voloshyn do you think we should rename this to `e_fog_hint` ?

transaction/core/src/encrypted_fog_hint.rs
garbageslam: @rjwalters i think 2 bytes is not a terrible price to pay, and it might give us another option to solve some problem later without creating breaking changes. i guess i have in mind some like compliance thing where we can't store some of the users' rng's in sgx the way want to, or must use different code or something. like basically to accommodate a need that isn't motivated by engineering

rjwalters: it seems like this adds a "version" field to fog but I don't think in practice we would need to add a version code to the public records and then ingest wouldn't have any reason to hide it... we should collect other opinions, but I am leaning towards `U82`

api/proto/external.proto
garbageslam: done

transaction/core/src/encrypted_fog_hint.rs
garbageslam: @rjwalters i guess i'm at a point now where its like, i have a few other things i'm trying to do before code freeze, and deleting these magic bytes completely is more work than just reducing it to two magic bytes, so from that point of view, i like this revision. but i also don't really want to fight over two bytes. but at the same time, no one else was saying they think we should get rid of these two bytes.

so at this point i would like to just merge with U84 and do other things. let me know if that sounds alright to you

rjwalters: I know you have a lot on your plate... Maybe leave this open for now and see if you can find time to remove the magic bytes after you finish up the other stuff?

This is the sort of cruft that ends up staying around forever.

sugargoat: In general, I'm happy to see future-proofing efforts, especially if they are low-cost. On the other hand, I can see the wariness of prematurely laying plumbing for something without a clear need for it.  

I would vote for keeping the two magic bytes, will seek another opinion from the team as well.

PR #414, opened 2020-09-02
Use minting ingest pubkey for account hint

sugargoat:
### Motivation

We currently have two separate bootstrap scripts for generating origin blocks for testing. This unifies the two so we can deprecate the other one.

### In this PR
* Use default for ingest minting key, rather than FakeHint

First part of [MCC-1732](https://mobilecoin.atlassian.net/browse/MCC-1732)

### Future Work
* Deprecate bootstrap in internal
* Set CD bootstrap to create multiple smaller blocks rather than one large block

### DIFF
--- a/util/generate-sample-ledger/src/lib.rs
+++ b/util/generate-sample-ledger/src/lib.rs
@@ -94,8 +94,10 @@ pub fn bootstrap_ledger(
     ).expect("File I/O");
 }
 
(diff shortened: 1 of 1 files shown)


sugargoat:
Chatted with @garbageslam - the way generate-sample-ledger is generating the hints is correct. Closing this PR.

PR #415, opened 2020-09-02
[MCC-1693] Removes is_valid check when reinserting 

mfaulk:
Soundtrack of this PR: [Be Together](https://www.youtube.com/watch?v=GZSsOEqgm0c)

### Motivation

TxManager should maintain a cache of well-formed transactions. Previously, it also checked for validity (which is funky and unnecessary) because other things incorrectly required the transactions in its cache to be valid.

### In this PR
* Removes is_valid check when reinserting a transaction into the cache.


rjwalters (approved):

eranrund (approved):
LGTM assuming this has passed deployment (unless you're confident we can skip that)

mfaulk:
Thanks Eran, will deploy before merging

PR #422, opened 2020-09-07
[MCC-1745] Consensus enclave uses compute_implied_merkle_root

mfaulk:
### Motivation

In the change to constant-time Merkle proof validation, we changed the data stored in a TxOutMemebershipProof. Previously, the last element in the proof was the root hash of the Merkle tree used to create the proof, while after the change to constant-time validation, the function compute_implied_merkle_root must be used to obtain the Merkle root hash. Parts of the consensus enclave still assume that the last element of a proof is the root hash, causing occasional panics.

### In this PR
* Consensus enclave uses compute_implied_merkle_root instead of incorrectly assuming that the last element of a proof is the root hash.


eranrund (approved):

garbageslam (approved):
LGTM, sorry i didn't think of this when we merged before

PR #429, opened 2020-09-09
Add bash function wrapper to log services and restart them if they die

nyetwurk:
Soundtrack of this PR: [https://www.youtube.com/watch?v=EgT_us6AsDg](I keep hitting repeat repeat repeat)

### Motivation

The entrypoint starts several backgrounded processes. All of their output goes to the same stdout/stderr. This makes it very difficult to distinguish which messages go with which component.

Also, if any of these processes die, they are not restarted.

### In this PR

This patch sends logs to `NODE_LOG_DIR` if it is set, and restarts any process that exits after 10 seconds.

Log rotation is left to the user; recommend specifying`copytruncate` if using `logrotate`.

### Future Work

* `aesm-service` in our evironment exits immediately with no stdout/stderr message, so it is not handled like the rest of the services. `&` is not needed.
* With `--no-daemon` it does the more expected thing - it stays in foreground and outputs stuff to stdout/stderr on startup.

### DIFF
--- a/ops/entrypoints/consensus_validator.sh
+++ b/ops/entrypoints/consensus_validator.sh
@@ -47,7 +47,7 @@ fi
 # If the ledgerdir is not already populated, copy the origin block.
 if [ ! -r "${NODE_LEDGER_DIR}/data.mdb" ]; then
     mkdir -p "${NODE_LEDGER_DIR}"
-    rsync -a --delete /var/lib/mobilecoin/origin_data/* ${NODE_LEDGER_DIR}/
+    rsync -a /var/lib/mobilecoin/origin_data/* ${NODE_LEDGER_DIR}/
 fi
 
 # Update the ledger to the current version if necessary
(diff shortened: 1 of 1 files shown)


ops/entrypoints/consensus_validator.sh
```diff
@@ -38,14 +56,14 @@ fi
...
+    --dest "${AWS_PATH}"
 fi
 
 # Clean old dump directory - consensus writes a new dir, which is owned by root due to docker volume ownership
 rm -rf /scp-debug-dump/${LOCAL_NODE_ID}
 
-/usr/bin/mc-admin-http-gateway --listen-host 0.0.0.0 --listen-port ${NODE_MANAGEMENT_PORT} --admin-uri insecure-mca://127.0.0.1:9091/ &
+launch /usr/bin/mc-admin-http-gateway --listen-host 0.0.0.0 --listen-port ${NODE_MANAGEMENT_PORT} --admin-uri insecure-mca://127.0.0.1:9091/
 
 exec env consensus-service $@
```
nyetwurk, line 69: that `env` no longer does anything.

sugargoat (approved):
LGTM - may want to incorporate the +e change in launch?

nyetwurk:
> LGTM - may want to incorporate the +e change in launch?

Updated MR

jcape (approved):

PR #436, opened 2020-09-10
Add key utility

sugargoat:
### Motivation

Simple python utility for reformatting account key contents.

### In this PR
* Adds key-util.py

### DIFF
--- a/mobilecoind/clients/python/key-util/key-util.py
+++ b/mobilecoind/clients/python/key-util/key-util.py
@@ -1,0 +1,18 @@
+#!/usr/bin/env python3
+
+# Copyright (c) 2018-2020 MobileCoin Inc.
+
+import argparse
+import json
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(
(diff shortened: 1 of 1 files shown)


eranrund (commented):
I don't think it needs to be in its own directory, and `mobilecoind/clients` is an odd place for it. How about `https://github.com/mobilecoinofficial/mobilecoin/tree/master/tools`?
Additionally, `key-util` is not a descriptive name. Maybe `entropy2hex.py`?
I also think it can just output to stdout, and you could redirect it into a file if you wanted to but I don't feel strongly about any of this.

rjwalters:
seems like `scratchpad` might be a better fit?

sugargoat:
Could be scratchpad - it's part of the offline transaction flow, so it should be in this repo if I add documentation that explicitly walks through offline transactions. I should add that documentation anyway, will do as part of this PR, and move it to tools in the meantime.

sugargoat:
Not needed at this time - can be reopened if we need the utility again.

PR #439, opened 2020-09-11
Ensure origin block contains at least 16 txos

sugargoat:
### Motivation

The origin block should not be created with fewer than 16 outputs. Technically 14 and 15 would work, but don't divide evenly into 250M MOB, so there would be fractional MOBs in the origin block. 

(250_000_000*1_000_000_000_000)/18_446_744_073_709_551_615 = 13.55

### In this PR
* Adds an assert, which enforces a comment in the docstring that it should panic if < 16 txs.

### DIFF
--- a/util/generate-sample-ledger/src/lib.rs
+++ b/util/generate-sample-ledger/src/lib.rs
@@ -38,6 +38,8 @@ pub fn bootstrap_ledger(
     let mut db = LedgerDB::open(path.clone()).expect("Could not open ledger_db");
 
     let num_outputs: u64 = (recipients.len() * outputs_per_recipient_per_block * num_blocks) as u64;
+    assert!(num_outputs >= 16);
+
     let picomob_per_output: u64 = (TOTAL_MOB / num_outputs) * 1_000_000_000_000;
 
     println!("recipients: {}", recipients.len());


util/generate-sample-ledger/src/lib.rs
```diff
@@ -38,6 +38,8 @@ pub fn bootstrap_ledger(
     let mut db = LedgerDB::open(path.clone()).expect("Could not open ledger_db");
 
     let num_outputs: u64 = (recipients.len() * outputs_per_recipient_per_block * num_blocks) as u64;
+    assert!(num_outputs >= 16);
```
kylefleming, line 41: Should this be `== 16` since `MAX_OUTPUTS` is 16?

rjwalters:
There's no requirement that all of the TXOs have the same value. We can make a tidy 250M as:

> (16) * 15_625_000	
> (14) * 16_666_666 + (1) 16_666_676	
> (13) * 17_857_142 + (1) 17_857_154

all of these combos stay until the max allowed value of `18_446_744.073_709_500_000` (u64 max)

But I think 16 is a fine choice.

rjwalters (approved):

util/generate-sample-ledger/src/lib.rs
sugargoat: The origin block is the only case where we can create more outputs than 16 (that's how we make 100K outputs for testing).

mfaulk (approved):

PR #443, opened 2020-09-11
Remove more unnecessary mutex (based on #441)

garbageslam:
Remove unnecessary mutex around peer_keepalive
    
    This follows up on the commit that removes the mutex around
    byzantine ledger. The peer_keepalive mutex is only there to allow
    stopping the service. We change this from `Arc<Mutex<PeerKeepalive>`
    to `Option<Arc<PeerKeepalive>>`, and replace to call to `lock().stop()`
    in the stop function with `= None` which simply drops the value.
    
    We also make the other Arc's that are holding this, in lambda functions,
    into weak references. This prevents strong reference cycles which would
    prevent the drop from happening and cause a resource leak.
    
    We also put a code comment about why not to remove the mutex around
    threaded broadcaster.

### DIFF
--- a/consensus/service/src/consensus_service.rs
+++ b/consensus/service/src/consensus_service.rs
@@ -100,14 +100,20 @@ pub struct ConsensusService<
     consensus_msgs_from_network: BackgroundWorkQueue<IncomingConsensusMsg>,
 
     peer_manager: ConnectionManager<PeerConnection<E>>,
+    // This mutex is required because ThreadedBroadcaster API cannot be used concurrently,
+    // the LRU cache is not thread-safe among othter reasons.
+    // But there is only one ByzantineLedger worker thread anyways, so this should be uncontended.
     broadcaster: Arc<Mutex<ThreadedBroadcaster>>,
     tx_manager: Arc<TXM>,
-    peer_keepalive: Arc<Mutex<PeerKeepalive>>,
+    // Option is only here because we need a way to drop the PeerKeepalive without mutex,
+    // if we want to implement Stop as currently concieved
+    peer_keepalive: Option<Arc<PeerKeepalive>>,
 
     admin_rpc_server: Option<AdminServer>,
     consensus_rpc_server: Option<Server>,
     user_rpc_server: Option<Server>,
-    // Option is only here because we need a way to drop the ByzantineLedger without mutex.
+    // Option is only here because we need a way to drop the ByzantineLedger without mutex,
+    // if we want to implement Stop as currently concieved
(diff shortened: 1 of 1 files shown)


garbageslam:
rebased on master (after 441 merge)

consensus/service/src/consensus_service.rs
```diff
@@ -471,8 +482,7 @@ impl<
                     );
 
                     // Keep track that we heard from the sender of this message.
-                    {
-                        let peer_keepalive = peer_keepalive.lock().expect("mutex poisoned");
+                    if let Some(peer_keepalive) = peer_keepalive_weak.upgrade() {
```
mfaulk, line 485: Is there a situation where consensus_service would be running after the peer_keepalive has been dropped? I'm wondering if `upgrade().expect()` would be better than `if let Some`.

mfaulk (approved):

consensus/service/src/consensus_service.rs
garbageslam: yeah good call, that seems reasonable

eranrund (approved):
LGTM once k8 passes.

consensus/service/src/consensus_service.rs
garbageslam: @mfaulk i guess the reason is, if someone calls `consensus_service.stop()` then this should get dropped, and that will cause the weak reference to return none. so if we panic when it becomes `none`, then the server will crash when we try to stop it

one way to deal with this might be, to try to change which of these objects are actually owning eachother. i think the only party that talks to the peer_keepalive variable may be, the `consensus_msgs_from_network` thread. so we could try to make that the exclusive owner, and just have it be an `Box` that gets moved into that thread

it seems to me that the main reason we have all these member variables of `consensus_service` is to support the `pub fn stop(&mut self)` thing, which stops all the threads without tearing down any of the objects. i'm not sure the `stop` thing is actually useful, it might be simpler to just stop the threads in the `drop` calls. That might reduce the amount of code overall, and in case like the peer_keepalive, it seems that we end up with mutexes only because we need exclusive access to stop the object, which doesn't seem like a good trade. `drop` has signature `drop(&mut self)` -- the language is ensuring that the call to the destructor is exclusive, so that's a good time to join things, IMO. that's just one opinion though, maybe there's a good use case for stopping without tearing down the objects.

garbageslam:
this is deployed as diogenes/97: https://deploybot.canadacentral.mobilecoin.com/job/mobilecoin-internal/job/deploy%2Fdiogenes/97/console

garbageslam:
very little change in slam perf:
pr443

PR #444, opened 2020-09-11
Moves ScpNode trait into separate file.

mfaulk:
### Motivation

I often find that I want to quickly get to the ScpNode trait, which is 150 lines down into node.rs, or I want to hop back and forth between the Node implementation and its unit tests, in which case the ScpNode trait is just 50 lines of clutter.

### In this PR
- Moves the ScpNode trait into a separate file. It may just be personal preference, but I feel like this is more pleasant to navigate.

### DIFF
--- a/consensus/scp/src/node/mod.rs
+++ b/consensus/scp/src/node/mod.rs
@@ -5,5 +5,5 @@
 mod node_impl;
 mod node_trait;
 
-pub use node_impl::*;
-pub use node_trait::*;
+pub use node_impl::Node;
+pub use node_trait::ScpNode;
--- a/consensus/scp/src/node/mod.rs
+++ b/consensus/scp/src/node/mod.rs
@@ -7,3 +7,6 @@ mod node_trait;
 
 pub use node_impl::Node;
 pub use node_trait::ScpNode;
+
+#[cfg(test)]
+pub use node_trait::MockScpNode;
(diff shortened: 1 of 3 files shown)


eranrund (approved):

consensus/scp/src/node/node_impl.rs
```diff
@@ -6,13 +6,13 @@ use crate::{
...
     slot::{ScpSlot, Slot, SlotMetrics},
+    ScpNode,
 };
 use mc_common::{
     logger::{log, Logger},
     NodeID,
 };
-#[cfg(test)]
-use mockall::*;
+
```
jcape, line 15:
```suggestion
```

consensus/scp/src/node/mod.rs
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2018-2020 MobileCoin Inc.
+
+//! A node of the consensus network.
+
+mod node_impl;
+mod node_trait;
+
+pub use node_impl::*;
+pub use node_trait::*;
```
jcape, line 9: Unless it's causing issues, I'd recommend this explicitly uses the elements you need, rather than this, because this will also export the results of use statements in these nodes, AFAIK.

jcape (approved):
Some formatting/export nitpicks.

PR #447, opened 2020-09-14
MCC-1790 Fix verifier not doing multiple MRENCLAVE/SIGNER

jcape:
### Motivation

Currently, we check all the given status verifiers to ensure they all pass, which defeats the purpose. This treats them separately, and says "if any of the status verifiers pass, consider that a pass, and if any of the other verifiers fail, consider that a fail."

### In this PR
* Put all the status verifiers into their own collection, which is evaluated independently.


garbageslam (approved):
LGTM

kylefleming (approved):

PR #448, opened 2020-09-14
[MCC-1791] Resolve Homebrew issues in CI

kylefleming:
### Motivation

PR https://github.com/mobilecoinofficial/mobilecoin/pull/326 added `brew update` to all CI builds to resolve an issue with using an old version of Homebrew on CircleCI. Running `brew update` solved the issue by updating Homebrew before using it to install any dependencies. This added ~33s to each CI run.

This PR assumes that CircleCI is using a newer version of Homebrew by now. As such, this PR removes the call to `brew update` for all CI runs, and instead only runs it when building on master, which is done to ensure that the lastest versions of dependencies are saved to the cache (which is saved as part of the CI run on master).

This PR also fixes a minor yaml schema issue where parameters that are meant to be strings are considered by yaml to be integers because they are parsable as such. This issue should be harmless but is fixed regardless.

### In this PR
* Updates CircleCI's macOS executor image to Xcode 11.7.0
* Only run `brew update` during CI setup on master
* Quote integers that are meant to be strings in CircleCI config

Fixes [MCC-1791]

[MCC-1791]: https://mobilecoin.atlassian.net/browse/MCC-1791

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,7 +37,7 @@ executors:
 
   macos:
     macos:
-      xcode: 11.6.0
+      xcode: 11.7.0
     environment:
       HOMEBREW_NO_AUTO_UPDATE: "1"
       HOMEBREW_NO_INSTALL_CLEANUP: "1"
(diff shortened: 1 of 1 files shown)


garbageslam (approved):
thank you!

voloshyn (approved):

PR #450, opened 2020-09-15
MCC-1790 Fix node-to-node failures introduced by last PR

jcape:
### Motivation

Node-to-node attestation began failing after the most recent commit related to this branch. There is a nuance with the `Iterator::any()` method: it will return false if there is nothing to iterate---which is the case for node-to-node attestation.

### In this PR
* Fix the case where a verifier has no status checking to perform, and add regression tests.
* Fix some issues with whitespace introduced in spite of commit hook.


sugargoat (approved):

voloshyn (approved):
LGTM

attest/core/src/ias/verifier.rs
```diff
@@ -531,14 +532,16 @@ impl IasReportVerifier {
 
         let report_data = VerificationReportData::try_from(report)?;
 
-        if self
-            .or_verifiers
-            .iter()
-            .any(|verifier| verifier.verify(&report_data))
-            && self
+        if (self.and_verifiers.is_empty()
```
kylefleming, line 535: `self.and_verifiers.is_empty()` should be necessary since `Iterator.all` returns `true` if the iterator is empty.

Reference: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.all

PR #452, opened 2020-09-16
[Minor] Comment Change

sugargoat:
### Motivation

A nice property of mobilecoind's monitor_id is that it is not secret material. This PR updates a comment to reflect this.


eranrund (approved):
To be fair, hashing sensitive data would not make monitor_id sensitive. If we can digest that, then that is equally good.

sugargoat:
@eranrund - good point

voloshyn (approved):

jcape (approved):

PR #453, opened 2020-09-16
Integration test for mobilecoind / json gateway / mirror

tsegaran:
### Motivation

We have been checking our binary releases for various client-side components manually. This script automates downloading, extracting and running all the executables and verifies that they works as expected.

### In this PR
* Creates `mirror-test-json.py`
* Downloads specified tarball, defaulting to latest
* Launches mobilecoind, mirror-public, mirror-private and mobilecoind-json
* Runs a series of requests against the REST services to ensure they act as expected

### Future Work
* More tests on mirror service
* Set up to run against fresh builds for CI testing

### DIFF
--- a/tools/release-tests/mirror-json-test.py
+++ b/tools/release-tests/mirror-json-test.py
@@ -1,0 +1,183 @@
+#!/usr/bin/env python3
+
+# This is an integration test that downloads the latest releases and attempts to run mobilecoind,
+# the mirror services and the JSON gateway in the exact manner that a user might run them, using the 
+# provided shell scripts and compiled binaries.
+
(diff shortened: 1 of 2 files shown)


tsegaran:
Note that this is designed to run against the current release, forthcoming API changes will require modification, I am happy to do those before merging this

sugargoat (approved):
Looks great. Adding the ability to run the integration test from current HEAD of both mobilecoin and mobilecoind-mirror will make this test valuable in local development and CD as well, but that can be in a future PR, which would also address the remaining TODOs outlined here.

rjwalters (approved):

PR #456, opened 2020-09-17
add TestNet management scripts

rjwalters:
Soundtrack of this PR: [trip hop is still a thing](https://soundcloud.com/jedischillout/hugo-kant-exclusive-live-for-jco)

### Motivation

We manage the TestNet with a Mailchimp database. These scripts might go into `util` but they do serve as additional examples of how to use the `mobilecoin` python package so `clients/python` seems like a reasonable choice. I think it's best to check these in since we've been maintaining them (and using them) for a few months now.

### In this PR
* `allocate_mobilecoins` - this looks for new TestNet signups, allocates TestNet tokens, and updates the mailchimp db
* `download_members` - this checks the balances for all users. Note that this gets impractically slow as the ledger grows since each user has a unique master key

### Future Work
* we could do a better job with gRPC error handling
* we could move some functions into the `mobilecoin` package [MCC-1804](https://mobilecoin.atlassian.net/browse/MCC-1804)


rjwalters:
pulling until library is updated

PR #461, opened 2020-09-18
improve pypi build documentation

rjwalters:
This marks the updated PyPi release after #460.


kylefleming (approved):
LGTM, although I think the title should be changed, right?

eranrund (approved):

PR #463, opened 2020-09-22
Return b58 encoding of public address as part of public address endpoint

tsegaran:
### Motivation
Requesting a printable public address currently requires two round-trips, we have a feature request to simplify this by including the b58 code in GetPublicAddress

This has been run through integration testing

### In this PR
* Adds b58-encoded Public Address in GetPublicAddress in mobilecoind
* Add to JSON interface

### Future Work
* Potentially allow b58 codes used directly in transaction creation functions, further reducing round-trips


rjwalters (approved):

sugargoat (approved):
LGTM! Would be good to add a public address example to the readme, showing the b58 rsponse as well!

PR #465, opened 2020-09-23
update account_info client example for mobilecoin 0.2.0

rjwalters:
### Motivation

We recently made a breaking change to the python API which requires updates for all python examples.

### In this PR
* get_<> --> create_<>
* add an arg for setting `first_block`


rjwalters:
I will reopen later

PR #468, opened 2020-09-24
Delete sgx backtrace

garbageslam:
This deletes the `sgx backtrace` crate and `sgx unwind`
- We aren't using it
- It isn't known to ever have worked in hardware mode, we need to fix something with the enclave loading address compensation
- it is likely a security problem even if it works since it reveals the contents of the stack
- The unwinding code is some pretty nasty C code that may have memory unsafety problems
- Several auditors raised questions about this, and we've been planning to remove it for a while

After this commit, `mc-sgx-panic` only has a `panic=abort` configuration

---

Note this is based on #467 but I couldn't find a way to show only the final commit in this PR


garbageslam:
this needs to be rebased now because of the "Squash merge" feature

garbageslam:
now rebased on master

sgx/urts/Cargo.toml
```diff
@@ -5,20 +5,12 @@ authors = ["MobileCoin"]
 
 [features]
 default = [ ]
```
jcape, line 7:
```suggestion
```

jcape (approved):

jcape (approved):

eranrund (approved):

mfaulk:
Screenshot from 2020-09-24 15-32-16

Nice

PR #470, opened 2020-09-24
adjust mc-util-build-info implementation details

garbageslam:
I have frequently noticed that the act of simply `git commit`'ing
your code often triggers many targets in the project to build.
I think this is because when the constants exposed by `mc-util-build-info`
change, all the reverse dependencies must be recompiled, because the
compiler cannot know that the optimizer in those translation units
doesn't start inlining those values and doing things like
`core::mem::size_of(GIT_COMMIT)` at build time.

Other projects use a different strategy where a function is used
that returns a static reference to a string, like so.
The point of this is that the optimizer cannot inline the function
across crate boundaries
unless we mark it inline, so even if the value returned by it changes,
nothing else needs to be recompiled because the API of this crate
didn't change.

So, I think this may help fix our too-frequent-rebuilds issue.
It may also make our build more cacheable and so make CI faster,
but that's speculative.

https://docs.rs/vergen/0.1.1/vergen/fn.vergen.html


jcape (approved):
LGTM, but also, `vergen` seems like it would be a great addition at a later date.

garbageslam:
I remember that we had a constraint that whatever happens in the enclave, you can't bake the build time and date in there or the MRENCLAVE won't be reproducible, but we could use something like `vergen` outside the enclave if we wanted

eranrund (approved):

PR #476, opened 2020-09-25
Adds moburl to returned proto when creating a request code

tsegaran:
### Motivation

Product spec includes QR codes for request codes in desktop clients, these should be encoded as mob_urls 

### In this PR
* Return mob_url  as part of CreateRequestCode in mobilecoind

### Future Work
* If we like this approach, also add to transfer codes?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2873,6 +2873,7 @@ dependencies = [
  "mc-util-repr-bytes",
  "mc-util-serial",
  "mc-util-uri",
+ "mc-util-url-encoding",
  "mc-watcher",
  "more-asserts",
  "num_cpus",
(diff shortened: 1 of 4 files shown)


jcape (approved):

rjwalters (approved):
Please create a Jira ticket to extend the moburl scheme to cover transfer codes. We will need this for the desktop client to export gift codes.

voloshyn (changes requested):
Please add mob_url property to Transfer and PublicAddress payloads. Those are needed for the mobile wallet.

tsegaran:
I have added the mob_url to the public address call, and created a ticket to do it for transfer codes

PR #478, opened 2020-09-26
MCC-1862 Consolidate mc-sgx-libc-types with mc-sgx-types, confirm errno values

garbageslam:
The mc-sgx-libc-types contains a bunch of unused types that are copied
from baidu, and some duplicated with mc-sgx-types, which I think is
our preferred thing now.

The `mc-sgx-types` crate is defining a `SysError` type which we are
using, but it isn't providing the values of `SysError` which is wierd.

This fixes it by putting the errno values under `libc::SysError` in
the `mc-sgx-types` crate. Then it makes the `sgx-sync` crate, which
was getting `SysError` from here, use the errno values also from there.
Then it deletes the unused `mc-sgx-libc-types` crate and deletes
documentation about it.

---

We are not using `condvar` or `rwlock` in our enclave AFAIK, so I think this is non-functional change.
It probably is not necessary for launch, also because I don't think we are actually changing any errno
values, it's just code movement and dead code deletion.


garbageslam:
there might be some reason that it's preferable to have the libc errno values in a separate crate, if like, to make them accessible in the `sgx-compat` layer. but i don't think we have such a need, because nothing in that crate is returning errno values AFAIK. So I think this PR makes sense

eranrund (approved):

sgx/README.md
```diff
@@ -67,20 +67,12 @@ Crates for outside the enclave:
 
 | Crate     | Description | Dependencies |
 | --------- |------------ | ------------ |
-| `mc-sgx-urts` | This crate suports untrusted code that creates an enclave. It also contains OCALL implementations. | rust `std`, `mc-sgx-libc-types`, and for backtraces, `rustc_demangle` and `backtrace_sys` |
-| `mc-sgx-build` | Shared code for build.rs scripts that link to SGX | |
+| `mc-sgx-urts` | This crate suports untrusted code that creates an enclave. It also contains OCALL implementations. | rust `std`, `slog` for logging.
```
jcape, line 70:
```suggestion
| `mc-sgx-urts` | This crate suports untrusted code that creates an enclave. It also contains OCALL implementations. | rust `std`, `slog` for logging. | |
```

jcape (approved):

PR #483, opened 2020-10-01
Update rust_decimal for testnet-client

sugargoat:
### Motivation

rust-decimal is out of date and throws 

```
error[E0599]: no method named `powf` found for type `f64` in the current scope
    --> /tmp/mobilenode/cargo/registry/src/github.com-1ecc6299db9ec823/rust_decimal-1.8.1/src/decimal.rs:2666:34
     |
2666 |             let round_to = 10f64.powf(self.scale() as f64);
     |                                  ^^^^ method not found in `f64`
```

### In this PR

* Updates rust-decimal package


jcape (changes requested):
I appreciate the need to re-enable `rust_decimal`'s `std` feature (`core::f64` does not support the APIs that `rust_decimal` needs), but that doesn't require changing the consensus enclave's `Cargo.lock` at all.

Recommend reverting both `Cargo.lock` files to what exists in `master` and doing `cargo clean && cargo build`, then updating git appropriately. As it stands now, it looks like both lock files were deleted and re-created, effectively upgrading all our dependencies at once.

sugargoat:
@jcape - updated the Cargo.lock to only include rust-decimal changes.

Also removed the testnet-client release scripts, as they can live with the other release scripts in internal.

kylefleming (approved):

jcape (approved):

PR #484, opened 2020-10-01
python mobilecoin package v0.2.1

rjwalters:
### Motivation

Further improvements to the python library are needed for API changes and convenience.

### In this PR
* reorganizes the package code and adds some tests using nose
* introduces python utility functions `wait_for_ledger`, `wait_for_monitor`, `display_as_MOB`, `parse_tx_status`
* exports some constants from the python library
* reimplements examples to use the new library
* MCC-1833 - cmd line tool for sending payments
* cli benchmark tool

### Future Work
* update the terminal wallet to use package v0.2.1


mobilecoind/clients/python/blockchain_explorer/blockchain_explorer.py
```diff
@@ -1,9 +1,14 @@
...
+
 # Copyright (c) 2018-2020 MobileCoin Inc.
 
-import argparse, datetime, sys
+import argparse, datetime
 from flask import Flask, render_template
 
-from mobilecoin import Client
+import os,sys
+sys.path.insert(1, os.path.realpath(os.path.join(os.path.pardir, "lib")))
```
rjwalters, line 10: This allows us to test the package code in `lib` without pushing a release to pypi.

tsegaran (approved):
Wow, this is a ton of work, thank you for adding test coverage

mobilecoind/clients/python/cli/README.md
```diff
@@ -0,0 +1,114 @@
+## Python CLI tools for `mobilecoind`
+
+### Getting your Public Address
+
+Every 32-byte Master Key defines a unique MobileCoin account. You can share the corresponding public address with other users by sending them the base-58 encoded "Address Code" or the "Address URL" generated by this CLI tool. The "Address URL" can be conveniently embedded as the payload of a QR code.
+
+Note that this command creates a monitor for the provided or generated Master Key that will remain active until removed.
+
+Help:
+```
```
sugargoat, line 10: I would suggest not including the help in the README, so that it's easier to maintain the README, and users can run help as needed.

mobilecoind/clients/python/cli/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+Send a payment.
+
+Help:
+```
+usage: send_payment.py [-h] --sender SENDER --recipient RECIPIENT [-v VALUE]
+                       [--sender-subaddress [SENDER_SUBADDRESS]]
+                       [--recipient-subaddress [RECIPIENT_SUBADDRESS]]
+
+Send all available funds (default) or VALUE pMOB from a sender's Master Key to a recipient, specified by Master Key or Address Code.
```
sugargoat, line 79: I'm not sure that draining the account should be the default. Would maybe suggest erroring if no value is provided.

mobilecoind/clients/python/mailchimp/allocate_mobilecoins.py
```diff
@@ -0,0 +1,189 @@
...
+            if blocks_per_second > 0:
+                time_remaining_seconds = blocks_remaining / blocks_per_second
+                print("#    {} blocks remain ({} seconds)".format(blocks_remaining, round(time_remaining_seconds, 1)))
+            else:
+                print("#    {} blocks remain (? seconds)".format(blocks_remaining))
+            (monitor_is_behind, next_block, remote_count, blocks_per_second) = mobilecoind.wait_for_monitor(monitor_id)
+        print("# sender's  monitor has processed all {} blocks\n#".format(local_count))
+    return remote_count
+
+# generates a new master key, allocates funds, stores data at Mailchimp and triggers the welcome email
```
sugargoat, line 33: Prefer [PEP-8 styling](https://www.python.org/dev/peps/pep-0008/#documentation-strings) for doc strings

sugargoat (changes requested):
Looks good - my one blocking request is the unexpected behavior of sending without a value would cause your account to be drained.

mobilecoind/clients/python/cli/README.md
rjwalters: I find myself wanting to fold one wallet into another fairly often, usually because I'm done with a key or because I moved funds to a key and then a script failed before the key was assigned etc... I can make the user add "--all" instead of having this as a default.

sugargoat: I think `--all` makes sense - the negative experience of accidentally sending all your tokens seems like it should have a protection against it.

sugargoat (approved):

PR #485, opened 2020-10-01
Create `api` module

mfaulk:
### Motivation

Some file reorg.

### In this PR
* Creates an `api` module and moves the various gRPC API files into it.
* A little cleanup of how things are imported.
* Renames some things from FetchFoo to GetFoo, which is shorter and (I think) more idiomatic.


consensus/service/src/lib.rs
```diff
@@ -12,12 +12,8 @@ pub mod consensus_service;
 pub mod tx_manager;
 pub mod validators; // Public so that it can be benchmarked by the `benchmarks` crate.
 
-mod attested_api_service;
+mod api;
```
mfaulk, line 15: This module is the main change.

eranrund (approved):

xoloki (approved):
LGTM

PR #486, opened 2020-10-02
Fix a minor issue in mc-util-keyfile crate

garbageslam:
The keyfile crate writes default keyfiles and has a function
to read default keyfiles from a directory. However, due to a subtle
sorting issue, the i'th file in the returned vector is not
`account_keys_{i}.json`. This fixes that.

If this was not already causing problems in tests, then I don't think
it will cause a problem to commit this, and it will help the fog
conformance tests which are being developed.

### DIFF
--- a/util/keyfile/src/keygen.rs
+++ b/util/keyfile/src/keygen.rs
@@ -8,7 +8,12 @@ use crate::{read_keyfile, read_pubfile, write_keyfile, write_pubfile};
 use mc_account_keys::{AccountKey, PublicAddress, RootIdentity};
 use rand::SeedableRng;
 use rand_hc::Hc128Rng as FixedRng;
-use std::{ffi::OsStr, fs, path::Path};
+use std::{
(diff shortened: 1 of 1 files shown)


mfaulk (approved):

rjwalters (approved):
I still think it might be better to just deterministically generate keys as they are used rather than rely on disk io in tests.

garbageslam:
@rjwalters yeah the main argument against that is, sometimes we need to test things that aren't in rust and can't call that function directly. we could expose it in libmobilecoin, but one use-case for testing is, we want to test if we broke sdks using old versions of libmobilecoin. if libmobilecoin is the only way that any code can get these keys, then there isn't a real versioning story

i think it's also useful to have pub files and json files as a concept, because for instance, `generate-sample-ledger` uses only the pub files. so you could in principle ask the investors to send pub files when you bootstrap the genesis block

PR #487, opened 2020-10-02
add shell script integration test for mobilecoind-json

rjwalters:
### Motivation

More tests are needed for `mobilecoind-json`.

### In this PR
* A shell script test that sends a transaction between two subaddresses.

### Future Work
* MCC-1888: we need to expose more API in order to check that the ledger is current
* MCC-1768: we should add more unit tests using the [rocket framework](https://rocket.rs/v0.4/guide/testing/)

### DIFF
--- a/mobilecoind-json/tests/test.sh
+++ b/mobilecoind-json/tests/test.sh
@@ -20,13 +20,21 @@ ACCOUNT_KEY=`curl -s localhost:9090/entropy/${ENTROPY}`
 
 echo "Sending a test payment for master key = $ENTROPY"
 
-MONITOR_ID=`curl -s localhost:9090/monitors \
-  -d '{"account_key": '"$ACCOUNT_KEY"', "first_subaddress": 0, "num_subaddresses": 10}' \
-  -X POST -H 'Content-Type: application/json' \
-  | jq -r '.monitor_id'`
+MONITOR_ID_RESPONSE=`curl -s localhost:9090/monitors \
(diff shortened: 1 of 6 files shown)


mobilecoind-json/tests/test.sh
```diff
@@ -0,0 +1,122 @@
...
+ACCOUNT_KEY=`curl -s localhost:9090/entropy/${ENTROPY}`
+
+echo "Sending a test payment for master key = $ENTROPY"
+
+MONITOR_ID=`curl -s localhost:9090/monitors \
+  -d '{"account_key": '"$ACCOUNT_KEY"', "first_subaddress": 0, "num_subaddresses": 10}' \
+  -X POST -H 'Content-Type: application/json' \
+  | jq -r '.monitor_id'`
+
+# wait until the ledger is current
```
rjwalters, line 28: I created a ticket to expose the API for `GetNetworkStatus` so that we can do this properly. As it is now, the transaction is likely to fail since the monitor will outrun the ledger download and the script will continue without utxos to spend.

sugargoat (changes requested):
Thanks for putting together this test for mobilecoind-json. As discussed in [mobilecoind-mirror #6](https://github.com/mobilecoinofficial/mobilecoind-mirror/pull/6), we want to design a testing strategy for mobilecoind-json using mocks and unittests, while working on incorporating the [existing integration test](https://github.com/mobilecoinofficial/mobilecoin/blob/master/tools/release-tests/mirror-json-test.py) into CI/CD. I don't think this test fits into either of those frameworks, so I'm not sure we should invest in a shell-script oriented testing strategy at this time.

PR #506, opened 2020-10-13
remove unused encodings

rjwalters:
### Motivation

We have settled on using b58-encoded protobuf encodings and decided to use "mob58://<b58>" as the equivalent URL. 

This PR removes `mob-url` and `b58-payload` since these are no longer used.

### In this PR
* [fixes MCC-1940](https://mobilecoin.atlassian.net/browse/MCC-1940)
* [fixes MCC-1959](https://mobilecoin.atlassian.net/browse/MCC-1959)


mobilecoind/clients/python/cli/check_balance.py
```diff
@@ -52,7 +52,7 @@
     print("    {:<18}{}".format("Master Key:", args.key))
     print("    {:<18}{}".format("Subaddress Index:", args.subaddress))
     print("    {:<18}{}".format("Address Code:", public_address.b58_code))
-    print("    {:<18}{}".format("Address URL:", public_address.mob_url))
+    print("    {:<18}{}".format("Address URL:", "mob58://"+ public_address.b58_code))
```
kylefleming, line 55: `mob58://`? Why not just have it print out the b58 code without the url prefix?

kylefleming: Actually, I see that it prints it out on the line before, so why not just remove this line entirely?

rjwalters: we've decided that our convention for deeplinks will be to pass a b58 code with the URI "mob58://"

kylefleming (approved):

eranrund (approved):

PR #507, opened 2020-10-13
Flip the order of the generators back to match cryptonote

xoloki:
Soundtrack of this PR: [https://www.youtube.com/watch?v=oTUSeac7IuQ](When Doves Cry)

### Motivation

Since integrating bulletproofs, our code uses dalek's default Pedersen generator.  This means that we swap the G and H values relative to most cryptonote implementations.   So this change flips them back.

### In this PR
* Change the order of the generators

### Future Work
This will allow us to more easily integrate future protocol changes from other cryptonote style currencies.


mfaulk:
Assuming that this works when deployed, looks good!

mfaulk (approved):

sugargoat (approved):

PR #511, opened 2020-10-15
Fix a documentation error in the ledger db, we had block height when …

garbageslam:
…it was really block index


eranrund (approved):

rjwalters:
image

do you want to also update these return values?

garbageslam:
@rjwalters i think the actual users of the API know what it is actually is, it was only a documentation issue, so i don't think anything else should be changed. but we could look and see if there are any code comments that need fixing at those sites

rjwalters (approved):
lgtm

PR #531, opened 2020-10-22
Cleanup generic array references, add more_lengths

jcape:
### Motivation

The `generic-array` crate does not include `ArrayLength<T>` implementations for all the types in `typenum`, and in fact only does `U0` through `U32` by default. This adds the `more_lengths` feature to the use of `generic-array` in the `mc-util-repr-bytes` crate, and removes `default-features = false` where it appears, as it is not necessary.

### In this PR
* Cleanup cargo inclusion of `generic-array`
* Enable the `more_lengths` feature in `mc-util-repr-bytes` use of `generic-array`.

### DIFF
--- a/crypto/digestible/Cargo.toml
+++ b/crypto/digestible/Cargo.toml
@@ -7,7 +7,7 @@ edition = "2018"
 [dependencies]
 cfg-if = "0.1"
 merlin = { version = "2.0", default-features = false }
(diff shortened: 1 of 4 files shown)


sugargoat (approved):

eranrund (approved):

garbageslam (approved):
thank you!

PR #534, opened 2020-10-22
MCC-2026 Make enclave errors always return PERMISSION_DENIED

jcape:
### Motivation

Restarting a node on the network deletes any peer session IDs they have. As a result, the next time the peer goes to perform an action, it will receive an enclave error. Unfortunately this is translated into the INVALID_ARGUMENT gRPC status code, which does not trigger a re-attestation. Without re-attestation, enclaves within the network cannot communicate with each other, leading to network collapse.

### In this PR
* Treat any RPC failure of an attested call as cause to re-attest.
* Explicitly return PERMISSION_DENIED for enclave errors.

### DIFF
--- a/connection/src/traits.rs
+++ b/connection/src/traits.rs
@@ -43,10 +43,8 @@ pub trait AttestedConnection: Connection {
 
         let result = func(self);
 
-        if let Err(GrpcError::RpcFailure(rpc_status)) = &result {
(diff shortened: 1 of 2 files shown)


eranrund (approved):

garbageslam (approved):
LGTM

PR #535, opened 2020-10-23
Build report.proto in mobilecoin python library, and expose

garbageslam:
This adds a `fog.py` module to the mobilecoin python library, and exposes
a simple `fetch_fog_reports` function which wraps the grpc API of the fog
report server.

This is useful because we can use this in the fog conformance tests, which
are also in python and need to know the fog public key.

That's not directly related to mobilecoind, but it is very convenient to be able to
`pip install mobilecoin` and have this functionality available.

Let me know what you think about this


garbageslam:
note: i haven't tested this properly yet, but I did test compiling the protos

garbageslam:
this might be a bad idea, it might be simpler to use grpccurl from python code

PR #536, opened 2020-10-23
Add grpcurl utility to build docker image

garbageslam:
The motivation is, this can be used to do health checks with the report server, and to actually get the fog pubkey, from a python script, using a CLI json interface. The utility is a few megabytes.


docker/Dockerfile
```diff
@@ -113,6 +113,13 @@ RUN wget https://github.com/SimonKagstrom/kcov/archive/v36.tar.gz && \
 COPY --chown=root:root install-bolt.sh /tmp
 RUN /bin/bash /tmp/install-bolt.sh
 
+# Install grpcurl
+RUN mkdir -p /tmp/grpcurl
+WORKDIR /tmp/grpcurl
+RUN wget https://github.com/fullstorydev/grpcurl/releases/download/v1.7.0/grpcurl_1.7.0_linux_x86_64.tar.gz && \
+    tar xvf grpcurl_1.7.0_linux_x86_64.tar.gz && \
+    mv grpcurl /usr/bin/
```
sugargoat, line 121: Do you still have to do things like set the GOPATH etc? Maybe a very short example usage, even just as a comment on this PR

garbageslam: i don't think i need to do that -- I think it's a self-contained binary and doesn't depend on go being installed, and i didn't have to do that in my experiments yet, in my local env or in the mob prompt itself

garbageslam:
i guess this might be a bad direction, I realized that the pubkey bytes themselves are no longer in the `Report` protobuf, so if I want to do this from python, I need to find a CBOR library and read the VerificationReport. I think it might be better to just make a rust cli program that just wraps the `fog-report-connection` functionality like I was originally going to do

PR #549, opened 2020-10-30
MCC-2012 Limit gRPC thread counts in mobilecoind, mobilecoind-json

jcape:
### Motivation

Some mobilecoind users have seen intermittent failures under load, when messages to an enclave are delivered out of order, which is possible with unary gRPC. This PR attempts to reduce the incidence of these errors by making `mobilecoind` and `mobilecoind-json` use single-threaded call/completion queues in their consensus-facing gRPC environments.

### In this PR
* Sets `cq_count(1)` on gRPC environments in `mobilecoind` and `mobilecoind-json`

### Future Work
* MCC-2092 Optional explicit nonces in attested call requests/responses.


eranrund (approved):
Hacky indeed but lets see where it gets us

sugargoat (approved):

PR #555, opened 2020-11-09
[MCC_2031] Slot::out_msg integer underflow

mfaulk:
### Motivation

Previously, the logic for "clamping" P and PP when creating an SCP message could panic with an integer underflow if the ballot counter B.N is zero. 

### In this PR
* Extracts a 'clamp' function that explicitly returns None when B.N is zero instead of panicking,
* Uses 'clamp' to remove some code duplication,
* Adds unit tests

This feels like an under-specified part of the SCP protocol, so I'm trying to fill in the gaps with something reasonable.


eranrund (approved):

rjwalters (approved):
nit... `clamp` makes me think that you are forcing a value into a range... like `max(min(...` but the logic here is def more subtle... maybe `limit` would be more clear.

jcape:
What needs to change for this to get merged?

PR #560, opened 2020-11-20
[MCC-1909] ConsensusService incorrectly reports blocks_behind in JSON info blob

mfaulk:
### Motivation

Users have reported that the consensus service incorrectly reports `blocks_behind` in the JSON info blob from the Admin gRPC service.

### In this PR
* Separates JSON formatting from the database queries for server status.
* Adds more explicit error handling.

### DIFF
--- a/consensus/service/src/consensus_service.rs
+++ b/consensus/service/src/consensus_service.rs
@@ -794,12 +794,6 @@ fn config_and_status_as_json(
     is_behind: bool,
     highest_peer_block_index: u64,
 ) -> String {
-    // Number of blocks in the local ledger.
-    let block_height = highest_block.index + 1;
-
-    // ID of the last block in the local ledger.
-    let highest_block_id = hex::encode(&highest_block.id.0);
-
     // The approximate time in which the block was signed, represented at seconds
     //  of UTC time since Unix epoch 1970-01-01T00:00:00Z.
     let highest_block_signed_at: Option<u64> =
@@ -808,7 +802,8 @@ fn config_and_status_as_json(
     let sync_status = if is_behind { "catchup" } else { "synced" };
 
     // Number of blocks that the local node is behind, or zero if the local node is not behind.
-    // TODO: it's funky that this is both an Option and a saturating value...
+    // (It's funky that this is both an Option and a saturating value. It would probably be simpler to
+    // return the local node's block index and the highest peers' block index.)
     let blocks_behind = Some(highest_peer_block_index.saturating_sub(highest_block.index));
 
     json!({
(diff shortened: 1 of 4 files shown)


consensus/service/src/consensus_service.rs
```diff
@@ -636,70 +646,165 @@ impl<
     /// Helper method for creating the get config json function needed by the GRPC admin service.
     fn create_get_config_json_fn(&self) -> GetConfigJsonFn {
         let ledger_db = self.ledger_db.clone();
-        let byzantine_ledger = self
+
+        let byzantine_ledger: Weak<OnceCell<ByzantineLedger>> = self
```
mfaulk, line 650: Added an explicit type because CLion was having trouble inferring it.

consensus/service/src/consensus_service.rs
```diff
@@ -636,70 +646,165 @@ impl<
     /// Helper method for creating the get config json function needed by the GRPC admin service.
     fn create_get_config_json_fn(&self) -> GetConfigJsonFn {
```
mfaulk, line 647: The idea here is to organize the function around getting the ledger status (highest_block, block_signature) and getting the sync status (is_behind, highest_peer_block_index).

consensus/service/src/consensus_service.rs
```diff
@@ -636,70 +646,165 @@ impl<
...
-                        .map_err(|e| log::error!(logger, "Error getting block {} {:?}", b - 1, e))
-                        .ok();
-
-                    latest_block_timestamp = match ledger_db.get_block_signature(b - 1) {
-                        Ok(x) => Some(x.signed_at()),
-                        // Note, a block signature will be missing if the corresponding block was not
+            // The highest block in the local ledger, and its BlockSignature if available.
+            let (highest_block, highest_block_signature): (Block, Option<BlockSignature>) = {
+                // Number of blocks in the local ledger.
+                let num_blocks: u64 = ledger_db.num_blocks().map_err(|e| {
```
mfaulk, line 662: This now returns an error if ledger_db returns an error. Previously, the function would still return JSON but the result would contain null-ish values for any fields that could not be computed.

consensus/service/src/consensus_service.rs
```diff
@@ -636,70 +646,165 @@ impl<
...
+/// Format the service's configuration and ledger status as JSON.
+///
+/// # Arguments
+/// * `config` - ConsensusService configurations.
+/// * `network_config` - Network configuration.
+/// * `highest_block` - Highest block in the local ledger.
+/// * `highest_block_signature` - Signature, if any, for the highest block in the local ledger.
+/// * `is_behind` - True if the local ledger is behind the peers' ledgers.
+/// * `highest_peer_block_index` -  Highest block index agreed upon by peers.
+fn config_and_status_as_json(
```
mfaulk, line 787: This is a pure function that renders the configuration and status data as JSON. It is currently difficult to unit test ConsensusService, so extracting some of the logic like this makes it easier to test.

consensus/service/src/consensus_service.rs
```diff
@@ -636,70 +646,165 @@ impl<
...
+) -> String {
+    // The approximate time in which the block was signed, represented at seconds
+    //  of UTC time since Unix epoch 1970-01-01T00:00:00Z.
+    let highest_block_signed_at: Option<u64> =
+        highest_block_signature.map(|signature| signature.signed_at());
+
+    let sync_status = if is_behind { "catchup" } else { "synced" };
+
+    // Number of blocks that the local node is behind, or zero if the local node is not behind.
+    // (It's funky that this is both an Option and a saturating value. It would probably be simpler to
```
mfaulk, line 803: Is it OK to change the JSON structure? So far, I've kept the JSON fields as-is, but I think there's room for improvement.

sugargoat: Yeah absolutely

PR #564, opened 2020-11-23
Set hint in bootstrap

sugargoat:
### Motivation

We would like to be able to set the hint arbitrarily in bootstrap.

### In this PR
* Takes an arbitrary string as input to bootstrap and writes the bytes to the hint field for the generated txos.


transaction/core/src/encrypted_fog_hint.rs
```diff
@@ -103,12 +103,31 @@ impl EncryptedFogHint {
...
+        } else {
+            GenericArray::<
+                u8,
+                Diff<
+                    EncryptedFogHintSize,
+                    <VersionedCryptoBox as CryptoBox<Ristretto>>::FooterSize,
+                >,
+            >::default()
+        };
+
```
garbageslam, line 130: can we just make this a separate function?
it seems like this alternate mode of fake one-time hint will almost never be used, and so making it part of the API for the version that it used all the time seems not so good

transaction/core/src/encrypted_fog_hint.rs
```diff
@@ -103,12 +103,31 @@ impl EncryptedFogHint {
...
-        >::default();
+        let plaintext = if let Some(pt) = plaintext_buf {
+            // Truncate the plaintext to a buffer of the correct size
+            GenericArray::<
+                u8,
+                Diff<
+                    EncryptedFogHintSize,
+                    <VersionedCryptoBox as CryptoBox<Ristretto>>::FooterSize,
+                >,
+            >::clone_from_slice(pt)
```
garbageslam, line 120: this will panic if plaintext_buf is not exactly the right size

transaction/core/src/encrypted_fog_hint.rs
sugargoat: sgtm 👍

garbageslam:
I think right now, this just won't work as intended, because, fake_one_time_hint creates a random public key to encrypt against. Because no one knows the private key, and no one can ever find the private key, no one will ever be able to decrypt the message.

Here's what I suggest:

(1) Make a new function, not `fake_onetime_hint`, called `personalized_fog_hint_insecure` or something like this, that takes a byte buffer, and copies it over the 84 byte buffer for the encrypted fog hint, so like, padding with 0's or spaces if the message isn't that long (and ideally return an error if the message is too long).

(2) Don't ecnrypt it at all, it will not satisfy the cryptographic requirements of the onetime hint field. And that's fine, since this is just to allow someone to decorate the origin block right? Fog doesn't need to read the origin block anyways.

transaction/core/src/encrypted_fog_hint.rs
sugargoat: I ran into that in testing as well - the len is 38 bytes, which was a bit hard to track down, as it's referencing [Ristretto(u32) + TagSize](https://github.com/mobilecoinfoundation/mobilecoin/blob/master/crypto/box/src/hkdf_blake2b_aes_128_gcm.rs#L21). Will see if I can derive the expected len and then slice to the correct len before sending to GenericArray. Otherwise, I can note in a comment that the new fn will panic if not passed the correct size.

sugargoat:
@garbageslam - great points. I think we could just use EncryptedFogHint::new(&[u8]) in that case, if that sounds alright. Then the changes are isolated to the bootstrap code.

util/generate-sample-ledger/src/lib.rs
```diff
@@ -52,14 +57,28 @@ pub fn bootstrap_ledger(
...
 
     for block_index in 0..num_blocks as u64 {
         println!("Creating block {} of {}.", block_index, num_blocks);
 
         let mut outputs: Vec<TxOut> = Vec::new();
         for recipient in recipients {
             for _i in 0..outputs_per_recipient_per_block {
-                outputs.push(create_output(recipient, picomob_per_output, &mut rng));
+                let hint_slice = if let Some(ht) = hint_text {
+                    let hint_chunk = ht.len() / num_outputs as usize;
```
garbageslam, line 69: might be better to use `hint_chunk = min(ENCRYPTED_FOG_HINT_LEN, ht.len() - hint_cursor)` ?

garbageslam: otherwise like, if you have a hint with 50 characters, and you have 1000 outputs, hint_chunk will be 0 every time due to rounding down, and no hints will be written at all

garbageslam: @sugargoat it might be a good idea to make some kind of unit test for this?

sugargoat: @garbageslam - I added an assert prior to making the outputs that the hint text divided among the txouts is > 0. We will need to decide whether we want all the hints to be the same in the origin block, or whether we want them all to be different, and likely padded. I think we might actually be leaning toward same.

util/generate-sample-ledger/src/bin/generate_sample_ledger.rs
```diff
@@ -20,11 +21,17 @@ struct Config {
...
+
+    #[structopt(long = "hint-text")]
+    pub hint_text: Option<String>,
 }
 
 fn main() {
     let config = Config::from_args();
 
+    mc_common::setup_panic_handler();
+    let (logger, _global_logger_guard) = create_app_logger(o!());
```
garbageslam, line 33: recommend to use `create_stderr_logger` for command line tools that are not servers, or you probably won't see the log messages (because the program will close on exit and not flush the buffer)

sugargoat: Interesting - I do actually see the log output with the current configuration

garbageslam: what we do elsewhere is like this:

```
use mc_common::logger::create_root_logger;

...
fn main() {
    // Logging must go to stderr to not interfere with STDOUT
    std::env::set_var("MC_LOG_STDERR", "1");
...
    mc_watcher::watcher_db::create_or_open_rw_watcher_db(config.watcher, &[], create_root_logger())
}
```

garbageslam (approved):
looks good other than app logger

util/generate-sample-ledger/src/bin/generate_sample_ledger.rs
garbageslam: you probably see some but not the parts at the end, you would have to put a `thread::sleep` before exiting main

sugargoat: Switched to create_root_logger (which wraps create_stderr_logger)

eranrund (approved):

PR #565, opened 2020-11-24
Revert "[MCC-1969] Removes pseudo_outputs from range proof"

mfaulk:
Reverts mobilecoinfoundation/mobilecoin#542

This will be included in a later enclave upgrade.

### DIFF
--- a/transaction/core/src/ring_signature/mod.rs
+++ b/transaction/core/src/ring_signature/mod.rs
@@ -4,7 +4,7 @@
 #![macro_use]
 extern crate alloc;
 
-use crate::{constants::MAX_OUTPUTS, domain_separators::HASH_TO_POINT_DOMAIN_TAG};
+use crate::domain_separators::HASH_TO_POINT_DOMAIN_TAG;
 use blake2::{Blake2b, Digest};
 use bulletproofs::{BulletproofGens, PedersenGens};
 pub use curve25519_dalek::scalar::Scalar;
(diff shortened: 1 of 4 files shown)


sugargoat (approved):

transaction/core/src/ring_signature/mod.rs
```diff
@@ -33,9 +33,9 @@ lazy_static! {
 
     /// Generators (base points) for Bulletproofs.
     /// The `party_capacity` is the maximum number of values in one proof. It should
-    /// be at least MAX_OUTPUTS to allow for all outputs in a transaction
+    /// be at least 2 * MAX_INPUTS + MAX_OUTPUTS, which allows for inputs, pseudo outputs, and outputs.
```
xoloki, line 36: Orthogonal to our immediate need, but were we ever including range proofs for the inputs?

transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -556,7 +581,7 @@ mod rct_bulletproofs_tests {
                 fee,
                 &mut rng,
             );
-            assert_eq!(result, Ok(()));
+            assert!(result.is_ok());
```
xoloki, line 584: It should be safe to leave this in place.

xoloki (approved):
A couple of comments, but nothing blocking.  Given the time contrainsts, I vote merge now and think about my comments later :)

transaction/core/src/ring_signature/mod.rs
mfaulk: Great question. Inputs are assumed to be in the valid because they are already in the ledger. If that is the "inductive step", the "base case" is that all outputs created by the origin block are valid. We prove that by publishing the view key for the origin block.

PR #569, opened 2020-12-01
[MCC-2138] Add custom fog hint support to TransactionBuilder

kylefleming:
### Motivation

This PR adds support for providing a separate `PublicAddress` to be used when making the fog hint for the output of a transaction. This allows the recipient of the `TxOut` itself and the public key contained inside the fog hint to be different.

The API added in this PR is intended to be used for transactions outputs addressed to the user creating the transaction, such as in the case of the change output of a transaction. This provides a way to use a single view key for fog while allowing the `TxOut` itself to be addressed to a different account or a different subaddress within the same account.

It should be noted that under normal circumstances, it would be considered griefing to send a transaction with an output that has mismatching recipient/fog hint. However, because only fog ingest can decrypt the fog hint once encrypted, there is no way for consensus to prevent this sort of griefing.

### In this PR
* Adds support for specifying the `PublicAddress` to use when creating the fog hint.

Fixes [MCC-2138]

[MCC-2138]: https://mobilecoin.atlassian.net/browse/MCC-2138

### DIFF
--- a/transaction/std/src/transaction_builder.rs
+++ b/transaction/std/src/transaction_builder.rs
@@ -56,7 +56,6 @@ impl TransactionBuilder {
     /// * `recipient` - The recipient's public address
     /// * `recipient_fog_ingest_key` - The recipient's fog server's public key
     /// * `rng` - RNG used to generate blinding for commitment
-    ///
     pub fn add_output<RNG: CryptoRng + RngCore>(
         &mut self,
         value: u64,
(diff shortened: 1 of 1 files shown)


garbageslam (approved):
LGTM

Maybe add a comment explaining that, it is usually a bug to use this except for special cases?

garbageslam:
Another approach that might be more user-friendly is, instead of adding this function, add a way to specify an alternate subaddress for the change

rjwalters:
> Another approach that might be more user-friendly is, instead of adding this function, add a way to specify an alternate subaddress for the change

The layer above this in the api/sdk abstraction is prob. the right place to put the more user-friendly options.

rjwalters (approved):

kylefleming:
> Another approach that might be more user-friendly is, instead of adding this function, add a way to specify an alternate subaddress for the change

It's needed for more than just change, e.g. self-payments.

voloshyn (approved):

sugargoat (approved):
Awesome - this also helps address https://mobilecoin.atlassian.net/browse/MCC-2135

PR #576, opened 2020-12-07
Encrypt account key at rest in mobilecoind

sugargoat:
### Motivation

Currently, the account key is stored in the mobillecoind-db in plaintext. This is fine in scenarios where the DB is treated with proper isolation and security, but ideally, the sensitive material should always be encrypted at rest.

### In this PR
* Adds ability to include password hash in call to AddMonitor, so that the account is stored encrypted at rest. It is up to the caller to construct the password hash according to their needs, e.g. with argon2.
* Adds an in-memory password hash table for passwords which are active for the session. This is required in order for mobilecoind to decrypt account keys from the DB during sync.

MCC-2146

### Future Work
* Expose password API in mobilecoind-json
* Utilize password API in desktop wallet


mobilecoind/src/monitor_store.rs
```diff
@@ -63,31 +73,95 @@ impl MonitorData {
...
-        Ok(Self {
-            account_key,
-            first_subaddress,
-            num_subaddresses,
-            first_block,
-            // The next block we need to sync is our first block.
-            next_block: first_block,
-            name: name.to_owned(),
-        })
+        if let Some(pw) = password_hash {
```
sugargoat, line 76: This is the main motivation for this PR.

mobilecoind/src/sync.rs
```diff
@@ -232,12 +241,41 @@ fn sync_thread_entry_point(
...
+                match mobilecoind_db.get_monitor_data(&monitor_id) {
+                    Ok(data) => {
+                        if let Some(_encrypted_account_key) = data.encrypted_account_key {
+                            // If we don't have this password yet, see whether it's been added.
+                            if !passwords.contains_key(&monitor_id) {
+                                // Drain the password channel and update the map
+                                for pw_msg in monitor_passwords_receiver.iter() {
+                                    match pw_msg {
+                                        MonitorPwMessage::MonitorPw(pw_hash) => {
+                                            passwords.insert(monitor_id, pw_hash);
```
garbageslam, line 262: do you need to check return value of this, or the idea is you don't care?

mobilecoind/src/monitor_store.rs
```diff
@@ -63,31 +76,117 @@ impl MonitorData {
...
+            let nonce_arr = GenericArray::from_slice(&nonce[..]);
+
+            // Get cipher from hash bytes
+            let cipher = Aes256Gcm::new(key_arr);
+
+            let plaintext_bytes = mc_util_serial::encode(&account_key);
+
+            let ciphertext = cipher
+                .encrypt(nonce_arr, &plaintext_bytes[..])
+                .expect("encryption failure!"); // NOTE: handle this error to avoid panics!
```
garbageslam, line 92: iirc, the aes-gcm encrypt function only panics if the plaintext bytes are like > 4GB or something

garbageslam: you could return the AEAD error if you want though

garbageslam (approved):
LGTM

sugargoat:
Closing in favor of: https://github.com/mobilecoinfoundation/mobilecoin/pull/577

PR #581, opened 2020-12-08
MCC-710 Final changes regarded FullyValidatedFogPubkey and TransactionBuilder

garbageslam:
There are a few problems with the TransactionBuilder API in regards
to fog ingest public keys.

(1) It makes the API user match public addresses with ingest public keys.
If they do this wrong, the money will not be found by the recipient.
(2) It doesn't enforce that the ingest public keys that are used are
fully validated using the type system.
(3) It doesn't enforce the rule that the tombstone block of the transaction
may not exceed the pubkey_expiry of any of the fog public keys that used.

This change makes the TransactionBuilder take a set of FullyValidatedPublicKeys
object when building transactions. This object can be populated by taking
a FogPubkeyResolver and passing that, with each recipient PublicAddress,
to the object, which caches them for a short period of time.
The cache is consumed by the TransactionBuilder, reflecting the idea
that these should be fetched again with each transaction.

This caching also automatically minimizes network traffic since we
skip the second request to fog report server if two recipients have
the same fog url and report id.

This is a good idea because it turns out that in testnet block 42,
we have all zeroes for fog hints, which suggests that some client
in the past was doing fog hints wrong.

The present change was proposed months ago to try to make it harder
to use the transaction builder wrongly with respect to fog, and
hopefully reduce the number of bugs that lead to lost/missing transactions
or bad fog hints, that are hard debug due to use of encryption.

An advantage of the proposed API is that it is compatible with a device that
wants to creates off-line transactions:
- They would fetch the FullyValidatedFogPubkeys first, online, and serialize that
- Then move that data offline and use the transaction builder, fully offline.
- Then move the resulting Tx to online machine to submit the transaction.

### DIFF
--- a/fog/report/connection/src/lib.rs
+++ b/fog/report/connection/src/lib.rs
@@ -1,172 +1,104 @@
 // Copyright (c) 2018-2020 MobileCoin Inc.
 
-//! Fog Report Connection handles resolving the public key from the report service
-//! on making grpc to the reports endpoint
+#![deny(missing_docs)]
 
-use core::str::FromStr;
+//! Fog Report Connection handles connecting to the fog report service and
+//! building up a FogReportResponses object needed to create a transaction
+//! with fog recipients.
+
 use displaydoc::Display;
 use grpcio::{ChannelBuilder, Environment};
-use mc_account_keys::PublicAddress;
-use mc_attest_core::{VerificationReport, Verifier};
 use mc_common::logger::{log, o, Logger};
 use mc_fog_api::report_grpc;
-use mc_fog_report_validation::ingest_report::{Error as IngestReportError, IngestReportVerifier};
-use mc_util_grpc::{BasicCredentials, ConnectionUriGrpcioChannel};
-use mc_util_uri::{ConnectionUri, FogUri, UriParseError};
+use mc_util_grpc::ConnectionUriGrpcioChannel;
+use mc_util_uri::FogUri;
 use std::sync::Arc;
 
-pub use mc_fog_report_validation::{FogPubkeyResolver, FullyValidatedFogPubkey};
+pub use mc_fog_report_validation::FogReportResponses;
 
-#[derive(Debug, Display)]
-pub enum Error {
-    /// Recipient doesn't have fog
-    RecipientHasNoFog,
-    /// Invalid fog url: {0}
-    InvalidFogUrl(UriParseError),
-    /// grpc failure: {0}
-    Rpc(grpcio::Error),
-    /// deserialization failed: {0}
-    DeserializationFailed(mc_util_serial::decode::Error),
-    /// report rejected: {0}
-    Rejected(IngestReportError),
-    /// Fog Report Server has no available reports
-    NoReports,
-    /// Matching report not found
-    NotFound,
+/// Fog report server connection based on grpcio
+///
+/// TODO: As an optimization, it might be good to make this like a connection
+/// pool that holds onto its grpc connections, so that they can be reused
+/// if a client is constructing transactions frequently.
+/// The user of this API would tear down this object when they want these
+/// connections to be closed.
+#[derive(Clone)]
+pub struct GrpcFogReportConnection {
+    /// grpc environment
+    env: Arc<Environment>,
+    /// The logging instance
+    logger: Logger,
 }
 
-impl From<UriParseError> for Error {
-    fn from(src: UriParseError) -> Self {
-        Self::InvalidFogUrl(src)
+impl GrpcFogReportConnection {
+    /// Create a new GrpcFogReportConnection object
+    pub fn new(env: Arc<Environment>, logger: Logger) -> Self {
+        Self { env, logger }
     }
+
+    /// Fetch fog reports corresponding to a series of FogUris.
+    /// This attempts to be efficient, not contacting a server twice if a FogUri appears twice.
+    pub fn fetch_fog_reports(
+        &self,
+        uris: impl Iterator<Item = FogUri>,
+    ) -> Result<FogReportResponses, Error> {
+        let mut responses = FogReportResponses::default();
+        for uri in uris {
+            self.fetch_fog_report(&mut responses, &uri)?;
+        }
+        Ok(responses)
+    }
+
+    /// Given a set of previously collected FogReportResponse's, and another Uri, make a corresponding
+    /// request and add it to the collection, if such a response is not already present.
+    pub fn fetch_fog_report(
+        &self,
+        responses: &mut FogReportResponses,
+        uri: &FogUri,
+    ) -> Result<(), Error> {
+        match responses.entry(uri.to_string()) {
+            std::collections::btree_map::Entry::Occupied(_) => Ok(()),
+            std::collections::btree_map::Entry::Vacant(ent) => {
+                let logger = self.logger.new(o!("mc.fog.cxn" => uri.to_string()));
+
+                // Build channel to this URI
+                let ch = ChannelBuilder::default_channel_builder(self.env.clone())
+                    .connect_to_uri(uri, &logger);
+                let report_grpc_client = report_grpc::ReportApiClient::new(ch);
+
+                // Request reports
+                let req = mc_fog_api::report::ReportRequest::new();
+                let resp = report_grpc_client.get_reports(&req)?;
+
+                if resp.reports.len() == 0 {
+                    log::warn!(
+                        self.logger,
+                        "Report server at {} has no available reports",
+                        uri
+                    );
+                    return Err(Error::NoReports);
+                }
+
+                // Store entire response, for later validation against measurement and public addresses
+                ent.insert(resp);
+                Ok(())
+            }
+        }
+    }
+}
+
+/// Errors that can occur during GrpcFogReportConnection operation
+#[derive(Debug, Display)]
+pub enum Error {
+    /// grpc failure: {0}
+    Rpc(grpcio::Error),
+    /// Fog Report Server has no available reports
+    NoReports,
 }
 
 impl From<grpcio::Error> for Error {
     fn from(src: grpcio::Error) -> Self {
         Self::Rpc(src)
-    }
-}
-
-impl From<mc_util_serial::decode::Error> for Error {
-    fn from(src: mc_util_serial::decode::Error) -> Self {
-        Self::DeserializationFailed(src)
-    }
-}
-
-impl From<IngestReportError> for Error {
-    fn from(src: IngestReportError) -> Self {
-        Self::Rejected(src)
-    }
-}
-
-/// Fog Pubkey resolver based on grpcio
-pub struct GrpcFogPubkeyResolver {
-    /// Ingest report verifier
-    verifier: IngestReportVerifier,
-    /// The logging instance
-    logger: Logger,
-    /// grpc environment
-    env: Arc<Environment>,
-}
-
-impl GrpcFogPubkeyResolver {
-    pub fn new(verifier: &Verifier, env: Arc<Environment>, logger: Logger) -> Self {
-        Self {
-            verifier: IngestReportVerifier::from(verifier),
-            env,
-            logger,
-        }
-    }
-}
-
-impl FogPubkeyResolver for GrpcFogPubkeyResolver {
-    type Error = Error;
-
-    fn get_fog_pubkey(&self, recipient: &PublicAddress) -> Result<FullyValidatedFogPubkey, Error> {
-        let fog_report_url_str = recipient.fog_report_url().ok_or(Error::RecipientHasNoFog)?;
-        let fog_report_url = FogUri::from_str(fog_report_url_str)?;
-        let fog_report_id_str = recipient.fog_report_id().unwrap_or("");
-
-        let logger = self
-            .logger
-            .new(o!("mc.fog.cxn" => fog_report_url.to_string()));
-
-        let creds = BasicCredentials::new(&fog_report_url.username(), &fog_report_url.password());
-
-        // Build channel to this URI
-        // FIXME: We must get the TLS fingerprints and do fingerprint checking with sig
-        let ch = ChannelBuilder::default_channel_builder(self.env.clone())
-            .connect_to_uri(&fog_report_url, &logger);
-        let report_grpc_client = report_grpc::ReportApiClient::new(ch);
-
-        // Request reports
-        let req = mc_fog_api::report::ReportRequest::new();
-        let resp = report_grpc_client.get_reports_opt(&req, creds.call_option()?)?;
-
-        if resp.reports.len() == 0 {
-            log::warn!(
-                self.logger,
-                "Report server at {} has no available reports",
-                fog_report_url_str
-            );
-            return Err(Error::NoReports);
-        }
-
-        for rep in resp.reports.iter() {
-            // Make sure the fog_url, which came from ingest --fqdn config param,
-            // matches Alice's fog_url in her public identity
-            // Note: Neither of these strings is expected to have scheme or port
-            // So we should possibly load them as url's and normalize them by
-            // removing those things before doing the test.
-            log::debug!(
-                logger,
-                "Comparing found fog_report_id: '{}', with recipient fog_report_id: '{}'",
-                rep.fog_report_id,
-                fog_report_id_str,
-            );
-            if rep.fog_report_id == fog_report_id_str {
-                // Get the pubkey from the attestation evidence
-                // NOTE: We are not doing a key exchange, so we do not care about Ingest's ResponderId
-                let remote_report: VerificationReport = mc_util_serial::deserialize(&rep.report)
-                    .map_err(|err| {
-                        log::error!(
-                            logger,
-                            "Failed deserializing fog ingest VerificationReport: report_id: '{}', err = {}",
-                            fog_report_id_str,
-                            err
-                        );
-                        err
-                    })?;
-
-                // Validate report
-                let pubkey = self.verifier.validate_ingest_ias_report(remote_report).map_err(|err| {
-                        log::error!(
-                            logger,
-                            "Failed validating fog ingest VerificationReport: report_id: '{}', err = {}",
-                            fog_report_id_str,
-                            err
-                        );
-                        err
-                    })?;
-                return Ok(FullyValidatedFogPubkey {
-                    pubkey,
-                    pubkey_expiry: rep.pubkey_expiry,
-                    fog_report_id: rep.fog_report_id.clone(),
-                });
-            }
-        }
-
-        let found_urls: Vec<String> = resp
-            .reports
-            .iter()
-            .map(|rep| format!("'{}'", rep.fog_report_id))
-            .collect();
-        log::error!(
-            logger,
-            "Could not find needed fog report among the returned reports: wanted id: '{}', found ids: [{}].",
-            fog_report_id_str,
-            found_urls.join(", ")
-        );
-        Err(Error::NotFound)
     }
 }
--- a/fog/report/connection/src/lib.rs
+++ b/fog/report/connection/src/lib.rs
@@ -9,7 +9,10 @@
 use displaydoc::Display;
 use grpcio::{ChannelBuilder, Environment};
 use mc_common::logger::{log, o, Logger};
-use mc_fog_api::report_grpc;
+use mc_fog_api::{
+    report::{ReportRequest, ReportResponse},
+    report_grpc,
+};
 use mc_util_grpc::ConnectionUriGrpcioChannel;
 use mc_util_uri::FogUri;
 use std::sync::Arc;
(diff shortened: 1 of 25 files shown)


fog/report/validation/src/lib.rs
```diff
@@ -57,3 +68,92 @@ pub struct FullyValidatedFogPubkey {
...
+    fn table_key(&self, addr: &PublicAddress) -> Option<(String, String)> {
+        addr.fog_report_url().map(|url| {
+            (
+                url.to_string(),
+                addr.fog_report_id().unwrap_or("").to_string(),
+            )
+        })
+    }
+
+    /// Get the smallest pubkey_expiry value of any of these fog pubkeys.
```
xoloki, line 121: Why do you want the smallest expiry?  Doesn't it make more sense to use the key that expires later?

xoloki: Oh, I think I see what's going on.  You aren't selecting between keys for an individual user, you're trying to determine an appropriate tombstone block based on all the keys needed for an individual transaction.

xoloki (approved):
LGTM

fog/report/validation/src/lib.rs
```diff
@@ -57,3 +68,92 @@ pub struct FullyValidatedFogPubkey {
...
+                Err(FogPubkeyError::FogPubkeyNotPrefetched(
+                    table_key.0,
+                    table_key.1,
+                ))
+            }
+        } else {
+            Err(FogPubkeyError::NoFogReportUrl)
+        }
+    }
+}
```
garbageslam, line 150: a problem with this is that we're not validating the fog sig of each user

it might be better to say, this class should hold the reports that passed IAS validation, but when we `get_fog_pubkey` for each respective user, that's when we check the sig

fog/report/validation/src/lib.rs
```diff
@@ -57,3 +68,92 @@ pub struct FullyValidatedFogPubkey {
     /// transaction larger than this limit if the fog pubkey is used.
     pub pubkey_expiry: u64,
 }
+
+/// Represents a set of fully validated fog public keys
+///
+/// These can be fetched before building a transaction, then the rest of the
+/// transaction building process can happen offline.
+#[derive(Default, Debug)]
+pub struct FullyValidatedFogPubkeys {
```
garbageslam, line 77: we could also call this class `PrefetchedFogPubkeys`

transaction/std/src/transaction_builder.rs
```diff
@@ -108,12 +111,16 @@ impl TransactionBuilder {
...
     /// # Arguments
     /// * `tombstone_block` - Tombstone block number.
-    pub fn set_tombstone_block(&mut self, tombstone_block: u64) {
-        self.tombstone_block = tombstone_block;
+    pub fn set_tombstone_block(&mut self, tombstone_block: u64) -> u64 {
+        self.tombstone_block = core::cmp::min(
+            tombstone_block,
+            self.fog_pubkeys.smallest_pubkey_expiry_value(),
+        );
+        self.tombstone_block
```
garbageslam, line 123: another option here might be to return an error if it exceeds smallest_pubkey_expiry_value

but the caller can also easily test if clamping occurred

fog/report/validation/src/lib.rs
garbageslam: idk, the code for checking the sig doesn't exist quite yet, so maybe we can do that later

fog/report/validation/src/lib.rs
```diff
@@ -1,17 +1,26 @@
+//! Logic for representing fog public keys from the fog-report server
+//! that have been fully validated, and the associated metadata.
+
 // Copyright (c) 2018-2020 MobileCoin Inc.
 
+#![deny(missing_docs)]
```
mfaulk, line 6: Thanks

fog/report/validation/src/lib.rs
mfaulk: Feels a little odd to have both a type for `FullyValidatedFogPubkey` and `FullyValidatedFogPubkeys`, but I'm OK with it if it works for you.

transaction/std/src/transaction_builder.rs
```diff
@@ -28,16 +29,21 @@ pub struct TransactionBuilder {
...
-    pub fn new() -> Self {
+    ///
+    /// # Arguments
+    /// * `fog_pubkeys` - Set of prefetched fog public keys to choose from
+    pub fn new(fog_pubkeys: FullyValidatedFogPubkeys) -> Self {
         TransactionBuilder {
             input_credentials: Vec::new(),
             outputs_and_shared_secrets: Vec::new(),
-            tombstone_block: u64::max_value(),
+            tombstone_block: fog_pubkeys.smallest_pubkey_expiry_value(),
```
mfaulk, line 44: We ought to enforce that this is not too large: there is a constant MAX_TOMBSTONE_BLOCKS, which is supposed to limit how long-lived a transaction may be, and transaction validation will reject a transaction that has a tombstone block too far in the future.

Tangentially, I always find "tombstone block" to be an awkward term. Something like "not_valid_after" (to borrow a term from TLS) or |expires_on" might make the code read more like idiomatic English. Just thinking out loud; now is probably not the time to rename it.

transaction/std/src/transaction_builder.rs
mfaulk: Ditto about  MAX_TOMBSTONE_BLOCKS

mfaulk (approved):
LGTM

transaction/std/src/transaction_builder.rs
kylefleming, line 44: In the event that `FullyValidatedFogPubkeys` contains no fog pubkeys and the user doesn't explicity call `TransactionBuilder::set_tombstone_block`, then the built `Tx` will have a `tombstone_block` that will cause it to be rejected by consensus for being too far in the future, meaning that in this situation, `TransactionBuilder::set_tombstone_block` _must_ be called in order to build a valid Tx (which was the case before this PR also).

Because of this, I'm not convinced there's utility in setting the default `tombstone_block` to anything other than `u64::max_value()`. My reasoning is, why default to a value that might or might not be a valid `tombstone_block` in certain circumstances and is definitely not a valid `tombstone_block` in other circumstances? Seems like it would be less confusing to just always set it to a non-valid tombstone block and then have the caller always call `TransactionBuilder::set_tombstone_block`.

transaction/std/src/transaction_builder.rs
```diff
@@ -108,12 +111,16 @@ impl TransactionBuilder {
...
+    /// Sets the tombstone block, clamping to smallest pubkey expiry value.
     ///
     /// # Arguments
     /// * `tombstone_block` - Tombstone block number.
-    pub fn set_tombstone_block(&mut self, tombstone_block: u64) {
-        self.tombstone_block = tombstone_block;
+    pub fn set_tombstone_block(&mut self, tombstone_block: u64) -> u64 {
+        self.tombstone_block = core::cmp::min(
+            tombstone_block,
+            self.fog_pubkeys.smallest_pubkey_expiry_value(),
```
kylefleming, line 121: I think it would be better to only enforce the expiries of the fog pubkeys that were used to construct fog hints, rather than enforcing all fog pubkeys contained in `FullyValidatedFogPubkeys`.

If `FullyValidatedFogPubkeys` contains more fog pubkeys than are actually used, then this technique runs the risk of unnecessarily restricting the `tombstone_block` to a lower value than it would otherwise need to.

transaction/std/src/transaction_builder.rs
kylefleming: Unless a network call is made prior to building the Tx to determine the current block count of the ledger, how would this check be performed?

fog/report/validation/src/lib.rs
```diff
@@ -57,3 +68,92 @@ pub struct FullyValidatedFogPubkey {
...
 }
+
+/// Represents a set of fully validated fog public keys
+///
+/// These can be fetched before building a transaction, then the rest of the
+/// transaction building process can happen offline.
+#[derive(Default, Debug)]
+pub struct FullyValidatedFogPubkeys {
+    /// Map from fog_url, fog_report_id strings to fully validated pubkeys
+    keys: BTreeMap<(String, String), FullyValidatedFogPubkey>,
```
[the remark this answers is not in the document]
garbageslam: the reason was that look up was based on looking for the url and report id, so we'd have to use a more complex hash function API based on iterators or something if we dropped report id here

fog/report/validation/src/lib.rs
```diff
@@ -57,3 +68,92 @@ pub struct FullyValidatedFogPubkey {
...
+                Entry::Occupied(_) => {}
+                Entry::Vacant(ent) => {
+                    ent.insert(resolver.get_fog_pubkey(recipient)?);
+                }
+            }
+        }
+        Ok(())
+    }
+
+    /// Alternative to add_key_for_recipient, this may be helpful for SDK code which doesn't want to make blocking calls into rust.
```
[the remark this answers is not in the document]
garbageslam: :+1:

transaction/std/src/transaction_builder.rs
```diff
@@ -28,16 +29,21 @@ pub struct TransactionBuilder {
...
+    fog_pubkeys: FullyValidatedFogPubkeys,
 }
 
 impl TransactionBuilder {
     /// Initializes a new TransactionBuilder.
-    pub fn new() -> Self {
+    ///
+    /// # Arguments
+    /// * `fog_pubkeys` - Set of prefetched fog public keys to choose from
+    pub fn new(fog_pubkeys: FullyValidatedFogPubkeys) -> Self {
```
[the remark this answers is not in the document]
garbageslam: ok -- i agree with that, but if we take by reference then the transaction builder has to take a lifetime parameter, which is annoying. we could add `Clone` to the manager object

transaction/std/src/transaction_builder.rs
garbageslam: interesting, i didn't know consensus will reject things for being too far in the future

does consensus just test it against `MAX_TOMBSTONE_BLOCKS` ?

yeah, it seems fine to require that the user set a tombstone value, and we enforce the limit at the time that they try to set it

transaction/std/src/transaction_builder.rs
garbageslam: yeah, okay, we can keep a vec of actually used expiry values in the transaction builder or something

garbageslam:
It seems like i can't reply on some of the github comments for some reason, so i'm replying down here:

> What do you think about calling it FogPubkeyManager?

For me, Manager usually has a very specific meaning in C++ code, of being like a "lifetime manager" for objects of one or several specific types, that owns them, and allows other code to get references to those objects by their id or something like this. Usually the manager is a static global variable.

What do you think about calling it like `FogPubkeySet` ?

transaction/std/src/transaction_builder.rs
garbageslam: when does the user want to set the tombstone block though? Is there a time that you don't want it to be the smallest of pubkey expiry values that is used? That might be a sane default

garbageslam:
> Unless a network call is made prior to building the Tx to determine the current block count of the ledger, how would this check be performed?

I guess I didn't follow the question, in connection to the line of code where this comment was placed

In that line of code, the function "smallest_pubkey_expiry_value" is just the smallest of the pubkey_expiry values of whatever is in the cache set, so computing that doesn't require network calls. But it is expected that these pubkeys (usually if not always?) get refreshed when we make a new transaction. If that isn't done and the expiry values get stale then the transaction will expire.

If there is a problem running fog, and ingest goes down and stops publishing keys, then the keys you have will expire, and I guess (hope?) that consensus will tell you your transaction is dead on arrival?

transaction/std/src/transaction_builder.rs
kylefleming: The smallest of pubkey expiry values won't always be available or be a valid default.

E.g.:
* When all of the recipients are non-fog users.
* When the fog pubkey expiry returned from the fog report server is too far in the future for it to be used as a valid tombstone block.

kylefleming:
> It seems like i can't reply on some of the github comments for some reason, so i'm replying down here:

That means the comment is a comment on a previous PR comment. If you look for the original comment thread, you can comment there.

kylefleming:
> > Unless a network call is made prior to building the Tx to determine the current block count of the ledger, how would this check be performed?
> 
> I guess I didn't follow the question, in connection to the line of code where this comment was placed

This was a comment in reply to a previous PR comment by @mfaulk, https://github.com/mobilecoinfoundation/mobilecoin/pull/581#discussion_r539598163

garbageslam:
i see, thanks, i will post my comments in the right place

fog/report/validation/src/lib.rs
garbageslam: For me, Manager usually has a very specific meaning in C++ code, of being like a "lifetime manager" for objects of one or several specific types, that owns them, and allows other code to get references to those objects by their id or something like this. Usually the manager is a static global variable.

What do you think about calling it like `FogPubkeySet` ?

transaction/std/src/transaction_builder.rs
garbageslam: @mfaulk can we make it more forgiving, like, if you submit a transaction too far in the future, it gets clamped, and the submitter gets told that it was clamped?

garbageslam: or is that all under the signature i guess

kylefleming: > interesting, i didn't know consensus will reject things for being too far in the future
> 
> does consensus just test it against `MAX_TOMBSTONE_BLOCKS` ?

Exactly. Based on the current ledger block count.

https://github.com/mobilecoinfoundation/mobilecoin/blob/9cf0f47ac841aba6b8e343695f2900a566c70ccf/transaction/core/src/validation/validate.rs#L353

https://github.com/mobilecoinfoundation/mobilecoin/blob/5472a0377a03faddb17c650fb5fd944832ff4cc8/consensus/enclave/impl/src/lib.rs#L408

fog/report/validation/src/lib.rs
kylefleming: Not sure what the right naming is. The object both fetches new pubkeys and stores them. To me, `FogPubkeySet` implies that it just stores them.

transaction/std/src/transaction_builder.rs
garbageslam: or just compute the min as we go.

in an early draft of this PR i made `create_output` a private member funciton of TxBuilder but then a bunch of unit tests would have had to change. but i can work through that.

fog/report/validation/src/lib.rs
garbageslam: i mean, the resolver fetches them

we could refactor the resolver trait and try to make a better division of responsibilities here

posix4e: Manager is a bad smell unless it means something specific. The GOOS book generally would recommend laying out something about the implementation in the name, to distinguish it from a nop testing object providing a facade functionality . What about FullyValidatedBTreeKeychain or just FullyValidatedKeychain if you don't like the goos book style classname

garbageslam:
After discussion with james and further development on the fog report signature scheme, we're going to make changes to this:

- `FogPubkeySet` type introduced in current revision is going to be changed to roughly a type alias `pub type FogReportResponses: HashMap<FogUrl, ReportResponse>`. The idea is, the user of the API makes a request to each fog url of a recipient in the system, then attaches the entire protobuf response object from the fog report server.

Then, all validation, including checking cert chains and user's fog authority signatures, happens when adding a txout for that user.

I'm going to post a modified version of this PR shortly

transaction/std/src/transaction_builder.rs
```diff
@@ -54,24 +71,14 @@ impl TransactionBuilder {
...
-        let (tx_out, shared_secret) =
-            create_output(value, recipient, recipient_fog_ingest_key, rng)?;
-
-        self.outputs_and_shared_secrets
-            .push((tx_out.clone(), shared_secret));
-
-        let confirmation = TxOutConfirmationNumber::from(&shared_secret);
-
-        Ok((tx_out, confirmation))
+        self.add_output_with_fog_hint_address(value, recipient, recipient, rng)
```
garbageslam, line 81: it seems better to avoid duplicating the implementation details

fog/report/connection/src/lib.rs
```diff
@@ -1,172 +1,104 @@
...
-    pub fn new(verifier: &Verifier, env: Arc<Environment>, logger: Logger) -> Self {
-        Self {
-            verifier: IngestReportVerifier::from(verifier),
-            env,
-            logger,
-        }
-    }
-}
-
-impl FogPubkeyResolver for GrpcFogPubkeyResolver {
```
garbageslam, line 82: the grpc object no longer implements `FogPubkeyResolver` after this commit, it just produces `FogReportResponses`

fog/report/validation/src/lib.rs
```diff
@@ -1,59 +1,150 @@
...
         recipient: &PublicAddress,
-    ) -> Result<FullyValidatedFogPubkey, Self::Error>;
+    ) -> Result<FullyValidatedFogPubkey, FogPubkeyError> {
+        if let Some(url) = recipient.fog_report_url() {
+            let url = url.to_string();
+            if let Some(result) = self.responses.get(&url) {
+                let report_id = recipient.fog_report_id().unwrap_or("").to_string();
+                for report in result.reports.iter() {
+                    if report_id == report.fog_report_id {
+                        // TODO validate x509 chain and recipient.fog_authority_sig here,
```
garbageslam, line 105: this is where the x509 chain and cert checking validation code should go, after #684 is merged

mobilecoind/src/config.rs
```diff
@@ -167,19 +169,42 @@ impl Config {
...
+    /// The string error should be mapped by invoker of this factory to Error::FogError
+    pub fn get_fog_resolver_factory(
+        &self,
+        logger: Logger,
+    ) -> Arc<dyn Fn(&[FogUri]) -> Result<FogResolver, String> + Send + Sync> {
+        let env = Arc::new(
+            grpcio::EnvBuilder::new()
+                .name_prefix("FogPubkeyResolver-RPC".to_string())
+                .build(),
+        );
```
garbageslam, line 188: it might be a good idea to limit the number of threads in this thread pool

voloshyn (approved):
LGTM

transaction/std/src/transaction_builder.rs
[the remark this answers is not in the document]
garbageslam: @kylefleming my resolution here is that `FogReportResponses` implements Clone, i'd rather not make the class hold a reference

garbageslam:
I guess I will try to change this to use MAX_TOMBSTONE_BLOCKS instead of u64::max_value()

fog/report/validation/src/lib.rs
```diff
@@ -1,59 +1,150 @@
...
+                        // Howeve, probably skip that if uri scheme is "insecure-fog"?
```
eranrund, line 106: Typo in `Howeve`

eranrund (commented):
I think this looks good, but I'd like the people that were involved in the redesign to take a look.

kylefleming:
> I guess I will try to change this to use MAX_TOMBSTONE_BLOCKS instead of u64::max_value()...
> still pondering it
> 
> it seems like its easier to let the client decide how to deal with that, since its inherently racy

I would avoid using something that changes in validity over time.

`current_ledger_block_count + MAX_TOMBSTONE_BLOCKS` changes over time, assuming this is what you meant?

E.g. what if company X wanted to create a Tx that is intended to be submitted a week later after board approval?

fog/report/connection/src/lib.rs
```diff
@@ -1,172 +1,104 @@
...
+        let mut responses = FogReportResponses::default();
+        for uri in uris {
+            self.fetch_fog_report(&mut responses, &uri)?;
+        }
+        Ok(responses)
+    }
+
+    /// Given a set of previously collected FogReportResponse's, and another Uri, make a corresponding
+    /// request and add it to the collection, if such a response is not already present.
+    pub fn fetch_fog_report(
```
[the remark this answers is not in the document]
garbageslam: ok, i will rejigger it a bit

garbageslam:
@kylefleming ok, will leave as is with `u64::max_value()`

garbageslam:
still need to resolve conflicts with master branch

garbageslam:
this is now rebased on master

transaction/std/src/transaction_builder.rs
```diff
@@ -225,27 +243,26 @@ impl TransactionBuilder {
...
 /// * `recipient` - Recipient's address.
-/// * `ingest_pubkey` - The public key for the recipients fog server, if any
-/// * `rng` -
-fn create_output<RNG: CryptoRng + RngCore>(
+/// * `fog_resolver` - Set of prefetched fog public keys to choose from
+/// * `rng` - Entropy for the encryption.
+///
+/// # Returns
+/// * A transaction output, and the shared secret for this TxOut.
+#[allow(unused)]
```
jcape, line 258: This function should just be nuked, if nobody is using it.

jcape (approved):
Looks good, with some minor cleanups that could be made (or JIRA'd).

transaction/std/src/transaction_builder.rs
garbageslam: It's used in tests, it could move to test module

fog/report/connection/src/lib.rs
```diff
@@ -1,172 +1,104 @@
...
-pub use mc_fog_report_validation::{FogPubkeyResolver, FullyValidatedFogPubkey};
+pub use mc_fog_report_validation::FogReportResponses;
+
+/// Fog report server connection based on grpcio
+///
+/// TODO: As an optimization, it might be good to make this like a connection
+/// pool that holds onto its grpc connections, so that they can be reused
+/// if a client is constructing transactions frequently.
+/// The user of this API would tear down this object when they want these
+/// connections to be closed.
```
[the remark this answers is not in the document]
garbageslam: but the implementation here creates a new grpc channel every time we fetch -- i think for grpc to provide connection reuse, we have to keep the channel object alive across calls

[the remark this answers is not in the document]
garbageslam: https://tikv.github.io/doc/grpcio/struct.Channel.html

> Channels are an abstraction of long-lived connections to remote servers. More client objects can reuse the same channel.

So like, ideally, if mobilecoind is sending to many fog users in succession, it would re-use the tcp connection each time it hits `fog.mobilecoin.com` or whatever. I think we have to keep the `Channel` object alive for that to happen, I think the grpc connection pool is in the channel object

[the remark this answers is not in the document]
garbageslam: i can rewrite the code comment

PR #586, opened 2020-12-08
switch to a smaller debian docker

posix4e:

### Motivation

< The motivation for the changes in this PR. "Currently we...", "This is needed because..." >

### In this PR

< Ticket status, e.g. "fixes #issue number" > 

### Future Work

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -15,7 +15,7 @@
 #  builder-install - Build environment  #
 # ##################################### #
 
-FROM ubuntu:18.04 AS builder-install
+FROM debian:stable-slim AS builder-install
 
 # Party like it's June 8th, 1989.
 #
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -15,7 +15,7 @@
 #  builder-install - Build environment  #
 # ##################################### #
 
-FROM ubuntu:18.04 AS builder-install
+FROM debian:stable-slim AS builder-install
```
garbageslam, line 18: i think the version of the SGX driver we install may only support ubuntu 18.04, we should check if that's right

posix4e: Wow this was supposed to be just on my private repo. Sorry to bug you with my experimenting

garbageslam: no worries!

PR #589, opened 2020-12-09
Black format on lint

posix4e:
Soundtrack of this PR: [link to song that really fits the mood of this PR]()

### Motivation

Python deserves some linting, just like rust!

### In this PR
* I added black to the dockerfile
* I added black check to the linter tool
* I black formatted all linted python

### Future Work
* Potentially use  mypy for further checks
* perhaps move from pip3 black to apt

PR #590, opened 2020-12-09
Add black for python linting

posix4e:

### Motivation
https://github.com/mobilecoinfoundation/mobilecoin/issues/592
https://github.com/mobilecoinfoundation/mobilecoin/pull/589 is going to require linter tools to be installed before we can lint 

### In this PR
We install the newest black. This will mean that we will have to relint as new black updates come out

### Future Work
* Verify we only want to use the newest black
* https://github.com/mobilecoinfoundation/mobilecoin/pull/589

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -76,7 +76,7 @@ RUN apt-get update -q -q && \
 # prometheus = helps with running slam scripts locally
 
 # For use in CI
-RUN pip3 install awscli
+RUN pip3 install awscli black
 
 # filebeat is used for logs when running slam scripts locally
 RUN curl -L -O https://artifacts.elastic.co/downloads/beats/filebeat/filebeat-7.4.2-amd64.deb && \


eranrund (approved):

rjwalters (approved):

posix4e:
I cannot merge the patch, so someone will have to do this for me. 

While I'm chatting, are there docs vis a vis updating what image gets used by CI and other required systems?

posix4e:
Sorry to be a bother but I can't merge this

garbageslam:
fyi: the way the infra works now, we're supposed to increment Dockerfile-version when we change this file:

https://github.com/mobilecoinfoundation/mobilecoin/blob/master/docker/Dockerfile#L3

kylefleming:
> fyi: the way the infra works now, we're supposed to increment Dockerfile-version when we change this file:
> 
> [https://github.com/mobilecoinfoundation/mobilecoin/blob/master/docker/Dockerfile#L3](https://github.com/mobilecoinfoundation/mobilecoin/blob/master/docker/Dockerfile?rgh-link-date=2020-12-18T22%3A51%3A21Z#L3)

Publishing a new version to Google Cloud is currently an internal process, so I'm not sure that the same policy applies. I believe the `Dockerfile-version` should only be incremented when it's also accompanied with a release to gcloud, no?

garbageslam:
the point is, ideally at each revision, the Dockerfile in the repo matches what it was used to build and test with, and that can also be fetched from the docker hub based on this version number tag

garbageslam:
so like, right now, the mob tool could get the wrong prompt (with or without black installed), because the image in dockerhub doesn't have black in it. but if you did mob image, you would get a docker image 1_14 with black installed. and then you can get non-reproducible build failures.

PR #594, opened 2020-12-10
Wallet service

sugargoat:
### Motivation

We are introducing a new wallet-service, which includes mobilecoin as a submodule. This PR introduces some changes necessary to support the wallet-service.

### In this PR
* Updates the mobilecoind-json data_types to derive Default and Debug
* Updates the builder image to include sqlite3


eranrund (approved):

tsegaran (approved):
Talked over with Sara, this looks good.

mobilecoind/api/proto/mobilecoind_api.proto
```diff
@@ -143,6 +143,9 @@ message TxProposal {
     /// A list of the confirmation numbers, in the same order
     /// as the outlays.
     repeated bytes outlay_confirmation_numbers = 6;
+
+    /// Value of the change output
+    uint64 change_value = 7;
```
rjwalters, line 148: isn't this redundant with (inputs - outputs)? what do we do about conflicts?

sugargoat: Parsing the TxProposal for useful information for a TransactionLog store requires a few assumptions about how the transactions are constructed. I think eventually we will move to a different format for the TxProposal, especially since this version relies on the UnspentTxOut, which is specific to mobilecoind's utxo store. For now, adding the change_value gets us all the information we need at the time of submission, whether or not the transaction was constructed on an offline machine or locally. For example, you may be submitting a TxProposal for an account that is not loaded in the wallet where you are submitting it, in which case the inputs in the Tx are opaque, and you cannot determine the value in the inputs.

rjwalters: I'm not sure I understand the use case... both `UnspentTxOut` and `Outlay` both contain value, and the change must be the difference of sums to balance the tx. I think this invites confusion.

Are you saying this is in case TxProposal is later modified so that you can't figure out the change?

sugargoat: I somehow missed that we have all the information that we need in the utxos (intput list). I updated the call site to calculate the change value from `sum(tx_proposal.utxos.value) - sum(tx_proposal.outlays.value) - tx_proposal.fee`

PR #596, opened 2020-12-11
Consensus crate test prep work

mfaulk:
### Motivation

The `consensus` crate tests serve to test the consensus protocol independently of concrete application implementations (i.e. MobileCoin-specific transaction, network, or persistence implementations). Currently, the crate tests use some custom test fixtures to test simulated SCP networks. That was previously the best approach to testing, but changes (dependency-injection) to the `consensus` now allow crate tests that more directly test the deployed code.

### In this PR
* Code cleanup and refactor in preparation for testing against structs in `consensus/scp'

### Future Work
* Replace usage of `mobilecoin/consensus/scp/tests/mock_network/scp_node.rs` with usage of `mobilecoin/consensus/scp/src/node/node_impl.rs`


consensus/scp/tests/mock_network/scp_network.rs
```diff
@@ -0,0 +1,178 @@
...
+    /// NodeID to node.
+    nodes: Arc<Mutex<HashMap<NodeID, SCPNode>>>,
+
+    /// NodeID to the node's thread handle.
+    handles: HashMap<NodeID, JoinHandle<()>>,
+
+    /// NodeID to node name.
+    pub names: HashMap<NodeID, String>,
+
+    /// NodeID to ...???
```
rjwalters, line 28: `SCPNodeSharedData` is basically just a mock ledger and could probably be cleaned up to that effect.

consensus/scp/tests/mock_network/scp_network.rs
```diff
@@ -0,0 +1,178 @@
...
+    shared_data: HashMap<NodeID, Arc<Mutex<SCPNodeSharedData>>>,
+
+    logger: Logger,
+}
+
+impl SCPNetwork {
+    // Creates a simulated network.
+    pub fn new(node_configs: &[NodeConfig], test_options: &TestOptions, logger: Logger) -> Self {
+        // Why is this check necessary?
```
rjwalters, line 37: I guess to avoid have a node self-broadcast? maybe we should allow that?

consensus/scp/tests/mock_network/scp_network.rs
```diff
@@ -0,0 +1,178 @@
...
+            .collect();
+
+        let nodes = Arc::new(Mutex::new(HashMap::default()));
+        let mut handles = HashMap::default();
+        let mut shared_data = HashMap::default();
+
+        for node_config in node_configs {
+            let node_id = &node_config.id;
+            let (node, join_handle) = {
+                let nodes_map = nodes.clone(); // ???
```
rjwalters, line 56: we move these into our weird broadcast function below

consensus/scp/tests/mock_network/scp_network.rs
mfaulk: Ah -- I missed the "!"

eranrund (approved):

remoun:

jcape:
I think so, but it needs to get rebased on master, and some of the stylistic changes (e.g. I don't care for `let foo: Vec<bar> = yaddayadda.collect();` vs `let foo = yaddayadda.collect::<Vec<bar>>();`, but it's also not really germane to test changes) get moved to a separate PR.

PR #667, opened 2020-12-18
Move some postgres setup commands into the docker image

garbageslam:
This reduces the number of commands someone has to do from `./mob prompt`
to just `service postgresql start` in order to be able to run `cargo test`.
This will also mean we don't need to put those commands in circle-ci config.


eranrund (approved):

joekottke (approved):
I think this is ok for now.  But as we get into more multi-service development, I think something like docker-compose would be a better way to go with this in general.  But thats definitely not a blocker here.

docker/Dockerfile-version
```diff
@@ -1 +1 @@
-1_14
+1_15
```
kylefleming, line 1: Could you rebase this on master first? Otherwise you're not including the change made in https://github.com/mobilecoinfoundation/mobilecoin/pull/590.

kylefleming (changes requested):

docker/Dockerfile-version
garbageslam: :+1: thank you, didn't see that

PR #678, opened 2021-01-12
Fix test_client usage of GetTxStatusAsSender

sugargoat:
### Motivation

In #674 we fixed status as sender, but the test_client also needed to incorporate the fix for continuous deployment

### In this PR
* Fixes up GetTxStatusAsSender in the test_client

[MCC-2183](https://mobilecoin.atlassian.net/browse/MCC-2183)


joekottke (approved):
LGTM

eranrund (approved):
I think we should move the parts of the receiver receipt that we need for sender verification into the sender receipt, the current API is convoluted.

sugargoat:
@eranrund - good point. I'll make a ticket to revisit the API. One nice thing is being able to take the output of SubmitTxResposne and feed it directly to the check, which is how mobilecoind-json is functioning with the change.

PR #679, opened 2021-01-12
Mixin

jadeydi:
Add POST /membership_proofs, to get the  membership_proofs and rings of txos.


jadeydi:
May JsonBlockDetailsResponse return full data, like JsonTxOut, key images etc?

https://github.com/mobilecoinfoundation/mobilecoin/blob/master/mobilecoind-json/src/data_types.rs#L1148

PR #686, opened 2021-01-21
Break dep fog-report-validation -> grpcio

garbageslam:
This dependency is bad because android bindings and libmobilecoin
don't really want to build grpcio.

We fix it by making a prost, no_std compatible version of the
report server `ReportResponse` object, and putting this in mc-fog-types.
The report connection returns the prosty version of the type,
and validation uses that type.

Tests are used, as in the other fog-api crate, to ensure that the
protobuf generated and prost versions are in-sync and can round-trip
correctly.


sugargoat (approved):

fog/types/src/lib.rs
```diff
@@ -0,0 +1,34 @@
...
+//! doesn't depend on grpcio, as `mc-fog-api` does.
+
+extern crate alloc;
+
+use alloc::{string::String, vec::Vec};
+use prost::Message;
+use serde::{Deserialize, Serialize};
+
+/// A fog report from the report server
+#[derive(Clone, Eq, PartialEq, Serialize, Deserialize, Message)]
```
kylefleming, line 15: Do these need to be `Serialize`/`Deserialize`? I believe the Protobuf ones we're switching away from in this PR wouldn't have been serde serializable. Assuming that's the case, what's the motivation for adding it?

kylefleming (commented):
(This is not a review, just adding a comment)

fog/types/src/lib.rs
garbageslam: just copying the pattern I used in other parts of fog
one reason things become `Serialize/Deserialize` is if they have to enter or leave an enclave
it seems like it doesn't hurt anything to add this and might help something later

kylefleming: Ok. I personally think we should avoid adding serde to prost struct unless there's a clear reason to. Why add the extra dependency just in case?

I believe the pattern is generally to create a separate struct specifically for the enclave boundary, no?

garbageslam: > I believe the pattern is generally to create a separate struct specifically for the enclave boundary, no?

no, the pattern is to have as few structs as possible.

As a rule, structs should implement standard traits like `Debug, Clone`. In this case, it's a struct meant for serialization, so it should implement `Serialize` and `Deserialize`unless there's a reason it can't.

For some earlier structures in fog, I tried avoiding serde entirely. Most of them got serde added by other people later (for instance, the fog-kex-rng crate types.)

jcape (approved):
This is fine, but this may need to move towards a "conversion" style API, akin to what we have in `mc-api`. I bring this up because I think we should simultaneously:

1. Expose the `VerificationReport` data as a protobuf object in the report server response object, as opposed to a "wrapped protobuf bytes" thing, because that's clearly superior protobuf usage.
2. Use Digestible over a `VerificationReport[]` slice to create the report server's signature, because that's clearly how that should go.

Doing both of those without using grpcio's `"prost"` feature implies that we should follow the "conversions" convention we created for `mc-api`.

PR #710, opened 2021-02-08
Make an error passable across enclave boundary

cbeck88:
This error should derive serde so that it can be passed across enclave boundary


jcape (changes requested):
This needs to use `serde::{Serialize, Deserialize};`

remoun:
This PR still seems relevant; do we want to (fix and) merge it?

garbageslam:
Yea that would be a good thing

nick-mobilecoin:
Closing it looks like this was done in https://github.com/mobilecoinfoundation/mobilecoin/pull/1658

PR #717, opened 2021-02-10
move mc-crypto-rand to be a dev-dependency of mc-transaction-core

garbageslam:
`mc-transaction-core` should be as portable as possible, but it had
a dependency on `mc-crypto-rand` which is not as portable.

Getting an RNG is very hard to do portably -- the approach we use
in `mc-crypto-rand` is like, use the standard library approach and
the getrandom crate, which still requires an OS, or if you have
RDRAND instruction, use that, and selecting between these
alternatives is what `mc-crypto-rand` does basically. But this will
not work on embedded devices that don't have an OS or intel chips.
It just happens to work for all our current enclave phone and server
targets. But it may pose a portability problem if someone wanted
to build `mc-transaction-core` for a more restricted device.

Fortunately the only thing that was creating an RNG directly in
`mc-transaction-core`, was the code in the `FogHint` which we removed
in https://github.com/mobilecoinfoundation/mobilecoin/pull/712
So at this point `mc-transaction-core` only needs `mc-crypto-rand`
for tests. Moving it to dev-dependency helps make the transaction core
more portable.

### DIFF
--- a/consensus/enclave/trusted/Cargo.lock
+++ b/consensus/enclave/trusted/Cargo.lock
@@ -1053,7 +1053,6 @@ dependencies = [
  "mc-crypto-digestible 1.0.0",
  "mc-crypto-hashes 1.0.0",
  "mc-crypto-keys 1.0.0",
- "mc-crypto-rand 1.0.0",
  "mc-util-from-random 1.0.0",
  "mc-util-repr-bytes 1.0.0",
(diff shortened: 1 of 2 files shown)


jcape (approved):
w00t!

kylefleming (approved):

PR #725, opened 2021-02-18
Replace Aes128 with Aes256 everywhere that was using it

garbageslam:
During review, we noticed that sometimes we select Aes128 and
sometimes Aes256. The mc-noise crate is using Aes256 to create
attested connections to enclaves. The mc-crypto-box is using Aes128,
and the message-cipher was also using Aes128.

Using two kinds of AES, especially in attested connections and
in cryptobox, means that an embedded device that wants to construct
mobilecoin transactions needs to support both forms of AES. Usually
it's better to just use one primitive everywhere to reduce code
foot print. AES256 is not appreciably slower than AES128, so it is
fine to replace Aes128 with Aes256 everywhere. That makes things
simpler and more consistent anyways.

The change is likely going to change the MRENCLAVE value of consensus,
if only because of the message cipher change.
We also noticed that there is an unused file in crypto/box/src.
The real implementation is `hkdf_box.rs` and this is generic over
the different primitives. The old file may have gotten committed
again by accident somehow, I'm not sure.

Note: This is a breaking change to crypto-box, but since Fog has
not been deployed in production anywhere, this is fine.


eranrund (approved):

UkoeHB:
> The change is likely going to change the MRENCLAVE value of consensus

What is the plan to handle this?

sugargoat:
@UkoeHB - the MRENCLAVE change will be handled as a regular network enclave upgrade, and can be rolled into our next release.

UkoeHB:
@sugargoat does this imply the next hardfork will occur before Fog, if Fog will depend on this change?

sugargoat:
@UkoeHB - chatted with @garbageslam, and it's possible that this change wouldn't require an update on consensus to deploy fog, but it would make things easier :)

UkoeHB:
@sugargoat not sure what your timeline and roadmap look like, but if there are no pressing reasons to hardfork aside from this change, then I feel it would be better go with the inconvenient solution for now. There is a risk of setting a precedent - hardforking without a big update to make.

jcape:
We need to upgrade the enclaves to handle the Dec. 15 Intel cutoff date, and I'd like to isolate all the pain points of a hard fork before we have fog, since the inevitable code changes required to make one happen seamlessly will take at least a round or two to find and fix.

jcape (approved):

UkoeHB:
> We need to upgrade the enclaves to handle the Dec. 15 Intel cutoff date

Is this a cutoff date from 2020 or 2021? Are you saying Intel issued a critical security update late last year but did not stop accepting attestations for enclaves active during the old SGX version before MobileCoin was launched? That would qualify as a reason to hardfork in the short term.

PR #728, opened 2021-02-19
FOG-373 Make sample-keys take a fog authority certificate PEM

jcape:
### Motivation

In order to perform integration testing of fog authority signatures, keys generated with keyfile need to include the proper bytes from the chain to be used.

### In this PR
* Removes the `--fog-authority-fingerprint` and adds `--fog-authority-root` which will take a pem certificate and extract the relevant bytes necessary for the fog authority signatures.


sugargoat (approved):

garbageslam (approved):
lgtm!

PR #735, opened 2021-02-25
Enforces invariant that pending_transactions are valid w.r.t the current ledger

mfaulk:
### Motivation

This PR more clearly states, and enforces, the invariant that transactions added to the `pending_values` collection must be valid with respect to the current ledger. While transactions are validated when they are received from clients, enforcing the invariant when the transactions are first consumed by the service worker clarifies the intention and prevents potential race conditions where a transaction expires between when it was received from a client and when it was consumed by the worker.


garbageslam (approved):
this makes sense to me, and the docu is very clear, thank you

eranrund (approved):
Awesome, thank you @mfaulk !
I have confirmed this fix solves the problem on a k8 network.

PR #745, opened 2021-03-08
SCP align vote to prepare ballot values with spec

UkoeHB:

### Motivation

The current implementation of SCP in MobileCoin is very close to the protocol description [in the IETF draft](https://tools.ietf.org/pdf/draft-mazieres-dinrg-scp-04.pdf) and [whitepaper](https://assets.website-files.com/5deac75ecad2173c2ccccbc7/5df2560fba2fb0526f0ed55f_stellar-consensus-protocol.pdf). However one nuance was overlooked.

When setting the ballot value `B`, which corresponds to the ballot the node will vote to prepare, `get_next_ballot_values()` uses this priority list:
- value of highest ballot confirmed prepared {with available messages from the network}
- composite value from nomination
- value of highest ballot accepted prepared {with available messages from the network}
- previous value of `B`
- `None`

However the available documentation on SCP does not indicate the {with available messages from the network} should be used. Instead they say
> [IETF pg. 15] Each time the ballot counter is changed, the value is also recomputed as follows: 
> *  If any ballot has been confirmed prepared, then "ballot.value" is taken to to be "h.value" for the highest confirmed prepared ballot "h".
> Otherwise, if no ballot is confirmed prepared and no value is confirmed nominated, but the node has accepted a ballot         prepared (because "prepare(b)" meets blocking threshold for some ballot "b"), then "ballot.value" is taken as the value of         the highest such accepted prepared ballot.

> [whitepaper pg. 24] If phase = PREPARE and m lets v confirm new higher ballots prepared, then raise h to the highest such ballot and set z = h:x.
> [whitepaper pg. 23] z: Value to use in next ballot. If h=0, then z is the composite value ... otherwise, z=h:x.

Since nodes only store very limited historical information in their state (only `B`, `P`, `PP`, `C:H`), it is possible for future messages from a node to 'forget' about old information that a node once used to set their `H` or `P`. This means after `H` is set, a node could see a future state of the network that would not permit it to confirm `H` is prepared, and `get_next_ballot_values()` will fail to return the highest ballot confirmed prepared previously.

### In this PR
* SCP: aligns setting of ballot value for vote to prepare a ballot with specification

### Future Work
* none


consensus/scp/src/slot.rs
```diff
@@ -1196,8 +1196,16 @@ impl<V: Value, ValidationError: Display> Slot<V, ValidationError> {
...
+            if let Some(h_old) = &self.H {
+                if h_old > h {
+                    return Some(h_old.X);
+                }
+            }
             return Some(h.X);
         }
+        else if let Some(h_old) = &self.H {
+            return Some(h_old.X);
+        }
```
UkoeHB, line 1208: A simpler alternative would be returning H if it exists.

consensus/scp/src/slot.rs
```diff
@@ -1214,8 +1222,16 @@ impl<V: Value, ValidationError: Display> Slot<V, ValidationError> {
...
+            if let Some(p_old) = &self.P {
+                if p_old > p {
+                    return Some(p_old.X);
+                }
+            }
             return Some(p.X);
         }
+        else if let Some(p_old) = &self.P {
+            return Some(p_old.X);
+        }
```
UkoeHB, line 1234: Similarly, simpler to return P if it exists.

consensus/scp/src/slot.rs
```diff
@@ -1214,8 +1222,16 @@ impl<V: Value, ValidationError: Display> Slot<V, ValidationError> {
...
 
         // Otherwise, values are unchanged.
         if !self.B.is_zero() {
```
UkoeHB, line 1237: I do not believe this will ever be executed.

UkoeHB, line 1237: I do not believe this conditional will ever be true. B can only be non-zero if: Z is set, H is set, or P is set. Whichever way it gets set, the earlier conditionals in this function will catch it.

PR #747, opened 2021-03-09
FOG-361: Fog resolver should normalize URIs on both sides

garbageslam:
Currently, Fog resolver normalizes uris when looking up queries,
but not when taking the set of report responses. After this commit,
it normalizes the strings to URIs on both sides.

Arguably, we should just change the `BTreeMap<String, ReportResponse>`  to `BTreeMap<FogUri, ReportResponse>`, but I think that is a more disruptive change so i was reluctant to do it that way


eranrund (approved):

voloshyn (approved):
LGTM

PR #752, opened 2021-03-11
Increment block version for next scheduled hardfork (v0 -> v1)

UkoeHB:

### Motivation

Scheduled hardforks should be tracked on-chain (note: any change to validator enclaves requires a scheduled hardfork). Currently the only version number available is the block header version. Originally this was intended to represent the block header format version, however block header format version changes can be subsumed under the overall hardfork version (i.e. the protocol version). Since the current header version is `0`, repurposing the version number to represent the hardfork version is semantically harmless.

This PR prepares a new block version for whenever the hardfork that goes v0 to v1 occurs. Note that this PR is a breaking change (will alter the validator enclave MRENCLAVE).

#### Block validation rule

This PR also adds a block validation rule. A block version must be no less than the previous block version, and may not exceed the maximum block version. [EDIT: Skipping block versions is allowed. It may be useful if block versions are consensed on but a certain version fails for one reason or another (e.g. a bug is found, or a large contingent of nodes reject it). In that case the only safe way to consense on a different hardfork (different validator code) is to use a different hardfork version.]

### In this PR
* Increments block header version (see MC-2258).
* Updates block validation to support version changes.

### Future Work
* Future scheduled hardforks may make use of `RFC - MobileCoin Scheduled Hardforks with SCP and Dual-Enclave Blocks` (unpublished draft at this time), which specifies a new block header format for tracking hardfork versions.
* Refactor semantics so instead of tracking 'hardfork versions' we track 'number of hardforks'.

### DIFF
--- a/ledger/db/src/lib.rs
+++ b/ledger/db/src/lib.rs
@@ -559,12 +559,8 @@ impl LedgerDB {
             let last_block = self.get_block(num_blocks - 1)?;
 
             // The block's version should be bounded by
-            // [prev block version, min(prev block version + 1, max block version)]
-            if block.version >= last_block.version && block.version <= BLOCK_VERSION {
-                if block.version > last_block.version + 1 {
-                    return Err(Error::InvalidBlock);
-                }
-            } else {
+            // [prev block version, max block version]
+            if block.version < last_block.version || block.version > BLOCK_VERSION {
                 return Err(Error::InvalidBlock);
             }
 
--- a/ledger/db/src/lib.rs
+++ b/ledger/db/src/lib.rs
@@ -539,9 +539,39 @@ impl LedgerDB {
         block: &Block,
         block_contents: &BlockContents,
     ) -> Result<(), Error> {
-        // Check that version is correct
-        if block.version != BLOCK_VERSION {
-            return Err(Error::InvalidBlock);
+        // Check version is correct
+        // Check if block is being appended at the correct place.
+        let num_blocks = self.num_blocks()?;
+        if num_blocks == 0 {
+            // This must be an origin block.
+
+            // The origin block is version 0
+            if block.version != 0 {
+                return Err(Error::InvalidBlock);
+            }
+
+            // The origin block is index '0' with default-initialized parent ID, by convention
+            if block.index != 0 || block.parent_id != BlockID::default() {
+                return Err(Error::InvalidBlock);
+            }
+        } else {
+            let last_block = self.get_block(num_blocks - 1)?;
+
+            // The block's version should be bounded by
+            // [prev block version, min(prev block version + 1, max block version)]
+            if block.version >= last_block.version && block.version <= BLOCK_VERSION {
+                if block.version > last_block.version + 1 {
+                    return Err(Error::InvalidBlock);
+                }
+            }
+            else {
+                return Err(Error::InvalidBlock);
+            }
+
+            // The block must have the correct index and parent.
+            if block.index != num_blocks || block.parent_id != last_block.id {
+                return Err(Error::InvalidBlock);
+            }
         }
 
         // A block must have outputs.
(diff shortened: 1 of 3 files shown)


jcape (commented):
Question about `initialize_ledger()`, it seems like the ledger it creates will be all-v0 which will break when we release v2, right?

transaction/core/test-utils/src/lib.rs
```diff
@@ -189,7 +189,7 @@ pub fn initialize_ledger<L: Ledger, R: RngCore + CryptoRng>(
                 let block_contents = BlockContents::new(key_images, outputs);
 
                 let block = Block::new(
-                    BLOCK_VERSION,
+                    0,
```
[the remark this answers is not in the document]
eranrund: In the tests, yes. I don't see a good alternative for this, since we use `BLOCK_VERSION` things will break once it moves to `2` since nothing will create a block with version `1` to bridge between `0` and `2`.

Why do you think it would break when we release v2?

ledger/db/src/lib.rs
```diff
@@ -539,36 +539,51 @@ impl LedgerDB {
...
+
+            // The block's version should be bounded by
+            // [prev block version, min(prev block version + 1, max block version)]
+            if block.version >= last_block.version && block.version <= BLOCK_VERSION {
+                if block.version > last_block.version + 1 {
+                    return Err(Error::InvalidBlock);
+                }
+            } else {
+                return Err(Error::InvalidBlock);
+            }
```
UkoeHB, line 569: Should we allow version skipping? Instead bound by
```
// [prev block version, max block version]
```

If hardfork versions are consensed on, then without version skipping the network would have a hard time navigating buggy releases. If there is a scheduled hardfork but the code has a bug, then it would be hard to replace the new code with different code that has the same version number. When nodes consense on the new version number, if half of them have the old buggy code and half have the fixed code, then the network will get stuck on the next block because nodes will have different enclaves and no quorum will exist on the same enclave.

Right now that is not a problem because upgrading the network is a 'point in time' event where node software can be actively coordinated.

eranrund: I think that right now, and for as long as upgrading happens in an all-nodes-or-none fashion, this is fine as is. If we push out a version that is unable to produce blocks then we will roll back to the old version and try again.

UkoeHB: If we might need to remove this in the future, is it better to never have it in the first place?

eranrund: I have no preference.

UkoeHB: Ok I will just remove it

eranrund: 👍 
Note that it will likely break the test I added... I'll fix it.

eranrund: Actually the test can't test for that yet because of the hardcoded bound on `BLOCK_VERSION` which is currently `1`. Oh well.

UkoeHB: Thank you!

jcape: In the event of a problem with the network where no new transactions have made it through, we would roll back.

transaction/core/test-utils/src/lib.rs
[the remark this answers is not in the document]
jcape: If we initialized a ledger at `0`, and tried to append a new block to it in the normal way when `BLOCK_VERSION >= 2`, it would blow up, no? This is `mc_consensus_enclave_impl::tests::form_block_refuses_duplicate_root_elements()`.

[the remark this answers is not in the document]
jcape: This appears moot at this point.

jcape (approved):

sugargoat (approved):

PR #758, opened 2021-03-15
MCC-2263 Ensure MonitorId remains stable across mobilecoind versions.

jcape:
### Motivation

Earlier changes to the `mc_account_keys::PublicAddress` structure have changed it's digestible hash (a semantic hash will include structural details, including field names). This hash is being used as part of the a `MonitorId` used by mobilecoind to track keys which are being monitored.

As a result, the resulting MonitorId for equvilent addresses has changed, and must be restored.

### In this PR
* Update the `From<&MonitorData> for MonitorId` implementation to include a duplicate of the original/legacy `PublicAddress` structure in it's hash, rather than the one from `mc-account-keys`.
* Patch the mobilecoin-1.0.1 release, which had the older hash, to generate a pair of `MonitorId`s.
* Test that the current system will create the same `MonitorId` as the previous one.


sugargoat (approved):

garbageslam (approved):
this looks reasonable to me

PR #770, opened 2021-03-26
[MCC-2257] Adds a regression test for rejecting expired transactions

mfaulk:
### Motivation
MCC-2200 fixed a bug where transactions that expired while queued for processing would be added to the list of pending transactions and would not be purged. That bugfix did not include an explicit regression test for that behavior.

### In this PR
* An additional regression unit test checking that an expired transaction is rejected when it is consumed from the queue of incoming messages.

### DIFF
--- a/consensus/service/src/byzantine_ledger/worker.rs
+++ b/consensus/service/src/byzantine_ledger/worker.rs
@@ -1346,6 +1346,68 @@ mod tests {
         assert_eq!(worker.pending_consensus_msgs.len(), 1);
     }
 
+    /// Should maintain the invariant that pending_values and pending_values map only contain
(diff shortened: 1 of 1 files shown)


eranrund (approved):
LGTM, although I think it would be better to submit some valid transactions as well to see that they are in fact making it into `pending_values`.

mfaulk:
@eranrund I believe what you're talking about is already done as part of `test_receive_tasks`. The unit test added by this PR covers some of the same functionality, and is mostly meant to be the minimal test case that would catch a regression.

garbageslam (approved):

garbageslam:
This is now moot

PR #817, opened 2021-05-12
Undo block version bump

jcape:
### Motivation

Bumping the block version represents a breaking change to clients, and the impetus to bump the block version is a semantic change to this field. This updates the release branch to revert this change.

### In this PR
* Revert the actual block version bump, while keeping the rest of PR 

### Future Work
* Re-designing the ledger to handle changes in an extensible way.


eranrund (approved):
LGTM pending tests

sugargoat (approved):

PR #837, opened 2021-06-14
Remove dead code from VerificationReport.

christian-oudard:
Removed functions used only in a unittest, along with that unittest.


jcape (approved):
LGTM, merge after CI passes.

eranrund (approved):

garbageslam (approved):

PR #855, opened 2021-07-13
Consensus enclaves don't run on sha-ni chips right now.

jcape:
### Motivation

The master branch is currently producing a `SIGILL` when trying to verify the quote given by a quoting enclave. None of the CPUs we expect to run an enclave on will support the SHA-NI instructions anyways.

### In this PR
* Disable the `asm` feature from the `sha2` crate (should never be used anyways)

### DIFF
--- a/consensus/enclave/trusted/Cargo.lock
+++ b/consensus/enclave/trusted/Cargo.lock
@@ -1566,16 +1566,6 @@ dependencies = [
  "cpufeatures",
  "digest",
  "opaque-debug",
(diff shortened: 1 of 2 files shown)


garbageslam (approved):

garbageslam:
shouldn't this be on a release branch and not master? we cannot cut a hotfix release from fog master right now

jcape:
The sha2 changes are on master.

eranrund (approved):

PR #857, opened 2021-07-15
Update cpufeatures fork

jcape:
### Motivation

Due to issues with cargo feature unification, which is seemingly only exposed (sadly) when running on an actual enclave, we need to make our patches to `cpufeatures` just use `target_features` directly.

### In this PR
* Update our crates.io patch of `cpufeatures` to one which works in our environment.

### Future Work
* Create test case and report issue to cargo
* Get a reasonable patch for cpufeatures upstreamed after cargo issue is fixed.

### DIFF
--- a/consensus/enclave/trusted/Cargo.toml
+++ b/consensus/enclave/trusted/Cargo.toml
@@ -39,7 +39,6 @@ mbedtls = { version = "0.8.1", default-features = false, features = ["no_std_dep
 mbedtls-sys-auto = { version = "2.26.1", default-features = false, features = ["custom_threading"] }
 
 # We include these here so we can select in-the-enclave features
-cpufeatures = { version = "0.1.5", features = ["compile-only"] }
 lazy_static = { version = "1.4", features = ["spin_no_std"] }
 sha2 = { version = "0.9.5", default-features = false }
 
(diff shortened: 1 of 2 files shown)


consensus/enclave/trusted/Cargo.toml
```diff
@@ -39,7 +39,6 @@ mbedtls = { version = "0.8.1", default-features = false, features = ["no_std_dep
 mbedtls-sys-auto = { version = "2.26.1", default-features = false, features = ["custom_threading"] }
 
 # We include these here so we can select in-the-enclave features
-cpufeatures = { version = "0.1.5", features = ["compile-only"] }
```
garbageslam, line 42: So we're changing the patch version and removing it?

jcape: cpufeatures is used directly by the RustCrypto crates (aes-gcm, sha2, maybe blake2, etc.).

This line was there to ensure the "compile-only" feature was turned on, via feature unification, but not actually used in the trusted crate.

Since feature unification doesn't work that way, there's no need for this line to exist.

We still patch in our fork in the enclave so our fork is the one used by the upstream RustCrypto crates.

eranrund (approved):

garbageslam (approved):

PR #860, opened 2021-07-23
Work around broken cargo trim -d

jcape:
### Motivation

PRs to master are still failing, due to a bug / change in `cargo trim` (the `--directory` argument doesn't appear to function)

### In this PR
* Run `cargo trim init` in the top-level and trusted enclave directories directly.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -138,8 +138,10 @@ commands:
 
             # Configure cargo-trim with the project's Cargo.lock files
             mkdir -p ~/.config
-            cargo trim --directory "$(pwd)"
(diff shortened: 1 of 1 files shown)


garbageslam (approved):
What is the point of this step? Is it like making the ci build cache smaller?

eranrund (approved):

jcape:
> What is the point of this step? Is it like making the ci build cache smaller?

Yeah, the thing is, I don't know where/when we'd use that cache... It's bifurcated by arch and git revision, it feels like it's only going to be used if you re-run the same check on the same revision without changing anything---but I just want merges to master to start passing CI again, TBH :-)

PR #866, opened 2021-08-10
fix documentation for onetime key vs target key

garbageslam:
It is very confusing to have two names for the same value used
in source code. In TxOut's we have a value called "target_key",
but in some functions in `onetime_keys.rs` this is called the
`onetime_public_key`.

"One-time public key" is not a good technical term for a key,
because many keys in cryptography are one-time use, and this name
does not disambiguate the key from any other one-time use key.

This commit unifies the source code and code comments around the
term "target_key" which is what we are already using in the TxOut
structure.

### DIFF
--- a/transaction/core/src/onetime_keys.rs
+++ b/transaction/core/src/onetime_keys.rs
@@ -3,16 +3,18 @@
 //! # MobileCoin transactions use CryptoNote-style `onetime keys` to protect
 //! recipient privacy.
 //!
-//! When creating a transaction, the sender computes a onetime public key for
-//! each output in such a way that only the sender and the recipient know who
-//! the recipient is, and only the recipient is able to compute the
-//! corresponding onetime private key that is required to spend the output.
+//! When creating a transaction, the sender computes a one-time use public key
+//! for each output in such a way that only the sender and the recipient know
+//! who the recipient is, and only the recipient is able to compute the
+//! corresponding private key that is required to spend the output.
 //!
 //! To further protect recipient privacy, an output's onetime key is computed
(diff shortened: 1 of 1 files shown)


transaction/core/src/onetime_keys.rs
```diff
@@ -3,17 +3,19 @@
...
+//! who the recipient is, and only the recipient is able to compute the
+//! corresponding private key that is required to spend the output.
 //!
 //! To further protect recipient privacy, an output's onetime key is computed
 //! for a `subaddress` that the recipient generated from their CryptoNote-style
 //! address. This makes it easy for a recipient to use different subaddresses
 //! for different purposes and keep track of how much MobileCoin was sent to
 //! each subaddress.
 //!
+//! This key is called the "target key" of the TxOut.
```
garbageslam, line 17: I guess that this is actually not the right thing to say:
the comment below about target key being the "onetime public key" is what was correct

So maybe we should just say, the public key is called "target key" of the TxOut, and the corresponding private key is the "onetime private key"? Since I think this commit isn't actually getting rid of the term `onetime_private_key` in source code

garbageslam:
Maybe this isn't the right approach for now -- my real goal here is to try to make it easier for developers to figure out if a TxOut matches a particular subaddress. So maybe I will just try to make a helper function for that and document that, and avoid messing with the code I was touching in this commit

jcape:
Isn't "one time public key" a term from the original cryptonote paper?

garbageslam:
@james That's true but I don't think it's a current term anymore.

We already went with `target_key` in the `TxOut`'s and that name is now baked into the hash

PR #884, opened 2021-08-27
Merge mfaulk 2247

garbageslam:
This is a merge and update of https://github.com/mobilecoinfoundation/mobilecoin/pull/770

The API of the `PendingValues` object changed since that PR was created, but this test is pretty important and we need to land it.
This merges that PR and then fixes up the issues.

### DIFF
--- a/consensus/service/src/byzantine_ledger/worker.rs
+++ b/consensus/service/src/byzantine_ledger/worker.rs
@@ -1231,8 +1231,8 @@ mod tests {
         // Initially, pending_values should be empty.
         assert!(worker.pending_values.is_empty());
         assert_eq!(worker.receive_tasks(), true);
-        // Should maintain the invariant that pending_values only contain tx_hashes
-        // corresponding to transactions that are valid w.r.t the current ledger.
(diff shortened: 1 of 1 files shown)


consensus/service/src/byzantine_ledger/worker.rs
```diff
@@ -1251,6 +1251,69 @@ mod tests {
         assert_eq!(worker.pending_consensus_msgs.len(), 1);
     }
 
+    /// Should maintain the invariant that pending_values and pending_values map
```
mfaulk, line 1254: Looks like the pending_values and pending_values_map have been encapsulated in the PendingValues struct, so I think this comment is a little out of date. Maybe something like this is better:

```
/// Should maintain the invariant that pending_values only contains tx_hashes corresponding to transactions that are valid w.r.t. the current ledger.
```

mfaulk (approved):
This looks good. I think some of this logic is also covered by the 'test_clear_invalid_values_discards_invalid_values' unit test in pending_values.rs, but I think it's ok to have both unit tests.

jcape (approved):

PR #892, opened 2021-09-02
Use rust-mbedtls that has linker fix backported from upstream

garbageslam:
See commit message:
https://github.com/mobilecoinofficial/rust-mbedtls/commit/d0c2aea8700bf151951bf48daa9ac4724df924a0

This is an alternative to the hackier fix that was removing object
code manually from the `libmobilecoin_stripped.a` in the libmobilecoin
makefile

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2236,7 +2236,7 @@ checksum = "60302e4db3a61da70c0cb7991976248362f30319e88850c487b9b95bbf059e00"
 [[package]]
 name = "mbedtls"
 version = "0.8.1"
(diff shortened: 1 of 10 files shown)


garbageslam:
@the-real-adammork I confirmed that on OSX this gets rid of the `psa` symbols using `rust-nm` on the `libmobilecoin_stripped.a`

eranrund (approved):
Nice

jcape:
> @jcape i can cherry-pick that commit in https://github.com/mobilecoinofficial/rust-mbedtls to `mc-develop` branch as well, I assume that is helpful to you with the rust uprev PR

Yeah, that would be useful---I'd like to get into the habit of locking down our forks and forcing any changes to them through a PR process to ensure external changes we're adding to them are getting properly reviewed.

jcape (approved):

the-real-adammork (approved):
works !

PR #893, opened 2021-09-03
Remove CompressedCommitment processing in `transaction.rs`

the-real-adammork:
## **Final**  

Soundtrack of this PR: [Roy Ayers & Wayne Henderson - For Real](https://www.youtube.com/watch?v=DaYKIcqIJdU)

### Motivation

TxOut no longer gets the CompressedCommitment in the protobuf. But it can be recreated from its constituent parts:

- Masked Value -> Value
- Shared Secret of `tx_out_public_key` & `view_private_key`

### In this PR

- Requires changes in two binding functions **for iOS**
- Add a new function to reconstruct an Amount's `commitment` from its constituent parts


UkoeHB:
> TxOut no longer gets the CompressedCommitment in the protobuf. it is now the tx_out_shared_secret that is computed by multiplying the tx_out_public_key & view_private_key.

Can you explain this? Isn't `TxOut` 1:1 with outputs in the ledger, which contain compressed commitments?

the-real-adammork:
@UkoeHB this is the context I was given for this change:

> Fog no longer stores a compressed commitment in a TxOutRecord

https://github.com/mobilecoinofficial/android-sdk/pull/34

UkoeHB:
It looks like so far this PR only touches `libmobilecoin`, are you planning to make changes to `transaction` as well? Specifically [this struct](https://github.com/mobilecoinfoundation/mobilecoin/blob/753ebb3863f89e0835cfc893b06f8264d23a5461/transaction/core/src/tx.rs#L233).

the-real-adammork:
@UkoeHB do you think that is necessary ? The android side of these changes didn't need todo that.

jcape:
It shouldn't remove that field from the transaction-core TxOut.

jcape:
It's not only unnecessary, it is a breaking change for the ledger.

UkoeHB:
@the-real-adammork I don't think it needs to change. I got a bit confused about the scope of this PR based on the semantics used.

libmobilecoin/src/transaction.rs
```diff
@@ -147,12 +166,14 @@ pub extern "C" fn mc_tx_out_get_value(
...
-        let amount = Amount::try_from_ffi(&tx_out_amount)?;
         let tx_out_public_key = RistrettoPublic::try_from_ffi(&tx_out_public_key)?;
         let view_private_key = RistrettoPrivate::try_from_ffi(&view_private_key)
             .expect("view_private_key is not a valid RistrettoPrivate");
 
         let shared_secret = get_tx_out_shared_secret(&view_private_key, &tx_out_public_key);
+        let value =
+            (tx_out_amount.masked_value as u64) ^ get_value_mask(&shared_secret);
+        let amount: Amount = Amount::new(value, &shared_secret)?;
         let (val, _blinding) = amount.get_value(&shared_secret)?;
```
UkoeHB, line 177: Btw I think `value == val` here.

the-real-adammork: Ya I agree, so maybe it could be optimized. Do you know what the `_blinding` field represents?

UkoeHB: It is the amount's Pedersen commitment's blinding factor.

the-real-adammork:
Changing the solution to return a `crc32` encoding of the `Amount.commitment`. Ill push a commit shortly.

the-real-adammork:
Updated solution includes a function to reconstruct TxOut commitments from their constituent parts.

libmobilecoin/src/transaction.rs
the-real-adammork: updated to pass out the `value` instead of creating an `Amount` object

libmobilecoin/src/transaction.rs
```diff
@@ -147,15 +173,15 @@ pub extern "C" fn mc_tx_out_get_value(
     out_error: FfiOptMutPtr<FfiOptOwnedPtr<McError>>,
 ) -> bool {
     ffi_boundary_with_error(out_error, || {
-        let amount = Amount::try_from_ffi(&tx_out_amount)?;
         let tx_out_public_key = RistrettoPublic::try_from_ffi(&tx_out_public_key)?;
         let view_private_key = RistrettoPrivate::try_from_ffi(&view_private_key)
             .expect("view_private_key is not a valid RistrettoPrivate");
 
         let shared_secret = get_tx_out_shared_secret(&view_private_key, &tx_out_public_key);
-        let (val, _blinding) = amount.get_value(&shared_secret)?;
```
garbageslam, line 156: so, because we are removing this line, we no longer have a failure mode that corresponds to "TransactionCrypto", I think this is why kyle wanted to have the `crc32` value that we check

OTOH I think we only get to this point after view key matching has succeeded?

You might want to update documentation that TransactionCrypto is not a failure mode anymore?

libmobilecoin/src/transaction.rs
```diff
@@ -18,22 +19,44 @@ use mc_util_ffi::*;
...
-    }
+        let tx_out_public_key = RistrettoPublic::try_from_ffi(&tx_out_public_key)?;
+
+        let shared_secret = get_tx_out_shared_secret(&view_private_key, &tx_out_public_key);
+        
+        let value =
+            (tx_out_amount.masked_value as u64) ^ get_value_mask(&shared_secret);
+
+        let amount: Amount = Amount::new(value, &shared_secret)
+            .expect("could not create amount object");
```
garbageslam, line 50: I think it's better not to use `.expect` here, because if the error happens, the app will crash. I think usually we are using `?` to pass the error back, ultimately as a human readable string error message?

the-real-adammork: good idea, Ill make that change.

As a side-note: how can I know what Error types the function may throw/return ? I see those enumerated in the documentation comments and Im wondering if there is an automated way for me to figure it out for a given function.

garbageslam: I think I'm just reading the code and looking at what appear to be the failure paths, so it's literally like, if there is `return Err(...);` or `?` which is a shorter syntax for `return Err(...);`, then those are the paths where it can return an error, and otherwise, it would only be like, if we have `return foo(); ` and `foo` returns `Result<..., ...>` and so could also be an error.

garbageslam: Usually like, if a function return `Result<Foo, Error>` then `Error` will be an enum and you can look at the cases -- in this code I think all the errors get stringified before going across FFI? not sure, would have to check

jcape: [Plug for CLion+Rust, with it's inline type hints, goes here]

In `Rust`, there are no exceptions, and `Result<T, E>` is actually just a type-parameterized enum:

```
pub enum Result<T: Sized, E: Sized> {
    Ok(T),
    Err(E),
}
```

So when you do `return Err(SomeError::ErrCode);`, you're really creating an concrete instance of that `Result` enum, and returning it by value---there's zero magic around returning errors in the usual case, you're just creating a new `Result::Err` enum value, which contains the `SomeError` value inside it, and returning that enum-in-an-enum to the caller, by value. LLVM is really good at optimizing this case, so it's not typically a real hit to performance.

And, as Chris mentioned, the `?` is syntactic sugar that resolves:

```
fn inner_foo() -> Result<u64, InnerError> { /* ... */ }

fn outer_foo() -> Result<i128, OuterError> {
    let value = do_foo()?;
    /* ... */
    value as i128
}
```
into this:
```
fn inner_foo() -> Result<u64, InnerError> { /* ... */ }

fn outer_foo() -> Result<i128, OuterError> {
    let value = {
        let result = do_foo();
        match result {
            Ok(value) => value,
            Err(error) => return Err(OuterError::from(error)), // early return
        }
    };
    /* ... */
    value as i128
}
```

libmobilecoin/src/transaction.rs
the-real-adammork: How would I check the crc32 of this amount ? Would I need to pass in the original TxOut object from the client ?

libmobilecoin/src/transaction.rs
```diff
@@ -18,22 +18,42 @@ use mc_util_ffi::*;
...
+/// # Preconditions
+///
+/// * `view_private_key` - must be a valid 32-byte Ristretto-format scalar.
+#[no_mangle]
+pub extern "C" fn mc_tx_out_reconstruct_commitment(
+    tx_out_amount: FfiRefPtr<McTxOutAmount>,
+    tx_out_public_key: FfiRefPtr<McBuffer>,
+    view_private_key: FfiRefPtr<McBuffer>,
+    out_tx_out_commitment: FfiMutPtr<McMutableBuffer>,
+    out_error: FfiOptMutPtr<FfiOptOwnedPtr<McError>>,
```
garbageslam, line 35: I think what I would suggest is, pass the `tx_out_commitment_crc32` as an argument to this perhaps?

Then you can compute the crc32 of the commitment bytes that this function produces, and confirm that they match, or return an error if they don't?

garbageslam: That would be simpler than trying to do it in the other function I think, sorry for the misguided advice earlier

the-real-adammork: ya that makes sense. I won't have time to implement and test that for this PR because it involves a decent amount of Swift changes. I will create a backlog ticket to add that functionality.

garbageslam (approved):
LGTM

voloshyn (approved):
LGTM

PR #895, opened 2021-09-07
Make transaction-core code call it "tx_target_key" and not "onetime_public_key"

garbageslam:
The documentation in `onetime_keys.rs` explains that the term
`onetime public key` comes from cryptonote papers, and is also called
the "target key". However, in the `TxOut` object, we have gone with
the name `target_key`, while in many functions in `onetime_keys.rs`,
it is still called `onetime_public_key`.

This makes it much more difficult to read and use this code, because
there is no type safety that ensures that `onetime_public_key` is not
used with `tx_out.public_key` rather than `tx_out.target_key`, and
searching for `onetime_public_key` only gives very few hits in the
source code, so you have to read all of the documentation to figure
out how to use any of these functions correctly.

By changing the function names, and input parameter names, to things
like `tx_target_key`, we make it much easier to figure out how to
use these functions correctly with `TxOut` objects.

It's fine to have multiple names for a key in documentation, and
to match the literature. It's not fine to have multiple names in code
for the same thing, and unifying the names in this manner makes the
code substantially more maintainable.

### DIFF
--- a/consensus/enclave/impl/src/lib.rs
+++ b/consensus/enclave/impl/src/lib.rs
@@ -50,7 +50,7 @@ use mc_sgx_report_cache_api::{ReportableEnclave, Result as ReportableEnclaveResu
 use mc_transaction_core::{
     constants::MINIMUM_FEE,
     membership_proofs::compute_implied_merkle_root,
-    onetime_keys::{create_shared_secret, create_tx_public_key, create_tx_target_key},
+    onetime_keys::{create_shared_secret, create_tx_out_public_key, create_tx_out_target_key},
     ring_signature::{KeyImage, Scalar},
     tx::{Tx, TxOut, TxOutMembershipProof},
     validation::TransactionValidationError,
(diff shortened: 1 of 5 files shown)


consensus/enclave/impl/src/lib.rs
```diff
@@ -50,7 +50,7 @@ use mc_sgx_report_cache_api::{ReportableEnclave, Result as ReportableEnclaveResu
 use mc_transaction_core::{
     constants::MINIMUM_FEE,
     membership_proofs::compute_implied_merkle_root,
-    onetime_keys::{create_onetime_public_key, create_shared_secret, create_tx_public_key},
+    onetime_keys::{create_shared_secret, create_tx_public_key, create_tx_target_key},
```
mfaulk, line 53: Suggestion: maybe these should be 'create_output_public_key' and 'create_output_target_key', or 'create_tx_out_public_key' and 'create_tx_out_target_key'. I think the function name 'create_tx_public_key' hearkens back to when a transaction had a single public key that was shared by every output. Now, each output has a distinct target key and public key.

mfaulk (approved):
Thanks, Chris. I think you're right that we should standardize on "target_key", given that this is what is used in TxOut. There are a lot of other "onetime" names and variables in the codebase that should probably be changed to agree with "target_key", but I think its fine to establish "target_key" as correct, and then enforce more consistency in the rest of the code as we go.

jcape (approved):

PR #907, opened 2021-09-14
Make Encrypted memo changes releasable

garbageslam:
In testing, we discovered that old clients are unable to submit
transactions to new servers, even if the MRENCLAVE value is up to date.
This is because the entire TxPrefix is under the ring signature in the
transaction math, and the TxPrefix is different whether or not a memo
is there. An old server has protobufs that don't include the memo field,
so they compute the hash of the TxPrefix differently from clients that
included an empty memo.

To make this go, we decided to release this change in two steps.
First, we make versions of the servers that tolerate the presence of
the memo and compute the TxPrefix hash in the same way as the clients,
but also accept the state of affairs that there are no memos.

Existing clients can thus continue to work with this server simply by
updating the MRENCLAVE value that they have hard coded.

In a subsequent server release, the memos will become mandatory.
In the transition period before that release, Clients will all start
using the memos, and reading from them, expecting to find data there
(but tolerating the absence of that data, so that these clients work
in the period before this secondary server release, during which
memos are optional.)

The changes are:
* Don't require every TxOut to have a memo during validation
  (but keep that functionality under test, and refer to a ticket to
  bring it back.)
* MemoFn in the TxOut::new, MemoBuilder, and TransactionBuilder pathways
  returns `Result<Option<MemoPayload>,...>` rather than `Result<MemoPayload, ...>`
  to allow that memo builders may omit the memo.
  (This modificaiton will be reverted when memos are made mandatory.)
* A `NoMemoBuilder` is added in the `transaction-std` crate, which always
  produces `None`. This is intended to be used during the transationary
  period when the servers have not yet started accepting memos.
* Make android-bindings, libmobilecoin, mobilecoind, use the `NoMemoBuilder`.
  They can be switched to `EmptyMemoBuilder` or `RTHMemoBuilder` as necessary
  during the next release.

### DIFF
--- a/libmobilecoin/src/transaction.rs
+++ b/libmobilecoin/src/transaction.rs
@@ -10,9 +10,9 @@ use mc_transaction_core::{
     onetime_keys::{recover_onetime_private_key, recover_public_subaddress_spend_key},
     ring_signature::KeyImage,
     tx::{TxOut, TxOutConfirmationNumber, TxOutMembershipProof},
-    Amount, CompressedCommitment, MemoPayload,
+    Amount, CompressedCommitment,
 };
-use mc_transaction_std::{InputCredentials, RTHMemoBuilder, TransactionBuilder};
+use mc_transaction_std::{InputCredentials, NoMemoBuilder, TransactionBuilder};
 use mc_util_ffi::*;
 
 /* ==== TxOut ==== */
@@ -301,8 +301,9 @@ pub extern "C" fn mc_transaction_builder_create(
                     FogResolver::new(fog_resolver.0.clone(), &fog_resolver.1)
                         .expect("FogResolver could not be constructed from the provided materials")
                 });
-        let memo_builder = RTHMemoBuilder::default();
-        // FIXME: Enable recoverable transaction history by configuring memo_builder
+        // TODO: After servers are deployed that are supporting the memos,
+        // Enable recoverable transaction history by configuring an RTHMemoBuilder
+        let memo_builder = NoMemoBuilder::default();
         let mut transaction_builder = TransactionBuilder::new(fog_resolver, memo_builder);
         transaction_builder
             .set_fee(fee)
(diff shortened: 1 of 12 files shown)


libmobilecoin/src/transaction.rs
```diff
@@ -441,44 +442,19 @@ pub extern "C" fn mc_transaction_builder_add_output(
...
-            &recipient_address,
-            &fog_hint_address,
-            |_| Ok(MemoPayload::default()),
-            &mut rng,
-        )?;
-
-        out_tx_out_confirmation_number.copy_from_slice(confirmation.as_ref());
-        Ok(mc_util_serial::encode(&tx_out))
-    })
+    unimplemented!("TransactionBuilder::add_output_with_fog_hint_address was removed, please use add_change_output");
```
jcape, line 457: Could you flag this with `#[deprecated]`---ideally the manually constructed headers are also flagged with `__attribute__(deprecated("Use mc_transaction_builder_add_change_output"))`, but you'd need some cross-platform macros since `deprecated` is apparently a GNU extension.

transaction/core/src/validation/validate.rs
```diff
@@ -194,6 +196,7 @@ fn validate_outputs_public_keys_are_unique(tx: &Tx) -> TransactionValidationResu
 }
 
 /// All outputs have a memo (old-style TxOuts are rejected)
+#[allow(unused)]
```
jcape, line 199: Why is this required?

transaction/std/src/memo_builder/rth_memo_builder.rs
```diff
@@ -159,7 +159,7 @@ impl MemoBuilder for RTHMemoBuilder {
             .checked_add(1)
             .ok_or(NewMemoError::LimitsExceeded("num_recipients"))?;
         self.last_recipient = ShortAddressHash::from(recipient);
-        Ok(if let Some(cred) = &self.sender_cred {
+        Ok(Some(if let Some(cred) = &self.sender_cred {
```
jcape, line 162: Nitpick (readability): It might be slightly less intimidating to pull the value going into the `Ok()` out and trust the compiler to do the right thing. I realize that was in the original PR, and it's not strictly necessary, but two layers of `if let Some(foo)` inside an `Ok()` is a bit much...

transaction/std/src/transaction_builder.rs
```diff
@@ -212,12 +208,12 @@ impl<FPR: FogPubkeyResolver> TransactionBuilder<FPR> {
     /// * `fog_hint_address` - The public address used to create the fog hint
     /// * `memo_fn` - The memo function to use (see TxOut::new_with_memo)
     /// * `rng` - RNG used to generate blinding for commitment
-    pub fn add_output_with_fog_hint_address<RNG: CryptoRng + RngCore>(
+    fn add_output_with_fog_hint_address<RNG: CryptoRng + RngCore>(
```
jcape, line 211: If this is no longer being called from source, it should also get marked with `#[deprecated]`.

If things still do call it, then that will break the build, unfortunately, so it's tricky. When things are broken up into crates, it will be easier (IMO) to move the `-D warnings` stuff into CI and start being more expressive with warnings.

jcape (approved):
LGTM, some small improvements can be done with marking functions as deprecated.

transaction/core/src/validation/validate.rs
garbageslam: it's required because it's not used except in tests, so during `cargo build` we get an unused function warning.

we intend to bring this back in resolution to #907 , so i figured it makes sense to just keep the code here and keep it under test, and then start using it in the main `validate()` function when we're ready to do that

transaction/std/src/transaction_builder.rs
garbageslam: i think it is being called from source, it's not actually deprecated

garbageslam: it's being used at line 181 above, it is called by `add_change_output`

transaction/std/src/memo_builder/rth_memo_builder.rs
garbageslam: yeah okay i can do that

libmobilecoin/src/transaction.rs
garbageslam: Why is `#[deprecated]` better than putting `unimplemented!()` ?

By putting `unimplemented!()` I have done more than just deprecate it, I have actually broken code that is calling it. This is necessary because we made the transaction builder API it was calling private and not public anymore.

jcape: Because unimplemented is a runtime file, deprecated is a compiler warning.

garbageslam: yes, i'm not attempting to create a deprecation cycle, i'm attempting to remove this function since I don't think Adam will be using it in the next release, and any form of testing will eventually hit this, since every transaction involves a change tx out

garbageslam: In review of the first PR, eran suggested to just delete it rather than do a deprecation cycle, if the only consumer is libmobilecoin

jcape: I guess I'm not sure what making this a runtime error accomplishes, if the goal is to remove the function anyways, I'd much rather break the build and just fix it...

transaction/core/src/validation/validate.rs
jcape: Why not `#[cfg(test)]`?

libmobilecoin/src/transaction.rs
garbageslam: @the-real-adammork I'm happy to do whatever is most convenient for you

transaction/core/src/validation/validate.rs
garbageslam: that should be fine too

libmobilecoin/src/transaction.rs
the-real-adammork: I don't have a preference. Seems like I'll need to make Swift changes either way.

the-real-adammork: iOS/Swift code does not call into this function directly

the-real-adammork (approved):

voloshyn (approved):
LGTM

PR #960, opened 2021-09-16
Group enclave and base dirs together, add reviewers.

jcape:
This PR is attempting to coerce dependabot into updating multiple cargo workspaces/lock files in a single PR. We'll see how well it goes...

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,15 +1,43 @@
 version: 2
 updates:
 - package-ecosystem: cargo
-  directory:
-    - "/"
-    - "/consensus/enclave/trusted/"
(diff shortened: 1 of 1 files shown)


eranrund (approved):
Nice (if it works :))

jcape:
Unfortunately it looks like it won't work: https://github.com/dependabot/dependabot-core/issues/2824

jcape:
https://github.com/dependabot/dependabot-core/issues/2178 is the issue to fix it, but it's just a bunch of plus-ones at this point.

jcape:
Just going to close this, the best option is simply to triage the PRs as they come in.

Issue #1024, opened 2021-09-17
Investigate using `signal-hook` instead of `nix` in `fog-load-testing`


garbageslam:

garbageslam:
fixed

PR #1032, opened 2021-09-22
simplify fog-load-testing implement, and fix it

garbageslam:
The original goal of this commit was to remove nix and libc from
being direct dependencies of fog-load-testing. When I first created
this, I knew that fundamentally what I needed to do was make the
load testing binary launch the binary under test, and then listen
for SIGCHLD to see if it dies. This is important because otherwise
the test continue even after that point. We also want to actually
stop and reap the child when the test ends, or it is very annoying
for the user to run the test -- often if you are changing config
params, the server will fail on startup etc. Or if the watcher
stuff is wrong, then the server won't make any progress. If the
child continues running after the test ends and you have to manually
find it and kill it that's just annoying.

However, upreving nix recently turned out to be annoying, because
of something where they didn't bump MSRV when everyone else did
and created wierd constraints on their dependencies.

I started looking at alternatives to nix / manually writing a signal
handler, and started investigating if the rust std process module
will actually do this.

Sure enough it turns out that the rust std::process stuff does listen
for SIGCHLD and does give you a way to check if the child died, and
does give you a way to call "waitpid" effectively and reap the child
like you are supposed to.

https://github.com/rust-lang/rust/blob/cfff31bc833070a00578bd6178160aeed56f28ba/library/std/src/sys/unix/process/process_unix.rs#L187

https://github.com/rust-lang/rust/blob/cfff31bc833070a00578bd6178160aeed56f28ba/library/std/src/sys/unix/process/process_unix.rs#L594

So in this commit we ripped out all the manual SIGCHLD handler stuff,
and we actually made a wrapper over `std::process:Child` that kills
the child on drop if it isn't dead yet, which is an improvement over
what previously existed, based on my testing.

We also learned, in testing, that the fog load testing target is broken,
because the test is not actually exercised in CI, and at some point
we changed the arguments to the ingest server, around peers etc.
Additionally we made it a requirement to activate the ingest server,
and load testing wasn't doing that, because it wasn't under test.

In this commit we also fixed that stuff.

In the future, it would be nice to run the load testing target in CI
or something, but we should probably also figure out how to harvest
the data, which is a larger project.


fog/load_testing/src/sig_child_handler.rs
```diff
@@ -1,75 +0,0 @@
-// Copyright (c) 2018-2021 The MobileCoin Foundation
```
garbageslam, line 1: removing all this is the main point of the PR

fog/load_testing/src/bin/ingest.rs
```diff
@@ -113,12 +114,44 @@ impl core::fmt::Display for TestResult {
...
+            }
+            Ok(None) => {}
+            Err(err) => {
+                panic!("error getting child status: {}", err)
+            }
+        }
+    }
+}
+
+impl Drop for AutoKillChild {
```
garbageslam, line 133: this improves the way the test works, because if the load testing binary panics (e.g. the test timed out), rust will kill this destructor and kill the child, so it doesn't spam your console and force you to `killall ...`

fog/load_testing/src/bin/ingest.rs
```diff
@@ -220,6 +262,21 @@ fn load_test(ingest_server_binary: &Path, test_params: TestParams, logger: Logge
...
+                Duration::from_secs(30),
+                grpcio_env,
+                logger.clone(),
+            );
+            ingest_client
+                .activate()
+                .expect("Could not activate ingest server");
+            std::thread::sleep(std::time::Duration::from_millis(10));
+            ingest_server.assert_not_stopped();
+        }
```
garbageslam, line 278: this is needed because this code hasn't been touched since the ingest server was required to be activating it, and nothing was exercising the load test

fog/load_testing/src/bin/ingest.rs
```diff
@@ -268,6 +325,33 @@ fn load_test(ingest_server_binary: &Path, test_params: TestParams, logger: Logge
...
+                            .expect("Could not add block signature");
+                    }
+                }
+                // Sanity check that watcher is behaving correctly
+                assert_eq!(
+                    watcher
+                        .highest_common_block()
+                        .expect("Could not get highest common block"),
+                    block.index,
+                );
```
garbageslam, line 353: similarly, the load test wasn't writing anything to the watcher, so we had to add this code to make it work

fog/load_testing/src/bin/ingest.rs
garbageslam: compare with `std::process::Child`, which per documentation, does not kill the child in the destructor: https://doc.rust-lang.org/std/process/struct.Child.html

garbageslam:
Note: This fixes #1024

jcape (approved):

eranrund (approved):

PR #1073, opened 2021-09-27
fix DestinationMemo code comment on the total outlay

cbeck88:
Sam pointed out that there is a discrepancy between how the
`total_outlay` is documented in code comments and how it is used in
code, including the memo builder.

The code comment indicates that the fee is not part of the total
outlay, but the memo builder inlcudes the fee in the total outlay.

The MCIP documents that total outlay means the "total amount that
deducted from the sender's balance" and so includes the fee, and
this is the reference material. This also matches the code.
This also matches the tests in the `fog-test-client` integration
test which is asserting that total outlay has the expected value.

This commit brings the documentation on `DestinationMemo` in line
with everything else, I think I was just confused when I wrote it.

### DIFF
--- a/transaction/std/src/memo/destination.rs
+++ b/transaction/std/src/memo/destination.rs
@@ -36,15 +36,16 @@ pub struct DestinationMemo {
     /// scenario.
     fee: u64,
     /// The sum of all outlays of the transaction (in picomob)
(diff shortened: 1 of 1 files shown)


samdealy (approved):
LGTM, thanks for updating!

PR #1086, opened 2021-09-30
minor fixes in fog test client

garbageslam:
For measuring elapsed time, the Instant (which corresponds to
monotonic time) is always preferable to using SystemTime
(which corresponds to the OS-configured wall clock)


wjuan-mob (approved):

garbageslam:
this was merged in fog-canary PR

PR #1091, opened 2021-10-04
Make fog-ingest-enclave `set_ingress_private_key` function return bool

samdealy:
### Motivation

[FOG-180](https://github.com/mobilecoinfoundation/fog/pull/116/files#diff-1bd75fa7d4e53379b9fe76ed5de3edd920c698f1d188a08ae34d2ba9fb414247R212) introduced logic to check if a user's private key had changed. This check introduced additional calls across the enclave border, which increases latency. 

### In this PR
* fog-ingest-enclave `set_ingress_private_key` function returns bool that indicates if the key changed or not.
* fog ingest server controller uses this returned value to decide whether or not to seal the private key to disk.

The `set_ingress_private_key` function runs inside the enclave, so the logic that decides if the key changed or not needs to be constant time. As such, I implemented `subtle::ConstantTimeEq `for `RistrettoPrivate`. 

Fixes this [ticket](https://app.asana.com/0/1200353042931237/1200719673733203/f).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2992,6 +2992,7 @@ dependencies = [
  "serde_json",
  "sha2 0.9.5",
  "signature",
+ "subtle 2.4.1",
  "tempdir",
  "x25519-dalek",
  "zeroize",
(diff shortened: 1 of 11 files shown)


garbageslam (approved):
LGTM thank you!

PR #1103, opened 2021-10-08
Reduce size of `libmobilecoin` at the expense of compile time (`codegen-units = 1`)

the-real-adammork:
Soundtrack of this PR: [Fatboy Slim - Mad Flava (Autograf Tribute Mix)](https://www.youtube.com/watch?v=hXv7ogsWY7c)

### Motivation

We needed to reduce the size of our iOS SDK

### In this PR

Add a new build profile `mobile-release` that uses `1` codegen unit instead of the default `16`. 

This reduces the `.a` file size by `~30%` at the expense of compile time. 

Based on [my understanding of codegen units](https://doc.rust-lang.org/cargo/reference/profiles.html#codegen-units), this is a low risk change.

#### Bottomline Savings

I verified the space savings carry to the App Store. These lines come from `Packaging.log` which Xcode can generate for our Example App. 

`mobile-release` for `arm64` (iPhone)
```
...
copying file ./Frameworks/LibMobileCoin.framework/LibMobileCoin ... 
2021-10-07 23:38:28 +0000  10727776 bytes for ./Frameworks/LibMobileCoin.framework/LibMobileCoin
...
```

`release` for `arm64` (iPhone)
```
...
copying file ./Frameworks/LibMobileCoin.framework/LibMobileCoin ... 
2021-10-08 00:26:01 +0000  14438976 bytes for ./Frameworks/LibMobileCoin.framework/LibMobileCoin
...
```

### Future Work

Update for `libmobilecoin-ios-artifacts` to pull from the new `mobile-release` profile.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -166,6 +166,10 @@ inherits = "release"
 debug = false
 lto = true
 
+[profile.mobile-release]
+inherits = "mobile"
+codegen-units = 1
+
 # Skip the need for LD_LIBRARY_PATH in `cargo test`
 [profile.test]
 rpath = true
--- a/libmobilecoin/Makefile
+++ b/libmobilecoin/Makefile
@@ -11,9 +11,11 @@ export IAS_MODE ?= DEV
 
 ### Build Configuration
 
-CARGO_PROFILE ?= release
+CARGO_PROFILE ?= mobile-release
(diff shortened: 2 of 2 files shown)


eranrund (approved):
Nice!

PR #1140, opened 2021-10-29
[🍒  pick] Bitcode for `release-1.2.0`

the-real-adammork:
### Motivation

Add support for Apple's `Bitcode` and reduce the size of our SDK & its dependencies. With this enabled the compressed "downloadable" size **shrinks by ~25%.**  

#### `ENABLE_BITCODE = YES`

```
App Thinning Size Report for All Variants of Example
App + On Demand Resources size: 7.2 MB compressed, 19.7 MB uncompressed
App size: 7.2 MB compressed, 19.7 MB uncompressed
```

#### `ENABLE_BITCODE = NO`

```
App Thinning Size Report for All Variants of Example
App + On Demand Resources size: 9.6 MB compressed, 35.6 MB uncompressed
App size: 9.6 MB compressed, 35.6 MB uncompressed
```

#### `rust-bitcode`

Shallow clone [`rust-bitcode`](https://github.com/mobilecoinofficial/rust-bitcode) to compile `rustup` toolchains that include LLVM compatible with Apple's Xcode build system.

#### No more symbol stripping

Remove "symbol" stripping step from the `Makefile` because it's not compatible with enabling `Bitcode`. While this step had large impact on the size of the archive `.a` (10mb vs 55mb) it has a small impact on the "downloadable" size (~0.2mb). This is because un-linked symbols are not included in final binary regardless of whether or not they get stripped.

#### Remove build-target `x86_64-apple-darwin`

To my knowledge MobileCoin and Signal are not using this target, and it does not support `Bitcode`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -74,3 +74,6 @@ mobilecoind/clients/python/mob_client/blockchain_pb2.py
 
 # css files (should not be committed)
 *.css
+
+# libmobilecoin/rust-bitcode build
+libmobilecoin/rust-bitcode/*
--- a/libmobilecoin/Makefile
+++ b/libmobilecoin/Makefile
@@ -14,8 +14,9 @@ export IAS_MODE ?= DEV
 CARGO_PROFILE ?= mobile-release
 CARGO_BUILD_FLAGS ?=
 CARGO_TARGET_DIR ?= ../target
+RUST_BITCODE_DIR ?= rust-bitcode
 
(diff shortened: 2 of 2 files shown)


jcape (approved):

garbageslam:
Is cherry-picking the right thing to do? this will make it hard to merge the release branch into master.

garbageslam (approved):

PR #1144, opened 2021-11-01
Update usage of grpc-rs to use new server-side headers API.

christian-oudard:


remoun:
This still seems useful. Do we want to try and merge it?

jcape:
Superseded by #1592

PR #1147, opened 2021-11-03
Ensure underlying SqlRecoveryDb is not dropped during sync_keys_from_remote test

samdealy:
### Motivation

The "sync_keys_from_remote" test, which was added in #1139, intermittently fails during CI runs. The error message is: `process didn’t exit successfully: /root/project/target/debug/deps/sync_keys_from_remote-df5d33ac3e128cd9 (signal: 11, SIGSEGV: invalid memory reference)`. 

### In this PR
* Retain the db test context after the test's set up method. This ensures that the underlying SQL db won't be dropped.

### Future work
We think that retaining this DB will mitigate the chance of this segfault, but if it doesn't then we'll consider adding (a) logging or (b) investigate using gdb on the CI machines.


garbageslam (approved):

garbageslam:
thank you!

eranrund (approved):

PR #1148, opened 2021-11-03
Store fog test core dumps as CircleCi build artifacts

samdealy:
### Motivation

When a flaky fog tests fail due to segfaults, there is no easy way to grab the core dumps from the CircleCi machine. To complicate matters further, even if you ssh into the CircleCi machine, the test may not fail during multiple runs. 

### In this PR
* Stores core dumps produced by the "Run fog tests" action as a CircleCi build artifact. Follows this implementation [guide](https://circleci.com/docs/2.0/artifacts/#uploading-core-files).

Fixes [ticket](https://app.asana.com/0/1200353042931237/1201321243869915/f).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -365,7 +365,16 @@ commands:
       - run:
           name: Run fog tests
           command: |
+            # tell the operating system to remove the file size limit on core dump files
+            ulimit -c unlimited
             cargo test --package "mc-fog-*" -j 4 --frozen --no-fail-fast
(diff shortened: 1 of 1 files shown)


garbageslam (approved):

eranrund (changes requested):
Nice! This should also get added to all the other `cargo test` executions we have here (e.g. `run-mc-tests`)

eranrund (approved):

PR #1150, opened 2021-11-05
add tests for maximum fog domain vs. 255 b58 pub address limit

garbageslam:
Some exchanges use `VARCHAR(255)` for cryptocurrency public addresses, and it is difficult for them to migrate their SQL DBs.

We want that the mobilecoin b58 printable wrapper address strings are always within 255 bytes.

This imposes a limit on how long the fog report url can be.

The purpose of this test is to experimentally determine exactly what the limit is.

The test uses repetition because b58 encoding length isn't determined completely by payload length, it depends on what bytes are in the payload. So we try it with several choices of the mobilecoin private keys.

The test ensures that:
* Any domain less than the limit leads to a public address less than the 255 limit, trying private keys ten times
* Any domain at or more than the limit does violate the 255 limit at least sometimes, so the domain length limit is tight.

We compute a domain limit of 34 characters, so `fog.mobilecoin.com` would have to be less than 34 characters.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4941,6 +4941,7 @@ dependencies = [
  "mc-crypto-rand",
  "mc-util-from-random",
  "mc-util-serial",
+ "mc-util-test-helper",
  "pem",
  "rand 0.8.4",
  "rand_hc 0.3.1",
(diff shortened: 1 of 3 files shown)


eranrund (approved):

jgreat (approved):
Ran tests locally LGTM

PR #1158, opened 2021-11-08
run fog conformance tests in release mode

garbageslam:
the tests run locally in 1 minute instead of 29 minutes when we
use --release, so this may improve performance in CI also. the
reason for the performance gain is that all the ORAM servers are
much faster when compiled in release mode

(see stats here: https://github.com/mobilecoinfoundation/mobilecoin/pull/1157)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -385,7 +385,7 @@ commands:
       - store_artifacts:
           path: /tmp/core_dumps
 
-  # A job that runs the fog-conformance-tests, building things in debug mode
+  # A job that runs the fog-conformance-tests, building things in release mode
   # Note: If we bring back the run-parallel-tests stuff, we could make this use --skip-build,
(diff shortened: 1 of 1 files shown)


garbageslam:
This appears to improve the time for this overall CI step from 40 min to 9 min

eranrund (approved):

jcape (approved):

.circleci/config.yml
```diff
@@ -388,7 +388,7 @@ commands:
   # A job that runs the fog-conformance-tests, building things in debug mode
```
wjuan-mob, line 388: This should probably say release?

wjuan-mob (approved):

.circleci/config.yml
garbageslam, line 388:
```suggestion
  # A job that runs the fog-conformance-tests, building things in release mode
```

PR #1163, opened 2021-11-09
Add retries to many fog enclave connections using retry crate

garbageslam:
This addresses many old tickets, and I developed this while trying
to figure out why the test client failed with seemingly retriable errors

There is a large amount of tech debt to resolve the discrepancy
between mc-connection and mc-fog-enclave-connection. They are mostly
supposed to be doing the same thing, but `mc-fog-enclave-connection`
is generic over a protobuf client object that could be ledger key image, merkle proof, or view
(or consensus user tx conceivably). OTOH the mc-connection objects
have been used in production and support more features like the credentials
provider, and before now, doing retries with the retry crate for
connection / attestation issues.

This does not attempt to resolve that issue for now, but it does port
forward the retry aspect of this from "sync_connection".

I decided not to do it in exactly the same way sync_connection, instead
I decided that `enclave_connection` should have a helper function that
converts its result into a `retry::OperationResult`, so that the caller
can use the `retry` call, and we don't need to pass the retry Iterator
into this call. To me this seems simpler and cleaner than what is done
in `mc-connection`, and does not involve any rust macros to get it to work.

This will hopefully make unit tests and CD more stable, as well as making
the test client more stable.

### DIFF
--- a/fog/ledger/connection/src/key_image.rs
+++ b/fog/ledger/connection/src/key_image.rs
@@ -13,6 +13,7 @@ use mc_fog_types::ledger::{
 use mc_fog_uri::FogLedgerUri;
 use mc_transaction_core::{ring_signature::KeyImage, BlockIndex};
 use mc_util_grpc::ConnectionUriGrpcioChannel;
+use retry::{delay::Fixed, retry};
 use std::sync::Arc;
 
 /// An attested connection to the Fog Key Image service.
@@ -54,8 +55,9 @@ impl FogKeyImageGrpcClient {
                 .collect(),
         };
 
-        let response: CheckKeyImagesResponse =
-            self.conn.encrypted_enclave_request(&request, &[])?;
+        let response: CheckKeyImagesResponse = retry(Fixed::from_millis(100).take(5), || {
+            self.conn.retriable_encrypted_enclave_request(&request, &[])
+        })?;
 
         Ok(response)
     }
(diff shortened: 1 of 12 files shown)


fog/ledger/connection/src/key_image.rs
```diff
@@ -54,8 +55,9 @@ impl FogKeyImageGrpcClient {
                 .collect(),
         };
 
-        let response: CheckKeyImagesResponse =
-            self.conn.encrypted_enclave_request(&request, &[])?;
+        let response: CheckKeyImagesResponse = retry(Fixed::from_millis(100).take(5), || {
```
samdealy, line 58: This value, 100, is used in the other clients as well. Consider making it a shared constant with a descriptive name that's defined in the retry crate.

samdealy (approved):

eranrund (approved):

fog/ledger/connection/src/key_image.rs
garbageslam: okay, we can make the duration iterator a parameter and it can be in the sample paykit or something

garbageslam: I'm going to fix this in follow up PR because i want to merge this, it will help with testing deployments and make fog test client work better in prod environments, and it's a lot of extra work if I do it right now, and that work is likely to conflict with other PRs that are inflight like yours I think.

I think this is important to do though and I have created a ticket to track it. Thanks for the suggestion

PR #1169, opened 2021-11-12
Prevent ingress keys from using start block higher than highest known

samdealy:
### Motivation

As described in this [ticket](https://app.asana.com/0/1200353042931237/1201342064475415/f), "when ingest creates a new ingress key, it can specify any start_block value, and it currently does this using ledger_db.num_blocks(). But if its ledger_db is slow (because of a race), then this value may be very low, much older than the actual current number of blocks."

### In this PR
* At the DB level, enforce that a new ingress key can't specify a start block that's higher than the highest known block index

Fixes requirement 1 of this [ticket](https://app.asana.com/0/1200353042931237/1201342064475415/f). 

### Future Work
* Investigate if fog view should retrieve the highest known block count from the DB

### DIFF
--- a/fog/ingest/server/src/controller.rs
+++ b/fog/ingest/server/src/controller.rs
@@ -716,11 +716,11 @@ where
                         .map(|x| x + 1)
                         .unwrap_or(0),
                 );
-                if !self.recovery_db.new_ingress_key(&our_pubkey, start_block)? {
-                    return Err(PeerBackupError::CreatingNewIngressKey.into());
-                };
 
-                start_block
+                match self.recovery_db.new_ingress_key(&our_pubkey, start_block) {
+                    Ok(accepted_start_block) => accepted_start_block,
+                    Err(_err) => return Err(PeerBackupError::CreatingNewIngressKey.into()),
+                }
             };
         // This unwrap is okay because we are idle right now.
         state
--- a/fog/sql_recovery_db/src/lib.rs
+++ b/fog/sql_recovery_db/src/lib.rs
@@ -178,7 +178,7 @@ impl SqlRecoveryDb {
         }
     }
 
-    fn get_highest_known_block_index_impl(&self, conn: &PgConnection) -> Result<Option<u64>, Error> {
+    fn get_highest_known_block_index_impl(conn: &PgConnection) -> Result<Option<u64>, Error> {
         Ok(schema::ingested_blocks::dsl::ingested_blocks
             .select(diesel::dsl::max(schema::ingested_blocks::dsl::block_number))
             .first::<Option<i64>>(conn)?
@@ -205,7 +205,7 @@ impl RecoveryDb for SqlRecoveryDb {
     ) -> Result<u64, Error> {
         let conn = self.pool.get()?;
         conn.build_transaction().read_write().run(|| -> Result<u64, Error> {
-            let highest_known_block_count: u64 = match self.get_highest_known_block_index_impl(&conn) {
+            let highest_known_block_count: u64 = match SqlRecoveryDb::get_highest_known_block_index_impl(&conn) {
                 Ok(block_index) => block_index.unwrap_or(0) + 1,
                 Err(_) => 0,
             };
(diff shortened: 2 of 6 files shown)


fog/ingest/server/src/controller.rs
```diff
@@ -716,11 +716,11 @@ where
...
                         .unwrap_or(0),
                 );
-                if !self.recovery_db.new_ingress_key(&our_pubkey, start_block)? {
-                    return Err(PeerBackupError::CreatingNewIngressKey.into());
-                };
 
-                start_block
+                match self.recovery_db.new_ingress_key(&our_pubkey, start_block) {
+                    Ok(accepted_start_block) => accepted_start_block,
+                    Err(err) => return Err(PeerBackupError::CreatingNewIngressKey.into()),
```
samdealy, line 722: Note that this line slightly modifies the behavior of how we reported errors from the `new_ingress_key` method. Previously, we used the `?` operator to pass along the two errors ([here](https://github.com/mobilecoinfoundation/mobilecoin/blob/25aad6e0ffef1f00e7fde8a161c7c7e72c4a9b74/fog/sql_recovery_db/src/lib.rs#L198) and [here](https://github.com/mobilecoinfoundation/mobilecoin/blob/25aad6e0ffef1f00e7fde8a161c7c7e72c4a9b74/fog/sql_recovery_db/src/lib.rs#L210)) reported by the method, both of which are r2d2 Errors. It would then pass a `PeerBackupError::CreatingNewIngressKey` when the key wasn't successfully inserted.

Instead of using a boolean to signify a failed key insertion, the code now emits a newly-defined `IngressKeyUnsuccessfulInsert` defined the sql_db crate. This method then takes any error emitted by the `new_ingress_key` method and returns `PeerBackupError::CreatingNewIngressKey`. So it will no longer pass along those r2d2 Errors. 

I investigated the usage of these errors, and couldn't see any specific use cases for the error type emitted here. Therefore, I think it's safe to make this small change in error handling, but just wanted to call it out here in case I'm missing something.

fog/sql_recovery_db/src/lib.rs
```diff
@@ -194,7 +195,13 @@ impl RecoveryDb for SqlRecoveryDb {
...
         key: &CompressedRistrettoPublic,
         start_block: u64,
-    ) -> Result<bool, Error> {
+    ) -> Result<u64, Error> {
+        let highest_known_block_count: u64 = match self.get_highest_known_block_index() {
+            Ok(block_count) => block_count.unwrap_or(0),
+            Err(_) => 0,
+        };
+
+        let accepted_start_block = min(start_block, highest_known_block_count);
```
garbageslam, line 204: shouldn't this be a max?

fog/sql_recovery_db/src/lib.rs
```diff
@@ -194,7 +195,13 @@ impl RecoveryDb for SqlRecoveryDb {
         &self,
         key: &CompressedRistrettoPublic,
         start_block: u64,
-    ) -> Result<bool, Error> {
+    ) -> Result<u64, Error> {
+        let highest_known_block_count: u64 = match self.get_highest_known_block_index() {
```
garbageslam, line 199: this should be transactional with the insert operation i think

fog/sql_recovery_db/src/lib.rs
```diff
@@ -209,7 +216,14 @@ impl RecoveryDb for SqlRecoveryDb {
             .on_conflict_do_nothing()
             .execute(&conn)?;
 
-        Ok(inserted_row_count > 0)
+        if inserted_row_count > 0 {
+            Ok(accepted_start_block)
+        } else {
+            Err(Error::IngressKeyUnsuccessfulInsert(format!(
+                "Unable to insert ingest key: {:?}",
```
garbageslam, line 223:
```suggestion
                "Unable to insert ingress key: {:?}",
```

fog/sql_recovery_db/src/lib.rs
samdealy: As discussed offline, yes this should be max.

fog/ingest/server/src/controller.rs
garbageslam: it looks good to me, thank you

fog/sql_recovery_db/src/lib.rs
samdealy: Done.

garbageslam (approved):

eranrund (approved):

garbageslam:
LGTM! thank you!

PR #1172, opened 2021-11-15
Add grpc gateway

garbageslam:
This moves the go-grpc-gateway, which was previously in internal repo only, into the mobilecoin repo. This will simplify releases in the future.


eranrund (approved):
Nice, good job on getting this properly CIed/tested.

voloshyn (approved):
LGTM

Issue #1160, opened 2021-11-09
Make a test that would have caught sample-paykit subaddress index issues


garbageslam:
In https://github.com/mobilecoinfoundation/mobilecoin/pull/1145 we fixed a bug where, the sample paykit assumed that all TxOut's were owned by the default subaddress for purposes of spending those TxOut's.

This bug existed for months because it would only surface when the sample paykit tried to spend a change TxOut from an earlier transaction, but none of our testing actually created this situation.

Proposal:
* Make a test in circleci that runs the fog-local-network script as described here (https://github.com/mobilecoinfoundation/mobilecoin/tree/master/tools/fog-local-network)
* Only bootstrap a small ledger with e.g. 1 account.
* Generate only e.g. 4 fog accounts.
* When running mc-fog-distribution should use ` --num-tx-to-send 1`
* Run test client with the 4 fog accounts, because each fog account only has one initial TxOut, it should be forced to spend change the second time.

Advantages:
* This puts the fog-local-network script under test
* It puts something similar to what we do in CD also in CI
* It would have caught the issue

Disadvantages:
* It's not a unit test
* It's harder and slower to run locally than a unit test

I think unfortunately would be very difficult to write a unit test that would have caught this issue. Part of the reason for the issue is that we started making the sample paykit handle the change subaddress, but we didn't realize at the time of writing the code TxOut's have to be spent differently if they are owned by the change subaddress. Without realizing that in advance we would not have realized that we need to validate built transactions when the change subaddress is involved. We could write after-the-fact unit tests but the more important scenario to prevent is that some future unforeseen change (perhaps involving memos) causes change TxOut's not to be spent correctly, and nothing in our integration testing is actually exercising that functionality before the test is declared successful. So I would rather see an integration test that we know is covering this scenario.

garbageslam:
I think Sam fixed this and it's now part of CI, thank you!

PR #1173, opened 2021-11-16
add fog local network to CI

garbageslam:
This adds fog local network to CI, with fog distro using very
few transactions. The point is to get test coverage that ensures
that test client is able to spend change outputs.

This fixes mobilecoin issue #1160
It also puts the fog local network script under test

### DIFF
--- a/consensus/api/proto/consensus_common.proto
+++ b/consensus/api/proto/consensus_common.proto
@@ -77,6 +77,7 @@ enum ProposeTxResult {
     UnsortedInputs = 39;
     MissingMemo = 40;
     MemosNotAllowed = 41;
+    UnsortedOutputs = 42;
 }
 
 /// Response from TxPropose RPC call.
--- a/consensus/api/src/conversions.rs
+++ b/consensus/api/src/conversions.rs
@@ -46,6 +46,7 @@ impl From<Error> for ProposeTxResult {
             Error::TxFeeError => Self::TxFeeError,
             Error::KeyError => Self::KeyError,
             Error::UnsortedInputs => Self::UnsortedInputs,
+            Error::UnsortedOutputs => Self::UnsortedOutputs,
             Error::MissingMemo => Self::MissingMemo,
             Error::MemosNotAllowed => Self::MemosNotAllowed,
         }
@@ -91,6 +92,7 @@ impl TryInto<Error> for ProposeTxResult {
             Self::TxFeeError => Ok(Error::TxFeeError),
             Self::KeyError => Ok(Error::KeyError),
             Self::UnsortedInputs => Ok(Error::UnsortedInputs),
+            Self::UnsortedOutputs => Ok(Error::UnsortedOutputs),
             Self::MissingMemo => Ok(Error::MissingMemo),
             Self::MemosNotAllowed => Ok(Error::MemosNotAllowed),
         }
--- a/transaction/core/src/validation/error.rs
+++ b/transaction/core/src/validation/error.rs
@@ -78,6 +78,9 @@ pub enum TransactionValidationError {
      */
     UnsortedInputs,
 
+    /// Outputs must be sorted by public_key, ascending.
+    UnsortedOutputs,
+
     /// Key Images must be sorted.
     UnsortedKeyImages,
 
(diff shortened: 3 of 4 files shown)


garbageslam:
There seems to be a reproducible bug in fog test client here, it seems like when the services are all local and everything goes fast enough, the balance checking result is wrong. Still investigating if it's server side or client side but I suspect there is an off by one in test client somewhere

tools/fog-local-network/README.md
```diff
@@ -23,7 +23,7 @@ In order to use it, the following steps are necessary.
 
 2) Generating a set of sample keys:
     ```
-    cargo run -p mc-util-keyfile --bin sample-keys --release --manifest-path ../Cargo.toml -- --num 1000 --fog-report-url 'insecure-fog://localhost:6200'
+    cargo run -p mc-util-keyfile --bin sample-keys --release -- --num 1000
```
samdealy, line 26: Thank you for modifying this, it works!

samdealy (approved):

tools/fog-local-network/README.md
```diff
@@ -33,7 +33,7 @@ In order to use it, the following steps are necessary.
 
 3) Bootstrap a ledger:
     ```
-    cargo run --manifest-path ../util/generate-sample-ledger/Cargo.toml --release -- --txs 100
+    cargo run mc-util-generate-sample-ledger --release -- --txs 100
```
samdealy, line 36: Running this command as-is fails with this error: `cargo run could not determine which binary to run`.

If you change it to `cargo run -p ...` then it works.

garbageslam: oh yeah that's a mistake thank you

jcape:
Next step here is to figure out the issue with local-network (@wjuan-mob has been able to repro in CI, possibly locally).

.circleci/config.yml
```diff
@@ -419,6 +419,72 @@ commands:
...
+            openssl genrsa -out Enclave_private.pem -3 3072
+            export SGX_MODE=SW
+            export IAS_MODE=DEV
+            export CONSENSUS_ENCLAVE_PRIVKEY=$(pwd)/Enclave_private.pem
+            export INGEST_ENCLAVE_PRIVKEY=$(pwd)/Enclave_private.pem
+            export LEDGER_ENCLAVE_PRIVKEY=$(pwd)/Enclave_private.pem
+            export VIEW_ENCLAVE_PRIVKEY=$(pwd)/Enclave_private.pem
+            export MC_LOG=debug
+
+            cargo build -p mc-util-keyfile -p mc-util-generate-sample-ledger -p mc-consensus-service -p mc-ledger-distribution -p mc-admin-http-gateway -p mc-util-grpc-admin-tool -p mc-mobilecoind -p mc-crypto-x509-test-vectors -p mc-fog-distribution -p mc-fog-test-client -p mc-fog-ingest-server -p mc-fog-ingest-client -p mc-fog-view-server -p mc-fog-report-server -p mc-fog-ledger-server -p mc-fog-sql-recovery-db --release
```
remoun, line 442: Optional/Equivalent: One line per package, and sort them.

tools/local-network/local_network.py
```diff
@@ -412,7 +412,7 @@ def build_binaries(self):
             )
 
         subprocess.run(
-            f'cd {PROJECT_DIR} && CONSENSUS_ENCLAVE_PRIVKEY="{enclave_pem}" cargo build -p mc-consensus-service -p mc-ledger-distribution -p mc-admin-http-gateway -p mc-util-grpc-admin-tool -p mc-slam -p mc-crypto-x509-test-vectors {CARGO_FLAGS}',
+            f'cd {PROJECT_DIR} && CONSENSUS_ENCLAVE_PRIVKEY="{enclave_pem}" cargo build -p mc-consensus-service -p mc-ledger-distribution -p mc-admin-http-gateway -p mc-util-grpc-admin-tool -p mc-mobilecoind -p mc-crypto-x509-test-vectors {CARGO_FLAGS}',
```
remoun, line 415: This doesn't look right -- `mc-mobilecoind` is built slightly differently in the next step

.circleci/config.yml
samdealy: Ack

PR #1183, opened 2021-11-19
Use slip10 accounts instead of legacy root entropy in internal testing

garbageslam:
This migrates the `mc-util-keyfile` crate to write private key json
files containing slip10 root entropies instead of the legacy root entropy.
The legacy root entropies are deprecated but they must still exist
in some capacity because there are accounts with money on them IRL.

This will simplify the effort to get uniform integration tests
because many wallet products may never implement root entropies.

The bootstrap program does not need to be touched because it only
reads the public key files.

The keygen program is touched, fog-distro is touched, the sample
paykit is touched, the test-client is touched, slam is touched,
but most of the changes are very localized.

The main thing that would make this not work is if someone somewhere
is storing keyfiles in a persistent way, however this is unlikely
and whatever system that is likely needs to be reworked anyways.

This also changes the conformance tests remote wallet API a bit, but
that should also be fine.

---

This WILL eventually impact SDKs because in dev nets we will now be funding the previous "root entropies" but interpreting
those 32 bytes as a slip10 key. If they continue to interpret those bytes as root entropies, then going forward they will have zero balances during testing, for any deployment after this commit.

### DIFF
--- a/mobilecoind/strategies/accounts.py
+++ b/mobilecoind/strategies/accounts.py
@@ -38,7 +38,7 @@ def register_account(key_data, stub) -> AccountData:
     # Generate an account key from this root entropy
     resp = stub.GetAccountKeyFromRootEntropy(
         mobilecoind_api_pb2.GetAccountKeyFromRootEntropyRequest(
-            root_entropy=bytes(key_data['root_entropy'])))
+            slip10_key=bytes.fromhex(key_data['slip10_key'])))
     account_key = resp.account_key
 
     # Add this account to the wallet
--- a/mobilecoind/strategies/accounts.py
+++ b/mobilecoind/strategies/accounts.py
@@ -37,7 +37,7 @@ def connect(host, port):
 def register_account(key_data, stub) -> AccountData:
     # Generate an account key from this root entropy
     resp = stub.GetAccountKeyFromRootEntropy(
-        mobilecoind_api_pb2.GetAccountKeyFromRootEntropyRequest(
+        mobilecoind_api_pb2.GetAccountKeyFromSlip10Request(
             slip10_key=bytes.fromhex(key_data['slip10_key'])))
     account_key = resp.account_key
 
--- a/tools/fog-local-network/fog_conformance_tests.py
+++ b/tools/fog-local-network/fog_conformance_tests.py
@@ -154,7 +154,7 @@ class RemoteWallet:
         with grpc.insecure_channel(self.remote_wallet_host_port) as channel:
             stub = remote_wallet_pb2_grpc.RemoteWalletApiStub(channel)
             response = self._retrying_grpc_request(stub.FreshBalanceCheck, remote_wallet_pb2.FreshBalanceCheckRequest(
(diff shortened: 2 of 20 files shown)


tools/fog-local-network/fog_conformance_tests.py
```diff
@@ -154,7 +154,7 @@ def start(self):
         with grpc.insecure_channel(self.remote_wallet_host_port) as channel:
             stub = remote_wallet_pb2_grpc.RemoteWalletApiStub(channel)
             response = self._retrying_grpc_request(stub.FreshBalanceCheck, remote_wallet_pb2.FreshBalanceCheckRequest(
-                root_entropy=bytes(key['root_entropy']),
+                slip10_key=bytes.fromhex(key['slip10_key']),
```
garbageslam, line 157: this is because we are encoding the slip10 key as hex in the slip10 key json files

mobilecoind/strategies/accounts.py
```diff
@@ -37,8 +37,8 @@ def connect(host, port):
 def register_account(key_data, stub) -> AccountData:
     # Generate an account key from this root entropy
     resp = stub.GetAccountKeyFromRootEntropy(
-        mobilecoind_api_pb2.GetAccountKeyFromRootEntropyRequest(
-            root_entropy=bytes(key_data['root_entropy'])))
+        mobilecoind_api_pb2.GetAccountKeyFromSlip10Request(
+            slip10_key=bytes.fromhex(key_data['slip10_key'])))
```
garbageslam, line 41: this change is needed for Sara's python strategies test to keep working in CD

util/keyfile/src/bin/main.rs
```diff
@@ -4,10 +4,12 @@
...
 //! stdout
 
-use mc_account_keys::{AccountKey, RootIdentity};
+use mc_account_keys::AccountKey;
+use mc_util_keyfile::Slip10IdentityJson;
+use std::convert::TryFrom;
 
 fn main() {
-    let root_id: RootIdentity = {
+    let root_id: Slip10IdentityJson = {
```
wjuan-mob, line 12: Should this be slip10_id

wjuan-mob (approved):
Closed mine because it seemed like this PR was mostly good and there were only small nits to fix and the merge conflict was minimal.

jcape (commented):
Why aren't we using mnemonics for this?

garbageslam:
> Why aren't we using mnemonics for this: #780

Because it's not a requirement to make progress towards improved integration testing

> Alex Voloshyn: Root entropies are deprecated, we should not keep using those in any new developments. Let's address the tech debt instead of making it larger. I propose to use either bip39 entropies or a list of mnemonic phrases.

jcape:
The current state of this PR is that it needs to be reworked to integrate the 24-word nmenomics/entropy from #780 instead of the SLIP10 derivation of mnemonics (in order to prevent future breakage if we change how 24 words turns into keys).

PR #1191, opened 2021-12-01
remove unused logging macros

garbageslam:
these macros, `log_assert`, `log_assert_eq`, etc. are not used
anywhere in the codebase.

these macros are harmful because the panic handler already logs
the panic to the logger, so using these macros means that you
actually log the assert failure twice. I believe that this may
be why no one is actually using this.

Since this is dead code we should delete it.


eranrund (approved):

jcape (approved):
LGTM, I didn't know these even existed.

PR #1226, opened 2021-12-08
try to fix build info crate

garbageslam:
the build info crate records build info, like environment variables
that configure the build, so that targets can log them and such

however, i notice that oftentimes the info can be wrong, especially
in local testing, e.g. SGX_MODE or IAS_MODE not matching the actual
build.

it turns out the build.rs did not emit cargo's rerun-if-changed
directives, so cargo doesn't regenerate the file if one of these
variables changes

this commit may reduce confusion


eranrund (approved):

jcape (approved):
LGTM

PR #1229, opened 2021-12-09
Move Verifier to its Own Crate

jcape:
### Motivation

The verifier code in `mc-attest-core` is currently a fairly complex set of structures which will each verify a particular field in an Attestation Verification Report provided by IAS. In it's current implementation, all verifiers (which are nested) return `bool`, and so are unable to log which specific tests caused an attestation failure, and instead relies on the caller printing the verifier being used and the report being examined, and manually comparing the differences.

In order to implement logging capable of being understood by those without a familiarity with the AVR structure, we need to change every verifier implementation to return `Option<Error>` (for some error type), add some error-collation types, and implement custom `Display` outputs. This will allow us to print a list of everything wrong with an AVR, which allows complete debugging of all possible mismatches in one shot.

Given the complexity of the current verifier (1200 lines of code, 1400 lines of unit tests), it makes sense to start breaking this up to ease maintenance tasks. Additionally, this code is the place where `mc-attest-core` uses `mbedtls`.

Therefore, the tasks to clean this up are:

1. Move the existing `verifier.rs` to it's own crate (this will remove the dependency on `mbedtls` from `mc-attest-core`, most notably allowing `VerificationReport` to be used directly in more places).
2. Split the `Verifier` code into modules: signature verification, IAS details, quote details, report body details, and MRSIGNER/MRENCLAVE details.
3. Implement the `Option<Error>` and `Display` as described above
4. Make the signature verification useful with multiple backends. This will allow us to remove `mbedtls` from non-enclave environments (e.g. Windows, thick clients, etc.)

### In this PR
* Move the existing `verifier.rs` into `mc-attest-verifier crate`

### Future Work
* Split the Verifier code into modules
* Implement `Option<Error>` returns and a better `Display` hierarchy.
* Backend-independent signature verification / only use mbedtls in the enclaves


garbageslam:
LGTM, looks like this is just code movement and I didn't see any problems.

If at all possible let's please do
* Implement Option<Error> returns and a better Display hierarchy.
Before
* Split the Verifier code into modules

since we would like to get the former in 1.2 release, which is soon.
This and the open telemetry are the last things we are holding the release for

garbageslam (approved):

jcape:
There are 25 verifier structures working on four different input data structures stuffed into one file with 2500 lines of code and tests, it needs to get split up in order to make maintenance (like changing the API and adding error-bundle structures which implement `Display`) easier.

eranrund (approved):

PR #1242, opened 2021-12-14
Use less severe log levels for server levels that aren't usually actionable

garbageslam:
Currently, both Error and Critical get forwarded to sentry and cause
sentry alerts. However, most attestation errors are logged at Error
level, as well as other client-triggered "errors" that usually
represent the normal operation of the system.

This commit makes attestation errors and "invalid arg" errors (that
don't necessarily indicate a server problem) logged at debug level.
It makes general "enclave errors" which get mapped to permission denied
normally, logged as warn level only. We should revisit this in the future
and try to break down those error levels in a more granular way.

This commit is intended for 1.1.2 maintenance release and is based off
the v1.1.0+ tag.


jgreat (approved):
Thanks, this will help clean up monitoring.

jcape (approved):

PR #1258, opened 2021-12-16
Feature/verifiers to modules

jcape:
### Motivation

The verifier code in `mc-attest-core` is currently a fairly complex set of structures which will each verify a particular field in an Attestation Verification Report provided by IAS. In it's current implementation, all verifiers (which are nested) return `bool`, and so are unable to log which specific tests caused an attestation failure, and instead relies on the caller printing the verifier being used and the report being examined, and manually comparing the differences.

In order to implement logging capable of being understood by those without a familiarity with the AVR structure, we need to change every verifier implementation to return `Option<Error>` (for some error type), add some error-collation types, and implement custom `Display` outputs. This will allow us to print a list of everything wrong with an AVR, which allows complete debugging of all possible mismatches in one shot.

Given the complexity of the current verifier (1200 lines of code, 1400 lines of unit tests), it makes sense to start breaking this up to ease maintenance tasks. Additionally, this code is the place where `mc-attest-core` uses `mbedtls`.

Therefore, the tasks to clean this up are:

1. ~Move the existing `verifier.rs` to it's own crate (this will remove the dependency on `mbedtls` from `mc-attest-core`, most notably allowing `VerificationReport` to be used directly in more places).~
2. Split the `Verifier` code into modules: signature verification, IAS details, quote details, report body details, and MRSIGNER/MRENCLAVE details.
3. Implement the `Option<Error>` and `Display` as described above
4. Make the signature verification useful with multiple backends. This will allow us to remove `mbedtls` from non-enclave environments (e.g. Windows, thick clients, etc.)

### In this PR
* Split the Verifier code into modules

### Future Work
* Implement `Option<Error>` returns and a better `Display` hierarchy.
* Backend-independent signature verification / only use mbedtls in the enclaves


attest/verifier/src/ias.rs
```diff
@@ -0,0 +1,244 @@
...
+                let mut newcert = src_cert.clone();
+                newcert
+                    .public_key_mut()
+                    .verify(HashType::Sha256, hash.as_slice(), report.sig.as_ref())
+                    .and(Ok(newcert))
+                    .ok()
+            })
+            // Then construct a set of chains, one for each signer certificate
+            .filter_map(|cert| {
+                let mut signer_chain = vec![cert];
```
jcape, line 147: This construct has been changed, in the original version, it was `let signer_chain: Vec<MbedtlsBox<Certificate>> = vec![cert];`, which is strange because cert here is of type `&MbedtlsBox<Certificate>`.

This should make the type `Vec<&MbedtlsBox<Certificate>>`, which feels like it would be closer to what we are looking for, but I'm very suspicious about all of this code and would appreciate a second set of eyes on it.

attest/verifier/src/ias.rs
```diff
@@ -0,0 +1,244 @@
...
+                || self
+                    .or_verifiers
+                    .iter()
+                    .any(|verifier| verifier.verify(&report_data)))
+        {
+            Ok(report_data)
+        } else {
+            Err(Error::Verification(report_data))
+        }
+    }
```
polsar88, line 243: This function does seem a bit too long. Also, the nested loops are too complex. Maybe break up the code a bit?

attest/verifier/src/quote.rs
```diff
@@ -0,0 +1,394 @@
...
+
+        assert!(verifier.verify(&quote));
+    }
+
+    /// When the report does not contain the EPID group ID we're expecting
+    #[test]
+    fn epid_group_id_fail() {
+        let quote =
+            Quote::from_base64(BASE64_QUOTE).expect("Could not parse quote from base64 file");
+        let epid_group_id = [0u8; 4];
```
polsar88, line 249: Can we turn these magic numbers into constants with descriptive names?

jcape: These functions will be updated in part 3, where verifiers return `Option<Error>` instead of `bool`, I can take a look at that then, but I'd like to keep the amount of code changes in this PR to a minimum.

attest/verifier/src/ias.rs
jcape: The signature chain validation will be moved to it's own "signature verifier" type in phase 4 of this, but the signature verification is complex because of the reasons discussed in the comment.

The chained iterator scheme is also meant to reduce the number of times we're doing a heap allocation to the bare minimum, and `clippy` (which is a hard CI requirement) complains if you're creating intermediate collections just to iterate over them again, which would be the obvious way to make this more readable.

attest/verifier/src/ias.rs
```diff
@@ -0,0 +1,244 @@
...
+                    }
+                }
+                None
+            })
+            .ok_or(Error::BadSignature)?;
+
+        let report_data = VerificationReportData::try_from(report)?;
+
+        if (self.and_verifiers.is_empty()
+            || self
```
garbageslam, line 229: I think this part that checks `is_empty` is unnecessary:
`.iter().all(...)` returns true if the set is empty, so we could delete the `is_empty` and the `||` part

https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.all

attest/verifier/src/ias.rs
```diff
@@ -0,0 +1,244 @@
...
+                .and_verifiers
+                .iter()
+                .all(|verifier| verifier.verify(&report_data)))
+            && (self.or_verifiers.is_empty()
```
garbageslam, line 233: here I think the special-casing is needed if you want for having no "or_verifiers" mean that no passing or_verifiers are required, because `any()` of an empty iterator is false

garbageslam (approved):
LGTM

attest/verifier/src/ias.rs
```diff
@@ -0,0 +1,244 @@
...
+                    .and(Ok(newcert))
+                    .ok()
+            })
+            // Then construct a set of chains, one for each signer certificate
+            .filter_map(|cert| {
+                let mut signer_chain = vec![cert];
+                'outer: loop {
+                    // Exclude any signing chains greater than our max depth
+                    if signer_chain.len() > MAX_CHAIN_DEPTH {
+                        return None;
```
garbageslam, line 151: Is it better to return `None` here or `Some(signer_chain)`? Even if the chain is too long if one of the intermediate parts is a trust anchor we could still accept it I think?

Don't really want to argue for changes in working code, just a question that occurred to me

attest/verifier/src/ias.rs
jcape: Fair point. I'll leave it for now because Part 3 will end up replacing the `all()` usage here in favor of `filter_map().collect()`.

attest/verifier/src/ias.rs
jcape: Yeah, same story as the `all()` usage above: this will need to get replaced with a `filter_map().collect()` and a check for the length of an `errors: Vec` against the length of `or_verifiers`.

attest/verifier/src/ias.rs
jcape: Sounds good, I'll leave this as-is for now to keep the changes in this PR to a minimum, but I'll throw it on the list for the next PR in this series.

attest/verifier/src/ias.rs
wjuan-mob: I'm not sure this is right. I think that the previous filter map makes cert a MbedtlsBox<Certificate> rather than an &MbedtlsBox<Certificate> because it is cloned, which should return the object rather than a reference? maybe? Sorry if I'm misunderstanding.

wjuan-mob (approved):
Nit: Copyright 2022

jcape:
> Nit: Copyright 2022

Thanks, will fix in PR3

PR #1259, opened 2021-12-17
minor quality of life improvements in fog

garbageslam:
* ingest server says "Checking up on peer..." instead of
  "peer checkup period exceeded" because the latter sounds like
  an error
* fog ingest client pretty prints its json results instead of
  dense json, because this is usually what an operator wants, and
  currently people are piping things to jq, but then you also have
  to filter out the log statements that the client emits

### DIFF
--- a/fog/ingest/client/src/main.rs
+++ b/fog/ingest/client/src/main.rs
@@ -10,13 +10,14 @@ use mc_fog_ingest_client::{
     ClientResult, FogIngestGrpcClient,
 };
 use mc_fog_uri::FogIngestUri;
-use serde_json::json;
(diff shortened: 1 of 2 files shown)


eranrund (approved):
LGTM but I do I wonder how the JSON is being used - is pretty printing going to affect it?
Perhaps @jgreat or @joekottke have an opinion on this?

garbageslam:
based on how the deployment went today, i would say that usually it looks like `fog_ingest_client --uri ... activate 2>/dev/null | jq . r`, to try to make the output easier to read. but after this the `jq` and the `2>/dev/null` are probably unnecessary

wjuan-mob (approved):

PR #1272, opened 2022-01-03
add another option to test client config

garbageslam:
this allows more control over what happens when a transaction
hangs and the test client gets blocked, in the continuous mode

### DIFF
--- a/fog/test-client/src/config.rs
+++ b/fog/test-client/src/config.rs
@@ -35,6 +35,20 @@ pub struct TestClientConfig {
     #[structopt(long, env)]
     pub continuous: bool,
 
+    /// If not set, the test is terminated if a deadline is passed. We fail
+    /// immediately, then start counting down for the next trial.
+    ///
+    /// If set, then we continue waiting after the deadline until the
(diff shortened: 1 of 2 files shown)


fog/test-client/src/config.rs
```diff
@@ -35,6 +35,20 @@ pub struct TestClientConfig {
     #[structopt(long, env)]
```
wjuan-mob, line 35: If this is required with the measure_after_deadline parameter, should we use structopt required_if? Also, I think the comment about failing fast should note that this is set by measure_after_deadline.

garbageslam: I don't think its actually required in the code.

Previously I tried tying these fields together more tightly and that turned out not to be what we wanted, so I'd rather have a new option that's completely separate from the continuous option.

It's not clear we're going to use `fail_fast_on_deadline: false` in any tests or deployment anymore, so it might eventually go away

wjuan-mob (approved):

jgreat (approved):
LGTM

PR #1285, opened 2022-01-06
make the fog sample paykit avoid making massive requests for key image checks

garbageslam:
this divides key image requests with more than 100 key images into
several requests, to avoid hitting any grpc maximums etc.

### DIFF
--- a/fog/sample-paykit/src/cached_tx_data/mod.rs
+++ b/fog/sample-paykit/src/cached_tx_data/mod.rs
@@ -658,7 +658,7 @@ impl CachedTxData {
         // Split up key images queries into several requests if there are a lot of key
         // images
         for key_images in key_images.chunks(MAX_KEY_IMAGES_PER_QUERY) {
-            match key_image_client.check_key_images(&key_images) {
+            match key_image_client.check_key_images(key_images) {
                 Ok(response) => {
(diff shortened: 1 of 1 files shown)


wjuan-mob (approved):

fog/sample-paykit/src/cached_tx_data/mod.rs
```diff
@@ -640,67 +655,77 @@ impl CachedTxData {
...
-                                match otxo.status {
-                                    KeyImageStatus::SpentAt(_) => panic!(
-                                        "We were not supposed to ask about already spent txo's"
-                                    ),
-                                    KeyImageStatus::NotSpent(not_spent_as_of) => {
-                                        if spent_at < u64::from(not_spent_as_of) {
-                                            log::error!(self.logger, "Inconsistency from server -- we earlier learned that a Txo was not spent as of {}, but now we have learned that it was spent at {}", not_spent_as_of, spent_at);
+        // Split up key images queries into several requests if there are a lot of key
+        // images
+        for key_images in key_images.chunks(MAX_KEY_IMAGES_PER_QUERY) {
```
garbageslam, line 660: all that happened was that i added this for loop and indented this block

jgreat (approved):
Does the trick with the canary instances.

PR #1298, opened 2022-01-08
Implement core Fog Overseer functionality

samdealy:
### Motivation
This PR contains the core functionality of the Fog Overseer service. The service monitors the active Fog Ingest cluster and makes sure that there is at least one active node. If there is no active node, then it performs the automatic failover logic. This logic is described in detail in the worker.rs documentation. 

Note that this PR contains the core logic: the monitoring logic is in worker.rs and the rest of the files basically just expose that functionality to clients via a Rocket HTTPS server. The tests test 5 different scenarios in an end to end fashion. Note that there are still some key features that need to be implemented before we can deploy this service. Namely, we'll need to implement the following items: 

1. Sentry alerting: in some cases, we need to send alerts to humans when Overseer encounters certain (unlikely) Fog Ingest cluster states.
2. Prometheus metrics: as noted in the design doc, we want to capture Fog ingest cluster state metrics like number of ingress keys, number of responsive Fog ingest nodes, etc. 
3. Get info api: in addition to storing Fog ingest cluster state in Prometheus, we want to expose a get info api that allows clients to get the current Fog Ingest cluster info in real time.
4. Config refresh: currently, we pass the list of Fog Ingest node uris to overseer through a CLI config. To allow for overseer to look at a new set of node urls during blue / green deployment without having to restart it, we'll move this list of uris to an environment variable that can be dynamically refreshed. This way, we can update the list of uris without having to restart overseer.

I figured that it would be best if I implemented these 4 features in future PRs, since this PR is already quite large. However, if you find that any of these features should be in this PR, I'm open to discussing. 

### In this PR
* Implements the core functionality of the Fog Overseer Service
* Creates 5 end to end tests that cover the core scenarios we expect Fog Overseer to help with.

### Future Work
* Implement the 4 features listed above.

### DIFF
--- a/fog/ingest/server/src/server.rs
+++ b/fog/ingest/server/src/server.rs
@@ -359,6 +359,19 @@ where
         self.get_ingest_summary().mode == IngestControllerMode::Active
     }
 
+    /// Ask if the server is idle
+    /// This is a convenience wrapper used in tests
+    pub fn is_idle(&self) -> bool {
+        self.get_ingest_summary().mode == IngestControllerMode::Idle
+    }
+
+    /// Set new keys.
+    /// This is used in tests when it would be simpler than making an RPC
+    /// client.
+    pub fn set_new_keys(&self) -> Result<IngestSummary, IngestServiceError> {
+        self.controller.new_keys()
+    }
+
     /// Tell the server to activate.
     /// This is used in tests when it would be simpler than making an RPC client
     pub fn activate(&self) -> Result<IngestSummary, IngestServiceError> {
--- a/fog/ingest/server/src/server.rs
+++ b/fog/ingest/server/src/server.rs
@@ -361,6 +361,7 @@ where
 
     /// Ask if the server is idle.
     /// This is a convenience wrapper used in tests.
+    #[cfg(test)]
     pub fn is_idle(&self) -> bool {
         self.get_ingest_summary().mode == IngestControllerMode::Idle
     }
@@ -368,6 +369,7 @@ where
     /// Set new keys.
     /// This is used in tests when it would be simpler than making an RPC
     /// client.
+    #[cfg(test)]
     pub fn set_new_keys(&self) -> Result<IngestSummary, IngestServiceError> {
         self.controller.new_keys()
     }
--- a/fog/ingest/server/src/server.rs
+++ b/fog/ingest/server/src/server.rs
@@ -359,8 +359,8 @@ where
         self.get_ingest_summary().mode == IngestControllerMode::Active
     }
 
-    /// Ask if the server is idle
-    /// This is a convenience wrapper used in tests
+    /// Ask if the server is idle.
+    /// This is a convenience wrapper used in tests.
     pub fn is_idle(&self) -> bool {
         self.get_ingest_summary().mode == IngestControllerMode::Idle
     }
--- a/fog/ingest/server/src/server.rs
+++ b/fog/ingest/server/src/server.rs
@@ -361,7 +361,6 @@ where
 
     /// Ask if the server is idle.
     /// This is a convenience wrapper used in tests.
-    #[cfg(test)]
     pub fn is_idle(&self) -> bool {
         self.get_ingest_summary().mode == IngestControllerMode::Idle
     }
@@ -369,7 +368,6 @@ where
     /// Set new keys.
     /// This is used in tests when it would be simpler than making an RPC
     /// client.
-    #[cfg(test)]
     pub fn set_new_keys(&self) -> Result<IngestSummary, IngestServiceError> {
         self.controller.new_keys()
     }
--- a/fog/overseer/server/Cargo.toml
+++ b/fog/overseer/server/Cargo.toml
@@ -5,12 +5,18 @@ authors = ["MobileCoin"]
 edition = "2018"
 license = "GPL-3.0"
 
+[[bin]]
+name = "start_overseer_server"
+path = "src/bin/main.rs"
+
 [dependencies]
 # third-party
 displaydoc = { version = "0.2", default-features = false }
 grpcio = "0.9.0"
 retry = "1.3"
 rocket = { version = "0.4.5", default-features = false }
+serde = "1"
+structopt = "0.3"
 
 # mc
 mc-api = { path = "../../../api"}
--- a/fog/overseer/server/Cargo.toml
+++ b/fog/overseer/server/Cargo.toml
@@ -1,0 +1,44 @@
+[package]
+name = "mc-fog-overseer-server"
+version = "1.2.0-pre0"
+authors = ["MobileCoin"]
+edition = "2018"
+license = "GPL-3.0"
+
+[dependencies]
+# third-party
+displaydoc = { version = "0.2", default-features = false }
+grpcio = "0.9.0"
+retry = "1.3"
+rocket = { version = "0.4.5", default-features = false }
+
+# mc
+mc-api = { path = "../../../api"}
+mc-common = { path = "../../../common", features = ["loggers"] }
+mc-crypto-keys = { path = "../../../crypto/keys"}
+mc-transaction-core = { path = "../../../transaction/core" }
+
+# fog
+mc-fog-api = { path = "../../api" }
+mc-fog-ingest-client = { path = "../../ingest/client" }
+mc-fog-recovery-db-iface = { path = "../../recovery_db_iface" }
+mc-fog-sql-recovery-db = { path = "../../sql_recovery_db" }
+mc-fog-uri = { path = "../../uri" }
+
+[dev-dependencies]
+# third-party
+rand_core = "0.6"
+rand_hc = "0.3"
+tempdir = "0.3"
+url = "2.2.2"
+
+# mc
+mc-attest-net = { path = "../../../attest/net" }
+mc-watcher = { path = "../../../watcher" }
+mc-ledger-db = { path = "../../../ledger/db" }
+mc-util-from-random = { path = "../../../util/from-random" }
+
+# fog
+mc-fog-ingest-server = { path = "../../ingest/server" }
+mc-fog-ingest-enclave = { path = "../../ingest/enclave" }
+mc-fog-test-infra = { path = "../../test_infra" }
--- a/fog/overseer/server/Cargo.toml
+++ b/fog/overseer/server/Cargo.toml
@@ -40,11 +40,11 @@ url = "2.2.2"
 
 # mc
 mc-attest-net = { path = "../../../attest/net" }
-mc-watcher = { path = "../../../watcher" }
 mc-ledger-db = { path = "../../../ledger/db" }
 mc-util-from-random = { path = "../../../util/from-random" }
+mc-watcher = { path = "../../../watcher" }
 
 # fog
-mc-fog-ingest-server = { path = "../../ingest/server" }
 mc-fog-ingest-enclave = { path = "../../ingest/enclave" }
+mc-fog-ingest-server = { path = "../../ingest/server" }
 mc-fog-test-infra = { path = "../../test_infra" }
--- a/fog/overseer/server/src/bin/main.rs
+++ b/fog/overseer/server/src/bin/main.rs
@@ -1,4 +1,5 @@
-// Copyright (c) 2018-2021 The MobileCoin Foundation
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
 use mc_common::logger::{log, o};
 use mc_fog_overseer_server::{config::OverseerConfig, server, service::OverseerService};
 use mc_fog_sql_recovery_db::SqlRecoveryDb;
--- a/fog/overseer/server/src/bin/main.rs
+++ b/fog/overseer/server/src/bin/main.rs
@@ -12,9 +12,7 @@ fn main() {
     // TODO: Enable sentry here.
     // See https://github.com/mobilecoinfoundation/mobilecoin/blob/master/fog/view/server/src/bin/main.rs#L16.
     let config = OverseerConfig::from_args();
-    let (logger, _global_logger_guard) = mc_common::logger::create_app_logger(
-        o!("mc.local_overseer_node_id" => config.local_overseer_node_id.to_string()),
-    );
+    let (logger, _global_logger_guard) = mc_common::logger::create_app_logger(o!());
 
     // Open the database.
     let recovery_db = SqlRecoveryDb::new_from_url(
--- a/fog/overseer/server/src/bin/main.rs
+++ b/fog/overseer/server/src/bin/main.rs
@@ -1,12 +1,13 @@
 // Copyright (c) 2018-2022 The MobileCoin Foundation
 
+//! Starts a Rocket server that allows clients to access Fog Overseer APIs
+//! over HTTP.
+
 use mc_common::logger::{log, o};
 use mc_fog_overseer_server::{config::OverseerConfig, server, service::OverseerService};
 use mc_fog_sql_recovery_db::SqlRecoveryDb;
 use structopt::StructOpt;
 
-/// Starts a Rocket server that allows clients to access Fog Overseer APIs
-/// over HTTP.
 fn main() {
     mc_common::setup_panic_handler();
     // TODO: Enable sentry here.
--- a/fog/overseer/server/src/bin/main.rs
+++ b/fog/overseer/server/src/bin/main.rs
@@ -1,0 +1,41 @@
+// Copyright (c) 2018-2021 The MobileCoin Foundation
+use mc_common::logger::{log, o};
+use mc_fog_overseer_server::{config::OverseerConfig, server, service::OverseerService};
+use mc_fog_sql_recovery_db::SqlRecoveryDb;
+use structopt::StructOpt;
+
+/// Starts a Rocket server that allows clients to access Fog Overseer APIs
+/// over HTTP.
+fn main() {
+    mc_common::setup_panic_handler();
+    let config = OverseerConfig::from_args();
+    let (logger, _global_logger_guard) = mc_common::logger::create_app_logger(
+        o!("mc.local_overseer_node_id" => config.local_overseer_node_id.to_string()),
+    );
+
+    // Open the database.
+    let recovery_db = SqlRecoveryDb::new_from_url(
+        &std::env::var("DATABASE_URL").expect("DATABASE_URL environment variable missing"),
+        config.postgres_config.clone(),
+        logger.clone(),
+    )
+    .expect("Failed connecting to database");
+
+    let mut overseer_service =
+        OverseerService::new(config.ingest_cluster_uris, recovery_db, logger.clone());
+
+    overseer_service
+        .start()
+        .expect("OverseerService failed to start");
+    log::info!(logger, "OverseerService successfully started.");
+
+    let overseer_state = server::OverseerState { overseer_service };
+
+    let rocket_config: rocket::Config =
+        rocket::Config::build(rocket::config::Environment::Development)
+            .address(config.listen_host)
+            .port(config.listen_port)
+            .unwrap();
+
+    server::initialize_rocket_server(rocket_config, overseer_state);
+}
--- a/fog/overseer/server/src/bin/main.rs
+++ b/fog/overseer/server/src/bin/main.rs
@@ -9,6 +9,8 @@ use structopt::StructOpt;
 /// over HTTP.
 fn main() {
     mc_common::setup_panic_handler();
+    // TODO: Enable sentry here.
+    // See https://github.com/mobilecoinfoundation/mobilecoin/blob/master/fog/view/server/src/bin/main.rs#L16.
     let config = OverseerConfig::from_args();
     let (logger, _global_logger_guard) = mc_common::logger::create_app_logger(
         o!("mc.local_overseer_node_id" => config.local_overseer_node_id.to_string()),
--- a/fog/overseer/server/src/config.rs
+++ b/fog/overseer/server/src/config.rs
@@ -1,0 +1,52 @@
+// Copyright (c) 2018-2021 The MobileCoin Foundation
+use mc_common::ResponderId;
+use mc_fog_sql_recovery_db::SqlRecoveryDbConnectionConfig;
+use mc_fog_uri::FogIngestUri;
+use serde::Serialize;
+use structopt::StructOpt;
+
+/// StructOpt configuration options for an Overseer Server
+#[derive(Clone, Serialize, StructOpt)]
+pub struct OverseerConfig {
+    /// Host to listen on.
+    #[structopt(long, default_value = "127.0.0.1")]
+    pub listen_host: String,
+
+    /// Port to start webserver on.
+    #[structopt(long, default_value = "9090")]
+    pub listen_port: u16,
+
+    /// Local Ingest Node ID
+    #[structopt(long)]
+    pub local_overseer_node_id: ResponderId,
+
+    /// gRPC listening URI for client requests.
+    #[structopt(long, use_delimiter = true)]
+    pub ingest_cluster_uris: Vec<FogIngestUri>,
+
+    /// Postgres config
+    #[structopt(flatten)]
+    pub postgres_config: SqlRecoveryDbConnectionConfig,
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    #[test]
+    fn ingest_server_config_example() {
+        let config = OverseerConfig::from_iter_safe(&[
+            "/usr/bin/start_overseer_server",
+            "--listen-host",
+            "www.mycoolhost.com",
+            "--listen-port",
+            "8080",
+            "--local-overseer-node-id",
+            "fogoverseer.svc.cluster.local:443",
+            "--ingest-cluster-uris",
+            "insecure-fog-ingest://0.0.0.0.3226/,insecure-fog-ingest://0.0.0.0.3227/",
+        ])
+        .expect("Could not parse command line arguments.");
+
+        assert_eq!(config.ingest_cluster_uris.len(), 2);
+    }
+}
--- a/fog/overseer/server/src/config.rs
+++ b/fog/overseer/server/src/config.rs
@@ -1,6 +1,5 @@
 // Copyright (c) 2018-2022 The MobileCoin Foundation
 
-use mc_common::ResponderId;
 use mc_fog_sql_recovery_db::SqlRecoveryDbConnectionConfig;
 use mc_fog_uri::FogIngestUri;
 use serde::Serialize;
@@ -16,10 +15,6 @@ pub struct OverseerConfig {
     /// Port to start webserver on.
     #[structopt(long, default_value = "9090")]
     pub listen_port: u16,
-
-    /// Local Ingest Node ID
-    #[structopt(long)]
-    pub local_overseer_node_id: ResponderId,
 
     /// TODO: Make this an environment variable that can be dynamically
     /// refreshed. This will allow ops to have one Fog Overseer instance that
@@ -46,8 +41,6 @@ mod tests {
             "www.mycoolhost.com",
             "--listen-port",
             "8080",
-            "--local-overseer-node-id",
-            "fogoverseer.svc.cluster.local:443",
             "--ingest-cluster-uris",
             "insecure-fog-ingest://0.0.0.0.3226/,insecure-fog-ingest://0.0.0.0.3227/",
         ])
--- a/fog/overseer/server/src/config.rs
+++ b/fog/overseer/server/src/config.rs
@@ -1,4 +1,5 @@
-// Copyright (c) 2018-2021 The MobileCoin Foundation
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
 use mc_common::ResponderId;
 use mc_fog_sql_recovery_db::SqlRecoveryDbConnectionConfig;
 use mc_fog_uri::FogIngestUri;
@@ -20,7 +21,12 @@ pub struct OverseerConfig {
     #[structopt(long)]
     pub local_overseer_node_id: ResponderId,
 
-    /// gRPC listening URI for client requests.
+    /// TODO: Make this an environment variable that can be dynamically
+    /// refreshed. This will allow ops to have one Fog Overseer instance that
+    /// can look at the new Fog Ingest cluster during the blue / green
+    /// deployment.
+    ///
+    /// gRPC listening URIs for client requests.
     #[structopt(long, use_delimiter = true)]
     pub ingest_cluster_uris: Vec<FogIngestUri>,
 
--- a/fog/overseer/server/src/config.rs
+++ b/fog/overseer/server/src/config.rs
@@ -33,6 +33,7 @@ pub struct OverseerConfig {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use mc_fog_uri::ConnectionUri;
     #[test]
     fn ingest_server_config_example() {
         let config = OverseerConfig::from_iter_safe(&[
@@ -42,10 +43,13 @@ mod tests {
             "--listen-port",
             "8080",
             "--ingest-cluster-uris",
-            "insecure-fog-ingest://0.0.0.0.3226/,insecure-fog-ingest://0.0.0.0.3227/",
+            "insecure-fog-ingest://0.0.0.0:3226/,insecure-fog-ingest://0.0.0.0:3227/",
         ])
         .expect("Could not parse command line arguments.");
 
         assert_eq!(config.ingest_cluster_uris.len(), 2);
+
+        assert_eq!(config.ingest_cluster_uris[0].port(), 3226);
+        assert_eq!(config.ingest_cluster_uris[1].port(), 3227);
     }
 }
--- a/fog/overseer/server/src/config.rs
+++ b/fog/overseer/server/src/config.rs
@@ -1,5 +1,7 @@
 // Copyright (c) 2018-2022 The MobileCoin Foundation
 
+//! Configuration parameters for Fog Overseer.
+
 use mc_fog_sql_recovery_db::SqlRecoveryDbConnectionConfig;
 use mc_fog_uri::FogIngestUri;
 use serde::Serialize;
--- a/fog/overseer/server/src/error.rs
+++ b/fog/overseer/server/src/error.rs
@@ -4,6 +4,7 @@
 use displaydoc::Display;
 use mc_fog_recovery_db_iface::RecoveryDbError;
 use mc_fog_sql_recovery_db::Error as SqlRecoveryDbError;
+use retry::Error as RetryError;
 
 /// An error returned by the overseer service
 #[derive(Debug, Display)]
@@ -22,10 +23,26 @@ pub enum OverseerError {
 
     /// There are multiple outstanding keys: {0}
     MultipleOutstandingKeys(String),
+
+    /// Unknown error associated with retries. Should not happen: {0}
+    GenericRetryError(String),
 }
 
 impl From<SqlRecoveryDbError> for OverseerError {
     fn from(src: SqlRecoveryDbError) -> Self {
         Self::RecoveryDb(Box::new(src))
+    }
+}
+
+impl From<RetryError<OverseerError>> for OverseerError {
+    fn from(src: RetryError<OverseerError>) -> Self {
+        match src {
+            RetryError::Operation {
+                error,
+                total_delay: _,
+                tries: _,
+            } => error,
+            RetryError::Internal(s) => Self::GenericRetryError(s),
+        }
     }
 }
--- a/fog/overseer/server/src/error.rs
+++ b/fog/overseer/server/src/error.rs
@@ -1,0 +1,31 @@
+// Copyright (c) 2018-2021 The MobileCoin Foundation
+//
+
+use displaydoc::Display;
+use mc_fog_recovery_db_iface::RecoveryDbError;
+use mc_fog_sql_recovery_db::Error as SqlRecoveryDbError;
+
+/// An error returned by the overseer service
+#[derive(Debug, Display)]
+pub enum OverseerError {
+    /// RecoveryDbError: {0}
+    RecoveryDb(Box<dyn RecoveryDbError>),
+    // TODO: Add more errors as appropriate
+    /// Reporting a lost key failed: {0}
+    ReportLostKey(String),
+
+    /// Setting a new key failed: {0}
+    SetNewKey(String),
+
+    /// Activating an idle node failed: {0}
+    ActivateNode(String),
+
+    /// There are multiple outstanding keys: {0}
+    MultipleOutstandingKeys(String),
+}
+
+impl From<SqlRecoveryDbError> for OverseerError {
+    fn from(src: SqlRecoveryDbError) -> Self {
+        Self::RecoveryDb(Box::new(src))
+    }
+}
--- a/fog/overseer/server/src/error.rs
+++ b/fog/overseer/server/src/error.rs
@@ -10,7 +10,7 @@ use retry::Error as RetryError;
 pub enum OverseerError {
     /// RecoveryDbError: {0}
     RecoveryDb(Box<dyn RecoveryDbError>),
-    // TODO: Add more errors as appropriate
+
     /// Reporting a lost key failed: {0}
     ReportLostKey(String),
 
--- a/fog/overseer/server/src/error.rs
+++ b/fog/overseer/server/src/error.rs
@@ -27,9 +27,6 @@ pub enum OverseerError {
 
     /// There are multiple outstanding keys: {0}
     MultipleOutstandingKeys(String),
-
-    /// Unknown error associated with retries. Should not happen: {0}
-    GenericRetryError(String),
 }
 
 impl From<SqlRecoveryDbError> for OverseerError {
@@ -46,7 +43,7 @@ impl From<RetryError<OverseerError>> for OverseerError {
                 total_delay: _,
                 tries: _,
             } => error,
-            RetryError::Internal(s) => Self::GenericRetryError(s),
+            RetryError::Internal(s) => panic!("Internal retry error: {}", s),
         }
     }
 }
--- a/fog/overseer/server/src/error.rs
+++ b/fog/overseer/server/src/error.rs
@@ -1,5 +1,7 @@
 // Copyright (c) 2018-2022 The MobileCoin Foundation
 
+//! Errors that can occur during Fog Overseer operation.
+
 use displaydoc::Display;
 use mc_fog_recovery_db_iface::RecoveryDbError;
 use mc_fog_sql_recovery_db::Error as SqlRecoveryDbError;
--- a/fog/overseer/server/src/error.rs
+++ b/fog/overseer/server/src/error.rs
@@ -11,6 +11,9 @@ pub enum OverseerError {
     /// RecoveryDbError: {0}
     RecoveryDb(Box<dyn RecoveryDbError>),
 
+    /// A Fog Ingest node did not respond to Fog Overseer: {0}
+    UnresponsiveNodeError(String),
+
     /// Reporting a lost key failed: {0}
     ReportLostKey(String),
 
--- a/fog/overseer/server/src/error.rs
+++ b/fog/overseer/server/src/error.rs
@@ -1,12 +1,11 @@
-// Copyright (c) 2018-2021 The MobileCoin Foundation
-//
+// Copyright (c) 2018-2022 The MobileCoin Foundation
 
 use displaydoc::Display;
 use mc_fog_recovery_db_iface::RecoveryDbError;
 use mc_fog_sql_recovery_db::Error as SqlRecoveryDbError;
 use retry::Error as RetryError;
 
-/// An error returned by the overseer service
+/// An error returned by the OverseerService.
 #[derive(Debug, Display)]
 pub enum OverseerError {
     /// RecoveryDbError: {0}
--- a/fog/overseer/server/src/lib.rs
+++ b/fog/overseer/server/src/lib.rs
@@ -2,6 +2,7 @@
 
 #![feature(proc_macro_hygiene, decl_macro)]
 
+pub mod config;
 pub mod server;
 pub mod service;
 
--- a/fog/overseer/server/src/lib.rs
+++ b/fog/overseer/server/src/lib.rs
@@ -2,5 +2,7 @@
 
 #![feature(proc_macro_hygiene, decl_macro)]
 
+pub mod service;
+
 mod error;
 mod worker;
--- a/fog/overseer/server/src/lib.rs
+++ b/fog/overseer/server/src/lib.rs
@@ -2,6 +2,7 @@
 
 #![feature(proc_macro_hygiene, decl_macro)]
 
+pub mod server;
 pub mod service;
 
 mod error;
--- a/fog/overseer/server/src/lib.rs
+++ b/fog/overseer/server/src/lib.rs
@@ -1,0 +1,6 @@
+// Copyright (c) 2018-2021 The MobileCoin Foundation
+
+#![feature(proc_macro_hygiene, decl_macro)]
+
+mod error;
+mod worker;
--- a/fog/overseer/server/src/lib.rs
+++ b/fog/overseer/server/src/lib.rs
@@ -1,4 +1,4 @@
-// Copyright (c) 2018-2021 The MobileCoin Foundation
+// Copyright (c) 2018-2022 The MobileCoin Foundation
 
 #![feature(proc_macro_hygiene, decl_macro)]
 
--- a/fog/overseer/server/src/lib.rs
+++ b/fog/overseer/server/src/lib.rs
@@ -2,7 +2,6 @@
 
 #![feature(proc_macro_hygiene, decl_macro)]
 #[deny(missing_docs)]
-
 pub mod config;
 pub mod server;
 pub mod service;
--- a/fog/overseer/server/src/lib.rs
+++ b/fog/overseer/server/src/lib.rs
@@ -1,6 +1,7 @@
 // Copyright (c) 2018-2022 The MobileCoin Foundation
 
 #![feature(proc_macro_hygiene, decl_macro)]
+#[deny(missing_docs)]
 
 pub mod config;
 pub mod server;
--- a/fog/overseer/server/src/service.rs
+++ b/fog/overseer/server/src/service.rs
@@ -1,0 +1,80 @@
+// Copyright (c) 2018-2021 The MobileCoin Foundation
+
+use crate::{error::OverseerError, worker::OverseerWorker};
+use mc_common::logger::{log, Logger};
+use mc_fog_recovery_db_iface::RecoveryDb;
+use mc_fog_uri::FogIngestUri;
+use std::sync::{
+    atomic::{AtomicBool, Ordering},
+    Arc,
+};
+
+// Implements core logic for the Fog Overseer HTTP server.
+pub struct OverseerService<DB: RecoveryDb + Clone + Send + Sync + 'static>
+where
+    OverseerError: From<DB::Error>,
+{
+    ingest_cluster_uris: Vec<FogIngestUri>,
+    logger: Logger,
+    overseer_worker: Option<OverseerWorker>,
+    recovery_db: DB,
+    stop_requested: Arc<AtomicBool>,
+}
+
+impl<DB: RecoveryDb + Clone + Send + Sync + 'static> OverseerService<DB>
+where
+    OverseerError: From<DB::Error>,
+{
+    pub fn new(ingest_cluster_uris: Vec<FogIngestUri>, recovery_db: DB, logger: Logger) -> Self {
+        Self {
+            ingest_cluster_uris,
+            logger,
+            overseer_worker: None,
+            recovery_db,
+            stop_requested: Arc::new(AtomicBool::new(false)),
+        }
+    }
+
+    /// Starts the Fog Overseer worker thread.
+    pub fn start(&mut self) -> Result<(), OverseerError> {
+        let ret = self.start_helper();
+        if let Err(ref err) = ret {
+            log::error!(self.logger, "Stopping ingest server due to {:?}", err);
+            self.stop()?;
+        }
+        ret
+    }
+
+    fn start_helper(&mut self) -> Result<(), OverseerError> {
+        assert!(self.overseer_worker.is_none());
+        log::info!(self.logger, "Starting overseer worker");
+
+        self.overseer_worker = Some(OverseerWorker::new(
+            self.ingest_cluster_uris.clone(),
+            self.recovery_db.clone(),
+            self.logger.clone(),
+            self.stop_requested.clone(),
+        ));
+
+        Ok(())
+    }
+
+    pub fn stop(&mut self) -> Result<(), OverseerError> {
+        // This blocks on teardown of overseer_worker
+        self.overseer_worker = None;
+
+        Ok(())
+    }
+
+    pub fn arm(&self) -> Result<String, String> {
+        log::trace!(self.logger, "Arming overseer worker");
+        self.stop_requested.store(true, Ordering::SeqCst);
+        Ok(String::from("Fog Overseer was successfully armed."))
+    }
+
+    pub fn disarm(&self) -> Result<String, String> {
+        log::trace!(self.logger, "Disarming overseer worker");
+        self.stop_requested.store(false, Ordering::SeqCst);
+        Ok(String::from("Fog Overseer was successfully disarmed."))
+    }
+}
--- a/fog/overseer/server/src/service.rs
+++ b/fog/overseer/server/src/service.rs
@@ -79,7 +79,7 @@ where
     }
 
     pub fn get_status(&self) -> Result<String, String> {
-        let is_enabled : bool = self.is_enabled.load(Ordering::SeqCst);
+        let is_enabled: bool = self.is_enabled.load(Ordering::SeqCst);
         let response_message = match is_enabled {
             true => "Fog Overseer is enabled.",
             false => "Fog Overseer is disabled.",
--- a/fog/overseer/server/src/service.rs
+++ b/fog/overseer/server/src/service.rs
@@ -74,7 +74,7 @@ where
             false => "Fog Overseer was successfully enabled",
         };
 
-        Ok(String::from(response_message))
+        Ok(response_message.to_string())
     }
 
     pub fn disable(&self) -> Result<String, String> {
@@ -85,7 +85,7 @@ where
             false => "Fog Overseer was already disabled",
         };
 
-        Ok(String::from(response_message))
+        Ok(response_message.to_string())
     }
 
     pub fn get_status(&self) -> Result<String, String> {
@@ -95,6 +95,6 @@ where
             false => "Fog Overseer is disabled.",
         };
 
-        Ok(String::from(response_message))
+        Ok(response_message.to_string())
     }
 }
--- a/fog/overseer/server/src/service.rs
+++ b/fog/overseer/server/src/service.rs
@@ -68,14 +68,26 @@ where
 
     pub fn enable(&self) -> Result<String, String> {
         log::info!(self.logger, "Enabling overseer worker");
-        self.is_enabled.store(true, Ordering::SeqCst);
-        Ok(String::from("Fog Overseer was successfully enabled."))
+        let was_enabled = self.is_enabled.swap(true, Ordering::SeqCst);
+        let response_message = if was_enabled {
+            "Fog Overseer was already enabled"
+        } else {
+            "Fog Overseer was successfully enabled"
+        };
+
+        Ok(String::from(response_message))
     }
 
     pub fn disable(&self) -> Result<String, String> {
         log::info!(self.logger, "Disabling overseer worker");
-        self.is_enabled.store(false, Ordering::SeqCst);
-        Ok(String::from("Fog Overseer was successfully disabled."))
+        let was_enabled = self.is_enabled.swap(false, Ordering::SeqCst);
+        let response_message = if was_enabled {
+            "Fog Overseer was successfully disabled"
+        } else {
+            "Fog Overseer was already disabled"
+        };
+
+        Ok(String::from(response_message))
     }
 
     pub fn get_status(&self) -> Result<String, String> {
--- a/fog/overseer/server/src/service.rs
+++ b/fog/overseer/server/src/service.rs
@@ -19,7 +19,6 @@ where
     overseer_worker: Option<OverseerWorker>,
     recovery_db: DB,
     is_enabled: Arc<AtomicBool>,
-    stop_requested: Arc<AtomicBool>,
 }
 
 impl<DB: RecoveryDb + Clone + Send + Sync + 'static> OverseerService<DB>
@@ -33,7 +32,6 @@ where
             overseer_worker: None,
             recovery_db,
             is_enabled: Arc::new(AtomicBool::new(true)),
-            stop_requested: Arc::new(AtomicBool::new(false)),
         }
     }
 
@@ -56,7 +54,7 @@ where
             self.recovery_db.clone(),
             self.logger.clone(),
             self.is_enabled.clone(),
-            self.stop_requested.clone(),
+            Arc::new(AtomicBool::new(false)),
         ));
 
         Ok(())
--- a/fog/overseer/server/src/service.rs
+++ b/fog/overseer/server/src/service.rs
@@ -69,10 +69,9 @@ where
     pub fn enable(&self) -> Result<String, String> {
         log::info!(self.logger, "Enabling overseer worker");
         let was_enabled = self.is_enabled.swap(true, Ordering::SeqCst);
-        let response_message = if was_enabled {
-            "Fog Overseer was already enabled"
-        } else {
-            "Fog Overseer was successfully enabled"
+        let response_message = match was_enabled {
+            true => "Fog Overseer was already enabled",
+            false => "Fog Overseer was successfully enabled",
         };
 
         Ok(String::from(response_message))
@@ -81,10 +80,9 @@ where
     pub fn disable(&self) -> Result<String, String> {
         log::info!(self.logger, "Disabling overseer worker");
         let was_enabled = self.is_enabled.swap(false, Ordering::SeqCst);
-        let response_message = if was_enabled {
-            "Fog Overseer was successfully disabled"
-        } else {
-            "Fog Overseer was already disabled"
+        let response_message = match was_enabled {
+            true => "Fog Overseer was successfully disabled",
+            false => "Fog Overseer was already disabled",
         };
 
         Ok(String::from(response_message))
--- a/fog/overseer/server/src/service.rs
+++ b/fog/overseer/server/src/service.rs
@@ -1,5 +1,12 @@
 // Copyright (c) 2018-2022 The MobileCoin Foundation
 
+//! Implements the service logic for the Fog Overseer. 
+//!
+//! This is encapsulated by Fog Oveseer Server, and it in turn encapsulates the 
+//! OverseerWorker, which contains the overseer busines logic. 
+//!
+//! HTTP Client -> Overseer Rocket Server -> *OverseerService* -> OverseerWorker
+
 use crate::{error::OverseerError, worker::OverseerWorker};
 use mc_common::logger::{log, Logger};
 use mc_fog_recovery_db_iface::RecoveryDb;
@@ -9,7 +16,7 @@ use std::sync::{
     Arc,
 };
 
-// Implements core logic for the Fog Overseer HTTP server.
+/// Implements core logic for the Fog Overseer HTTP server.
 pub struct OverseerService<DB: RecoveryDb + Clone + Send + Sync + 'static>
 where
     OverseerError: From<DB::Error>,
--- a/fog/overseer/server/src/service.rs
+++ b/fog/overseer/server/src/service.rs
@@ -68,13 +68,13 @@ where
 
     pub fn arm(&self) -> Result<String, String> {
         log::trace!(self.logger, "Arming overseer worker");
-        self.stop_requested.store(true, Ordering::SeqCst);
+        self.stop_requested.store(false, Ordering::SeqCst);
         Ok(String::from("Fog Overseer was successfully armed."))
     }
 
     pub fn disarm(&self) -> Result<String, String> {
         log::trace!(self.logger, "Disarming overseer worker");
-        self.stop_requested.store(false, Ordering::SeqCst);
+        self.stop_requested.store(true, Ordering::SeqCst);
         Ok(String::from("Fog Overseer was successfully disarmed."))
     }
 }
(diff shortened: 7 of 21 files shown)


fog/overseer/server/src/service.rs
```diff
@@ -0,0 +1,80 @@
...
+    pub fn start(&mut self) -> Result<(), OverseerError> {
+        let ret = self.start_helper();
+        if let Err(ref err) = ret {
+            log::error!(self.logger, "Stopping ingest server due to {:?}", err);
+            self.stop()?;
+        }
+        ret
+    }
+
+    fn start_helper(&mut self) -> Result<(), OverseerError> {
```
wjuan-mob, line 48: Why split this out from start?

fog/overseer/server/src/worker.rs
```diff
@@ -0,0 +1,546 @@
...
+    node_index: usize,
+
+    ingest_summary: IngestSummary,
+}
+
+impl<DB: RecoveryDb> OverseerWorkerThread<DB>
+where
+    OverseerError: From<DB::Error>,
+{
+    /// Poll the Fog Ingest cluter every 5 seconds.
```
wjuan-mob, line 130: Nit: Typo

fog/overseer/server/src/worker.rs
```diff
@@ -0,0 +1,546 @@
...
+            recovery_db,
+            stop_requested,
+            logger,
+        };
+        thread.run();
+    }
+
+    fn run(self) {
+        while !self.stop_requested.load(Ordering::SeqCst) {
+            log::trace!(self.logger, "Overser worker start of thread.");
```
wjuan-mob, line 154: Nit: typo

fog/overseer/server/src/worker.rs
```diff
@@ -0,0 +1,546 @@
...
+    ) -> Result<Vec<CompressedRistrettoPublic>, OverseerError> {
+        // An outanding key is one that Fog Ingest is still obligated to be
+        // scanning blocks with on behalf of users.
+        let outstanding_keys_filters = IngressPublicKeyRecordFilters {
+            // A lost key is not outstanding because it will never again be used
+            // to scan blocks.
+            should_include_lost_keys: false,
+            // A retired key can be outstanding if public expiry is greater than
+            // its last scanned block.
+            should_include_retired_keys: true,
```
wjuan-mob, line 334: The comment on should include retired keys and should_only_include_unexpired_keys is a bit confusing. Could we clarify this a little bit? How are these 2 different?

fog/overseer/server/src/worker.rs
```diff
@@ -0,0 +1,546 @@
...
+                Ok(())
+            }
+            // TODO: Add alerting that says we need to manually report these keys
+            // as lost.
+            Err(err) => Err(err.into()),
+        }
+    }
+
+    fn set_new_key_on_a_node(&self) -> Result<usize, OverseerError> {
+        for (i, ingest_client) in self.ingest_clients.iter().enumerate() {
```
wjuan-mob, line 463: I think we could assert that we're only calling this on idle nodes as a precaution.

fog/overseer/server/src/worker.rs
```diff
@@ -0,0 +1,546 @@
...
+            }
+        }
+
+        Err(OverseerError::SetNewKey(
+            "New keys were not successfully set on any of the idle nodes.".to_string(),
+        ))
+    }
+
+    fn activate_a_node(&self, activated_node_index: usize) -> Result<(), OverseerError> {
+        let result = retry_with_index(Fixed::from_millis(200), |current_try| {
```
wjuan-mob, line 507: This should probably also only be called on idle nodes.

wjuan-mob (approved):
I think it might be nice to move more of the boilerplate in the tests into the test utils or have a macro. But I don't think it's necessary. 

Is it possible for a concurrency issue to occur where the overseer queries the servers in sequence for their state and one becomes active after being queried but before the others are done. So then the overseer believes that they are all idle and attempts to do something.

fog/overseer/server/src/service.rs
samdealy: It helps keep the start function small and focused- it deals with handling any errors that might arise from starting the service without getting into the nitty gritty details of what threads etc. are started. Moreover, splitting the logic out between a `start` and `start_helper` is the convention for Fog services.

fog/overseer/server/src/worker.rs
samdealy: The two filters are different in that the `should_include_retired_keys` set to `true` includes keys that have been marked retired in the recovery DB (setting it to `false` discards any key that is marked retired). The second filter, `should_only_include_unexpired_keys` discards any key that is expired- an expired key is one in which it's pubkey_expiry (i.e. the last block for which the key must be used to scan) is <= the last scanned block. 

 We need retired keys because they can still be outstanding, meaning they might still have work to do because their pubkey_expiry (i.e. the last block for which they are responsible for scanning) might be greater than the last scanned block.

 I added to the `should_include_retired_keys` comment to try and clarify this point. Let me know if that helps or if you'd like to see anything else written in this comment.

eranrund (changes requested):
Nice work.
I have not looked at the tests, but I did have some (relatively minor) suggestions, and an open question about the use of stop_trigger.
I think what should've happened is that arming starts the worker thread or errors if it is already started, and disarming stops it?

fog/overseer/server/src/worker.rs
samdealy: I need to think about putting in an assert for idle state. I agree that it might be good for precaution's sake, but this method is only called when the main method, `perform_automatic_failover`, has determined that every node for which it received a response, is idle. 

I guess the question is: Is it possible for one of these idle nodes to become active in between when `perform_automatic_failover` determines that it's idle and when this method is called? In a perfect world, the answer is no. I'll think more on it to see if there any cases that might violate this assumption. If there are, then the assert makes perfect sense. 

@garbageslam any thoughts on this?

fog/overseer/server/src/worker.rs
samdealy: Same as above.

fog/overseer/server/src/worker.rs
samdealy: Done.

fog/overseer/server/src/worker.rs
samdealy: Done.

samdealy:
> I think it might be nice to move more of the boilerplate in the tests into the test utils or have a macro. But I don't think it's necessary.
I think this is a good point, and we should start having conversations as a team to see what we should and shouldn't include in tests / test utils. I tend to follow the convention that tests should be DAMP rather than DRY (https://testing.googleblog.com/2019/12/testing-on-toilet-tests-too-dry-make.html). I try to include enough stuff in a test such that the reader doesn't need to go to another file to understand what is going on in the test. For some things, like setting up the Ingest node, I make a method for it because the reader doesn't need to know how the node is made. 

> Is it possible for a concurrency issue to occur where the overseer queries the servers in sequence for their state and one becomes active after being queried but before the others are done. So then the overseer believes that they are all idle and attempts to do something.

I believe it's pretty unlikely that this would happen because the overseer is started after the three nodes are started, and the first node is activated. However, if we wanted to be very certain about this sequencing, we could implement some sort of async / await or condition variable that allows the test to wait until the nodes are started and activated before starting overseer.

wjuan-mob:
> > I think it might be nice to move more of the boilerplate in the tests into the test utils or have a macro. But I don't think it's necessary.
> 
> I think this is a good point, and we should start having conversations as a team to see what we should and shouldn't include in tests / test utils. I tend to follow the convention that tests should be DAMP rather than DRY (https://testing.googleblog.com/2019/12/testing-on-toilet-tests-too-dry-make.html). I try to include enough stuff in a test such that the reader doesn't need to go to another file to understand what is going on in the test. For some things, like setting up the Ingest node, I make a method for it because the reader doesn't need to know how the node is made.
> 
> > Is it possible for a concurrency issue to occur where the overseer queries the servers in sequence for their state and one becomes active after being queried but before the others are done. So then the overseer believes that they are all idle and attempts to do something.
> 
> I believe it's pretty unlikely that this would happen because the overseer is started after the three nodes are started, and the first node is activated. However, if we wanted to be very certain about this sequencing, we could implement some sort of async / await or condition variable that allows the test to wait until the nodes are started and activated before starting overseer.

I generally agree with damp over dry. I just think that some amount of the boilerplate in these tests is irrelevant to the actual condition that is being tested, and extracting it might make it easier to read. Basically everything prior to when nodes start being set for example.

fog/overseer/server/Cargo.toml
```diff
@@ -0,0 +1,50 @@
...
+rand_core = "0.6"
+rand_hc = "0.3"
+tempdir = "0.3"
+url = "2.2.2"
+
+# mc
+mc-attest-net = { path = "../../../attest/net" }
+mc-watcher = { path = "../../../watcher" }
+mc-ledger-db = { path = "../../../ledger/db" }
+mc-util-from-random = { path = "../../../util/from-random" }
```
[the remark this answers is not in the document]
samdealy: Done.

fog/overseer/server/src/bin/main.rs
```diff
@@ -0,0 +1,42 @@
...
+use structopt::StructOpt;
+
+/// Starts a Rocket server that allows clients to access Fog Overseer APIs
+/// over HTTP.
+fn main() {
+    mc_common::setup_panic_handler();
+    let config = OverseerConfig::from_args();
+    let (logger, _global_logger_guard) = mc_common::logger::create_app_logger(
+        o!("mc.local_overseer_node_id" => config.local_overseer_node_id.to_string()),
+    );
```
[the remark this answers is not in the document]
samdealy: Added a comment.

fog/overseer/server/src/config.rs
```diff
@@ -0,0 +1,58 @@
...
+
+    /// Port to start webserver on.
+    #[structopt(long, default_value = "9090")]
+    pub listen_port: u16,
+
+    /// Local Ingest Node ID
+    #[structopt(long)]
+    pub local_overseer_node_id: ResponderId,
+
+    /// TODO: Make this an environment variable that can be dynamically
```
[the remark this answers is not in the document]
samdealy: For context, this idea came up in a discussion with ops about how to keep one Fog Overseer instance running between blue-green deployments. 

I had, perhaps naively, thought that there'd be a widely-used, OOTB way to achieve this in Rust. After digging in a little more, I found that this isn't the case, but there is this crate, [spirit](https://docs.rs/spirit/0.3.7/spirit/), that appears to allow for config reloading. What do you think about this crate? Not sure if we've looked into this before? 

@jgreat Feel free to add anything re: our discussion on dynamic config reloading.

fog/overseer/server/src/worker.rs
```diff
@@ -0,0 +1,546 @@
...
+                    );
+                    match self.ingest_clients[ingest_summary_node_mapping.node_index].activate() {
+                        Ok(_) => {
+                            log::info!(self.logger, "Successfully activated ingest client.");
+                            return Ok(());
+                        }
+                        Err(_) => {
+                            // TODO: See if we can make an id field for ingest client. This
+                            // will allow us to better identify ingest clients.
+                            log::warn!(self.logger, "Could not activate idle ingest client.");
```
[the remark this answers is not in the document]
samdealy: IIUC, yes, I opted to treat the node's URI as it's identifier. If using the URI as an ID is fine, I can just remove this comment.

fog/overseer/server/src/worker.rs
```diff
@@ -0,0 +1,546 @@
...
+                        }
+                    }
+                }
+            }
+        }
+
+        Err(OverseerError::ActivateNode("Could not activate a node because no node has an ingress key that matches the outstanding key.".to_string()))
```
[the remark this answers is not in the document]
samdealy: Done.

fog/overseer/server/src/config.rs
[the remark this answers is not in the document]
eranrund: Why not send configuration changes over an HTTP API?

fog/overseer/server/src/worker.rs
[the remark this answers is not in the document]
eranrund: I think the URI is a good unique identifier for ingest nodes.

fog/overseer/server/src/config.rs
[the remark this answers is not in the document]
samdealy: I tend to think that configuration shouldn't be modified over HTTP and that we should try to keep configs in a text file. That way, anyone can look at that text file at any given time to see what the configs are. However, I'd be happy to do HTTP given it's simplicity / if ops is ok with it.

fog/overseer/server/src/worker.rs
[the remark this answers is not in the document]
samdealy: SG, removed the comment.

fog/overseer/server/src/config.rs
```diff
@@ -0,0 +1,58 @@
...
+    #[structopt(long, default_value = "127.0.0.1")]
+    pub listen_host: String,
+
+    /// Port to start webserver on.
+    #[structopt(long, default_value = "9090")]
+    pub listen_port: u16,
+
+    /// Local Ingest Node ID
+    #[structopt(long)]
+    pub local_overseer_node_id: ResponderId,
```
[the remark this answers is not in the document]
samdealy: Ah, I didn't realize that ResponderId corresponded to an enclave. I still want to have some sort of ID that corresponds to the Fog Overseer instance (one day, we'll be able to run multiple overseer instances at once), so perhaps a simple String would suffice instead of a ResponderId?

fog/overseer/server/src/config.rs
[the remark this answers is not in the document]
eranrund: Text file is different than an env variable though. I view env variable as "dynamic" in the same way I'd consider HTTP dynamic, and a text file as "static"

fog/overseer/server/src/config.rs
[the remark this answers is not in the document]
eranrund: Is the hostname of the machine not good enough?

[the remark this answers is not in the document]
eranrund: I'd also suggest that there's no need to future-proof this at the moment. Multiple instances will be a big enough change that this can be introduced at that point.

fog/overseer/server/src/worker.rs
```diff
@@ -0,0 +1,546 @@
...
+    }
+
+    fn report_lost_ingress_key(
+        &self,
+        inactive_outstanding_key: CompressedRistrettoPublic,
+    ) -> Result<(), OverseerError> {
+        let result = retry_with_index(Fixed::from_millis(200), |current_try| {
+            if current_try as u8 > Self::NUMBER_OF_TRIES {
```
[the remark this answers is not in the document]
samdealy: I used it because I'd like the error message to include how many tries remain. I don't think the take option enables this? If it does, then great- it's more compact and would change to it.

Relevant [documentation](https://docs.rs/retry/latest/retry/#usage):
`If your operation needs to know how many times it’s been tried, use the retry_with_index function. This works the same as retry, but passes the number of the current try to the closure as an argument.`

fog/overseer/server/src/config.rs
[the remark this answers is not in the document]
wjuan-mob: Could you not save the active configs to a text file even if it's loaded over an http api?

fog/overseer/server/src/worker.rs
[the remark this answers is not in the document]
eranrund: You can use a combination of `retry_with_index` and `take`, which will allow you to include the number of retries remaining. It will save you from having to implement the limiting yourself.

fog/overseer/server/src/config.rs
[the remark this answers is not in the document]
samdealy: Ah, I was using text file == env variable because most env variables are set in some sort of .env file. The way I envision it, updates to configs would be made by editing the text file (making it dynamic and not static). Then a worker would check for an updated config. 

You could also save the active configs to a text file, but at that point why not just have the person changing the config edit the text file directly?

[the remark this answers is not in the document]
samdealy: I'm curious as to what @jgreat thinks about this. Would you prefer to update the list of Fog Ingest nodes that Overseer watches through (a) HTTP request or (b) updating a config file and having Overseer check for these updates?

fog/overseer/server/src/worker.rs
```diff
@@ -0,0 +1,546 @@
...
+                }
+                Err(_) => {
+                    let number_of_remaining_tries = Self::NUMBER_OF_TRIES - current_try as u8;
+                    let error_message = format!("The following key  was not successfully reported as lost: {}. Will try {} more times", inactive_outstanding_key, number_of_remaining_tries);
+                    OperationResult::Retry(OverseerError::ReportLostKey(error_message))
+                }
+            }
+        });
+
+        match result {
```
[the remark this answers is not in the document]
samdealy: Done. 

Changing the match statement as suggested, didn't work, and I had to use map_err` to get the method to return the correct type of E in Result<T, E>. Let me know if there's an even sleeker way of achieving this.

fog/overseer/server/src/worker.rs
```diff
@@ -0,0 +1,546 @@
...
+                }
+            }
+        }
+
+        Err(OverseerError::SetNewKey(
+            "New keys were not successfully set on any of the idle nodes.".to_string(),
+        ))
+    }
+
+    fn activate_a_node(&self, activated_node_index: usize) -> Result<(), OverseerError> {
```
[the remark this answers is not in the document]
samdealy: Done.

fog/overseer/server/src/worker.rs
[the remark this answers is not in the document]
eranrund: Does `Ok(result?)` work?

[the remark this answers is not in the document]
samdealy: Yes. Changed it to that. Thanks for suggestion, makes it sleek.

fog/overseer/server/src/config.rs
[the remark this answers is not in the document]
samdealy: SG, just removed the usage of ResponderId.

samdealy:
> Nice work. I have not looked at the tests, but I did have some (relatively minor) suggestions, and an open question about the use of stop_trigger. I think what should've happened is that arming starts the worker thread or errors if it is already started, and disarming stops it?

Ah I see your confusion. The way I've implemented it, the worker thread is started when OverseerService#start is called, which occurs when the "start_overseer_service" binary is executed. Arm is really only used after the service is disarmed- i.e. to restart the service after it's been stopped. In this sense, perhaps "arm" should be renamed to rearm? I can also modify it to emit an error if the worker thread is already running. 

That said, if you think the executable should not start the worker thread automatically let me know. I picked up this convention from other fog services, but perhaps it doesn't make sense here given that we are exposing an "arm" or "rearm" api.

eranrund:
> > Nice work. I have not looked at the tests, but I did have some (relatively minor) suggestions, and an open question about the use of stop_trigger. I think what should've happened is that arming starts the worker thread or errors if it is already started, and disarming stops it?
> 
> Ah I see your confusion. The way I've implemented it, the worker thread is started when OverseerService#start is called, which occurs when the "start_overseer_service" binary is executed. Arm is really only used after the service is disarmed- i.e. to restart the service after it's been stopped. In this sense, perhaps "arm" should be renamed to rearm? I can also modify it to emit an error if the worker thread is already running.
> 
> That said, if you think the executable should not start the worker thread automatically let me know. I picked up this convention from other fog services, but perhaps it doesn't make sense here given that we are exposing an "arm" or "rearm" api.

This still doesn't add up since I don't see where the thread is restarted... it appears its only started when the process starts.
Why do we need this if we are auto-starting it?

I also think that given what you said, the API names might just need to be "start" and "stop"?

eranrund:
And potentially include a status API endpoint that indicates if it is currently started...

fog/overseer/server/src/service.rs
```diff
@@ -66,15 +66,15 @@ where
         Ok(())
     }
 
-    pub fn arm(&self) -> Result<String, String> {
-        log::trace!(self.logger, "Arming overseer worker");
-        self.stop_requested.store(false, Ordering::SeqCst);
+    pub fn enable(&self) -> Result<String, String> {
+        log::trace!(self.logger, "Enabling overseer worker");
```
eranrund, line 70: Recommend `info!` so that there's a log of this happening.

fog/overseer/server/src/service.rs
```diff
@@ -66,15 +66,15 @@ where
...
+        self.is_enabled.store(true, Ordering::SeqCst);
         Ok(String::from("Fog Overseer was successfully armed."))
     }
 
-    pub fn disarm(&self) -> Result<String, String> {
-        log::trace!(self.logger, "Disarming overseer worker");
-        self.stop_requested.store(true, Ordering::SeqCst);
+    pub fn disable(&self) -> Result<String, String> {
+        log::trace!(self.logger, "Disabling overseer worker");
```
eranrund, line 76: Recommend `info!` so that there's a log of this happening.

fog/overseer/server/src/worker.rs
```diff
@@ -32,8 +32,8 @@ pub struct OverseerWorker {
     /// Join handle used to wait for the thread to terminate.
     join_handle: Option<JoinHandle<()>>,
 
-    /// Stop request trigger, used to signal the thread to stop.
-    stop_requested: Arc<AtomicBool>,
+    /// When true, overseer performs its monitoring logic.
+    is_enabled: Arc<AtomicBool>,
```
eranrund, line 36: Is there a need to store this here? If it comes from the outside (the caller of `OverseerWorker::new`) and is immediately passed to the thread, my thinking is no.

fog/overseer/server/src/worker.rs
```diff
@@ -78,14 +78,14 @@ impl OverseerWorker {
...
-            stop_requested,
+            is_enabled,
         }
     }
 
     /// Stop and join the db poll thread
     pub fn stop(&mut self) -> Result<(), ()> {
         if let Some(join_handle) = self.join_handle.take() {
-            self.stop_requested.store(true, Ordering::SeqCst);
+            self.is_enabled.store(false, Ordering::SeqCst);
```
eranrund, line 88: Why would this cause the thread to stop?

fog/overseer/server/src/worker.rs
```diff
@@ -137,23 +137,28 @@ where
...
-            stop_requested,
+            is_enabled,
             logger,
         };
         thread.run();
     }
 
     fn run(self) {
-        while !self.stop_requested.load(Ordering::SeqCst) {
+        loop {
```
eranrund, line 153: The thread is never going to terminate if the loop doesn't have an exit condition.

fog/overseer/server/src/worker.rs
```diff
@@ -137,23 +137,28 @@ where
...
             log::trace!(self.logger, "Overseer worker start of thread.");
             std::thread::sleep(Self::POLLING_FREQUENCY);
 
+            if !self.is_enabled.load(Ordering::SeqCst) {
+                log::info!(self.logger, "Overseer worker is currently disabled.");
```
eranrund, line 158: This will be very spammy in the logs.

eranrund (changes requested):
It appears to me that the thread management is still buggy.
I am also worried that there are no tests that cover this, since if I am right something should've caught that - and if I'm wrong then I am misunderstanding something.

I also recommend adding a `/status` endpoint to check if the thing is enabled or disabled. This will likely be useful in the future.

fog/overseer/server/src/worker.rs
samdealy: I moved it out. However, for the new AtomicBool, stop_requested, I keep it in this struct because the stop method needs  to set it to true.

fog/overseer/server/src/worker.rs
samdealy: Fixed in most recent commit.

fog/overseer/server/src/worker.rs
```diff
@@ -0,0 +1,546 @@
...
+struct OverseerWorkerThread<DB: RecoveryDb> {
+    /// The list of FogIngestClients that Overseer uses to communicate with
+    /// each node in the Fog Ingest cluster that it's monitoring.
+    ingest_clients: Vec<FogIngestGrpcClient>,
+
+    /// The database that contains, among other things, info on the Fog Ingest
+    /// cluster's ingress keys.
+    recovery_db: DB,
+
+    /// If this is true, the worker will not perform it's monitoring logic.
```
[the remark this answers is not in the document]
samdealy: Fixed in most recent commit.

fog/overseer/server/src/worker.rs
samdealy: Fixed in most recent commit.

fog/overseer/server/src/worker.rs
samdealy: Fixed in most recent commit.

fog/overseer/server/src/service.rs
samdealy: Done.

fog/overseer/server/src/service.rs
samdealy: Done.

fog/overseer/server/src/bin/main.rs
```diff
@@ -0,0 +1,42 @@
...
+use mc_common::logger::{log, o};
+use mc_fog_overseer_server::{config::OverseerConfig, server, service::OverseerService};
+use mc_fog_sql_recovery_db::SqlRecoveryDb;
+use structopt::StructOpt;
+
+/// Starts a Rocket server that allows clients to access Fog Overseer APIs
+/// over HTTP.
+fn main() {
+    mc_common::setup_panic_handler();
+    // TODO: Enable sentry here.
```
eranrund, line 12: Why not just stick that piece of code here? That's all that is needed to enable it.

fog/overseer/server/src/config.rs
```diff
@@ -0,0 +1,58 @@
...
+
+    /// TODO: Make this an environment variable that can be dynamically
+    /// refreshed. This will allow ops to have one Fog Overseer instance that
+    /// can look at the new Fog Ingest cluster during the blue / green
+    /// deployment.
+    ///
+    /// gRPC listening URIs for client requests.
+    #[structopt(long, use_delimiter = true)]
```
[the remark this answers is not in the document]
eranrund: Still an open issue/question.

fog/overseer/server/src/error.rs
```diff
@@ -0,0 +1,47 @@
...
+use mc_fog_recovery_db_iface::RecoveryDbError;
+use mc_fog_sql_recovery_db::Error as SqlRecoveryDbError;
+use retry::Error as RetryError;
+
+/// An error returned by the OverseerService.
+#[derive(Debug, Display)]
+pub enum OverseerError {
+    /// RecoveryDbError: {0}
+    RecoveryDb(Box<dyn RecoveryDbError>),
+    // TODO: Add more errors as appropriate
```
eranrund, line 13: I don't think we need this comment.

fog/overseer/server/src/service.rs
```diff
@@ -0,0 +1,93 @@
...
+pub struct OverseerService<DB: RecoveryDb + Clone + Send + Sync + 'static>
+where
+    OverseerError: From<DB::Error>,
+{
+    ingest_cluster_uris: Vec<FogIngestUri>,
+    logger: Logger,
+    overseer_worker: Option<OverseerWorker>,
+    recovery_db: DB,
+    is_enabled: Arc<AtomicBool>,
+    stop_requested: Arc<AtomicBool>,
```
eranrund, line 22: No need to store this in this struct since it doesn't look like it's ever being accessed.

fog/overseer/server/src/service.rs
```diff
@@ -0,0 +1,93 @@
...
+    fn start_helper(&mut self) -> Result<(), OverseerError> {
+        assert!(self.overseer_worker.is_none());
+        log::info!(self.logger, "Starting overseer worker");
+
+        self.overseer_worker = Some(OverseerWorker::new(
+            self.ingest_cluster_uris.clone(),
+            self.recovery_db.clone(),
+            self.logger.clone(),
+            self.is_enabled.clone(),
+            self.stop_requested.clone(),
```
eranrund, line 59: This can live entirely inside `OverseerWorker` and doesn't need to be exposed.

fog/overseer/server/src/worker.rs
```diff
@@ -0,0 +1,546 @@
...
+                Err(err) => log::error!(
+                    self.logger,
+                    "There was an issue activating a node for the outstanding key: {:?}",
+                    err
+                ),
+            }
+        }
+
+        // If there's an outstanding key, we need to report it as lost.
+        if let Some(inactive_outstanding_key) = inactive_outstanding_keys.get(0) {
```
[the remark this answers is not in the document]
eranrund: Still an open question

fog/overseer/server/src/worker.rs
```diff
@@ -0,0 +1,544 @@
...
+                // we don't want a server to be active with a lost key.
+                //
+                // Since we always expect the node to come back in some
+                // time, we should just keep trying here.
+                //
+                // TODO: Add alerting if this is taking too long.
+                loop {
+                    log::info!(
+                        self.logger,
+                        "Trying to activate node at index {}",
```
eranrund, line 417: I suggest logging the URI from the client as well, it will make it easier for anyone looking at the logs.

fog/overseer/server/tests/inactive_outstanding_key_idle_does_not_have_key.rs
```diff
@@ -0,0 +1,216 @@
...
+    // TODO: Consider testing the CLI here instead.
+    let rocket_config: rocket::Config =
+        rocket::Config::build(rocket::config::Environment::Development)
+            // TODO: Make these either passed from CLI or in a Rocket.toml.
+            .address("127.0.0.1")
+            .port(80)
+            .unwrap();
+    let rocket = server::initialize_rocket_server(rocket_config, overseer_state);
+    let client = Client::new(rocket).expect("valid rocket instance");
+    let _req = client.post("/arm").dispatch();
```
eranrund, line 147: I'm surprised to see this code in the tests. Why can't it use the stuff inside `fog/overseer/server/src/server.rs`, or just operate on `overseer_service` directly?

fog/overseer/server/tests/inactive_outstanding_key_idle_does_not_have_key.rs
```diff
@@ -0,0 +1,216 @@
...
+    overseer_service.start().unwrap();
+
+    // Set up the Rocket instance
+    let overseer_state = OverseerState { overseer_service };
+    // TODO: Consider testing the CLI here instead.
+    let rocket_config: rocket::Config =
+        rocket::Config::build(rocket::config::Environment::Development)
+            // TODO: Make these either passed from CLI or in a Rocket.toml.
+            .address("127.0.0.1")
+            .port(80)
```
eranrund, line 143: This is fishy, ports lower than 1024 requires running as root.

fog/overseer/server/tests/inactive_outstanding_key_idle_has_key.rs
```diff
@@ -0,0 +1,208 @@
...
+    overseer_service.start().unwrap();
+
+    // Set up the Rocket instance
+    let overseer_state = OverseerState { overseer_service };
+    // TODO: Consider testng the CLI here instead
+    let rocket_config: rocket::Config =
+        rocket::Config::build(rocket::config::Environment::Development)
+            // TODO: Make these either passed from CLI or in a Rocket.toml.
+            .address("127.0.0.1")
+            .port(80)
```
eranrund, line 130: Using the same port in multiple tests can have undesirable effects such as one of the tests failing to bind the port, or the requests going to the wrong place.

Like in the previous comment, I am not sure there's a need to bother with rocket here at all.

fog/overseer/server/tests/utils/mod.rs
```diff
@@ -0,0 +1,172 @@
...
+    let mut node = IngestServer::new(config, ra_client, db, watcher, ledger_db, logger);
+    node.start().expect("couldn't start node");
+
+    (node, state_tmp, client_listen_uri)
+}
+
+// Add an arbitrary block to ledger and a timestamp for it
+// This method is used in integration tests, but cargo still complains that it's
+// not in use.
+#[allow(dead_code)]
```
eranrund, line 108: Is this needed?

eranrund (changes requested):
Nice progress, thanks for dealing with the thread stuff!

fog/overseer/server/tests/inactive_outstanding_key_idle_does_not_have_key.rs
samdealy: I'd like to make it a proper end to end test. Why not exercise all of the code paths, rather than a layer beneath it? Without this, we can't be certain that the rocket logic is working correctly.

fog/overseer/server/tests/utils/mod.rs
samdealy: Yes, CI complains that it's not used (I don't think cargo checks that it's being used in the tests/ directory).

fog/overseer/server/tests/inactive_outstanding_key_idle_has_key.rs
samdealy: See above.

fog/overseer/server/src/config.rs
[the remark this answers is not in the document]
garbageslam: Maybe it could do something with like, a shared volume i guess? That might be durable? Lets see what @jgreat says

fog/overseer/server/src/bin/main.rs
samdealy: I haven't set up sentry alerting yet and would like to do it all in one PR.

fog/overseer/server/src/config.rs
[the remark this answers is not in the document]
samdealy: It seems easier to use a delimited set of words rather than typing in --peer multiple times, but can change it if you think that's better.

fog/overseer/server/src/error.rs
samdealy: Done.

fog/overseer/server/src/service.rs
samdealy: What can live inside it? (GitHub isn't showing me what you highlighted / are specifically referring to).

samdealy: I see now- you're referring to self.stop_requested.clone(). Changed accordingly. Done.

fog/overseer/server/src/service.rs
samdealy: Nevermind, we can remove it.

fog/overseer/server/src/service.rs
```diff
@@ -0,0 +1,80 @@
...
+            self.logger.clone(),
+            self.stop_requested.clone(),
+        ));
+
+        Ok(())
+    }
+
+    pub fn stop(&mut self) -> Result<(), OverseerError> {
+        // This blocks on teardown of overseer_worker
+        self.overseer_worker = None;
```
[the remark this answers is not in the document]
samdealy: Removed the struct field. Done.

fog/overseer/server/src/service.rs
```diff
@@ -0,0 +1,80 @@
...
+
+        Ok(())
+    }
+
+    pub fn arm(&self) -> Result<String, String> {
+        log::trace!(self.logger, "Arming overseer worker");
```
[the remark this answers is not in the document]
samdealy: This change was made in subsequent commit.

fog/overseer/server/src/service.rs
```diff
@@ -0,0 +1,80 @@
...
+        self.stop_requested.store(true, Ordering::SeqCst);
```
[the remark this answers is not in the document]
samdealy: Fixed in later commit.

fog/overseer/server/src/service.rs
```diff
@@ -0,0 +1,80 @@
...
+        Ok(String::from("Fog Overseer was successfully armed."))
+    }
+
+    pub fn disarm(&self) -> Result<String, String> {
+        log::trace!(self.logger, "Disarming overseer worker");
```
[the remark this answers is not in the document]
samdealy: Done.

fog/overseer/server/src/service.rs
```diff
@@ -0,0 +1,80 @@
...
+        self.stop_requested.store(false, Ordering::SeqCst);
```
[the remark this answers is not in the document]
samdealy: Fixed in later commit. Done.

fog/overseer/server/src/worker.rs
samdealy: Done.

fog/overseer/server/tests/inactive_outstanding_key_idle_does_not_have_key.rs
samdealy: Changed to 8080 and above.

fog/overseer/server/src/service.rs
```diff
@@ -56,7 +54,7 @@ where
             self.recovery_db.clone(),
             self.logger.clone(),
             self.is_enabled.clone(),
-            self.stop_requested.clone(),
+            Arc::new(AtomicBool::new(false)),
```
eranrund, line 57: Why pass this as an argument? It's an internal implementation detail of `OverseerWorker`

samdealy: Agreed. Done.

samdealy:
> I see, that makes sense. Does `Self::NUMBER_OF_TRIES as usize` doesn't work?

It does, and that's the way I had it written earlier. I find it a little bit cleaner to just write the `try_into` rather than the as usize during the subtraction. Happy to change it though, especially if it is more idiomatic that way.

samdealy:
@eranrund Github isn't letting me reply directly to the "use_delimiter" comment of yours, so will do so here

I'd prefer to use a delimiter because I think it's easier / less typing than putting` --ingest-cluster-uris` before each cluster uri. However, I think it's reasonable to use a delimiter if only because that's the convention in Fog. Happy to change it if you feel strongly about it.

eranrund:
> @eranrund Github isn't letting me reply directly to the "use_delimiter" comment of yours, so will do so here
> 
> I'd prefer to use a delimiter because I think it's easier / less typing than putting` --ingest-cluster-uris` before each cluster uri. However, I think it's reasonable to use a delimiter if only because that's the convention in Fog. Happy to change it if you feel strongly about it.

I dont feel strongly about it so whatever works for you.

samdealy:
> And potentially include a status API endpoint that indicates if it is currently started...

Done

eranrund:
@samdealy if you don't mind, I'm going to wait with my review until CI is green. Can you please ping me when its happy?

fog/overseer/server/src/service.rs
```diff
@@ -0,0 +1,90 @@
...
+    pub fn enable(&self) -> Result<String, String> {
+        log::info!(self.logger, "Enabling overseer worker");
+        self.is_enabled.store(true, Ordering::SeqCst);
+        Ok(String::from("Fog Overseer was successfully enabled."))
+    }
+
+    pub fn disable(&self) -> Result<String, String> {
+        log::info!(self.logger, "Disabling overseer worker");
+        self.is_enabled.store(false, Ordering::SeqCst);
+        Ok(String::from("Fog Overseer was successfully disabled."))
```
garbageslam, line 78: It might be nice to make it like,

```
let old_value = self.is_enabled.swap(false, ...);
if old_value { "Fog Overseer was successfully disabled" } else { "Fog Overseer was already disabled" }
```

might not be super important but it might help if the user is having trouble using it

fog/overseer/server/src/worker.rs
```diff
@@ -0,0 +1,544 @@
...
+                    // keys are active at once. This is
+                    // unexpected.
+                }
+            }
+        }
+    }
+
+    /// Returns the latest round of ingest summaries for each
+    /// FogIngestGrpcClient that communicates with a node that is online.
+    fn retrieve_ingest_summary_node_mappings(&self) -> Vec<IngestSummaryNodeMapping> {
```
garbageslam, line 240: I think this should return the error, so that the caller can abort this pass if some of the nodes are not reachable / not responding?

fog/overseer/server/src/worker.rs
```diff
@@ -0,0 +1,544 @@
...
+                break;
+            }
+
+            if !self.is_enabled.load(Ordering::SeqCst) {
+                log::trace!(self.logger, "Overseer worker is currently disabled.");
+                continue;
+            }
+
+            let ingest_summary_node_mappings: Vec<IngestSummaryNodeMapping> =
+                self.retrieve_ingest_summary_node_mappings();
```
garbageslam, line 176: I would want to make it so that if some nodes are unreachable, we back off until the next polling cycle -- so this would check if any node gave an error, then we log it, and `continue` the loop

fog/overseer/server/src/worker.rs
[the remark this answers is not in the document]
garbageslam: Why do we repeat this `if` test, can't we fold it into the earlier `if` ?

fog/overseer/server/src/worker.rs
garbageslam: if any of the nodes are not reachable then the overseer should not do anything, otherwise if the network partitions bad things can happen

fog/overseer/server/src/worker.rs
```diff
@@ -0,0 +1,544 @@
...
+                        self.logger,
+                        "A node was activated, so there is no need to report the key as lost."
+                    );
+                    return Ok(());
+                }
+                Err(err) => log::error!(
+                    self.logger,
+                    "There was an issue activating a node for the outstanding key: {:?}",
+                    err
+                ),
```
garbageslam, line 324: i think perhaps we should not fall through here?

fog/overseer/server/src/worker.rs
```diff
@@ -0,0 +1,544 @@
...
+            1 => log::info!(self.logger, "Found 1 outstanding key."),
+            _ => {
+                log::error!(self.logger, "Found multiple outstanding keys. This requires manual intervention. Disabling overseer.");
+                self.is_enabled.store(false, Ordering::SeqCst);
+                return Err(OverseerError::MultipleOutstandingKeys("Multiple outstanding keys found. This is unexpected and requires manual intervention. As such, we've disabled overseer. Take action and then enable overseer.".to_string()));
+            }
+        }
+
+        if let Some(inactive_outstanding_key) = inactive_outstanding_keys.get(0) {
+            match self.try_to_activate_node_for_inactive_outstanding_key(
```
garbageslam, line 309: I think this `try_to_activate_node` function is a little too busy -- there are errors that can occur if contacting a node fails, and errors that can occur if finding the key fails it seems. This seems bad for clear error handling.

What I would suggest is that we try to move more logic from that function to this function, and possibly eliminate that function:

* You could iterate across the `ingest_summary_node_mappings`, using `iter().find()`, or a for loop, whatever you think is more clear
* Try to activate the first node that has the desired key. If that fails, return an error, and we'll re-evaluate everything and try again in the next pass
* If no node has the desired key, then we report a lost key, and try to activate the first node

In general any errors that we get when talking to ingest should not be just "log and fall through", they should cause us to abort, and return an error to the caller, who logs it. This will make it much easier to figure out in what circumstances the overseer may attempt to activate something. We want that circumstance to be as simple and as restrictive as possible. It should only do this when we are as certain as we can be that this is the right thing to do.

fog/overseer/server/src/worker.rs
```diff
@@ -0,0 +1,544 @@
...
+                // active key as lost. But then the previously active server
+                // could be restarted by Kubernetes, have this lost key, and
+                // restart in the active state. This would be bad because
+                // we don't want a server to be active with a lost key.
+                //
+                // Since we always expect the node to come back in some
+                // time, we should just keep trying here.
+                //
+                // TODO: Add alerting if this is taking too long.
+                loop {
```
garbageslam, line 414: I think it would be better not to do this in a loop, we could just return the error to the caller?

Because the information we used to decide to activate this node could become stale -- what if the operator is also doing something in background for example. If it is in this loop for 20 minutes then the overseer might be attempting to do something that no longer makes sense

fog/overseer/server/src/worker.rs
garbageslam: I think the server already checks this basically -- it won't allow you to call new_keys if the ingest server is in the active state. I think that's the right layer to guard against this.

That being said, it probably can't hurt to put an assert here though. But I don't think it's essential.

garbageslam (changes requested):
This looks pretty good but there are some things that should change:

* We want the overseer to be very conservative about when it decides to activate things. But there are some conditions that it doesn't handle well right now. If the network partitions and some of the ingests are not reachable, it logs those errors but it doesn't stop it from going forward and trying to compute the number of active nodes. That's dangerous because if you can't reach any of the servers then you don't actually know the number of active nodes.
* The `try_to_activate_node_for_inactive_outstanding_key` function is too complicated, it's returning basically two very different types of errors that have to be handled differently. One of them is that the key we need doesn't exist anymore, and one of them is that, asking the server that had it to activate has failed. These errors need to be handled differently by the caller but the current code isn't doing that. I think this code would be simpler if we essentially inlined this function into the caller and then handled these errors correctly.
* There are several places where we handle errors by logging them and falling through. I think this is wrong in all the places that I saw, and it makes it too complicated to figure out in what cases the overseer will activate something. All these places should be fixed by returning an error to the caller.
* I think the infinite loop where we try to activate something forever is also wrong. We could retry it 2 or 3 times, but after that, we should return an error to caller, who logs it, and re-evaluate the whole situation in the next cycle.

fog/overseer/server/src/service.rs
samdealy: Done.

fog/overseer/server/src/worker.rs
samdealy: Yes good point. Originally, I was considering the case in which one of the nodes is continually unresponsive. In this scenario, there could be no active nodes and this unresponsive node could be an idle node. In this case, we'd want overseer to respond, but it wouldn't because of this one unresponsive node.

However, I agree with you that it's too dangerous to proceed with the logic if a node is unresponsive because it could be the active node and then come back online and there'd be two active nodes (a second one would be activated by overseer in the interim). 

I also think that because we're assuming that the unresponsive node will become responsive soon enough (i.e. Azure will bring the node back online in a relatively reasonable time frame), then the probability of this "infinite loop of inaction" is pretty low. 

I changed the method.

fog/overseer/server/src/worker.rs
samdealy: Done.

fog/overseer/server/src/worker.rs
samdealy: Agreed. As mentioned in an above comment, I moved this logic to the `perform_automatic_failover `method, and removed this loop. A failure to activate the node results in an error being propagated to the caller and no activation logic proceeds.

fog/overseer/server/src/worker.rs
```diff
@@ -0,0 +1,546 @@
...
+            .collect())
+    }
+
+    /// Tries to activate a node with the given inactive outstanding key.
+    fn try_to_activate_node_for_inactive_outstanding_key(
+        &self,
+        inactive_outstanding_key: &CompressedRistrettoPublic,
+        ingest_summary_node_mappings: &[IngestSummaryNodeMapping],
+    ) -> Result<(), OverseerError> {
+        log::debug!(
```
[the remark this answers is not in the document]
samdealy: Done.

fog/overseer/server/src/worker.rs
```diff
@@ -0,0 +1,546 @@
...
+                ),
+            }
+        }
+
+        // If there's an outstanding key, we need to report it as lost.
+        if let Some(inactive_outstanding_key) = inactive_outstanding_keys.get(0) {
+            self.report_lost_ingress_key(*inactive_outstanding_key)?;
+        }
+
+        // Regardless of whether or not there's an outstanding key, we need
```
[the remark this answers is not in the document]
samdealy: Good point. I changed the behavior to now return early if the key activation fails. So this code won't be reached in that case anymore.

fog/overseer/server/src/worker.rs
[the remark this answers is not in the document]
samdealy: Good point. Done.

fog/overseer/server/src/worker.rs
samdealy: Agreed. Changed.

fog/overseer/server/src/worker.rs
samdealy: Agreed. Thank you for expressing this so clearly. Implemented it as you described.

fog/overseer/server/src/worker.rs
samdealy: Yeah, agreed that the server layer is the right layer to check this. I'll leave as is for now, but if we decide to change in the future, will do.

samdealy:
> This looks pretty good but there are some things that should change:
> 
> * We want the overseer to be very conservative about when it decides to activate things. But there are some conditions that it doesn't handle well right now. If the network partitions and some of the ingests are not reachable, it logs those errors but it doesn't stop it from going forward and trying to compute the number of active nodes. That's dangerous because if you can't reach any of the servers then you don't actually know the number of active nodes.
> * The `try_to_activate_node_for_inactive_outstanding_key` function is too complicated, it's returning basically two very different types of errors that have to be handled differently. One of them is that the key we need doesn't exist anymore, and one of them is that, asking the server that had it to activate has failed. These errors need to be handled differently by the caller but the current code isn't doing that. I think this code would be simpler if we essentially inlined this function into the caller and then handled these errors correctly.
> * There are several places where we handle errors by logging them and falling through. I think this is wrong in all the places that I saw, and it makes it too complicated to figure out in what cases the overseer will activate something. All these places should be fixed by returning an error to the caller.
> * I think the infinite loop where we try to activate something forever is also wrong. We could retry it 2 or 3 times, but after that, we should return an error to caller, who logs it, and re-evaluate the whole situation in the next cycle.

Agreed with the points here. I implemented the changes listed in the above comments that address these points. I moved the `try_to_activate_node_for_inactive_oustanding_key` logic to the `perform_automatic_failover` method, and return early if a node that has the inactive outstanding key can't be activated successfully. 

I just double-checked, and I believe that this removes all of the situations in which we were logging and error and falling through (will check again on Monday). 

For the last point, I modified it to try and activate the node once, and if it doesn't work then it returns to the main thread loop. I can modify it to explicitly retry the activate call 2 to 3 times, but my sense is that it accomplishes basically the same the behavior as if we return to the loop (the loop will pause for 5 seconds according to it's sleep and then will retry activation). I'd be happy to modify it to explicilty retry 2 to 3 times though (which might make sense given that the rest of our node method calls explicitly do this too).

fog/overseer/server/src/lib.rs
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
+#![feature(proc_macro_hygiene, decl_macro)]
```
garbageslam, line 3: do we really need nightly features for this?

can we avoid using unstable features if they aren't needed?

samdealy: Yes, we need these to enable [rocket_codegen](https://docs.rs/rocket_codegen/latest/rocket_codegen/), which provides the macros for `#[post(...)] `and `#[get(...)]`. [Full Service](https://github.com/mobilecoinofficial/full-service/blob/main/full-service/src/lib.rs#L5) uses these declarations as well.

fog/overseer/server/src/config.rs
```diff
@@ -0,0 +1,51 @@
...
+    #[test]
+    fn ingest_server_config_example() {
+        let config = OverseerConfig::from_iter_safe(&[
+            "/usr/bin/start_overseer_server",
+            "--listen-host",
+            "www.mycoolhost.com",
+            "--listen-port",
+            "8080",
+            "--ingest-cluster-uris",
+            "insecure-fog-ingest://0.0.0.0.3226/,insecure-fog-ingest://0.0.0.0.3227/",
```
eranrund, line 45: This test is incomplete - notice how the port number is not delimited by `:` but the test did not actually catch that.

fog/overseer/server/src/service.rs
```diff
@@ -0,0 +1,102 @@
...
+            "Fog Overseer was successfully enabled"
+        };
+
+        Ok(String::from(response_message))
+    }
+
+    pub fn disable(&self) -> Result<String, String> {
+        log::info!(self.logger, "Disabling overseer worker");
+        let was_enabled = self.is_enabled.swap(false, Ordering::SeqCst);
+        let response_message = if was_enabled {
```
eranrund, line 84: Nit: a match statement like in `get_status` would be a bit shorter and easier to read (in my opinion)

fog/overseer/server/src/service.rs
```diff
@@ -0,0 +1,102 @@
...
+        // This blocks on teardown of overseer_worker
+        self.overseer_worker = None;
+
+        Ok(())
+    }
+
+    pub fn enable(&self) -> Result<String, String> {
+        log::info!(self.logger, "Enabling overseer worker");
+        let was_enabled = self.is_enabled.swap(true, Ordering::SeqCst);
+        let response_message = if was_enabled {
```
eranrund, line 72: Nit: a match statement like in `get_status` would be a bit shorter and easier to read (in my opinion)

fog/overseer/server/src/worker.rs
```diff
@@ -0,0 +1,511 @@
...
+    pub fn new<DB: RecoveryDb + Clone + Send + Sync + 'static>(
+        ingest_cluster_uris: Vec<FogIngestUri>,
+        recovery_db: DB,
+        logger: Logger,
+        is_enabled: Arc<AtomicBool>,
+    ) -> Self
+    where
+        OverseerError: From<DB::Error>,
+    {
+        let thread_is_enabled = is_enabled;
```
eranrund, line 52: Nit: I dont think this is needed, can just use `is_enabled` inside the thread

fog/overseer/server/src/worker.rs
```diff
@@ -0,0 +1,511 @@
...
+        let inactive_outstanding_keys: Vec<CompressedRistrettoPublic> =
+            self.get_inactive_outstanding_keys()?;
+
+        match inactive_outstanding_keys.len() {
+            0 => log::info!(self.logger, "Found 0 outstanding keys."),
+            1 => log::info!(self.logger, "Found 1 outstanding key."),
+            _ => {
+                log::error!(self.logger, "Found multiple outstanding keys. This requires manual intervention. Disabling overseer.");
+                self.is_enabled.store(false, Ordering::SeqCst);
+                return Err(OverseerError::MultipleOutstandingKeys("Multiple outstanding keys found. This is unexpected and requires manual intervention. As such, we've disabled overseer. Take action and then enable overseer.".to_string()));
```
eranrund, line 314: Nit: Maybe be worth including the keys?

eranrund: Another nit is I've seen a mixture of `.to_string()`, `.to_owned()` and `String::from()`. They all do the same, I think `to_owned()` or `to_string()` are both used in the codebase, I would pick one and stick to it.

fog/overseer/server/src/worker.rs
```diff
@@ -0,0 +1,511 @@
...
+                            log::info!(
+                                self.logger,
+                                "Successfully activated node {}.",
+                                node.get_uri()
+                            );
+                            return Ok(());
+                        }
+                        Err(_) => {
+                            let error_message =
+                                format!("Tried activating node {}, but it failed.", node.get_uri());
```
eranrund, line 346:
```suggestion
                        Err(err) => {
                            let error_message =
                                format!("Tried activating node {}, but it failed: {}", node.get_uri(), err);
```

(not tested)

fog/overseer/server/src/worker.rs
```diff
@@ -0,0 +1,511 @@
...
+            0 => log::info!(self.logger, "Found 0 outstanding keys."),
+            1 => log::info!(self.logger, "Found 1 outstanding key."),
+            _ => {
+                log::error!(self.logger, "Found multiple outstanding keys. This requires manual intervention. Disabling overseer.");
+                self.is_enabled.store(false, Ordering::SeqCst);
+                return Err(OverseerError::MultipleOutstandingKeys("Multiple outstanding keys found. This is unexpected and requires manual intervention. As such, we've disabled overseer. Take action and then enable overseer.".to_string()));
+            }
+        }
+
+        if let Some(inactive_outstanding_key) = inactive_outstanding_keys.get(0) {
```
eranrund, line 318: Nit: I personally am finding this trick a bit confusing. I would change the match statement above to do `        let activated_node_index = self.set_new_key_on_a_node()?; self.activate_a_node(activated_node_index)?;` on the zero case, and then have the rest of the logic under the `1` case. I understand it will result in repeating the two lines mentioned here but I think overall it will be easier to follow.

fog/overseer/server/src/worker.rs
```diff
@@ -0,0 +1,511 @@
...
+                        "The following key was successfully reported as lost: {}",
+                        inactive_outstanding_key
+                    );
+                    OperationResult::Ok(())
+                }
+                Err(_) => {
+                    let number_of_remaining_tries = Self::NUMBER_OF_TRIES - current_try;
+                    let error_message = match number_of_remaining_tries {
+                        0 => format!("Did not succeed in reporting lost ingress key {} within {} tries.", inactive_outstanding_key, Self::NUMBER_OF_TRIES),
+                        _ => format!("The following key  was not successfully reported as lost: {}. Will try {} more times", inactive_outstanding_key, number_of_remaining_tries),
```
eranrund, line 427:
```suggestion
                        _ => format!("The following key was not successfully reported as lost: {}. Will try {} more times", inactive_outstanding_key, number_of_remaining_tries),
```

eranrund: Might be worth to also include the error in the formatting.

eranrund (approved):
Bunch of small nits but nothing that I feel too strongly about. Thank you!a

fog/overseer/server/src/config.rs
samdealy: Added a test to address this.

fog/overseer/server/src/service.rs
samdealy: Agreed. Done.

fog/overseer/server/src/service.rs
samdealy: Agreed. Done.

fog/overseer/server/src/worker.rs
samdealy: I included the keys and standardized on to_string.

fog/overseer/server/src/worker.rs
samdealy: Done.

fog/overseer/server/src/worker.rs
samdealy: I agree that this is confusing. I implemented as you mentioned and pulled out the one outstanding key logic into its own method. Thank you for the suggestion!

fog/overseer/server/src/worker.rs
samdealy: Done. Added underlying error message to retry error messages.

fog/overseer/server/src/worker.rs
eranrund: Nice, that looks great. Thank you!

fog/overseer/server/src/lib.rs
garbageslam: ok

fog/overseer/server/src/lib.rs
```diff
@@ -0,0 +1,10 @@
...
+
```
garbageslam, line 4: nit: This crate should have `#[deny(missing_docs)]`, ideally all our crates would have that

samdealy: Done.

fog/overseer/server/src/error.rs
```diff
@@ -0,0 +1,52 @@
...
+    SetNewKey(String),
+
+    /// Activating an idle node failed: {0}
+    ActivateNode(String),
+
+    /// There are multiple outstanding keys: {0}
+    MultipleOutstandingKeys(String),
+
+    /// Unknown error associated with retries. Should not happen: {0}
+    GenericRetryError(String),
```
garbageslam, line 32: I think we probably don't need this? The internal retry errors actually never happen:

https://github.com/jimmycuadra/retry/issues/38

fog/overseer/server/src/error.rs
```diff
@@ -0,0 +1,52 @@
...
+
+impl From<RetryError<OverseerError>> for OverseerError {
+    fn from(src: RetryError<OverseerError>) -> Self {
+        match src {
+            RetryError::Operation {
+                error,
+                total_delay: _,
+                tries: _,
+            } => error,
+            RetryError::Internal(s) => Self::GenericRetryError(s),
```
garbageslam, line 49: i think its okay to just panic in this case

fog/overseer/server/src/worker.rs
```diff
@@ -0,0 +1,543 @@
...
+                    ingest_summary_node_mappings.push(IngestSummaryNodeMapping {
+                        node_index: ingest_client_index,
+                        ingest_summary,
+                    });
+                }
+                Err(_) => {
+                    let error_message = format!(
+                        "Unable to retrieve ingest summary for node: {}",
+                        ingest_client.get_uri()
+                    );
```
garbageslam, line 275: it would be better to capture the actual error and format it into this string, so like:

```
                Err(err) => {
                    let error_message = format!(
                        "Unable to retrieve ingest summary for node ({}): {}",
                        ingest_client.get_uri(),
                        err
                    );
```

or something like this

garbageslam (approved):
This all looks good to me, thank you! I had only minor comments

fog/overseer/server/src/error.rs
samdealy: Done.

fog/overseer/server/src/error.rs
samdealy: Done. Thanks for the context!

fog/overseer/server/src/worker.rs
samdealy: Done.

fog/overseer/server/src/config.rs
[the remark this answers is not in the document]
samdealy: @joekottke this is the comment I mentioned offline.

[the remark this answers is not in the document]
joekottke: @garbageslam @samdealy We can very easily map in a config into the running environment.  It will also follow a given running instance around even if k8s moves the running instance to another machine.  The file effectively gets mapped in as an external volume.  Another nice effect of this is that we can _change_ the running config file, k8s will update that file on the running instance(s) (after a few seconds as changes get pushed out).  The on-disk semantics of this are handled through symlinks:  The running config is mapped via a symlink to a file, and then the new file gets mapped in, and the symlink gets updated.  One caveat: if the 'config file watcher' is only watching a specifc on-disk inode (after following the symlink reference) then it _might not_ notice the updated file (because it is at a new inode).  Just something to keep in mind.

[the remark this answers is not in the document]
joekottke: Some other things to keep in mind:
* k8s will have to reference the same config file contents for both the active and inactive nodes, at least the way we have things deployed now.
* The config file is read-only

fog/overseer/server/src/worker.rs
```diff
@@ -0,0 +1,544 @@
...
+                    .filter(|ingest_summary_node_mapping| {
+                        ingest_summary_node_mapping.ingest_summary.mode
+                            == IngestControllerMode::Active
+                    })
+                    .collect();
+
+            let active_node_count = active_ingest_summary_node_mappings.len();
+            match active_node_count {
+                0 => {
+                    log::error!(
```
jcape, line 206: Recommend `warn!` here, at best. Situation is abnormal, but no human intervention is required (yet).

fog/overseer/server/src/worker.rs
```diff
@@ -0,0 +1,544 @@
...
+impl<DB: RecoveryDb> OverseerWorkerThread<DB>
+where
+    OverseerError: From<DB::Error>,
+{
+    /// Poll the Fog Ingest cluster every 5 seconds.
+    const POLLING_FREQUENCY: Duration = Duration::from_secs(5);
+
+    /// Try a request to Fog Ingest node this many times if you encounter an
+    /// error.
+    const NUMBER_OF_TRIES: u64 = 3;
```
jcape, line 148: Is there any reason for this to be a `u64` instead of a `usize`? It would let you remove all the `try_into().unwrap()` calls later...

jcape (approved):
LGTM generally, though in the interests of time, I didn't review the integration tests, and most of my comments are purely stylistic.

One suggestion I would have for this and future (large) PRs is adding a writeup of the design as a section in `README.md`---like, the code seems reasonable, but I wasn't involved in the design beyond a preliminary sketch many months ago, so there's always going to be a "fit for purpose" gap without that getting written down.

fog/ingest/server/src/server.rs
```diff
@@ -359,6 +359,19 @@ where
         self.get_ingest_summary().mode == IngestControllerMode::Active
     }
 
+    /// Ask if the server is idle.
+    /// This is a convenience wrapper used in tests.
```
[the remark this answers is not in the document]
samdealy: That sounds good to me- didn't realize you could add this to a method. I went ahead and added it to other Fog Ingest sever methods that are only meant for tests.

fog/ingest/server/src/server.rs
```diff
@@ -359,6 +359,19 @@ where
...
+    pub fn is_idle(&self) -> bool {
+        self.get_ingest_summary().mode == IngestControllerMode::Idle
+    }
+
+    /// Set new keys.
+    /// This is used in tests when it would be simpler than making an RPC
+    /// client.
```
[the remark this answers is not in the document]
samdealy: Done.

fog/overseer/server/src/worker.rs
```diff
@@ -0,0 +1,536 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
+use crate::error::OverseerError;
+
```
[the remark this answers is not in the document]
samdealy: Done.

fog/overseer/server/src/worker.rs
```diff
@@ -0,0 +1,536 @@
...
+    logger: Logger,
+}
+
+/// This associates an IngestSummary with an IngestClient. This makes it easy
+/// to query a given node based on its IngestSummary.
+struct IngestSummaryNodeMapping {
+    node_index: usize,
+
+    ingest_summary: IngestSummary,
+}
```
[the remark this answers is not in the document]
samdealy: yeah that would be more compact, and earlier I wasn't sure if I should be adding these derives to IngestSummary.

That said, I weakly prefer keeping it as is because it allows us to name what the `usize` parameter is. Without this named `node_index` field, I could foresee readers having a hard time understanding what the usize means. That said, I'd be happy to change it. LMK.

fog/overseer/server/src/lib.rs
```diff
@@ -0,0 +1,10 @@
...
+pub mod config;
+pub mod server;
+pub mod service;
```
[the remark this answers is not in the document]
samdealy: I agree. However, I think the `pub` modifier (which i think == export?) is needed here because the integration tests in the `test/` need to access these modules. If I remove `pub` for any of these modules it doesn't compile.

fog/overseer/server/src/worker.rs
```diff
@@ -0,0 +1,544 @@
...
+                    let active_node_ingress_pubkeys: Vec<&external::CompressedRistretto> =
+                        active_ingest_summary_node_mappings
+                            .iter()
+                            .map(|active_ingest_summary_node_mapping| {
+                                active_ingest_summary_node_mapping
+                                    .ingest_summary
+                                    .get_ingress_pubkey()
+                            })
+                            .collect();
+                    log::warn!(
```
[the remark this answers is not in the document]
samdealy: Yes good point. I updated this downstream in my forthcoming Sentry PR. Makes sense to just do it here. Done.

fog/overseer/server/src/worker.rs
samdealy: SGTM, done.

fog/overseer/server/src/worker.rs
samdealy: Yes, the reason is that the `current_try` parameter is `u64` (kinda weird to me that the retry library defines the types this way). 

I was talking with eran about this very concern and weakly prefer this to writing `as` when I do the subtraction to see how many calls remain (see line 523). that said, will change it if you'd like.

jcape: I'm ambivalent, but there's definitely a strong preference against `unwrap()` dating back to the early days of our codebase ;-)

I suppose the main thing is: `usize as u64` is infallible on all platforms today, whereas (obviously) `try_into().unwrap()` is not.

fog/overseer/server/src/worker.rs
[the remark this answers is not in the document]
jcape: I don't have a strong preference, how often are queries (map) vs. iterations (vec) is really the determining factor...

fog/overseer/server/src/worker.rs
samdealy: Thanks for the context. Didn't realize that last point. Implemented it as you suggested. Done.

fog/overseer/server/src/worker.rs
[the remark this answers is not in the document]
samdealy: Good point. We're doing more iterations than queries, so will leave it as such.

fog/ingest/server/src/server.rs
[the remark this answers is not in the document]
garbageslam: I think its okay, we have similar conventions in fog ingest to expose some state queries like this so that you don't need to use RPC for absolutely everything.

If we make it `cfg(feature = "under_test")` then you will have to rebuild the entire server to run the test which costs developer time. Meanwhile in the release mode binary this will get stripped out by the linker since it's not called in release mode.

garbageslam:
FWIW I also like james' request to add a README.md

fog/ingest/server/src/server.rs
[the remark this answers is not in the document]
samdealy: Good call. Can confirm that the overseer integration tests fail if you add this attribute to the methods (compiler says it can't find the methods). I'm going to take them out.

samdealy:
> FWIW I also like james' request to add a README.md that gives some kind of overview of the project and its goals

I just added a README.

samdealy:
> LGTM generally, though in the interests of time, I didn't review the integration tests, and most of my comments are purely stylistic.
> 
> One suggestion I would have for this and future (large) PRs is adding a writeup of the design as a section in `README.md`---like, the code seems reasonable, but I wasn't involved in the design beyond a preliminary sketch many months ago, so there's always going to be a "fit for purpose" gap without that getting written down.

Just added a README.

PR #1307, opened 2022-01-13
Extract dependency installation steps into a dedicated script.

remoun:
### Motivation
Make it easier to set up a Linux machine with the same steps as our `Dockerfile`.

### In this PR
Extract the package setup into a shell script, and invoke that from `Dockerfile`

Tested via `docker run` and direct invocation from a Linux VM.

### Future Work
There are other cleanups and optimizations that can be made, but this is a mostly direct copy.


docker/Dockerfile
```diff
@@ -8,134 +8,35 @@
...
-# Install go 1.16 release
-RUN wget https://golang.org/dl/go1.16.4.linux-amd64.tar.gz && \
-    mv go1.16.4.linux-amd64.tar.gz go.tgz && \
-    tar -C /usr/local -xzf go.tgz  && \
-    rm -rf go.tgz
-ENV PATH="/usr/local/go/bin:$PATH"
+# Unfortunately we cannot update ENV from the init script.
+# See https://github.com/docker/docker/issues/29110
 ENV GOPATH=/opt/go/
-ENV PATH=$PATH:$GOPATH/bin
```
garbageslam, line 85: Please test that building this and go grpc gateway in the `./mob prompt` still works after this change. We should probably keep the docker ENV directive.

docker/init_debian.sh
```diff
@@ -0,0 +1,126 @@
+#!/usr/bin/env bash
+
+# !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+# Note: When modifying this file, increment the Dockerfile-version minor version number
+# !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+#
+# This is needed for mob tool to be able to pull the right image from the farm,
+# if it isn't done, bad things will happen to local builds and CI using mob tool
+# See mob tool comments for extended discussion
```
garbageslam, line 9: Let's keep this warning in the dockerfile please, or in both places

docker/Dockerfile
garbageslam: I guess CI will test this as long as the new image is pushed

garbageslam (changes requested):
we should not remove the note about incrementing Dockerfile-version

docker/init_debian.sh
remoun: It's in both.

remoun:
> we should not remove the note about incrementing Dockerfile-version

It's in both files.

remoun:
Relatedly, when should we update the `1_24` references in `jenkins/build-pod.yaml` and `.circleci/config.yml`?

docker/init_debian.sh
```diff
@@ -0,0 +1,125 @@
...
+
+cd /tmp
+
+# Certain Installers make 'installations' easier by having a nice front-end. While this is great when you have a manual install, this becomes an issue during automated installations.
+export DEBIAN_FRONTEND=noninteractive
+
+# Install build tools and dependencies
+apt-get update -q -q
+apt-get upgrade --yes
+apt-get install --yes \
```
garbageslam, line 20: is there a good reason to remove the `--force-yes` lines?

docker/init_debian.sh
```diff
@@ -0,0 +1,125 @@
...
+# For use in CI
+# TODO: Do we need this outside CI?
+pip3 install awscli black
+
+# filebeat is used for logs when running slam scripts locally
+# via https://www.elastic.co/guide/en/beats/filebeat/current/setup-repositories.html
+wget -qO - https://artifacts.elastic.co/GPG-KEY-elasticsearch | apt-key add -
+echo "deb https://artifacts.elastic.co/packages/7.x/apt stable main" >> /etc/apt/sources.list.d/elastic-7.x.list
+apt-get update && apt-get install filebeat
+systemctl enable filebeat
```
garbageslam, line 77: it looks like you are changing the way that filebeat is installed, and configuring it to start on boot.

is there a reason to add `systemctl enable filebeat`?

it seems like this isn't helpful to most developers or to CI or to the release scripts, so we should not add this and keep it like it was?

docker/init_debian.sh
garbageslam: ok i misread, thank you

garbageslam:
> Relatedly, when should we update the 1_24 references in jenkins/build-pod.yaml and .circleci/config.yml?

The circle ci for this should be updated right away, so that we know CI wasn't broken by this
We should also update the jenkins build pod and see that deploy passes, usually we try to update the submodule internal concurrently with, or shortly after, merging stuff to master that could affect CD

garbageslam:
thanks for doing this

docker/init_debian.sh
remoun: Good catch. I removed the `enable filebeat` call.

docker/init_debian.sh
remoun: It is both deprecated and dangerous, e.g. https://serverfault.com/questions/1073792/apt-get-throwing-warning-w-force-yes-is-deprecated-use-one-of-the-options

I removed it because it was disabled in Ubuntu 20.04, which I have on the cloud VM I'm testing with.

remoun:
It seems CircleCI cannot build the new image yet, so I guess we should update that reference in a follow-up?

```
Starting container gcr.io/mobilenode-211420/builder-install:1_25
  image cache not found on this host, downloading gcr.io/mobilenode-211420/builder-install:1_25

Error response from daemon: manifest for gcr.io/mobilenode-211420/builder-install:1_25 not found: manifest unknown: Failed to fetch "1_25" from request "/v2/mobilenode-211420/builder-install/manifests/1_25".
```

garbageslam:
I think we have to follow the steps to build and push the new image, I can help you with that

garbageslam (approved):
Thank you!

jcape (approved):
LGTM

PR #1336, opened 2022-01-21
Add support for k8s readiness checks in fog view and fog ledger

garbageslam:
In kubernetes, services have both a "liveliness" status and a
"readiness" status. Servers may be live if they are running, but
they are not ready serve requests until they have loaded a bunch
of data. This well represents the fog servers, since loading data
into ORAM takes time.

https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-grpc-liveness-probe

We already have readiness checks in our k8s configs, but fog view
and fog ledger aren't exposing this information to k8s right now.
This commit makes them expose this by returning the "NOT_SERVING"
status until such point as they they think they have loaded all
(or nearly all) data into ORAM.

This modifies the way these `fog` servers are using the
`mc-util-grpc` health service. It creates a handle `ReadinessIndicator`
which can be passed around in servers so that they can signal when
they are ready.

Some parts of this are not perfect -- for `fog-ledger` it would be
better to couple the readiness check with a readiness check to `mobilecoind`,
so that we don't indicate that we are ready to receive traffic until
`mobilecoind` thinks it has synced the ledger at some point. It seems
like it is a substantial improvement that will get us most of the benefit
of implementing readiness checks though.

Note: I have not found any documentation that says that `NOT_SERVING`
is the way to indicate to k8s that we are live but not ready. This is just
a guess.


util/grpc/src/admin_server.rs
```diff
@@ -58,7 +61,14 @@ impl AdminServer {
...
     pub fn shutdown(&mut self) -> ShutdownFuture {
         self.server.shutdown()
     }
 }
+
+impl Drop for AdminServer {
+    fn drop(&mut self) {
+        block_on(self.shutdown()).expect("Could not shutdown admin server")
+    }
+}
```
garbageslam, line 74: I think it was a bug that we weren't shutting this down when it was dropped

remoun (approved):
Thanks!

util/grpc/src/health_service.rs
```diff
@@ -97,3 +103,38 @@ impl Health for HealthService {
...
+        self.ready.store(false, Ordering::SeqCst);
+    }
+
+    /// Check the status
+    pub fn is_ready(&self) -> bool {
+        self.ready.load(Ordering::SeqCst)
+    }
+
+    /// Make service health check callback
+    pub fn to_service_health_check_callback(self) -> ServiceHealthCheckCallback {
```
garbageslam, line 131: I guess I can make this more idiomatic: This can be done using `From` / `Into` traits

Hold my beer

jgreat:
Not sure when we will get to k8s 1.23 so we'll need to go down the `grpc-health-probe` route.  It looks like we already include a copy of this binary in the fog containers. :tada: 

Looking at https://github.com/grpc-ecosystem/grpc-health-probe only returns success if the response is `SERVING`.  So setting it to `NOT_SERVING` will work just fine.

jgreat (approved):

garbageslam:
@jgreat here's the thing i didn't understand: In their example:

https://github.com/grpc-ecosystem/grpc-health-probe#example-grpc-health-checking-on-kubernetes

they have a pod spec like this:

```
spec:
  containers:
  - name: server
    image: "[YOUR-DOCKER-IMAGE]"
    ports:
    - containerPort: 5000
    readinessProbe:
      exec:
        command: ["/bin/grpc_health_probe", "-addr=:5000"]
      initialDelaySeconds: 5
    livenessProbe:
      exec:
        command: ["/bin/grpc_health_probe", "-addr=:5000"]
      initialDelaySeconds: 10
```

Doesn't this mean we would fail both readiness and liveliness if we return `NOT_SERVING` ?

jgreat:
I think we can start with only doing a readiness probe and maybe a startup probe. If I understand it correctly readiness will control when traffic gets routed to the container.  

It would be nice if this grpc-health-probe tool allowed you to pick your healthy state, so we could transition states.

garbageslam:
@jgreat one possibility might be to use the exit code? https://github.com/grpc-ecosystem/grpc-health-probe#exit-codes

So like, the liveness test could be, is the exit code `0` or `4`? they say not to do that but it's probably fine?

garbageslam:
it seems like it deployed alright

fog/ledger/server/src/db_fetcher.rs
```diff
@@ -133,25 +139,51 @@ impl<DB: Ledger, E: LedgerEnclaveProxy + Clone + Send + Sync + 'static> DbFetche
                 log::info!(self.logger, "Db fetcher thread stop requested.");
```
samdealy, line 139: If we request the thread to be stopped, then should we also set the server to unready? I would guess that this isn't necessary since we normally stop this thread when we call `stop`, which seems to only be called when the ledger server that encapsulates the `db_fetcher` is dropped. 

However, I just want to make sure that we wouldn't expect Kubernetes to display a ready status for a server that is being / has just been dropped?

fog/view/server/src/db_fetcher.rs
```diff
@@ -331,6 +349,8 @@ impl<DB: RecoveryDb + Clone + Send + Sync + 'static> DbFetcherThread<DB> {
                         block_index,
                         err
                     );
+                    // We might have more work to do, we aren't sure because of the error
+                    has_more_work = true;
```
samdealy, line 353: nit: Would it make sense to rename this variable to `might_have_more_work`? Right now, the naming suggests to me that there is definitely more work.

samdealy (approved):

fog/ledger/server/src/db_fetcher.rs
garbageslam: I think the server is probably going to drop before K8s has a chance to react to the signal, because k8s works by periodically polling the health end point, and not that frequently

fog/view/server/src/db_fetcher.rs
garbageslam: ok

fog/ledger/server/src/db_fetcher.rs
```diff
@@ -133,25 +139,51 @@ impl<DB: Ledger, E: LedgerEnclaveProxy + Clone + Send + Sync + 'static> DbFetche
...
+                Err(err) => {
+                    log::error!(self.logger, "Could not get num blocks from db: {}", err);
+                }
+            };
+
             // Each call to load_block_data attempts to load one block for each known
             // invocation. We want to keep loading blocks as long as we have data to load,
             // but that could take some time which is why the loop is also gated
             // on the stop trigger in case a stop is requested during loading.
             while self.load_block_data() && !self.stop_requested.load(Ordering::SeqCst) {}
```
wjuan-mob, line 159: I wonder if there should be a check in here to potentially unready. It seems like it's possible that more data is added during this loop that would indicate that we should unready ourselves?

wjuan-mob (approved):

fog/ledger/server/src/db_fetcher.rs
garbageslam: yeah that seems possible

the unready thing is kind of a hack but i think its correct that loading into ORAM is slower than downloading data. So maybe the unready check should be in this loop

util/grpc/src/health_service.rs
```diff
@@ -97,3 +103,39 @@ impl Health for HealthService {
         ctx.spawn(resp);
     }
 }
+
+/// A "global" readiness indicator can be used when your server has an initial
+/// period in which it is unready.
+#[derive(Default, Clone)]
+pub struct ReadinessIndicator {
+    ready: Arc<AtomicBool>,
```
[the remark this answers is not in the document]
garbageslam: i guess for me, functions have names like `is_ready` that sound like a question with a yes or no answer, but booleans are like states or flags. so if the "ready" flag is true it means we are ready

i have changed this as you requested though, so the flag is `is_ready` and the function is `ready()`

eranrund (approved):
LGTM (I didn't look at past reviews, only at the current state of things and it all made sense to me). Thanks for adding all the missing docstrings.

PR #1345, opened 2022-01-25
Update cryptography dependencies

remoun:
### Motivation

There are some breaking changes in `digest`, which have cascaded to other crates so we have to update all the crates that depend on it transitively.

Builds on:
https://github.com/mobilecoinfoundation/KDFs/pull/1
https://github.com/mobilecoinfoundation/bulletproofs/pull/2
https://github.com/mobilecoinfoundation/curve25519-dalek/pull/2
https://github.com/mobilecoinfoundation/ed25519-dalek/pull/2
https://github.com/mobilecoinfoundation/x25519-dalek/pull/2
https://github.com/mobilecoinfoundation/schnorrkel/pull/2

### In this PR
* Update the forked versions we use with their respective dependency updates.
* Some mechanical updates around breaking changes.

Fixes #1089.

### Future Work
* Update other dependencies to latest.
* Dedupe deps.


Cargo.toml
```diff
@@ -179,39 +179,43 @@ codegen-units = 1
 rpath = true
 
 [patch.crates-io]
-# Fork and rename to use "OG" dalek-cryptography.
-bulletproofs-og = { git = "https://github.com/mobilecoinfoundation/bulletproofs.git", rev = "675330c754f28876dbf94fc303fe73666cf8f8f4" }
+# Fork to use latest dependencies.
+bulletproofs = { git = "https://github.com/remoun/bulletproofs.git", branch = "mobilecoin" }
```
garbageslam, line 183: This should be done in the mobilecoin fork not your personal fork

Cargo.toml
```diff
@@ -179,39 +179,43 @@ codegen-units = 1
...
 
 # This version contains iOS build fixes
 cmake = { git = "https://github.com/alexcrichton/cmake-rs", rev = "5f89f90ee5d7789832963bffdb2dcb5939e6199c" }
 
 # Fix issues with recent nightlies
-datatest = { git = "https://github.com/mobilecoinfoundation/datatest.git", rev = "dc3521f692228b846a2312e7ceece50e7664be6e" }
+datatest = { git = "https://github.com/mobilecoinfoundation/datatest.git", branch = "mobilecoin" }
```
garbageslam, line 189: Please pin to revisions rather than branches so that we can have repeatable builds
 If we cannot repeatably build the enclave then it makes remote attestation worthless. Branches are lightweight and can be changed

garbageslam (changes requested):
I did not review the changes in forks so in that sense my review is not complete. The code changes here all looked fine, just the changes in cargo toml looked problematic

* Don't use personal forks
* Pin to revisions not branches or else repeatable builds are undermined

crypto/noise/src/symmetric_state.rs
```diff
@@ -159,12 +159,12 @@ where
         &mut self,
         input_key_material: KexAlgo::Secret,
     ) -> Result<(), SymmetricError> {
-        let chaining_key_len = DigestType::OutputSize::to_usize();
-        let hash_len = DigestType::OutputSize::to_usize();
+        let chaining_key_len = DigestAlgo::OutputSize::to_usize();
+        let hash_len = DigestAlgo::OutputSize::to_usize();
```
jcape, line 163:
```suggestion
        let chaining_key_len = DigestAlgo::output_size();
        let hash_len = DigestAlgo::output_size();
```

account-keys/Cargo.toml
```diff
@@ -8,7 +8,7 @@ readme = "README.md"
 [dependencies]
 # External dependencies
 displaydoc = { version = "0.2", default-features = false }
-hkdf = { version = "0.9.0", default-features = false }
+hkdf = { version = "0.12.1-pre.0", default-features = false }
```
jcape, line 11: HKDF has no default features, so we can simplify this.
```suggestion
hkdf = "0.12.1-pre.0"
```

ledger/distribution/Cargo.toml
```diff
@@ -19,6 +19,7 @@ dirs = "4.0"
 displaydoc = "0.2"
 protobuf = "2.22.1"
 retry = "1.3"
+# TODO: Replace with https://github.com/awslabs/aws-sdk-rust when it is ready.
```
jcape, line 22: 👍🏻

peers/test-utils/src/lib.rs
```diff
@@ -22,7 +22,7 @@ use mc_util_from_random::FromRandom;
 use mc_util_uri::{ConnectionUri, ConsensusPeerUri as PeerUri};
 use rand::SeedableRng;
 use rand_hc::Hc128Rng as FixedRng;
-use sha2::{digest::Digest, Sha512Trunc256};
+use sha2::{digest::Digest, Sha512};
```
jcape, line 25: These are not the same, you want `Sha512_256`

transaction/core/Cargo.toml
```diff
@@ -10,13 +10,13 @@ aes = { version = "0.7.5", default-features = false, features = ["ctr"] }
 displaydoc = { version = "0.2", default-features = false }
 generic-array = { version = "0.14", features = ["serde", "more_lengths"] }
 hex_fmt = "0.3"
-hkdf = { version = "0.9", default-features = false }
+hkdf = { version = "0.12.1-pre.0", default-features = false }
```
jcape, line 13:
```suggestion
hkdf = "0.12.1-pre.0"
```

transaction/core/src/membership_proofs/mod.rs
```diff
@@ -12,16 +10,14 @@ use crate::{
...
-use mc_crypto_hashes::Blake2b256;
+use mc_crypto_hashes::{Blake2b256, Digest};
 use subtle::{Choice, ConditionallySelectable, ConstantTimeEq};
 
 mod errors;
-mod range;
-
 pub use errors::Error as MembershipProofError;
+
+mod range;
```
jcape, line 20: I'd like to keep the `mod`s and `pub use` grouped together (that other crate was wrong, this was correct :-)). It makes it much easier to switch to the preferred crate-based export:
```
pub use crate::{
    errors::Error as MembershipProofError,
    range::{Range, RangeError},
};
```
See also #1346

transaction/core/src/tx.rs
```diff
@@ -567,7 +564,7 @@ mod tests {
     use mc_util_from_random::FromRandom;
     use prost::Message;
     use rand::{rngs::StdRng, SeedableRng};
-    use std::convert::TryFrom;
+    use core::convert::TryFrom;
```
jcape, line 567: 👍🏻  (but also this should be sorted)

transaction/std/Cargo.toml
```diff
@@ -7,11 +7,11 @@ edition = "2018"
 [dependencies]
 # External dependencies
 displaydoc = { version = "0.2", default-features = false }
-hmac = { version = "0.11", default-features = false }
+hmac = { version = "0.12", default-features = false }
```
jcape, line 10: As with `hkdf`, there's no default-features in `hmac`.
```suggestion
hmac = "0.12"
```

jcape (changes requested):
Mostly looks good, I agree with Chris' comments as well, and there are a couple show-stoppers that have crept in.

Cargo.toml
remoun: Fair enough. I will update this PR before merging it, after those PRs are merged.

Cargo.lock
```diff
@@ -915,40 +921,29 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a2f4a431c5c9f662e1200b7c7f02c34e91361150e382089a8f2dec3ba680cbda"
 
 [[package]]
-name = "crypto-mac"
```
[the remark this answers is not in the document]
remoun: Agreed it looks weird. All our `crypto-mac` deps are due to `hmac`, and `hmac` dropped the `crypto-mac` dependency  with 0.11: https://github.com/RustCrypto/MACs/commit/e8ca014e4c565c441071edc2a204addaa14c9dea#diff-3ce764644a476d2d9ccc39be9b21191ba088c4ed7077e5303958f287d823c594

peers/test-utils/src/lib.rs
garbageslam: if this was passing tests despite getting a different hash, that is not ideal

but maybe this hash was just being used for identifiers in the peers stuff and not in the blockchain?

remoun: Yeah, the tests were passing. I'm not sure how to test the correctness of that hash.

jcape: This is in `test-utils`, so it shouldn't be ending up writing things to anything persistent, but `Sha512Trunc256`/`Sha512_256` is the correct variant to use when you're truncating the hash---IIRC it has a different IV.

jcape: Yeah, it looks like it's constructing a `TxHash` from message bytes, as opposed to actual messages.

remoun: Great, thanks!

transaction/core/src/membership_proofs/mod.rs
jcape: The reason it didn't work is it was `pub use {` instead of `pub use crate::{`

jcape (commented):
Looks good, will give formal approval once the other PRs have been merged and we've branched for 1.2.

garbageslam (approved):
approved based on reading all the dependent PRs

jcape:
It looks like this will need to be rebased.

garbageslam (approved):

transaction/core/Cargo.toml
```diff
@@ -10,13 +10,13 @@ aes = { version = "0.7.5", default-features = false, features = ["ctr"] }
 displaydoc = { version = "0.2", default-features = false }
 generic-array = { version = "0.14", features = ["serde", "more_lengths"] }
 hex_fmt = "0.3"
-hkdf = { version = "0.9", default-features = false }
+hkdf = "0.12"
```
jcape, line 13: This is fine to stay as `0.12`, because transactions are (currently) using `Hkdf<Sha512>`.

jcape (changes requested):
This mostly LGTM, the main thing is making sure we're explictly requiring `hkdf = "0.12.3"`, because we need at least 0.12.1, but there were various bugs fixed going into 0.12.3 that we need.

jcape (approved):

garbageslam (approved):
LGTM thank you

PR #1349, opened 2022-01-25
Simplify target-cpu usage

### LINKED ISSUE #580: An alternative fix for .cargo/config.toml
I noticed that in https://github.com/mobilecoinfoundation/mobilecoin/blob/24c9dcdf14440d6d285eef79dab611871646aaad/.cargo/config#L4

we hard code the architecture flags to some very particular things
I wonder if instead we could take an approach like 
https://github.com/shadowsocks/shadowsocks-rust/commit/4fd4ac5bbe29d9294b15abb422d82108e8a3a2fc
with envvars. We've already done so for things like SGX. Happy to do the work if this sounds good.

remoun:
Tested: `cargo build -v`, `cargo build --release -v` on my dev machine running Ubuntu.

I verified all the `rustc` commands include `-D warnings -C target-cpu=skylake` 

Fixes #580

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -608,7 +608,7 @@ jobs:
       <<: *default-build-environment
       OPENSSL_BIN: /usr/local/opt/openssl/bin/openssl
       SCCACHE_CACHE_SIZE: 450M
-      RUSTFLAGS: -D warnings -C target-cpu=penryn
+      RUSTFLAGS: -D warnings -C target-cpu=skylake
       CONSENSUS_ENCLAVE_CSS: /Users/distiller/project/sgx/css/src/valid.css
     steps:
       - prepare-for-build: { os: macos }
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,6 +25,7 @@ defaults:
     IAS_MODE: DEV
     SGX_MODE: SW
     RUST_BACKTRACE: "1"
+    RUSTFLAGS: -D warnings -C target-cpu=skylake
     SKIP_SLOW_TESTS: "1"
 
 executors:
@@ -608,7 +609,7 @@ jobs:
       <<: *default-build-environment
       OPENSSL_BIN: /usr/local/opt/openssl/bin/openssl
       SCCACHE_CACHE_SIZE: 450M
-      RUSTFLAGS: -D warnings -C target-cpu=skylake
+      RUSTFLAGS: -D warnings -C target-cpu=penryn
       CONSENSUS_ENCLAVE_CSS: /Users/distiller/project/sgx/css/src/valid.css
     steps:
       - prepare-for-build: { os: macos }
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,22 +1,7 @@
 [build]
 rustflags = ["-D", "warnings"]
 
-# Using 'cfg` is broken, see https://github.com/rust-lang/cargo/issues/6858
-# [target.'cfg(target_arch = "x86_64")']
-# rustflags = ["-D", "warnings", "-C", "target-cpu=skylake"]
-
-# ...so instead we list all target triples (Tier 1 64-bit platforms)
-
-[target.x86_64-unknown-linux-gnu]
-rustflags = ["-D", "warnings", "-C", "target-cpu=skylake"]
-
-[target.x86_64-pc-windows-gnu]
-rustflags = ["-D", "warnings", "-C", "target-cpu=skylake"]
-
-[target.x86_64-pc-windows-msvc]
-rustflags = ["-D", "warnings", "-C", "target-cpu=skylake"]
-
-[target.x86_64-apple-darwin]
+[target.'cfg(target_arch = "x86_64")']
 rustflags = ["-D", "warnings", "-C", "target-cpu=skylake"]
 
 [cargo-new]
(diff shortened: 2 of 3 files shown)


garbageslam:
It looks that this is failing in the build container at mc crypto rand picking up no `+rdrand` feature.
That is implied by CPU = Skylake so that suggests this fix isn't working.

Is it possible your local environment has RUSTFLAGS already set? That would explain why the build worked for you locally but not in CI

garbageslam:
in the issue you wrote this

> It seems removing the spaces makes the buildflags override work.

but i dont see that in the diff, so i was a little confused

jcape:
> but i dont see that in the diff, so i was a little confused

Same.

remoun:
It turns out the failure is due to the [enclave builder specifying the `RUSTFLAGS` env var](https://github.com/mobilecoinfoundation/mobilecoin/blob/c35bae9e312a76bd7f86d3e425d7e35baa0af79b/util/build/enclave/src/lib.rs#L249), which [pre-empts the `.cargo/config` values](https://doc.rust-lang.org/cargo/reference/config.html#buildrustflags). I couldn't figure out why the build scripts are not seeing the `target_feature` additions from `target-cpu=skylake`, so I'm inclined to abandon this approach.

remoun:
> in the issue you wrote this
> 
> > It seems removing the spaces makes the buildflags override work.
> 
> but i dont see that in the diff, so i was a little confused

... Yeah, I was wrong about this.

jcape:
OK. Given that state, it may be possible to simply remove the cargo configs, or set it to "native" by default (and simply override for release builds).

remoun:
> OK. Given that state, it may be possible to simply remove the cargo configs, or set it to "native" by default (and simply override for release builds).

This worked. Tested and added to this PR.

garbageslam:
Can we make CI also use Skylake override please?

garbageslam (approved):

garbageslam:
We need to make a note to ensure that both CD (jenkins) and the release scripts are building with the skylake override

remoun:
> We need to make a note to ensure that both CD (jenkins) and the release scripts are building with the skylake override

I'm not sure I follow. Do you mean adding `-C target-cpu=skylake` to `jenkins/Jenkinsfile`? What about the release scripts?

Updating all the release scripts seems cumbersome; that makes me wonder if this cleanup is worth it.

garbageslam:
So what I'm worried is, if we start building for CD with different architecture flags, then different instructions may or may not be available and it may not be predictive of real network performance anymore.

For the release scripts, for the same reason, we should make sure we keep building with the skylake flags, or else there could be a performance regression in prod.

The release scripts are not that big, I can take that if you want. We have some changes for the release scripts in flight anyways.
But the CD jenkins part is in internal repo.

I don't think it matters for like, clients, but for the backend services like consensus and fog we don't want to make it so CD or releases isn't building with the skylake arch anymore, because we want all those things to match and we don't want a performance regression.

.circleci/config.yml
```diff
@@ -608,7 +608,7 @@ jobs:
       <<: *default-build-environment
       OPENSSL_BIN: /usr/local/opt/openssl/bin/openssl
       SCCACHE_CACHE_SIZE: 450M
-      RUSTFLAGS: -D warnings -C target-cpu=penryn
+      RUSTFLAGS: -D warnings -C target-cpu=skylake
```
jcape, line 611: This should stay as penryn, since Mac builds are client-only, `RUSTFLAGS` needs to be added to, e.g. `build-and-lint-debug` and the test runners.

jcape (changes requested):
the `build-macos` stage is fine as-is, it's more about adding it to the `build-and-lint-debug` and `

.circleci/config.yml
remoun: Oops, I misunderstood. Added `RUSTFLAGS` to `default-build-environment`

.circleci/config.yml
```diff
@@ -25,6 +25,7 @@ defaults:
     IAS_MODE: DEV
     SGX_MODE: SW
     RUST_BACKTRACE: "1"
+    RUSTFLAGS: -D warnings -C target-cpu=skylake
```
jcape, line 28: Will this be overridden by line 612, or concatinated?

remoun: Overridden

remoun:
I tried omitting `target-cpu` from `.cargo/config` and hardcoding the list of `target-feature` flags to the enclave builder, but that broke local builds (while CI continued to work because of the `RUSTFLAGS` in `default-build-env`).

I've run out of ideas to try, so I'm inclined to abandon this PR...

remoun:
@jcape and I had discussed pinning a list of features rather than target-cpu; according to https://www.nickwilcox.com/blog/target_cpu_vs_target_feature/#tldr there is a meaningful difference between passing `target-cpu` vs a bunch of `target-feature`s

jcape:
Good find, and yeah, obviously `target-cpu=skylake` is superior for the enclave builds, so we can just use that.

remoun:
> Good find, and yeah, obviously `target-cpu=skylake` is superior for the enclave builds, so we can just use that.

IIUC we get that with the current `.cargo/config` syntax (without this PR), so I'm closing this PR.

Issue #1357, opened 2022-01-27
Handle when an ingest node has a relevant key but overseer has already activated a different node and marked missed blocks.


wjuan-mob:
When a node has an old ingest key for blocks marked as missed, we should try to recover the missed blocks in the db if possible by updating the db.

PR #1358, opened 2022-01-27
Fixups/statefilerecovery

wjuan-mob:
Soundtrack of this PR: [link to song that really fits the mood of this PR](https://www.youtube.com/watch?v=IgFwiCApH7E)

### Motivation

Better state file management to avoid losing a relevant key. Based on this PR by @garbageslam: https://github.com/mobilecoinfoundation/mobilecoin/pull/1085

### In this PR

1. Add functionality to delete the statefile (except instead of deleting, we rename it to .bak, in case it will help debug something.)
2. Delete the statefile if initializing the ingest enclave fails. This is generally going to indicate that we couldn't unseal the private key. (Enclave creation failing is a separate issue, and that could be caused by an SGX daemon not being woken up yet)
3. If we did successfully load a key, don't panic even if restoring from state file failed.
4. Besides this, we generally improved error handling around the creation of the IngestSgxEnclave object
5. Adding a initialization status to state to indicate when partially initialized for future error handling.

This resolve issue #1082

### Future Work
* [Do not handle the case where you have a valid key but overseer has already activated a new node.](https://github.com/mobilecoinfoundation/mobilecoin/issues/1357)
* Do not handle or inspect the initialization status in this PR. Will be dealt with after overseer PR.

### DIFF
--- a/fog/ingest/enclave/src/lib.rs
+++ b/fog/ingest/enclave/src/lib.rs
@@ -10,15 +10,19 @@ pub use mc_fog_ingest_enclave_api::{
     Error, IngestEnclave, IngestEnclaveInitParams, IngestEnclaveProxy, Result, SealedIngestKey,
 };
 
+use displaydoc::Display;
 use mc_attest_core::{
     IasNonce, Quote, QuoteNonce, Report, SgxError, TargetInfo, VerificationReport,
 };
 use mc_attest_enclave_api::{EnclaveMessage, PeerAuthRequest, PeerAuthResponse, PeerSession};
 use mc_attest_verifier::DEBUG_ENCLAVE;
-use mc_common::ResponderId;
+use mc_common::{
+    logger::{log, Logger},
+    ResponderId,
+};
 use mc_crypto_keys::{CompressedRistrettoPublic, RistrettoPublic, X25519Public};
 use mc_enclave_boundary::untrusted::make_variable_length_ecall;
-use mc_fog_ingest_enclave_api::{EnclaveCall, SetIngressPrivateKeyResult};
+use mc_fog_ingest_enclave_api::{EnclaveCall, Error as EnclaveError, SetIngressPrivateKeyResult};
 use mc_fog_kex_rng::KexRngPubkey;
 use mc_fog_recovery_db_iface::ETxOutRecord;
 use mc_fog_types::ingest::TxsForIngest;
@@ -31,6 +35,15 @@ use std::{path, result::Result as StdResult, sync::Arc};
 
 /// The default filename of the fog ingest's SGX enclave binary.
 pub const ENCLAVE_FILE: &str = "libingest-enclave.signed.so";
+
+/// An error that occurs when calling IngestSgxEnclave::new
+#[derive(Display, Debug)]
+pub enum NewEnclaveError {
+    /// When creating the new enclave: {0}
+    Create(SgxError),
+    /// When initializing the new enclave: {0}
+    Init(EnclaveError),
+}
 
 /// A handle to an ingest enclave, on the untrusted side
 #[derive(Clone)]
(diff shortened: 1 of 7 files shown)


fog/ingest/enclave/src/lib.rs
```diff
@@ -101,9 +119,9 @@ impl IngestSgxEnclave {
 
         sgx_enclave
             .enclave_init(params)
-            .expect("enclave_init failed");
+            .map_err(NewEnclaveError::Init)?;
```
jcape, line 122: For this kind of thing, if you implement `From<EnclaveError> for NewEnclaveError`, you can do `sgx_enclave.enclave_init(params)?;` and it will do the right thing. This is generally the preferred method for this unless there's some hairy thing where it's not always a 1:1 mapping between the error that you're getting and the one you want to return.

fog/ingest/server/src/controller.rs
```diff
@@ -159,21 +206,47 @@ where
...
+            let mut retry_seconds = 1;
+            loop {
+                match result.restore_state_from_summary(summary) {
+                    Ok(_) => {
+                        let mut state = result.get_state();
+                        state.set_initialized();
+                        break;
+                    }
+                    Err(err @ RestoreStateError::Connection(_)) => {
+                        log::error!(
```
jcape, line 225: This should be `warn!`, since it's recoverable.

fog/ingest/server/src/controller_state.rs
```diff
@@ -63,6 +80,9 @@ pub struct IngestControllerState {
     /// Whether the server is idling, actively scanning the blockchain and
     /// publishing reports, or retiring
     mode: IngestMode,
+    /// Whether the server is unintialized, partially initialized, or fully
+    /// initialized
+    status: IngestStatus,
```
garbageslam, line 85: I guess I don't see the value in adding this to the controller state

Is anything going to proceed differently if it was partially or fully initialized?

If not then this is probably dead code?

wjuan-mob: Currently no. But I think it's important to flag the state difference so that overseer can handle these cases differently in the future. It is also useful for debugging, since if a user were to attempt to activate a node, it would be useful to know that the node is only in a partially initialized state.

fog/ingest/server/src/controller_state.rs
```diff
@@ -263,6 +313,19 @@ impl IngestControllerState {
         self.update_metrics()
     }
 
+    /// Sets the current status and update relevant metrics.
+    fn set_status(&mut self, status: IngestStatus) {
+        log::info!(
+            self.logger,
+            "Status switching from {:?} to {:?}",
```
jcape, line 320: can use `Display` here

jcape (approved):
LGTM, some stylistic issues.

fog/ingest/server/src/controller_state.rs
garbageslam: > But I think it's important to flag the state difference so that overseer can handle these cases differently in the future.

I think we should punt on this, and add this state in the future when we've actually designed what the overseer might do. Otherwise this is just dead code which increases the complexity of the system and the maintenance of it.

garbageslam (changes requested):
Don't add the partially initialized state stuff

* It seems like this state should not be leaving the ingest node and be communciated to overseer or anyone else
* If more work needs to happen to get it into a healthy state, then it should do that work itself and then become fully initialized. If no one else needs to know about it then don't put it in controller state.
* Adding more and more states to the system increases its complexity, this should not be done casually and especially not when the additional state is dead code. There need to be wide-ranging discussions including all fog eng's and ops if ingest is going to have an increased number of states and behaviors.

wjuan-mob:
> Don't add the partially initialized state stuff
> 
> * It seems like this state should not be leaving the ingest node and be communciated to overseer or anyone else
> * If more work needs to happen to get it into a healthy state, then it should do that work itself and then become fully initialized. If no one else needs to know about it then don't put it in controller state.
> * Adding more and more states to the system increases its complexity, this should not be done casually and especially not when the additional state is dead code. There need to be wide-ranging discussions including all fog eng's and ops if ingest is going to have an increased number of states and behaviors. So adding this now may be premature

I agree with this after consideration. I had been initially concerned about the possibility of being in a partially initialized state. But I feel like the code has sufficient safety around the scenario where it fails to set_peers or otherwise that it is not necessary to add this code, which I agree is currently dead. It can be added in the future as needed with Fog Overseer. I think it is safe to be just set to idle with no other flags.

garbageslam (approved):
looks good, thank you!

PR #1389, opened 2022-01-28
update aead to 0.4.3 everywhere

garbageslam:
Goal is to try to get the same versions for all this stuff in and out of the enclave, before branching for release

this PR was made by doing `cargo update -p aead` in each workspace, then building and confirming it still works


remoun (approved):
LGTM, including https://github.com/RustCrypto/traits/compare/aead-v0.4.1...aead-v0.4.3

jcape (approved):

PR #1390, opened 2022-01-28
update siphasher everywhere

garbageslam:
this commit was made by doing `cargo update -p siphasher` in all
workespaces

there is a bugfix in that which is relevant to fog


remoun (approved):
LGTM

I also reviewed https://github.com/jedisct1/rust-siphash/compare/0.3.7...0.3.9

jcape (approved):

PR #1404, opened 2022-01-31
Origin/fixups/slip10

wjuan-mob:
clone of https://github.com/mobilecoinfoundation/mobilecoin/pull/1183


garbageslam:
it may be a good idea to rebase this on master

jcape:
How does this related to https://github.com/mobilecoinfoundation/mobilecoin/pull/780 ?

garbageslam:
@jcape not sure, that one is really old and I completely forgot about it.

i think there's some overlap, but I think your version has some extra stuff that probably isn't needed very urgently, like, I'm not sure we need protobuf representations of mnemonic accounts right now.

I made the first version of this PR in about 5 hours, when we were discussing having an automated client test tool to do integration testing against dev networks: https://docs.google.com/document/d/1zA5V23Q2YxHtdjR1WoXjnrM0-w943Hu63yD9idpUb4o/edit?disco=AAAAR0F4EZM

> Alex Voloshyn: (12:27 PM Nov 19) how long/difficult is it to fix this tech debt?
> Chris Beck: (12:45 PM Nov 19) i'm not really sure but we can start working on it and find out
> Chris Beck: (4:55 PM Nov 19) I have a PR here now, I'm gonna try to deploy it and see what happens: https://github.com/mobilecoinfoundation/mobilecoin/pull/1183

If this is passing tests then I think we should merge it. We can triage that older PR afterwards and port forward anything we need. It looks like it has a lot of conflicts now, so it would probably be a lot of work to make it merge-able.

garbageslam:
@wjuan-mob it looks like a bunch of ingest status enum commits got in this branch

PR #1423, opened 2022-02-01
Refactor some rct bulletproofs code for performance

cbeck88:
When we sign rct bulletproofs, we always use the `check_balance = true`
flag in prod, because that is the only public API that is exposed.

If there are `n` inputs and `m` outputs, this code causes us to do
at least `2n - 1 + 2m - 1` elliptic curve operations, since we
do one such operation to form n + m pedersen commitments, and then
add all of them together.

Since we have the values and blindings, we can add them together
in a more efficient way, just adding the underlying numbers together
and only converting to a commitment at the very end. That way we only
do 2 elliptic curve operations ever.

This may save many elliptic curve operations when there are many inputs.
Since elliptic curve operations are generally the bottleneck, this may
reduce the work that a mobile device needs to do to build a transaction.

It is likely not very noticeable because there are way more such
operations in bulletproofs, but it is still probably worth it to
minimize elliptic curve ops when its easy to do so.

For example, on fast Intel hardware, these ops take about .1 millis.
On ARM arch, it might be more like 1 millisecond to do an elliptic
curve operation. If we remove 20 elliptic curve operations, it might
translate to like .02 seconds savings. So probably still not noticeable
but it's an easy win.

### DIFF
--- a/transaction/core/src/ring_signature/rct_bulletproofs.rs
+++ b/transaction/core/src/ring_signature/rct_bulletproofs.rs
@@ -292,15 +292,23 @@ fn sign_with_balance_check<CSPRNG: RngCore + CryptoRng>(
     };
 
     if check_value_is_preserved {
-        let sum_of_output_commitments: RistrettoPoint = output_values_and_blindings
+        let sum_of_output_values: u64 = output_values_and_blindings
             .iter()
-            .map(|(value, blinding)| GENERATORS.commit(Scalar::from(*value), *blinding))
+            .map(|(value, _)| value)
             .sum();
 
-        let sum_of_pseudo_output_commitments: RistrettoPoint = pseudo_output_values_and_blindings
+        let sum_of_output_commitments =
+            GENERATORS.commit(Scalar::from(sum_of_output_values), sum_of_output_blindings);
+
+        let sum_of_pseudo_output_values: u64 = pseudo_output_values_and_blindings
             .iter()
-            .map(|(value, blinding)| GENERATORS.commit(Scalar::from(*value), *blinding))
+            .map(|(value, _)| value)
             .sum();
+
+        let sum_of_pseudo_output_commitments = GENERATORS.commit(
+            Scalar::from(sum_of_pseudo_output_values),
+            sum_of_pseudo_output_blindings,
+        );
 
         // The implicit fee output.
         let fee_commitment = GENERATORS.commit(Scalar::from(fee), *FEE_BLINDING);


garbageslam:
not sure what's wrong here, gonna investigate

mfaulk:
Good observation, Chris.

For context, the check_value_is_preserved flag allows normal transaction signing to be "fail-fast", while also allowing code within the crate to sign malformed transactions needed for unit tests.

The best thing to do is probably to add more detailed comments about why the check is there and what it is supposed to do, and then implement the check as you have suggested without building the pedersen commitments. Maybe something like this:

```
// Check that this transaction does not attempt to create or destroy value.
    // Specifically, the sum of pseudo_output_commitments must equal the sum of output commitments
    // plus the (implicit) fee commitment.
    // This check is optional, but clients are encouraged to perform it because it "fails faster" than
    // submitting a malformed transaction and waiting for the network to reject it.
    if check_value_is_preserved {
        // Sum of pseudo_output values should equal sum of output values (including fee).
        let sum_of_pseudo_output_values : u128 = pseudo_output_values_and_blindings
            .iter()
            .map(|(value, _)| *value as u128)
            .sum();

        let sum_of_output_values: u128 = output_values_and_blindings
            .iter()
            .map(|(value, _)| *value as u128)
            .sum();

        if sum_of_pseudo_output_values != sum_of_output_values + fee as u128 {
            return Err(Error::ValueNotConserved);
        }

        // Sum of pseudo_output blinding should equal sum of output blindings (including fee).
        let sum_of_pseudo_output_blindings: Scalar = pseudo_output_values_and_blindings
            .iter()
            .map(|(_, blinding)| blinding)
            .sum();

        let sum_of_output_blindings: Scalar = output_values_and_blindings
            .iter()
            .map(|(_, blinding)| blinding)
            .sum();

        if sum_of_pseudo_output_blindings != sum_of_output_blindings + *FEE_BLINDING {
            return Err(Error::ValueNotConserved);
        }
}
```

transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -292,16 +292,24 @@ fn sign_with_balance_check<CSPRNG: RngCore + CryptoRng>(
     };
 
     if check_value_is_preserved {
-        let sum_of_output_commitments: RistrettoPoint = output_values_and_blindings
+        let sum_of_output_values: u64 = output_values_and_blindings
```
mfaulk, line 295: This could experience overflow

garbageslam: yeah -- i think in the tests it does, and in the other branch, i was converting to scalar first. thanks

jcape:
Is this something that we want to get merged, or can this be closed?

cbeck88:
I think its low priority, its okay to close it but let's keep the branch

PR #1438, opened 2022-02-02
confidential asset types

garbageslam:
This PR implements the transaction math part of confidential token ids.
MCIP (still draft) is here: https://github.com/mobilecoinfoundation/mcips/pull/25

Still TODO:
- [x] Add the proof of knowledge of opening
- [x] Rebase on master
- [x] Use the mob id constant and token id types in master
- [ ] Rename Amount -> MaskedAmount (that may come in a separate PR afterwards)
- [x] More tests

---

More TODO
- [x] Remove proof of knowledge of opening
- [x] Undo breaking changes to protos
- [x] Use the block version feature test in transaction builder and validation
- [x] Make sample paykit support multiple asset balances
- [x] Make mobilecoind track token ids
- [x] Make slam and fog distro respect token ids
- [x] rebase again
- [x] Make it possible to bootstrap with token ids, and test in CD


api/proto/external.proto
```diff
@@ -224,10 +227,19 @@ message TxPrefix {
     repeated TxOut outputs = 2;
 
     // Fee paid to the foundation for this transaction
-    uint64 fee = 3;
+    fixed64 fee = 3;
```
jcape, line 230: Is this a breaking change?

transaction/core/src/amount/proof_of_opening.rs
```diff
@@ -0,0 +1,286 @@
...
+use curve25519_dalek::{
+    ristretto::{CompressedRistretto, RistrettoPoint},
+    scalar::Scalar,
+};
+use mc_crypto_digestible::{Digestible, MerlinTranscript};
+use mc_crypto_keys::CompressedRistrettoPublic;
+use prost::Message;
+use serde::{Deserialize, Serialize};
+
+/// This object is the transcript of a non-interactive "proof of opening"
```
jcape, line 21: Reading through the slides, is this better described as a proof of openability?

jcape:
As a general comment, the the `AmountData {}` struct seems really awkward to use, maybe just wrap it up entirely with `Amount`?

Like this *should* be represented as:
```
pub struct Amount<T: Token> {
    value: u64,
    _type: PhantomData<T>,
}
```

And just use the token ID when serializing/deserializing. I realize this is more complex than that, but I worry that we're opening the door to a lot of type confusion bugs by not letting the compiler enforce this kind of thing wherever possible.

garbageslam:
> I dunno, I feel like this should be represented as:

One of the design requirements is, if we ship this functionality, it should be possible to stand up new token ids in prod without signing a new enclave.

If we make AmountData a template over the token id as you describe, then there is no way the enclave can create fee outputs for tokens that didn't exist when it was signed, because those types were not part of the build at that time.. So this proposal is incompatible with the requirements. Let me know if you still have questions about this.

garbageslam:
> And just use the token ID when serializing/deserializing. I realize our uses are more complex than that, but I worry that we're opening the door to a lot of type confusion bugs by not letting the compiler enforce this kind of thing wherever possible.

I think actually, the best mechanism we have for catching those bugs is the proof of opening, which is checked both by the signer after they construct the Tx, and by the consensus verifier. Even if there is a bug or a weak API in the TransactionBuilder where tokens of different types get mixed up, build the transaction will fail because they will be unable to create a valid proof of opening.

transaction/core/src/amount/proof_of_opening.rs
garbageslam: Yeah I'm open to other names for this thing

I think "proof of opening" jibes with other things I've read?

We could also call it like "proof of generator" or "proof of pedersen base" or something like this? That's sort of more to the point, it proves that the prover built this commitment over a particular set of pedersen generators?

api/proto/external.proto
```diff
@@ -224,10 +227,19 @@ message TxPrefix {
...
+
+    // Token id for this transaction
+    fixed32 token_id = 4;
 
     // The block index at which this transaction is no longer valid.
-    uint64 tombstone_block = 4;
+    uint64 tombstone_block = 5;
```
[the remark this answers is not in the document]
garbageslam: Also intentional

api/proto/external.proto
jcape: That doesn't require changing the types or tags of existing fields.

garbageslam: I think the organization is much better if the fee is next to the token id, it makes it easier to read and understand. Is there a problem with that?

eranrund:
On the topic of breaking changes, I am wondering what the plan is for getting this deployed. If we deploy this to the network before SDKs upgrade, then they will no longer be able to submit transactions.
If we want to update SDKs first, they will need to support constructing both old and new txs and decide based on the node they connect to whether to construct v1 or v2.

It would be nice if we could somehow make the `Tx`/`TxPrefix` change non-breaking so that for MOB transactions stuff just continues to work. Not sure how feasible that is with the signature change. But if we are making a breaking change I do think we should plan for either:
1) Having Consensus support accepting both old and new tx types for a time period (which might allow us to deploy this change prior to any SDK updates?)
2) Having the SDKs support both current and new Tx construction, and figure out how they decide which version to submit.

garbageslam:
I don't know right now how we can make this not breaking. If we allow people to omit proof of opening, then they could mix token ids. So once we start having multiple token ids, the old transaction types are not valid.

We could maybe like, duplicate the transaction-core crate and have a `transaction-core v2` crate. But it would be very painful.

I think it would be simpler to make the clients figure out based on the MRENCLAVE they get during attestation whether to use the v1 transaction builder or the v2 transaction builder

garbageslam:
We could try to make the enclave depend on two different revisions of transaction core, that might avoid actually copy pasting all the code. But it's still going to be very complicated.

eranrund:
That does mean that we need to figure out how to expose two versions of the TransactionBuilder for the SDKs.

garbageslam:
@eranrund they might be able to just link an old and new version of libmobilecoin?

garbageslam:
maybe there is a better way, maybe we can accept the old Txs as long as all the inputs and outputs are missing the masked_token_id, and then skip the proof of opening.

garbageslam:
and then in a future revision, we require the masked token id and proof of opening?

that might be a better direction for this PR, gonna think on it some more

eranrund:
> @eranrund they might be able to just link an old and new version of libmobilecoin?

I would be surprised, due to symbol name collision. Might be able to do some horrible linker tricks to get around it but I am not excited about that...

I also have no idea how the Java JNI stuff would behave.

transaction/core/src/amount/mod.rs
```diff
@@ -34,31 +54,47 @@ pub struct Amount {
...
+        let generator = generators(*data.token_id);
+
+        // Pedersen commitment `v*H_i + b*G`.
+        let commitment = CompressedCommitment::new(data.value, blinding, &generator);
+
+        // The value is XORed with the first 8 bytes of the mask.
+        // `v XOR_8 Blake2B(value_mask | shared_secret)`
+        let masked_value: u64 = data.value ^ get_value_mask(shared_secret);
 
         // The value is XORed with the first 8 bytes of the mask.
```
mfaulk, line 89: This comment looks copy-pasted. Should be something like "The token_id is XORed...", no?

transaction/core/src/amount/proof_of_opening.rs
mfaulk: The best description I have so far is "a proof of knowledge of a Pedersen commitment opening". That terminology is used here, https://eprint.iacr.org/2020/061.pdf

So, ProofOfOpening seems ok to me.

transaction/core/src/encrypted_fog_hint.rs
```diff
@@ -25,15 +25,17 @@ use prost::{
...
-// The length of the encrypted fog hint field in the ledger.
-// Must be at least as large as mc_crypto_box::VersionedCryptoBox::FooterSize.
-// Footersize = 50, + 32 for one curve point, + 2 bytes of magic / padding space
-// for future needs
+/// The length of the encrypted fog hint field in the ledger.
+/// Must be at least as large as mc_crypto_box::VersionedCryptoBox::FooterSize.
+/// Footersize = 50, + 32 for one curve point, + 2 bytes of magic / padding
+/// space for future needs
 pub type EncryptedFogHintSize = U84;
+/// Lenght of encrypted fog hint as a usize
```
mfaulk, line 33: Length

transaction/core/src/lib.rs
```diff
@@ -1,7 +1,10 @@
 // Copyright (c) 2018-2021 The MobileCoin Foundation
 
+//! MobileCoin transaction data types, transaction construction and validation
+//! routines
+
 #![no_std]
-// #![deny(missing_docs)]
+#![deny(missing_docs)]
```
mfaulk, line 7: Thank you

transaction/core/src/ring_signature/mlsag.rs
```diff
@@ -416,20 +463,10 @@ mod mlsag_tests {
...
-                &ring_mlsag_parameters.ring,
-                ring_mlsag_parameters.real_index,
-                &ring_mlsag_parameters.onetime_private_key,
-                ring_mlsag_parameters.value,
-                &ring_mlsag_parameters.blinding,
-                &ring_mlsag_parameters.pseudo_output_blinding,
-                &mut rng,
-            )
-            .unwrap();
+            let signature = params.sign(&mut rng).unwrap();
```
mfaulk, line 469: Nice

transaction/core/src/ring_signature/mod.rs
```diff
@@ -22,26 +22,68 @@ mod key_image;
...
+    let mut hasher = Blake2b::new();
+    hasher.update(&HASH_TO_POINT_DOMAIN_TAG);
+
+    // This step xors the token id bytes on top of the "base point" bytes
+    // used prior to the introduction of token ids.
+    //
+    // This ensures:
+    // * The function is constant-time with respect to token id
+    // * The behavior for id 0 is the same as before
+    // * For different id values, the set of B points are orthogonal.
```
mfaulk, line 54: I haven't been able to find a reference for "orthogonal" base points. I pulled up some definitions of Pedersen commitments and found a couple of good descriptions:

"Given the two generators G and H, no one knows the discrete log for second generator H with respect to first generator G, meaning no one knows a y such that y ∗ G = H"

or 

Computational soundness (binding): if the generators g, h are independently generated, then finding a discrete logarithm relation between them is as hard as breaking the Discrete Log Problem.

transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -346,32 +410,35 @@ fn sign_with_balance_check<CSPRNG: RngCore + CryptoRng>(
         ring_signatures,
         pseudo_output_commitments,
         range_proof_bytes,
+        proof_of_opening,
     })
 }
 
-/// Concatenates [message || pseudo_output_commitments || range_proof].
+/// Computes a merlin digest of message, pseudo_output_commitments, range proof,
```
mfaulk, line 417: Thanks. This looks much better.

transaction/core/src/amount/mod.rs
garbageslam: yes thank you

api/proto/external.proto
garbageslam: we're going to redesign per discussion so that this isn't a breaking change, and to the extent there is, navigate that via the block version bumping

transaction/core/src/ring_signature/rct_bulletproofs.rs
garbageslam: I'm now not sure if I should actually do this, since it's a breaking change and if we remove proof of opening this may not have to be a breaking change...

pondering it

garbageslam: I guess that since we are doing the block version thing and it's a hard cut-over anyways we might as well keep this

transaction/core/src/amount/mod.rs
```diff
@@ -81,12 +115,92 @@ impl Amount {
...
+    /// * masked_value: u64
+    /// * masked_token_id: &[u8], either 0 or 4 bytes
+    /// * shared_secret: The shared secret curve point
+    ///
+    /// Returns:
+    /// * Amount
+    /// * AmountData (token id and value)
+    /// or
+    /// * An amount error
+    pub fn reconstruct(
```
garbageslam, line 144: @voloshyn i noticed that in a lot of the SDKs, this function is basically being reimplemented from scratch using `get_value_mask` etc, i think it will be more maintainable if we try to use this version instead

libmobilecoin/src/transaction.rs
```diff
@@ -81,22 +83,22 @@ pub extern "C" fn mc_tx_out_commitment_crc32(
...
             .expect("view_private_key is not a valid RistrettoPrivate");
 
         let mut matches = false;
-        if let Ok(public_key) = RistrettoPublic::try_from_ffi(&tx_out_public_key) {
-            let shared_secret = get_tx_out_shared_secret(&view_private_key, &public_key);
-            let value = (tx_out_amount.masked_value as u64) ^ get_value_mask(&shared_secret);
-            let amount: Amount =
-                Amount::new(value, &shared_secret).expect("could not create amount object");
-            matches = amount.get_value(&shared_secret).is_ok()
+        if let Ok(_public_key) = RistrettoPublic::try_from_ffi(&tx_out_public_key) {
```
eranrund, line 96: This feels very fishy to me... is this PR the right place to change the behavior? 
Is this used by the SDKs? Maybe its better to remove it and have the SDK deal with it when they uprev?

mobilecoind/src/service.rs
```diff
@@ -559,11 +559,14 @@ impl<T: BlockchainConnection + UserTxConnection + 'static, FPR: FogPubkeyResolve
         let shared_secret =
             get_tx_out_shared_secret(account_key.view_private_key(), &tx_public_key);
 
-        let (value, _blinding) = tx_out
+        let (amount_data, _blinding) = tx_out
             .amount
             .get_value(&shared_secret)
             .map_err(|err| rpc_internal_error("amount.get_value", err, &self.logger))?;
 
+        // FIXME: We should not discard amount_data.token_id
```
eranrund, line 567: What should we do with it?

slam/src/main.rs
```diff
@@ -501,9 +500,13 @@ fn build_tx(
     let block_version = BlockVersion::try_from(BLOCK_VERSION.load(Ordering::SeqCst))
         .expect("Unsupported block version");
 
+    // Assume MOB token
+    let token_id = Mob::ID;
```
eranrund, line 504: Shouldn't it get it from the SpendableTxOuts? and panic if they are mixing multiple token ids?

slam/src/main.rs
```diff
@@ -327,22 +329,19 @@ fn main() {
...
+
+                    let value = amount_data.value;
+                    log::trace!(logger, "amount of {} is {}", tx_out.public_key, value);
 
                     // Push to queue
                     spendable_txouts_sender
                         .send(SpendableTxOut {
                             tx_out: tx_out.clone(),
-                            amount: input_amount,
+                            amount: value,
```
eranrund, line 344: Maybe a good idea to add to the token id to SpendableTxOut

slam/src/main.rs
garbageslam: yeah that seems better

mobilecoind/src/service.rs
garbageslam: I guess the `UnspentTxOut` object should store it

garbageslam: we could store an `AmountData` instead of a value to try to keep the token id with the value

transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -461,40 +552,75 @@ mod rct_bulletproofs_tests {
...
         fn sign_rejects_empty_rings(
             num_inputs in 1..8usize,
             num_mixins in 1..17usize,
             seed in any::<[u8; 32]>(),
+            block_version in 2..=3u32,
         ) {
+            let block_version: BlockVersion = block_version.try_into().unwrap();
             let mut rng: StdRng = SeedableRng::from_seed(seed);
-            let mut params = SignatureParams::random(num_inputs, num_mixins, &mut rng);
+            let mut params = SignatureParams::random(block_version,num_inputs, num_mixins, &mut rng);
```
eranrund, line 620:
```suggestion
            let mut params = SignatureParams::random(block_version, num_inputs, num_mixins, &mut rng);
```

transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -509,20 +635,14 @@ mod rct_bulletproofs_tests {
...
```
eranrund, line 642:
```suggestion
            let mut params = SignatureParams::random(block_version, num_inputs, num_mixins, &mut rng);
```

eranrund (commented):
I went over this, and it all seems good to me with the exception of a relatively long list of TODOs that we need to keep track of in the SDKs/mobilecoind/sample-paykit/etc. Are you keeping track of it in Asana or a GH issue that summarizes all the followup work needed?

Has the most recent version of this been deployed and performance looked at? There have been concerns about not caching the generators if I recall correctly.

Thank you!

transaction/core/src/ring_signature/rct_bulletproofs.rs
garbageslam: i wonder why cargo fmt isn't doing this

eranrund: I think because it's inside a macro call

libmobilecoin/src/transaction.rs
garbageslam: we chatted about this with the SDK team, the current code here is a bug. they may have fixed it and perhaps i need to rebase. will follow up with them

garbageslam: @mobilecoinfoundation/sdk ^lmk what you think about this

garbageslam:
> Are you keeping track of it in Asana or a GH issue that summarizes all the followup work needed?

I'm mostly updating the PR description with new TODO items. I need to make a list of things to happen after this in GH i guess

transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -350,7 +387,41 @@ fn sign_with_balance_check<CSPRNG: RngCore + CryptoRng>(
     })
 }
 
+/// Toggles between old-style and new-style extended message
```
garbageslam, line 390: @mfaulk @isis-mc lmk how you guys feel about this, Matt I think you already commented positively on switching to Merlin. To keep compat though we have to use the old version until block version = 3 lands.

libmobilecoin/src/transaction.rs
garbageslam: according to Adam this function is never actually called

transaction/core/src/range_proofs/mod.rs
```diff
@@ -1,5 +1,12 @@
 // Copyright (c) 2018-2021 The MobileCoin Foundation
 
+//! Range proofs are used to prove that a set of committed values are all
+//! in a well-defined range, without revealing the values.
+//!
+//! A range proof is relative to a Pedersen Generator. If a prover can construct
+//! a range proof relative to one generator, they cannot construct a range proof
+//! relative to another generator, if those generators are orthogonal.
```
garbageslam, line 8: @mfaulk this is where i added note about the generator playing the role of the proof of opening, we could expand on this I guess

garbageslam:
Screenshot at 2022-03-02 16-01-12

garbageslam:
Screenshot at 2022-03-02 16-01-38

garbageslam:
tx/s looks about the same,
tx completion time is a bit worse, but this is also with a slam that has twice as many transactions, so that may be a reason

fog/distribution/src/main.rs
```diff
@@ -125,11 +151,13 @@ fn main() {
 
     let ledger_db = LedgerDB::open(ledger_dir.path()).expect("Could not open ledger_db");
 
+    let mut block_version = ledger_db.get_latest_block().unwrap().version;
+    if let Some(config_block_version) = config.block_version.as_ref() {
+        block_version = *config_block_version;
+    }
```
eranrund, line 157: Nit: This should not require block_version to be mut - unwrap_or_else can be used to fallback to the latest block version.

fog/distribution/tests/bootstrap.rs
```diff
@@ -0,0 +1,73 @@
...
+
+// Test that fog distro can find the spendable tx outs from the bootstrap
+#[test]
+fn test_find_spendable_tx_outs() {
+    let me = PathBuf::from(args().next().unwrap());
+    let dir = me.parent().unwrap().parent().unwrap();
+    println!("{:?}", dir);
+    set_current_dir(dir).unwrap();
+
+    assert!(Command::new("rm")
```
eranrund, line 15: This should use https://doc.rust-lang.org/std/fs/fn.remove_dir_all.html

fog/distribution/tests/bootstrap.rs
```diff
@@ -0,0 +1,73 @@
+use std::{
+    env::{args, set_current_dir},
+    path::PathBuf,
+    process::Command,
+};
+
+// Test that fog distro can find the spendable tx outs from the bootstrap
+#[test]
+fn test_find_spendable_tx_outs() {
```
eranrund, line 9: This is a pretty funky test. Is it generating all the keys and stuff in the target directory?
It should create a temporary directory that gets cleaned up after the test is done, and not rely on deleting stuff that might happen to exist in the current directory.

util/generate-sample-ledger/tests/bootstrap.rs
```diff
@@ -0,0 +1,50 @@
+use std::{
```
eranrund, line 1: Similar to the other test, this should not have to delete stuff that might've been left over from a previous run. It also doesn't do any validation that the binary actually produced anything useful, so personally I am not finding much value here.

eranrund (commented):
I went over the recent changes and made some comments. Did not spot any major blockers.
I am ready to hit approve once there are no more changes - but I would also like more than one other approval for this particular PR since it touches very critical paths. I think three reviewers who are familiar with the nature of this change would be appropriate.

Thank you for taking the time to document stuff and improve the utilities, thats super nice and will make our lives better!

fog/distribution/tests/bootstrap.rs
garbageslam: Ok we can do that

util/generate-sample-ledger/tests/bootstrap.rs
garbageslam: It is needed to have some kind of integration test in the util generate sample keys crate, otherwise cargo test will not build the binary, and then the other test will fail

garbageslam: It's testing that the binary runs without crashing -- we could do more validation, any obvious thing worth testing?

mobilecoind/src/sync.rs
```diff
@@ -384,7 +384,7 @@ fn match_tx_outs_into_utxos(
         let shared_secret =
             get_tx_out_shared_secret(account_key.view_private_key(), &tx_public_key);
 
-        let (value, _blinding) = tx_out
+        let (amount, _blinding) = tx_out
```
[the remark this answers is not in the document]
garbageslam: I was going to try to do it by filtering at each call like that, but maybe the api is too large to do it that way sanely. I'm just not sure what happens if we filter when adding things to database, then they restart the program with a different token id. Probably they won't compute the right balance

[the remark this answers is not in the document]
garbageslam: They would have to delete the ledger and resync I guess

mobilecoind/src/service.rs
```diff
@@ -322,6 +329,14 @@ impl<T: BlockchainConnection + UserTxConnection + 'static, FPR: FogPubkeyResolve
                 rpc_internal_error("mobilecoind_db.get_utxos_for_subaddress", err, &self.logger)
             })?;
 
+        // Filter out those that don't have the right token id
+        // Note: This is meant to avoid breaking changes for users who are using
+        // mobilecoind with just once token id
```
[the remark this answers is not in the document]
garbageslam: They could restart the program with a different token id I think

fog/distribution/tests/bootstrap.rs
garbageslam: I added this test to try to reduce my iteration times in CD

util/generate-sample-ledger/tests/bootstrap.rs
garbageslam: I guess we could also merge these crates into one

eranrund: Relying on having a unit test just so that the binary gets built and can be used by a unit test in another crate is exactly what I would call funky...
If we want to do this, the right approach is to move the functionality into a library, and call that through a `dev-dependencies` dependency.

mobilecoind/src/sync.rs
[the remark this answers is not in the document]
eranrund: Good point on restarting with a different token id. Okay, with that in mind I think it does make sense to to store all the matched txouts, and have the API filter out to the configured token id (until we add proper support for multiple token ids from a single mobilecoind).

fog/distribution/tests/bootstrap.rs
eranrund: At the very least it is worth explaining the purpose of this.

util/generate-sample-ledger/tests/bootstrap.rs
```diff
@@ -0,0 +1,50 @@
...
+
+// Test that the bootstrap binary works with basic config
+#[test]
+fn test_exercise_bootstrap() {
+    let me = PathBuf::from(args().next().unwrap());
+    let dir = me.parent().unwrap().parent().unwrap();
+    println!("{:?}", dir);
+    set_current_dir(dir).unwrap();
+
+    assert!(Command::new("rm")
```
jcape, line 15: Same comment re: future introduction of `remove_dir_all` for these.

jcape (commented):
Mostly LGTM, there are a few minor things noted in the comments.

transaction/core/src/range_proofs/mod.rs
```diff
@@ -21,13 +28,16 @@ use error::Error;
 /// # Arguments
 /// `values` - Secret values that we want to prove are in [0,2^64).
 /// `blindings` - Pedersen commitment blinding for each value.
+/// `pedersen_generators` - Pedersen generators on which the commitments are
+/// based `rng` - randomness
```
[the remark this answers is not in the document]
garbageslam: i think the formatter reflowed my comments, will try to figure out how to get it not to do that

api/proto/external.proto
```diff
@@ -127,6 +127,14 @@ message ViewKey {
 // `trasaction/core` crate
 ///////////////////////////////////////////////////////////////////////////////
 
+/// A list of "known" token id's and their names.
+///
+/// Note that this is not an exhaustive list and clients should gracefully handle
+/// the scenario that they receive a tx out with a token id they don't know about yet.
```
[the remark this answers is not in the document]
garbageslam: i can probably make some kind of test for that, i think its possible to enumerate over the protobuf enum

fog/sample-paykit/src/cached_tx_data/mod.rs
```diff
@@ -853,10 +856,10 @@ pub struct OwnedTxOut {
     /// The tx_out that we recovered from the view server, or from view-key
     /// scanning a missed block.
     pub tx_out: TxOut,
-    /// The value of the TxOut, computed when we matched this tx_out
+    /// The value of the TxOut, computed when we matched this tx out
     /// successfully against our account key.
-    pub value: u64,
-    // The subaddress index this tx_out was sent to.
+    pub amount: AmountData,
```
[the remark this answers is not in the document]
garbageslam: i guess i'm anticipating that Eran wants to rename `AmountData -> Amount` and `Amount -> MaskedAmount`

mobilecoind/api/proto/mobilecoind_api.proto
```diff
@@ -126,6 +126,9 @@ message UnspentTxOut {
     // The tombstone block used when we attempted to spend the UTXO.
     uint64 attempted_spend_tombstone = 6;
 
+    // The token id of the TxOut
+    uint32 token_id = 7;
```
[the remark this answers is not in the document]
garbageslam: i don't think so, i think the field #10 is not actually serialized

util/generate-sample-ledger/tests/bootstrap.rs
garbageslam: i wish that cargo would give me a way to depend on binaries from other crates :(

i really prefer testing the actual binary if i can because the whole point of this is to try to catch issues before waiting for a CD deploy cycle

garbageslam: maybe this just shouldn't be a cargo test if cargo can't accomodate this, maybe should build the binaries and then just run a script

mobilecoind/api/proto/mobilecoind_api.proto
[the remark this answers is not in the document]
jcape: Yeah, my concern is about the jump from tag 7 to tag 10... it feels like there were tags 7-9 at some point in the past, and they were removed but not documented.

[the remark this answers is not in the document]
garbageslam: i see, so maybe this should be #11 ? I can look in git history.

thanks

libmobilecoin/src/transaction.rs
```diff
@@ -192,11 +194,11 @@ pub extern "C" fn mc_tx_out_get_value(
...
 
         let shared_secret = get_tx_out_shared_secret(&view_private_key, &tx_out_public_key);
-        let value = (tx_out_amount.masked_value as u64) ^ get_value_mask(&shared_secret);
-        let amount: Amount = Amount::new(value, &shared_secret)?;
-        let (val, _blinding) = amount.get_value(&shared_secret)?;
+        let (_amount, amount_data) =
+            Amount::reconstruct(tx_out_amount.masked_value, &[], &shared_secret)?;
 
-        *out_value.into_mut() = val;
+        // FIXME: This should also return the amount_data.token_id
```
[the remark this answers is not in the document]
garbageslam: So, I agree, but I don't maintain the libmobilecoin API, and usually when i make changes that impact libmobilecoin, my goal is, do not change the libmobilecoin header and do not break their build. Instead just leave comments that would help Adam adjust it when he gets to it. I can make a github ticket for him.

isis-mc (approved):
I didn't look at the changes to the test suite code super closely, but the changes to the transaction maths look fine to me.

transaction/core/src/amount/commitment.rs
```diff
@@ -1,5 +1,5 @@
 use crate::{
-    ring_signature::{Error, Scalar, GENERATORS},
+    ring_signature::{Error, PedersenGens, Scalar},
```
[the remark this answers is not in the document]
garbageslam: We are just importing this name from the Bulletproofs crate that you worked on:

```
pub use bulletproofs_og::{BulletproofGens, PedersenGens};
```

I'm inclined to just use the upstream name to avoid confusion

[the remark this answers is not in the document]
garbageslam: https://github.com/dalek-cryptography/bulletproofs/blob/1a10ce1a5b87299014658770346b376a858e7691/src/generators.rs#L30

garbageslam:
This has passed deploy (build network #46), I think we should merge it if there are no outstanding issues.

I didn't get the bootstrap with custom token ids to work, that seems to break deployment and I'm not sure why. But I think we can test that functionality end to end once minting is merged.

libmobilecoin/src/transaction.rs
garbageslam: @the-real-adammork feel free to request that I do something different, I'm just trying to avoid doing something that would break your build, and making sure that there are tickets for follow up work

eranrund: If its not called can we remove it? Having incorrect code is worse than having a breaking build in my opinion.

transaction/core/src/token.rs
```diff
@@ -34,6 +37,26 @@ impl Deref for TokenId {
...
 
+impl FromStr for TokenId {
+    type Err = ParseIntError;
+    fn from_str(src: &str) -> Result<Self, Self::Err> {
+        let src = u32::from_str(src)?;
+        Ok(TokenId(src))
+    }
+}
+
+impl PartialEq<u32> for TokenId {
```
eranrund, line 48: Nice

eranrund (approved):
LGTM, thank you for all your hard work on this.
I would prefer if this one gets more than two approvals before merging, preferably two more approvals from people intimately familiar with the codebase.

mfaulk (approved):

mfaulk:
Thank you for all of your work on this, Chris!

sugargoat (approved):
This is awesome - thanks for the great work and prototype. Will make sure the MCIP is up to date with this as well!

mfaulk (approved):

PR #1440, opened 2022-02-02
KeyImage Generation for Change Subaddress

dolanbernard:
### Motivation

Updating KeyImage generation for change subaddress

### In this PR
* KeyImage generation fix


account-keys/src/account_keys.rs
```diff
@@ -403,6 +403,11 @@ impl AccountKey {
         self.subaddress_spend_private(DEFAULT_SUBADDRESS_INDEX)
     }
 
+    /// The private spend key for the change subaddress.
+    pub fn change_subaddress_spend_private(&self) -> RistrettoPrivate {
```
samdealy, line 407: Using "change" as the first word in this method name makes it seem like it's meant to be changing the subaddress. I realize that your following the convention set by the `subaddress_spend_private` for the DEFAULT_SUBADDRESS_INDEX, but I think it might be clearer if you changed the name to `get_change_subaddress_spend_private`. That way it's clear that this method is a getter.

account-keys/src/account_keys.rs
```diff
@@ -426,6 +431,11 @@ impl AccountKey {
         self.subaddress_view_private(DEFAULT_SUBADDRESS_INDEX)
     }
 
+    /// The private view key for the change subaddress.
+    pub fn change_subaddress_view_private(&self) -> RistrettoPrivate {
```
samdealy, line 435: Same as above.

samdealy (changes requested):

account-keys/src/account_keys.rs
dolanbernard: I named it this way to be consistent with default_subaddress_spend_private. Should I update that too for consistency? Or do you think I should leave it as is?

account-keys/src/account_keys.rs
dolanbernard: Same as above.

account-keys/src/account_keys.rs
samdealy: I would prefer you update it.

samdealy: Thanks!

samdealy (approved):

jcape:
Getters should be `foo()`, not `get_foo()`: https://github.com/rust-lang/rfcs/blob/master/text/0344-conventions-galore.md#gettersetter-apis

dolanbernard:
Reverted account_keys refactor

garbageslam (approved):
LGTM

PR #1444, opened 2022-02-02
Sam/fog distro

### LINKED ISSUE #1382: Guaranteed payouts in fog distro utility
The fog-distribution tool is a tool that runs when we make a deployment in order to fund a series of test accounts with fog.

The fog-distribution tool was derived from a modified version of `slam`. We decided to do it this way like 2 years ago to help stress test the network.

The problem with this is that sometimes not all accounts get paid, and then the tests fail with zero balance errors.

The code is very ugly and hacky with lots of threads trying to slam in parallel, and it's typically hard to figure out what is actually expected to happen after fog distro runs. (i.e. how many accounts are expected to be paid how many times)

I think we don't want to completely remove the slam part because when we look in grafana, this is the easiest way that we can see tx/s at any particular revision.

I think we should make it so that before the `slam` part starts, the program deterministically pays each target account once in a single thread, then uses the rest of the outputs to do whatever it's currently doing. This should not be too much work and it will make releases easier because fog distribution will be more reliable (there shouldn't be target accounts with zero balances afterwards) and it will be easier to test the clients.

samdealy:
### Motivation
As described in #1382, we want to debit each Fog Account with a TxOut prior to running the "slam" phase, which involves multiple threads and introduces race conditions in which a Fog Account might not be debited. This is bad because we have tests that fail if one or more of these Fog accounts has zero balance after the script is run.

### In this PR
* Debits each Fog account with one TxOut prior to the "slam" phase.

Fixes #1382.

### Tested
I ran a local fog network and ran the script with sensible defaults. The script ran with no errors, and the logs suggest that the accounts were debited as expected prior to the slam phase.

### DIFF
--- a/fog/distribution/src/main.rs
+++ b/fog/distribution/src/main.rs
@@ -424,7 +424,7 @@ fn worker_thread_entry(
 /// Builds and submits a transaction to a given FogAccount.
 ///
 /// If a transaction submit errors, then we get and use a new FogResolver
-/// to build and submit transactions. In this case, we return this new 
+/// to build and submit transactions. In this case, we return this new
 /// FogResolver to the caller so that it can be used in subsequent transactions.
 fn build_and_submit_transaction(
     txs_created: usize,
--- a/fog/distribution/src/main.rs
+++ b/fog/distribution/src/main.rs
@@ -246,21 +246,19 @@ fn main() {
 
     log::info!(logger, "Seeding Fog Accounts with initial TxOuts.");
     for (i, fog_account) in fog_accounts.iter().enumerate() {
-        if let Some(new_fog_resolver) = build_and_submit_transaction(
+        seed_fog_resolver = build_and_submit_transaction(
             i,
             // For this seed phase, only use one TxOut for each transaction.
             vec![spendable_txouts[i].clone()],
             fog_account,
             &config,
             &ledger_db,
-            &seed_fog_resolver,
+            seed_fog_resolver,
             &logger,
             &conns,
             &env,
             &fog_uri,
-        ) {
-            seed_fog_resolver = new_fog_resolver;
-        }
+        );
     }
 
     // Don't use spendable_txouts that were used in the seed step.
@@ -403,20 +401,18 @@ fn worker_thread_entry(
         )
         .expect("Could not parse fog url");
 
-        if let Some(new_fog_resolver) = build_and_submit_transaction(
+        fog_resolver = build_and_submit_transaction(
             txs_created,
             pending_spendable_txouts,
             to_account,
             &config,
             &ledger_db,
-            &fog_resolver,
+            fog_resolver,
             &logger,
             &conns,
             &env,
             &fog_uri,
-        ) {
-            fog_resolver = new_fog_resolver;
-        }
+        );
         txs_created += 1;
     }
 }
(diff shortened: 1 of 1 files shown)


fog/distribution/src/main.rs
```diff
@@ -206,13 +208,11 @@ fn main() {
...
 
                 // Push to queue
-                spendable_txouts_sender
-                    .send(SpendableTxOut {
-                        tx_out: tx_out.clone(),
-                        amount: input_amount,
-                        from_account_key: account.clone(),
-                    })
-                    .expect("failed sending to spendable_txouts_sender");
+                spendable_txouts.push(SpendableTxOut {
```
samdealy, line 211: This is the main idea- instead of sending the SpendableTxOuts using the sender / receiver structs for the slam threads, create a vector of SpendableTxOuts to be used in two distinct phases (1) Pre-slam debit and (2) slam phase.

fog/distribution/src/main.rs
```diff
@@ -238,6 +238,39 @@ fn main() {
...
+            &conns,
+            &env,
+            &fog_uri,
+        ) {
+            seed_fog_resolver = new_fog_resolver;
+        }
+    }
+
+    // Don't use spendable_txouts that were used in the seed step.
+    spendable_txouts = spendable_txouts[fog_accounts.len()..].to_vec();
```
samdealy, line 267: Perhaps there is a slicker way to do not use the SpendableTxOuts that were used in the pre-slam debit phase?

wjuan-mob: It might be cleaner to split this into a seed txouts and slam txouts before you use them for seeding instead of after.

fog/distribution/src/main.rs
```diff
@@ -369,32 +402,68 @@ fn worker_thread_entry(
...
+) -> Option<FogResolver> {
+    // Sometimes transactions can not be submitted before the tombstone block
+    // has passed, so loop until transactions can be submmitted
+    loop {
+        let tx = build_tx(
+            &pending_spendable_txouts,
+            to_account,
+            config,
+            ledger_db,
+            fog_resolver.clone(),
```
wjuan-mob, line 449: Should this be using the new_fog_resolver.clone()? As is I think the loop just keeps using the old fog resolver. Unless I'm missing something.

fog/distribution/src/main.rs
samdealy: I prefer the way it's written to that option because in that option we'd have to introduce logic that's like "For the first x spendable tx outs, where x == fog acocunts length, add to this one vec, and then for every spendable tx out after this x add to another vec." It seems cleaner to me to work off one vec of spendable txouts.

But this seems to just be based on a preference of mine and not grounded in any hard evidence etc.

fog/distribution/src/main.rs
samdealy: Yes. I reformatted the method slightly to achieve this.

fog/distribution/src/main.rs
wjuan-mob: I'm fine either way.

fog/distribution/src/main.rs
wjuan-mob: Thanks. Do you think it would make sense to add a test for a fog resolver failing just to make sure that this loop works?

wjuan-mob (approved):

garbageslam (approved):

fog/distribution/src/main.rs
samdealy: I think it's fine to not have a test here- using a new fog resolver is an implementation detail that has the intended behavior of making transfers possible after a failure scenario. IMO we shouldn't be testing implementation details.

PR #1450, opened 2022-02-03
Fix a couple of minor issues

remoun:
Fixes some issues I found by running `cargo hack check --each-feature`

* Fix an error when mc-util-telemetry is built without the jaeger feature. Follow-up to #1340.
* Add missing `blocking` feature to `reqwest` dependency.
* Bonus cleanup: Move conditional mods into the associated `cfg_if` block.

### DIFF
--- a/attest/net/Cargo.toml
+++ b/attest/net/Cargo.toml
@@ -26,7 +26,7 @@ displaydoc = "0.2"
 mbedtls = "0.8.1"
 pem = "1.0"
 percent-encoding = "2.1.0"
-reqwest = { version = "0.10" , default-features = false, features = ["rustls-tls", "gzip"] }
+reqwest = { version = "0.10" , default-features = false, features = ["blocking", "rustls-tls", "gzip"] }
 serde_json = "1.0"
 sha2 = "0.9.8"
 
--- a/util/telemetry/src/lib.rs
+++ b/util/telemetry/src/lib.rs
@@ -9,7 +9,6 @@ pub use opentelemetry::{
 
 use opentelemetry::{
     global::{tracer_provider, BoxedTracer},
-    sdk::{trace::Config, Resource},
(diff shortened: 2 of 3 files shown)


eranrund (approved):

wjuan-mob (approved):

util/telemetry/src/lib.rs
```diff
@@ -117,7 +116,9 @@ cfg_if::cfg_if! {
 
             opentelemetry_jaeger::new_pipeline()
                 .with_service_name(service_name)
-                .with_trace_config(Config::default().with_resource(Resource::new(tags)))
+                .with_trace_config(
+                    sdk::trace::Config::default()
+                        .with_resource(sdk::Resource::new(tags)))
```
garbageslam, line 121: Another way to fix this is to move the use statements into the cfg if block with the other use statements for this block

attest/net/Cargo.toml
```diff
@@ -26,7 +26,7 @@ displaydoc = "0.2"
 mbedtls = "0.8.1"
 pem = "1.0"
 percent-encoding = "2.1.0"
-reqwest = { version = "0.10" , default-features = false, features = ["rustls-tls", "gzip"] }
+reqwest = { version = "0.10" , default-features = false, features = ["blocking", "rustls-tls", "gzip"] }
```
garbageslam, line 29: I think this doesn't do anything unless you also pick the blocking module at the call sites

garbageslam (changes requested):
Make sure the change to reqwest is actually doing something

attest/net/Cargo.toml
remoun: Right, we use the `reqwest::blocking::Client` without the corresponding feature.

garbageslam: oh i see thanks

garbageslam (approved):

PR #1451, opened 2022-02-04
Make test client submit a consistent number of transfers per day

### LINKED ISSUE #1445: Make `test-client` submit a consistent number of transactions per day
## Background
Currently, the test client performs the following logic in a continuous loop in CI:
1. Send a transaction
1. Confirm the transaction was sent.
1. Wait 60 seconds.

Since the time between step 1 and 2, the number of transactions that a test client sends per day is variable. This makes it hard for MobileCoin  to estimate the number of transactions that real (i.e. non-test) users are making per day.

## Criteria
Use "adaptive" sleep statements to make the test client submit the same number of transactions per day. Adaptive in this context means set the sleep duration to account for how long the transaction took. 

E.g:
```
let tx_start = now();
send_tx();
while !got_tx_back() { }
let interval = now() - tx_start;
// ideally store that to a histogram or something, at minimum log it
sleep(60 - interval);
```

samdealy:
### Motivation
As described in #1445, we want the continuous run of the test client to submit a consistent number of transactions per day. This makes it easier to analyze transaction volume for "real" network participants rather than those executed on behalf of MobileCoin tests.

### In this PR
* Configures the sleep statement in the continous run's transfer loop to account for how long the previous transfer took. This means that a consistent number of transfers will be executed per day.

Fixes #1445

### DIFF
--- a/fog/test-client/src/test_client.rs
+++ b/fog/test-client/src/test_client.rs
@@ -730,7 +730,7 @@ impl TestClient {
                 Some(duration) => duration,
                 None => {
                     let excess_transaction_time = transfer_duration.sub(period);
-                    log::error!(
+                    log::warn!(
                         self.logger,
                         "Transfer took {} seconds. This is {} seconds more than the allotted transfer time.",
                         transfer_duration.as_secs(),
--- a/fog/test-client/src/test_client.rs
+++ b/fog/test-client/src/test_client.rs
@@ -25,6 +25,7 @@ use more_asserts::assert_gt;
 use once_cell::sync::OnceCell;
 use serde::Serialize;
 use std::{
+    ops::Sub,
     sync::{
         atomic::{AtomicBool, AtomicUsize, Ordering},
         Arc, Mutex,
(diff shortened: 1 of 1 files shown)


fog/test-client/src/test_client.rs
```diff
@@ -710,10 +725,24 @@ impl TestClient {
                     }
                 }
             }
+            let transfer_duration = transfer_start.elapsed();
+            let sleep_duration = match period.checked_sub(transfer_duration) {
+                Some(duration) => duration,
+                None => {
+                    let excess_transaction_time = transfer_duration.sub(period);
+                    log::error!(
+                        self.logger,
```
garbageslam, line 734: Should this be error? It will trigger an alert, right? Maybe warn instead?

garbageslam (approved):

fog/test-client/src/test_client.rs
samdealy: Done.

jcape (approved):

PR #1453, opened 2022-02-04
Feature/increasememosize

wjuan-mob:
Soundtrack of this PR: [link to song that really fits the mood of this PR](https://www.youtube.com/watch?v=vgP0ZKfNIv0)

### Motivation

The motivation for this PR is in this mcip: https://github.com/mobilecoinfoundation/mcips/pull/24

### In this PR

Updates the memo field to be 66 bytes instead of 46 bytes. 
Update some comments for clarity. 
Update tests that include the hash of block contents because block contents with a 66 byte e_memo do not hash to the same value as a 46 byte e_memo.


transaction/core/src/blockchain/block.rs
```diff
@@ -344,17 +344,17 @@ mod block_tests {
         assert_eq!(
             block.id.as_ref(),
             &[
-                54, 149, 234, 64, 10, 166, 9, 176, 203, 154, 166, 123, 16, 98, 112, 231, 21, 243,
-                198, 135, 186, 238, 195, 29, 166, 59, 85, 50, 215, 137, 171, 104
+                118, 205, 187, 34, 207, 104, 52, 137, 97, 124, 79, 205, 112, 204, 146, 217, 128,
+                178, 169, 214, 231, 120, 46, 237, 17, 93, 59, 136, 101, 131, 197, 217
```
garbageslam, line 348: I wonder if we should change the way this test is working, and make the block contents be "old" block contents so that we don't change the hashes here.

We could separately test block hashes that do have memos

garbageslam (approved):
LGTM!

garbageslam (changes requested):
following discussion offline:

the point of these "consistent hashing" tests is that we will never change the test vectors, because it is a breaking change if we do so, and will break the clients.

the correct way to fix the test is to make the test data not generate the memos, and go back to the old hash. then we could add a new test that also tests the data with the memos.

I think I didn't do this correctly in the first PR that merged the memos, but we should do it right this time.

transaction/core/tests/digest-test-vectors.rs
```diff
@@ -217,22 +217,22 @@ fn block_contents_digestible_test_vectors() {
...
+            171, 237, 164, 209, 9, 170, 109, 54, 133, 143, 110, 215, 199, 69
         ]
     );
     assert_eq!(
         results[2].1.hash().0,
         [
-            228, 130, 116, 249, 30, 143, 159, 34, 37, 156, 76, 25, 254, 136, 205, 157, 69, 167,
-            179, 156, 223, 149, 164, 110, 140, 105, 169, 233, 46, 235, 98, 194
+            215, 120, 30, 77, 126, 121, 230, 151, 225, 205, 247, 66, 101, 113, 221, 32, 122, 41,
+            29, 53, 98, 48, 72, 183, 219, 255, 228, 161, 47, 46, 245, 221
```
garbageslam, line 235: these probably also should not be changing

garbageslam: it might be okay because the origin block has no memos and those hashes are not changing

garbageslam: nm, i didn't read far enough down in the file

garbageslam (approved):
LGTM

samdealy (approved):

PR #1460, opened 2022-02-08
Fixup transaction validation for 1.2.0

garbageslam:
Per James, I actually got the release plan wrong for the RTH stuff.
The idea was supposed to be, update all readers first, in 1.2.0,
and only in the next release may writing memos be allowed.

This means that, transaction validation should enforce for now
that memos do not exist in the chain.

In [MCIP 26](https://github.com/mobilecoinfoundation/mcips/pull/26) we may introduce a mechanism to allow turning on the
memos. That is not yet implemented. We will continue to work on that
but by merging this to master now, we make it possible to branch
for release conceivably, and create clients capable of reading memos, while
we could later make a release that accepts writing memos in the
future.

We can discuss separately if the scope for 1.2.0 should be to
include a `block_version`-aware `TransactionBuilder`, it's possible
that it should be and then we aren't ready to branch. (We also still
need to adjust some of the memo schemas.) But then this 
commit is still a fine basis for future iteration.

### DIFF
--- a/transaction/core/src/validation/validate.rs
+++ b/transaction/core/src/validation/validate.rs
@@ -63,8 +63,8 @@ pub fn validate<R: RngCore + CryptoRng>(
 
     // In the 1.2 release, it is planned that clients will know to read memos,
     // but memos will not be allowed to exist in the chain until the next release.
-    // If we implement "block-version-based protocol evolution" (MCIP 26), then this function
-    // would become block-version aware and this coudl become a branch.
+    // If we implement "block-version-based protocol evolution" (MCIP 26), then this
+    // function would become block-version aware and this could become a branch.
     validate_no_memos_exist(tx)?;
     // validate_memos_exist(tx)?;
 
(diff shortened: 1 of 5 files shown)


transaction/core/src/validation/validate.rs
```diff
@@ -65,6 +61,13 @@ pub fn validate<R: RngCore + CryptoRng>(
     // Note: The transaction must not contain a Key Image that has previously been
     // spent. This must be checked outside the enclave.
 
+    // In the 1.2 release, it is planned that clients will know to read memos,
+    // but memos will not be allowed to exist in the chain until the next release.
+    // If we implement "block-version-based protocol evolution, then this function
```
remoun, line 66: Nit: Missing quote. Could also use an MCIP #.
```suggestion
    // If we implement "block-version-based protocol evolution" (MCIP 26), then this function
```

remoun (approved):

eranrund (approved):

remoun (approved):

PR #1461, opened 2022-02-08
Fog passes `latest_block_version` to clients in its responses

garbageslam:
This makes fog-ledger pass `latest_block_version` to clients in
its responses when they do `KeyImage` checks or get merkle proofs.

This will mean that they will get an update on this value when
they do a balance check or before they submit a transaction,
giving them a chance to potentially:

* Pass the `latest_block_version` to the transaction builder, as
  envisioned in MCIP https://github.com/mobilecoinfoundation/mcips/pull/26
* Maybe signal the need for an update if in the course of a balance
  check we learn that the block version has advanced beyond what
  our software was built to support.

This is not a breaking change, this is just the part where fog
communicates the `block_version` which is backwards compatible,
so it is safe to merge at any time.

The non-fog clients can already ask the local ledger for the latest
block version directly.

We could add an API to consensus as well to get the latest block
header, but it doesn't seem that it's a requirement right now
so we didn't.

We could also have piped this number through fog-view but we
decided not to, since it's more plumbing and it seems that as long
as the user gets this number at least once when they do a balance
check, then it's enough.

### DIFF
--- a/fog/types/src/view.rs
+++ b/fog/types/src/view.rs
@@ -20,8 +20,13 @@ pub use mc_fog_kex_rng::KexRngPubkey;
 // These are synced with types in fog_api view.proto, and tests enforce that
 // they round trip These are NOT expected to be synced with Db schema types
 
+/// The QueryRequestAAD structure, which should be passed as AAD when making
+/// an attested fog view request.
 #[derive(Clone, Eq, PartialEq, Message)]
 pub struct QueryRequestAAD {
+    /// The first id of a user event to return. If you set this larger than 0,
+    /// it means that you already saw some of the events and you don't want
+    /// to be sent them again.
     #[prost(int64, tag = "1")]
     pub start_from_user_event_id: i64,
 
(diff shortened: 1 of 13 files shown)


ledger/db/src/ledger_trait.rs
```diff
@@ -81,4 +81,13 @@ pub trait Ledger: Send {
...
     /// Get the tx out root membership element from the tx out Merkle Tree.
     fn get_root_tx_out_membership_element(&self) -> Result<TxOutMembershipElement, Error>;
+
+    /// Get the latest block header, if any. NotFound if the ledger-db is empty.
+    fn get_latest_block(&self) -> Result<Block, Error> {
+        let num_blocks = self.num_blocks()?;
+        if num_blocks == 0 {
+            return Err(Error::NotFound);
+        }
+        Ok(self.get_block(num_blocks - 1)?)
```
remoun, line 91: Optional/Equivalent: Return the result of `get_block` directly.

```suggestion
        if num_blocks == 0 {
            return Err(Error::NotFound);
        }
        self.get_block(num_blocks - 1)
```
or even
```suggestion
        return if num_blocks == 0 {
            Err(Error::NotFound)
        } else {
            self.get_block(num_blocks - 1)
```

remoun (approved):
Thanks for improving the documentation, too!

ledger/db/src/ledger_trait.rs
garbageslam: yeah clippy said the same thing :)

```
error: question mark operator is useless here
  --> ledger/db/src/ledger_trait.rs:91:9
   |
91 |         Ok(self.get_block(num_blocks - 1)?)
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try: `self.get_block(num_blocks - 1)`
   |
   = note: `-D clippy::needless-question-mark` implied by `-D warnings`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_question_mark
```

garbageslam:
Yeah i am a big believer in `#[deny(missing_docs)]`

eranrund (approved):
LGTM but there is one thing I am wondering about:
We will be announcing the latest block version in the blockchain, but maybe we should also announce the maximum possible block version (e.g. there might be a period where only v1 blocks are getting into the chain, but v2 has already been implemented and is awaiting activation by a configuration flag and we want clients to start updating so that when v2 starts showing up they are aware of it).

I'm not sure if there's a non-janky way to expose that, since it's just a constant that will depend on a redeploy of the view server.

jcape (approved):
LGTM

wjuan-mob (approved):

fog/types/src/view.rs
```diff
@@ -20,8 +20,13 @@ pub use mc_fog_kex_rng::KexRngPubkey;
 // These are synced with types in fog_api view.proto, and tests enforce that
 // they round trip These are NOT expected to be synced with Db schema types
 
+/// The QueryRequestAAD structure, which should be passed as AAD when making
+/// an attested fog view request.
 #[derive(Clone, Eq, PartialEq, Message)]
 pub struct QueryRequestAAD {
+    /// The first id of a user event to return. If you set this larger than 0,
```
the-real-adammork, line 27: love the new comments

PR #1466, opened 2022-02-09
Sam/fog overseer get ingest summaries

### LINKED ISSUE #1465: Implement GET /ingest_summaries endpoint to Fog Overseer
As discussed in the Fog Overseer design doc, we'd like to expose a `GET /ingest_summaries `endpoint that returns the ingest summaries for each node in the Fog Ingest Cluster (or, at least, each node that is responsive to gRPC requests). 

The intended user for this endpoint is an ops engineer who wants to quickly inspect the current state of the active Fog Ingest cluster. While they could check the prometheus metrics captured by Fog Overseer, this API offers real-time updates on the status of the cluster.

samdealy:
### Motivation
As discussed in #1465, we'd like to expose an endpoint that returns an ingest summary for each node in the active Fog Ingest cluster. This will allow ops even more visibility into the current state of Fog Ingest. 

### In this PR
* Creates an "internal" non-protobuf version of the `IngestSummary` struct in order to enable us to derive the serde::Serialize trait for it, which is needed in order to send the struct as Json through rocket.
* Creates a thread-safe list of FogIngestGrpcClient in the OverseerService and injects a thread-safe FogIngestGrpcClient in the OverseerWorker. This allows the worker use the client for it's logic and the OverseerService to respond to GET /ingest_summaries requests as they come in. (Previously, the worker was the only component in overseer that needed to use the FogIngestGrpcClient).
* Implements the HTTPS `GET /ingest_summaries` endpoint using the rocket framework. Returns the ingest summaries in JSON format.
* Implements an end to end test that ensures that the expected IngestSummaries are returned by the endpoint.

Fixes #1465


fog/overseer/server/src/service.rs
```diff
@@ -118,4 +139,31 @@ where
...
+                    let ingest_summary: IngestSummary = IngestSummary::from(&proto_ingest_summary);
+                    ingest_summaries.push(ingest_summary);
+                }
+                Err(err) => {
+                    let error_message = format!(
+                        "Unable to retrieve ingest summary for node ({}): {}",
+                        ingest_client.get_uri(),
+                        err
+                    );
+                    return Err(error_message);
```
samdealy, line 162: I opted to fail fast here instead of returning a partial slate of IngestSummaries. It's cleaner this way, but I understand that it might be useful to have some of the IngestSummaries returned, and we could also attach error messages for those that aren't returned. 

If you have an opinion on this, let me know!

fog/api/src/conversions.rs
```diff
@@ -14,3 +16,31 @@ impl From<&common::BlockRange> for fog_common::BlockRange {
...
+        let ingest_controller_mode = match proto_ingest_summary.mode {
+            ingest_common::IngestControllerMode::Idle => {
+                mc_fog_types::ingest_common::IngestControllerMode::IDLE
+            }
+            ingest_common::IngestControllerMode::Active => {
+                mc_fog_types::ingest_common::IngestControllerMode::ACTIVE
+            }
+        };
+        let ingress_pubkey: CompressedRistrettoPublic =
+            CompressedRistrettoPublic::try_from(proto_ingest_summary.get_ingress_pubkey()).unwrap();
```
eranrund, line 33: This should not call `unwrap()`. If it can fail, this should be a `try_from`.

fog/overseer/server/src/service.rs
eranrund: I think it would be useful to actually include error per client. If the clients have an identifying URL, maybe the correct response type is something along the lines of (making this up) `Json<HashMap<Uri, JsonResult>>`, where JsonResult is something like `{error: Option<String>, success: Option<IngestSummary>}`

eranrund (changes requested):

fog/api/src/conversions.rs
samdealy: Done.

fog/overseer/server/src/responses.rs
```diff
@@ -0,0 +1,16 @@
...
+use mc_fog_uri::FogIngestUri;
+use serde::Serialize;
+use std::collections::HashMap;
+
+/// The ingest summaries for the Fog Ingest cluster that Fog Overser is
+/// monitoring. If an ingest summary can't be retrieved for a given node, then
+/// an error message, rather than an ingest summary, is returned for that node.
+#[derive(Serialize)]
+pub struct GetIngestSummariesResponse {
+    pub ingest_summaries: HashMap<FogIngestUri, Result<IngestSummary, String>>,
```
eranrund, line 15: Ha, I didn't know `Result` can be serialized to JSON this easily. Neat. Would you mind sharing how the JSON ends up looking like?

eranrund (approved):

fog/overseer/server/src/worker.rs
```diff
@@ -254,7 +238,9 @@ where
         &self,
     ) -> Result<Vec<IngestSummaryNodeMapping>, OverseerError> {
         let mut ingest_summary_node_mappings: Vec<IngestSummaryNodeMapping> = Vec::new();
-        for (ingest_client_index, ingest_client) in self.ingest_clients.iter().enumerate() {
+        for (ingest_client_index, ingest_client) in
+            self.ingest_clients.lock().unwrap().iter().enumerate()
```
remoun, line 242: Ditto (RPC while holding lock)

fog/overseer/server/tests/get_ingest_summaries.rs
```diff
@@ -0,0 +1,121 @@
...
+    let db_test_context = SqlRecoveryDbTestContext::new(logger.clone());
+    let recovery_db = db_test_context.get_db_instance();
+    let mut rng: Hc128Rng = SeedableRng::from_seed([0u8; 32]);
+
+    // Set up the watcher db
+    let blockchain_path =
+        TempDir::new("blockchain").expect("Could not make tempdir for blockchain state");
+    let watcher_path = blockchain_path.path().join("watcher");
+    std::fs::create_dir(&watcher_path).expect("couldn't create dir");
+    WatcherDB::create(&watcher_path).unwrap();
```
remoun, line 33: Optional/Equivalent: Replace `unwrap` with `expect` for this setup.

fog/overseer/server/src/service.rs
samdealy: Done. Implemented using Result rather than a bespoke struct.

fog/overseer/server/src/responses.rs
samdealy: Yea, I was pleasantly surprised. I guess it makese sense: rocket's Json<T> accepts any T that implements serde::Serialize and Result does. 

Here's a snippet of the output:

```
{
  "ingest_summaries": {
    "insecure-fog-ingest://0.0.0.0:5021/": {
      "Ok": {
        "ingest_controller_mode": "IDLE",
        "next_block_index": 0,
        "pubkey_expiry_window": 10,
        "ingress_pubkey": [
            ...
        ],
        "egress_pubkey": [ 
         ...
        ],
        "kex_rng_version": 0,
        "peers": [
          "insecure-igp://0.0.0.0:5002/",
          "insecure-igp://0.0.0.0:5012/",
          "insecure-igp://0.0.0.0:5022/"
        ],
        "ingest_invocation_id": 0
      }
    },
   ...
}
```

fog/overseer/server/src/server.rs
```diff
@@ -50,6 +60,15 @@ pub fn initialize_rocket_server(
...
     rocket::custom(rocket_config)
-        .mount("/", routes![enable, disable, get_status, get_metrics])
+        .mount(
+            "/",
+            routes![
+                enable,
+                disable,
+                get_status,
+                get_metrics,
+                get_ingest_summaries
```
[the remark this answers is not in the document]
samdealy: For context. the endpoint methods are grouped logically and not alphanumerically. This order is the same as that logical ordering As such, I'd prefer to keep them in this order.

fog/overseer/server/tests/get_ingest_summaries.rs
samdealy: This is used throughout the codebase, so will leave as is. We can revisit this in a small follow-up PR if we so desire.

fog/overseer/server/src/service.rs
```diff
@@ -118,4 +141,46 @@ where
 
         Ok(response)
     }
+
+    pub fn get_ingest_summaries(&self) -> Result<Json<GetIngestSummariesResponse>, String> {
+        let mut ingest_summaries: HashMap<FogIngestUri, Result<IngestSummary, String>> =
+            HashMap::new();
+        for ingest_client in self.ingest_clients.lock().unwrap().iter() {
```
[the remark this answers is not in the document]
samdealy: I ended up removing the Mutex because we don't need to mutate the client. Thanks for pointing this out.

fog/overseer/server/src/worker.rs
samdealy: See comment above.

fog/overseer/server/src/service.rs
```diff
@@ -33,9 +41,24 @@ impl<DB: RecoveryDb + Clone + Send + Sync + 'static> OverseerService<DB>
...
+            .iter()
+            .map(|fog_ingest_uri| {
+                FogIngestGrpcClient::new(
+                    fog_ingest_uri.clone(),
+                    Self::GRPC_RETRY_SECONDS,
+                    grpcio_env.clone(),
+                    logger.clone(),
+                )
+            })
+            .collect();
```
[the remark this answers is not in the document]
samdealy: I think it's fair point that Dependency Injection is normally preferred. However, for the tests associated with this code, they need real FogIngestGrpcClients, and it's easier to just pass the URIs and have the service create them. So I'll opt to keep it as is, but if we write a test that needs to mock the clients, then we can add this.

fog/overseer/server/src/service.rs
```diff
@@ -118,4 +140,46 @@ where
 
         Ok(response)
     }
+
+    pub fn get_ingest_summaries(&self) -> Result<GetIngestSummariesResponse, String> {
+        let mut ingest_summaries: HashMap<FogIngestUri, Result<IngestSummary, String>> =
+            HashMap::new();
+        for ingest_client in self.ingest_clients.iter() {
```
jcape, line 147: I trust Rust to handle `Vec` mutability safely, but I'm it's worth looking into grpc-rs's generated clients to ensure there isn't some unsafe interior mutability that's being exposed due to FFI or whatever.

We ended up doing `mc_connection::SyncConnection`, to wrap the `Connection` object in a lock, so I am curious as to whether this is required here.

eranrund: I thought we did that because of enclave stuff / something to do with encryption/nonce behaving properly?

eranrund: Yeah it might have something to do with retries - e.g. a piece of code constructs a Tx and encrypts, then attempts to send it 3 times since `SyncConnection` implements retrying. Can't have anything else send something during that time or decryption will fail.

fog/types/src/ingest_common.rs
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) 2018-2021 The MobileCoin Foundation
+
+//! Contains "native" rust structs that correspond to structs generated by  
+//! protobuf.
+use alloc::{string::String, vec::Vec};
+use mc_crypto_keys::CompressedRistrettoPublic;
+use serde::{Deserialize, Serialize};
+
+/// Mirrors the proto definition of IngestSummary. We have to define this
+/// "native" Rust type to be able to add traits like Serialize to this data.
```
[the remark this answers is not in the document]
samdealy: I think it's a good idea to use the built-in protobuf JSON once v3 becomes stable. The protobuf-json library seems to have few users and hasn't had a new commit since 2019. IMO, I'd rather have this boilerplate for now then pull a stale third-party dep.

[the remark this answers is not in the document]
remoun: Fair enough.

fog/overseer/server/src/service.rs
```diff
@@ -118,4 +140,46 @@ where
...
+            match ingest_client.get_status() {
+                Ok(proto_ingest_summary) => {
+                    log::trace!(
+                        self.logger,
+                        "Ingest summary retrieved: {:?}",
+                        proto_ingest_summary
```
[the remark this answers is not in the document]
samdealy: Done.

fog/overseer/server/src/service.rs
remoun: FWIW I grepped the `grpcio` code and there were some mentions of thread-safety and thread-compatibility, so I'm cautiously optimistic there aren't any lurking thread safety issues with this approach.

remoun (approved):

fog/overseer/server/src/service.rs
jcape: I'm fine with this as-is, I just wanted to raise the issue since I didn't know why we had `SyncConnection`, and my strong suspicion was that whatever gRPC framework we were using simply wasn't getting tagged as `Sync` by `rustc` (presumably due to a raw pointer).

PR #1468, opened 2022-02-09
block-version-based protocol evolution

garbageslam:
This makes transaction builder and validation aware of BlockVersion.
It also plumbs this config into consensus.

This commit adds support for the change in fog sample paykit
and mobilecoind

See also: https://github.com/mobilecoinfoundation/mcips/pull/26

Checklist:
- [x] Rebase on master
- [x] Do the responder id parts
- [x] I made the block version be a parameter to things like `form_block`, but it might be better to do it differently, e.g. block version is announced on enclave startup, and can change after `form_block` but not in between calls to form block?
- [x] There is a place where I am taking a max of `latest_block.version` and configured block version, but perhaps it should only ever use the configured block version.
- [x] In many tests, I decided it was worth testing all acceptable values of block version, and in other places, I only tested `BlockVersion(1)` or `BlockVersion::MAX` because that seemed like the highest value testing. Review these decisions.


transaction/core/src/blockchain/block_version.rs
```diff
@@ -0,0 +1,43 @@
...
+
+use core::{fmt, hash::Hash, ops::Deref};
+use mc_crypto_digestible::Digestible;
+use serde::{Deserialize, Serialize};
+
+/// A block version number
+#[derive(
+    Clone, Copy, Debug, Eq, PartialEq, PartialOrd, Ord, Serialize, Deserialize, Digestible, Hash,
+)]
+pub struct BlockVersion(pub u32);
```
jcape, line 11: This should be private.

transaction/core/src/blockchain/block_version.rs
```diff
@@ -0,0 +1,43 @@
...
+
+impl From<u32> for BlockVersion {
+    fn from(src: u32) -> Self {
+        Self(src)
+    }
+}
```
jcape, line 17: This should be `TryFrom<u32>`, not `From<u32>`

This would gate supported versions from unsupported, rather than having to re-implement that test everywhere you're reading an integer. If you want an infallible constructor, it should be `impl BlockVersion { fn from_unchecked(src: u32) -> Self { Self(src) } }`

```suggestion
pub struct UnsupportedVersion;

impl TryFrom<u32> for BlockVersion {
    type Err = UnsupportedVersion;

    fn try_from(src: u32) -> Result<Self, Self::Err> {
        if src < Self::MIN.0 || src > Self::MAX.0 {
            Err(UnsupportedVersion)
        } else {
            Ok(Self(src))
        }
    }
}
```

transaction/core/src/blockchain/block_version.rs
```diff
@@ -0,0 +1,43 @@
...
+
+impl fmt::Display for BlockVersion {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{:?}", self)
```
jcape, line 21: Nit: Display -> Debug chains are not great for internationalization, so this should just print the version as digits and not worry about adding labels. We have debug for debug :-)

```suggestion
        write!(f, "{}", self.0)
```

transaction/std/src/transaction_builder.rs
```diff
@@ -65,10 +67,11 @@ impl<FPR: FogPubkeyResolver> TransactionBuilder<FPR> {
     /// * `memo_builder` - An object which creates memos for the TxOuts in this
     ///   transaction
     pub fn new<MB: MemoBuilder + 'static + Send + Sync>(
+        block_version: BlockVersion,
```
garbageslam: This is one of the other main points of the PR

transaction/core/src/blockchain/block_version.rs
```diff
@@ -0,0 +1,43 @@
...
+    }
+}
+
+impl BlockVersion {
+    /// The maximum value of block_version that this build of
```
jcape, line 26:
```suggestion
    /// The minimum value of block_version that this build of
    /// mc-transaction-core has support for
    pub const MIN: Self = Self(1);

    /// The maximum value of block_version that this build of
```

transaction/std/src/memo_builder/mod.rs
```diff
@@ -34,15 +38,15 @@ pub trait MemoBuilder: Debug {
         value: u64,
         recipient: &PublicAddress,
         memo_context: MemoContext,
-    ) -> Result<Option<MemoPayload>, NewMemoError>;
+    ) -> Result<MemoPayload, NewMemoError>;
```
garbageslam, line 41: I changed this because in the new world it doesn't make sense for a memo builder to sometimes omit the memo and sometimes not -- the block version controls if the memo is present or not, and the memo builder can just write `MemoPayload::default` if they don't want to write any content. This also means `NoMemoBuilder`goes away and is replaced by `EmptyMemoBuilder` everywhere it was used, since it will do the right thing based on Block version.

transaction/std/src/transaction_builder.rs
```diff
@@ -126,11 +131,18 @@ impl<FPR: FogPubkeyResolver> TransactionBuilder<FPR> {
...
             .take()
             .expect("memo builder is missing, this is a logic error");
+        let block_version = self.block_version.clone();
         let result = self.add_output_with_fog_hint_address(
             value,
             recipient,
             recipient,
-            |memo_ctxt| mb.make_memo_for_output(value, recipient, memo_ctxt),
+            |memo_ctxt| {
+                if block_version.e_memo_feature_is_supported() {
```
garbageslam: This is where the transaction builder uses block version to decide whether or not to filter out memos

transaction/core/src/blockchain/block_version.rs
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) 2018-2021 The MobileCoin Foundation
+
+use core::{fmt, hash::Hash, ops::Deref};
+use mc_crypto_digestible::Digestible;
+use serde::{Deserialize, Serialize};
+
+/// A block version number
+#[derive(
+    Clone, Copy, Debug, Eq, PartialEq, PartialOrd, Ord, Serialize, Deserialize, Digestible, Hash,
```
jcape, line 9: It's easier to check whether something is already implemented later when these are sorted...
```suggestion
    Clone, Copy, Debug, Deserialize, Digestible, Hash, Eq, Ord, PartialEq, PartialOrd, Serialize,
```

jcape (commented):
Comments on BlockVersion (namely, keep it private, and use try_from to check whether a version is supported -- this is where the "this version is not supported" check happens, rather than making everyone that's consuming a chain implement their own version).

transaction/core/src/blockchain/block_version.rs
garbageslam: The thing is, if it is private then I can't use the nice syntax `BlockVersion(1)` to construct it

transaction/core/src/blockchain/block_version.rs
garbageslam: This has some drawbacks: for instance, in Eran's idea: https://github.com/mobilecoinfoundation/mobilecoin/pull/1461#pullrequestreview-876465436

We might tell people about block versions that are coming soon before they actually happen. If that manifests as a deserialization error that would make it harder for the app to do the right thing.

transaction/core/src/blockchain/block_version.rs
garbageslam: I just copied the code from `TokenId` here

transaction/core/src/blockchain/block_version.rs
garbageslam: That is not the semantic of this field. If my transaction core has max block version 3, my software will handle any txout from versions 1 up to 3. Calling this a min implies that we won't be compatible with e.g. version 1 tx outs

transaction/core/src/blockchain/block_version.rs
garbageslam: I just copied the code from token id, if that's what you want let's fix botj

transaction/core/src/blockchain/block_version.rs
garbageslam: Oh I misunderstood

garbageslam: I think this isn't actually the min, the origin block seems to have version 0

transaction/core/src/blockchain/block_version.rs
garbageslam: Also this wouldn't actually work for e.g. the enclave checking this number when it is passed in, because it uses serde deserialize and not try from. So it's potentially a footgun to do this check "implicitly" as you suggest. I feel like it's safer to check the number against max everywhere that it's important to do so

transaction/core/src/blockchain/block_version.rs
garbageslam: Lmk if you want that to change also

transaction/core/src/blockchain/block_version.rs
garbageslam: I'm okay to do it this way if you want, I'll have to figure out a nicer way to iterate over block versions though

transaction/core/src/blockchain/block_version.rs
jcape: Right, I'm saying you can have BlockVersion > MAX if you do that, which probably isn't valid.

transaction/core/src/blockchain/block_version.rs
jcape: Actually it was the iteration that led me to this, hence `BlockVersion::MIN..BlockVersion::MAX`

jcape: Serde is the good point, but I'd think that `BlockVersion::is_valid()` is the better work-around for the lack of proper data-access serializable objects.

transaction/core/src/blockchain/block_version.rs
jcape: Ideally, yeah, I've gone back and forth on this, and really I just want the display to show human readable contents, and let me fill in the details.

transaction/core/src/blockchain/block_version.rs
jcape: OK, this was intended to replace `BlockVersion(1)`.

transaction/core/src/blockchain/block_version.rs
jcape: For the comment in #1461, I think this is fine, that's a good "time to upgrade" indicator (when `try_from()` fails).

eranrund:
Tip for reviewers: consider pulling locally and doing `git diff --ignore-space-change master`. A lot of the changes here are indentation and result in an enormous diff in GitHub.

transaction/core/src/blockchain/block_version.rs
eranrund: Works for me

transaction/core/src/blockchain/block_version.rs
eranrund: This might get ugly in various protobuf<->prost conversions, if we have them. E.g. I am imagining a world where some grpc (protobuf) response returns a bunch of fields and we try to convert it into our Prosty struct, but now this fails because the struct has a block_version in it and the `try_from` fails. We could use the `from_unchecked` variant there, but then that kind of means we can't rely on whats inside `BlockVersion` to be valid.

Separately, I am not sure the concept of valid/invalid makes sense here. One way I am thinking about it is all block versions are valid, we might just not know what to do with a given version yet so it is unsupported for certain purposes (like, we cant hash it, and we cant build it)

garbageslam: @jcape another alternative may be, `BlockVersion` objects are not valid and invalid, but attempting to determine if a feature is supported looks like:

```
pub fn feature_memos_is_supported(&self) -> Result<bool> {
    if self > BlockVersion::MAX { return Err(...); }
    Ok(self.0 >= 2)
}
```

so using it for transaction building or validation fails if this version is too new and you don't know what it the rules are, but deserializing a `BlockVersion` doesn't fail.

transaction/core/src/blockchain/block_version.rs
jcape: Either way is fine by me.

fog/sample-paykit/src/cached_tx_data/mod.rs
```diff
@@ -662,6 +676,8 @@ impl CachedTxData {
                 Ok(response) => {
                     self.latest_global_txo_count =
                         core::cmp::max(self.latest_global_txo_count, response.global_txo_count);
+                    self.latest_block_version =
+                        core::cmp::max(self.latest_block_version, response.latest_block_version);
```
garbageslam, line 680: TODO: document the rationale why we should take a max here

ledger/db/src/lib.rs
```diff
@@ -1358,22 +1367,11 @@ mod ledger_db_test {
                 (0..5).map(|_i| KeyImage::from(rng.next_u64())).collect();
 
             let block_contents = BlockContents::new(key_images.clone(), outputs);
-            assert_eq!(last_block.version, BLOCK_VERSION);
-
-            let invalid_block = Block::new_with_parent(
-                last_block.version + 1,
```
garbageslam, line 1364: this test is not possible to run anymore because `BlockVersion::MAX + 1` is enforced not to exist as a `BlockVersion`

garbageslam: we could try testing that getting an `ArchiveBlock` that has `BlockVersion::MAX + 1` fails with a block version error

transaction/core/src/blockchain/block.rs
```diff
@@ -122,7 +123,7 @@ impl Block {
     /// * `root_element` - The root element for membership proofs
     /// * `block_contents` - Contents of the block.
     pub fn new(
-        version: u32,
+        version: BlockVersion,
```
garbageslam, line 126: we could consider that `Block` doesn't use a `BlockVersion` and uses a `u32` instead, since it is using u32 internally. that would make the ledger db test around adding too recent blocks still possible.

transaction/core/src/validation/validate.rs
```diff
@@ -450,19 +450,26 @@ mod tests {
...
-    fn create_test_tx() -> (Tx, LedgerDB) {
+    fn create_test_tx(block_version: BlockVersion) -> (Tx, LedgerDB) {
         let mut rng: StdRng = SeedableRng::from_seed([1u8; 32]);
-        let sender = mc_transaction_core_test_utils::AccountKey::random(&mut rng);
+        let sender = AccountKey::random(&mut rng);
         let mut ledger = create_ledger();
         let n_blocks = 1;
-        initialize_ledger(&mut ledger, n_blocks, &sender, &mut rng);
+        initialize_ledger(
+            adapt_hack(&block_version),
```
garbageslam, line 459: I would really like to refactor to kill off this `adapt_hack` stuff, I think we just have to reorganize the crates a little bit.

garbageslam:
I think this may be ready to go, I fixed up all the consensus-service plumbing:

* `enclave_init` takes the `BlockVersion` and it is not an argument to `form_block` or the tx validation routine, instead it is stored in the enclave and can't be changed after init.
* That's better because the `BlockVersion` needs to go into the responder id, and the idea right now is that the network needs to be restarted to change the `BlockVersion`. (We can investigate protocol version scp statements in a later PR)
* `BlockVersion` now has strong invariants and uses `#[serde(try_from = "u32")]` to prevent skirting them
* Most of the changes are now in tests. There is some question if `Block::new(` should actually take a `BlockVersion`.

ledger/db/src/lib.rs
```diff
@@ -1358,10 +1367,12 @@ mod ledger_db_test {
...
 
             let block_contents = BlockContents::new(key_images.clone(), outputs);
-            assert_eq!(last_block.version, BLOCK_VERSION);
+            assert_eq!(last_block.version, *MAX_BLOCK_VERSION);
 
+            // Note: unsafe transmute is being used to skirt the invariant that BlockVersion
+            // does not exceed MAX_BLOCK_VERSION
             let invalid_block = Block::new_with_parent(
-                last_block.version + 1,
+                unsafe { core::mem::transmute(last_block.version + 1) },
```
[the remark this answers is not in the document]
garbageslam: i guess another option is to make `add_block` function of ledger db not take a `BlockVersion`

consensus/enclave/impl/src/lib.rs
```diff
@@ -173,9 +181,17 @@ impl ConsensusEnclave for SgxConsensusEnclave {
...
     ) -> Result<(SealedBlockSigningKey, Vec<String>)> {
-        // Inject the fee into the peer ResponderId.
-        let peer_self_id = fee_map.responder_id(peer_self_id);
+        // Inject the fee map and block version into the peer ResponderId.
+        let peer_self_id = ResponderId::from_str(&format!(
+            "{}-block-version-{}",
+            fee_map.responder_id(peer_self_id),
+            block_version
+        ))
+        .unwrap();
```
garbageslam, line 192: lmk if you guys think this responder-id part should be done differently

transaction/core/src/validation/validate.rs
```diff
@@ -30,6 +30,7 @@ use rand_core::{CryptoRng, RngCore};
 pub fn validate<R: RngCore + CryptoRng>(
     tx: &Tx,
     current_block_index: u64,
+    block_version: BlockVersion,
     root_proofs: &[TxOutMembershipProof],
     minimum_fee: u64,
     csprng: &mut R,
```
garbageslam, line 36: it might be a good idea to have explicit checks in this function comparing `BlockVersion` with `BlockVersion::MAX`, even though it's supposed to be a static invariant. that is what we are doing in ledger db right now. lmk what you think about that.

consensus/enclave/impl/src/lib.rs
eranrund: Hmm, has this been tested on a deployed network?
If you update the responder id it should get updated in all other places below, which would result in a bunch of repetition of this string formatting stuff. For example, https://github.com/mobilecoinfoundation/mobilecoin/blob/8250b9ac242a72f025c2d77e6ffb53dba1832b8a/consensus/enclave/impl/src/lib.rs#L267

I think a better approach at this point would be to add a `responder_id` method to this object that takes a responder id and appends whatever it is we decided we need to append to it.

consensus/enclave/impl/src/lib.rs
```diff
@@ -108,6 +108,13 @@ pub struct SgxConsensusEnclave {
     /// Logger.
     logger: Logger,
 
+    /// Block version
+    ///
+    /// This is the value we were configured with on startup, and it goes into
+    /// the responder id. The enclave only forms blocks using this block
+    /// version.
+    block_version: Mutex<BlockVersion>,
```
eranrund, line 116: I am feeling a bit uncomfortable with the growing amount of mutexes we are piling up here. Maybe we should put all of the configuration data that comes from initialization in a single struct and put that behind a mutex?

eranrund: If we put this struct inside `mc-consensus-enclave-api` then we wont have to extend the init api call every time we add more configuration (which involves changing the untrusted crate, two mock enclaves, this crate, the messages file, etc...)

eranrund: and we can Digestible it and use that for the responder-id :)

consensus/service/src/bin/main.rs
```diff
@@ -68,6 +70,7 @@ fn main() -> Result<(), ConsensusServiceError> {
         &config.client_responder_id,
         &cached_key,
         &fee_map,
+        BlockVersion::try_from(config.block_version).unwrap(),
```
eranrund, line 73: Hmm, is it not possible to have `config.block_version` be of type `BlockVersion`?

consensus/service/src/config/mod.rs
```diff
@@ -101,6 +101,10 @@ pub struct Config {
     /// Allow extreme (>= 1MOB, <= 0.000_000_01 MOB).
     #[structopt(long)]
     pub allow_any_fee: bool,
+
+    /// The configured block version
+    #[structopt(long, env = "MC_BLOCK_VERSION", default_value = "1")]
+    pub block_version: u32,
```
eranrund, line 107: What happens if you change this to be of type `BlockVersion`?

mobilecoind/src/payments.rs
```diff
@@ -763,8 +765,13 @@ impl<T: BlockchainConnection + UserTxConnection + 'static, FPR: FogPubkeyResolve
             fog_resolver_factory(&fog_uris).map_err(Error::Fog)?
         };
 
+        // FIXME: We should get the block version from
+        // `LedgerDb::get_last_block()?.version`
+        let block_version = BlockVersion::ONE;
```
eranrund, line 770: I would not leave this as a TODO. While libmobilecoin and the android bindings are used by the SDKs and changing them requires coordinating with them, this is a standalone change in code that entirely lives in this repo and nothing externally would be affected by it getting changed.

consensus/service/src/config/mod.rs
garbageslam: Will try

eranrund (changes requested):
Overall this looks good to me and a very nice improvement in leveraging the type system to save us from ourselves.
My main concerns are around passing configuration to the enclave, the increasing amount of mutexes there and the ResponderId stuff.
This should go through a k8 deployment since I think it is currently broken.
I am also wondering about what effect it has on CI times since we're now doing a bunch of the block tests multiple times, and some of those are on the slower side.
Thank you!

mobilecoind/src/payments.rs
garbageslam: ok i'll figure out how to do it

libmobilecoin/src/transaction.rs
```diff
@@ -342,10 +342,19 @@ pub extern "C" fn mc_transaction_builder_create(
                     FogResolver::new(fog_resolver.0.clone(), &fog_resolver.1)
                         .expect("FogResolver could not be constructed from the provided materials")
                 });
-        // TODO: After servers are deployed that are supporting the memos,
-        // Enable recoverable transaction history by configuring an RTHMemoBuilder
-        let memo_builder = NoMemoBuilder::default();
-        let mut transaction_builder = TransactionBuilder::new(fog_resolver, memo_builder);
+        // FIXME: block version should be a parameter, it should be the latest
+        // version that fog ledger told us about, or that we got from ledger-db
+        let block_version = BlockVersion::ONE;
```
remoun, line 347: This seems necessary to implement for building transactions with block version > 1

transaction/core/src/blockchain/block.rs
remoun: +1 for keeping BlockVersion and using the hack in test

transaction/core/src/blockchain/block_version.rs
```diff
@@ -0,0 +1,110 @@
...
+}
+
+#[derive(Clone, Display, Debug, Deserialize, Eq, Hash, Ord, PartialEq, PartialOrd, Serialize)]
+pub enum BlockVersionError {
+    /// Unsupported block version: {0} > {1}. Try upgrading your software
+    UnsupportedBlockVersion(u32, u32),
+}
+
+#[cfg(test)]
+mod tests {}
```
garbageslam, line 110: i want to add a test here that the invariant is actually being enforced

fog/sample-paykit/src/cached_tx_data/mod.rs
```diff
@@ -129,6 +133,7 @@ impl CachedTxData {
             owned_tx_outs: Default::default(),
             key_image_data_completeness: BlockCount::MAX,
             latest_global_txo_count: 0,
+            latest_block_version: 1u32,
```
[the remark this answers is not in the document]
garbageslam: you are probably right

libmobilecoin/src/transaction.rs
garbageslam: yeah, but i can't change that without changing the `libmobilecoin.h` header which impacts the swift build, so usually i put fixme's without breaking the SDK teams' build

consensus/enclave/impl/src/lib.rs
```diff
@@ -122,6 +129,7 @@ impl SgxConsensusEnclave {
             )),
             logger,
             fee_map: Mutex::new(FeeMap::default()),
+            block_version: Mutex::new(BlockVersion::ONE),
```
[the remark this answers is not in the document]
garbageslam: maybe, that's a good point
ideally it would enforce that you initialized it, so maybe we should make this `MutexOption` or something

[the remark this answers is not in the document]
garbageslam: i'm going to rework this a bit due to Eran's comment

consensus/service/src/bin/main.rs
garbageslam: looks like it works, i just need `BlockVersion` to implement `FromStr`

fog/sample-paykit/src/client.rs
```diff
@@ -173,6 +168,12 @@ impl Client {
         self.tx_data.get_last_memo()
     }
 
+    /// Get the latest block version that we heard about from fog
+    /// Note that this may not be a "valid" block version if our software is old
+    pub fn get_latest_block_version(&mut self) -> u32 {
```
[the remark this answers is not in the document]
garbageslam: fixed

slam/src/main.rs
```diff
@@ -491,8 +491,15 @@ fn build_tx(
     // Sanity
     assert_eq!(utxos_with_proofs.len(), rings.len());
 
+    // TODO: This should probably load block version from the ledger
```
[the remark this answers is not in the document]
garbageslam: i was just trying to work one step at a time :)

i can figure out how to get the block version from the ledger in slam and fog distro and plumb it through

crypto/digestible/src/lib.rs
```diff
@@ -446,6 +446,25 @@ impl DigestibleAsBytes for [u8; 32] {}
...
+        &self,
+        context: &'static [u8],
+        transcript: &mut DT,
+    ) {
+        transcript.append_agg_header(context, b"Tuple");
+        self.0.append_to_transcript(b"0", transcript);
+        self.1.append_to_transcript(b"1", transcript);
+        transcript.append_agg_closer(context, b"Tuple");
+    }
+}
```
garbageslam, line 466: this part is new in the latest commit, it is added to make it easy to digest `BTreeMap`

I should add some tests of digesting `BTreeMap`

consensus/enclave/api/src/config.rs
```diff
@@ -0,0 +1,146 @@
...
+use mc_transaction_core::BlockVersion;
+use serde::{Deserialize, Serialize};
+
+/// Configuration for the enclave which is used to help determine which
+/// transactions are valid.
+///
+/// (This can be contrasted with things like responder id and sealed block
+/// signing key)
+#[derive(Clone, Deserialize, Debug, Digestible, Eq, Hash, PartialEq, Serialize)]
+pub struct BlockchainConfig {
```
garbageslam, line 14: It might be that a better name for this is `TxValidityConfig` or something like this

slam/src/main.rs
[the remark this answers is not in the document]
garbageslam: done

mobilecoind/src/payments.rs
garbageslam: it turned out to be pretty easy, just a few lines :)

consensus/enclave/impl/src/lib.rs
```diff
@@ -262,8 +280,13 @@ impl ConsensusEnclave for SgxConsensusEnclave {
         peer_id: &ResponderId,
         msg: PeerAuthResponse,
     ) -> Result<(PeerSession, VerificationReport)> {
-        // Inject the if fee map hash passing off to the AKE
-        let peer_id = self.fee_map.lock()?.responder_id(peer_id);
+        // Inject the blockchain config hash passing off to the AKE
```
eranrund, line 283:
```suggestion
        // Inject the blockchain config hash before passing off to the AKE
```

consensus/enclave/impl/src/lib.rs
```diff
@@ -424,6 +453,12 @@ impl ConsensusEnclave for SgxConsensusEnclave {
         encrypted_txs_with_proofs: &[(WellFormedEncryptedTx, Vec<TxOutMembershipProof>)],
         root_element: &TxOutMembershipElement,
     ) -> Result<(Block, BlockContents, BlockSignature)> {
+        let blockchain_config = self.blockchain_config.lock()?;
+        let config = blockchain_config
```
eranrund, line 457: It might be worth adding a check that `parent_block.version <= config.block_version`?

consensus/enclave/impl/src/lib.rs
```diff
@@ -564,7 +599,7 @@ impl ConsensusEnclave for SgxConsensusEnclave {
 
         // Form the block.
         let block = Block::new_with_parent(
```
eranrund, line 601: Maybe this should actually be the place to check that `block_version >= parent_block.version`?
Not necessarily in this PR since it's an annoying change to have it return a `Result`.

crypto/digestible/tests/basic.rs
```diff
@@ -480,3 +521,32 @@ fn test_generic_array() {
         garray.digest32::<MerlinTranscript>(b"test"),
     );
 }
+
+// Test that hashing BTreeSet of strings is the same as hashing Vec of (sorted)
+// strings
+#[test]
+fn test_btree_set_vs_vec() {
```
eranrund, line 528: Nice

fog/distribution/src/main.rs
```diff
@@ -563,8 +568,14 @@ fn build_tx(
     // Sanity
     assert_eq!(utxos_with_proofs.len(), rings.len());
 
+    // This max occurs because the bootstrapped ledger has block version 0,
```
eranrund, line 571: Really? Our bootstrap block version is different than all the other blocks?

garbageslam: I thought it was true in mainnet and testnet also, I think the bootstrap program used block version zero for the blocks it made. Maybe code changed between now and then

mobilecoind/src/payments.rs
```diff
@@ -259,11 +261,17 @@ impl<T: BlockchainConnection + UserTxConnection + 'static, FPR: FogPubkeyResolve
         };
         log::trace!(logger, "Tombstone block set to {}", tombstone_block);
 
+        // Come up with a block version
+        let block_version =
+            BlockVersion::try_from(max(self.ledger_db.get_latest_block()?.version, 1))
```
eranrund, line 266: Can you explain (in a comment) why this is needed?

consensus/enclave/impl/src/lib.rs
garbageslam: Yea that seems fine, if we test in multiple places then we probably won't lose the test if lmdb gets replaced or something

fog/distribution/src/main.rs
eranrund: Yeah, looks like that is how the code is right now. I forgot about that.

eranrund (commented):
LGTM. Did this go through k8 and reviewed on Grafana?

garbageslam:
it just finished in k8s, deploy/build 7, i'm looking at grafana now

mobilecoind/src/payments.rs
garbageslam: I guess we could get rid of it if we make the `TransactionBuilder` support using `BlockVersion(0)`, and it just treats it like block version 1. That might be fine

garbageslam:
Screenshot at 2022-02-18 10-47-59
Screenshot at 2022-02-18 10-48-12

garbageslam:
Screenshot at 2022-02-18 10-50-30
Screenshot at 2022-02-18 10-50-41
Screenshot at 2022-02-18 10-50-49

remoun (approved):

transaction/core/src/blockchain/block_version.rs
```diff
@@ -0,0 +1,147 @@
...
+    type Target = u32;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl fmt::Display for BlockVersion {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", self)
```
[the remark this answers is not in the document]
garbageslam: it was actually a bug, it was causing stack overflow
`self.0` is the way it should be written i think

crypto/digestible/src/lib.rs
```diff
@@ -608,6 +627,32 @@ cfg_if! {
                 }
             }
         }
+
+        // Treat a BTreeMap as a (sorted) sequence
+        // This implementation should match that for &[(T, U)]
+        //
+        // Note: We don't currently implement digestible for tuples, but we should.
```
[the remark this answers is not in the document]
garbageslam: thanks, will fix

garbageslam:
Here's the comparison with deployment of current master to the same network
Screenshot at 2022-02-18 14-28-29
Screenshot at 2022-02-18 14-28-52
Screenshot at 2022-02-18 14-29-00

garbageslam:
it looks maybe a little worse to me towards the end of the slam, but peak is still ~40 tps, not sure if the rest of the graph is statistically significant

garbageslam:
i'm going to try deploying this PR a second time

eranrund (approved):
LGTM. I think it's okay to merge even if performance is a bit degraded because this is an important piece in many upcoming changes and releases. It would be great if we could find the time to do some profiling of stuff in the enclave and consensus-service. It's been awhile since we went through that exercise..

garbageslam:
Here are the results from the second deploy (of this PR):
Screenshot at 2022-02-18 16-05-06
Screenshot at 2022-02-18 16-05-14
Screenshot at 2022-02-18 16-05-28

eranrund:
20s tx completion time is a pretty big difference than 15s I am used to seeing...

garbageslam:
@eranrund yeah i'm not sure what could be causing this, maybe its because we are holding the mutex for the fees for a longer period of time?

garbageslam:
maybe we should try replacing the `mutex<option` for the config with a `oncecell`

eranrund:
> @eranrund yeah i'm not sure what could be causing this, maybe its because we are holding the mutex for the fees for a longer period of time?

Interesting, maybe?
You could try moving this https://github.com/garbageslam/mobilecoin/blob/block-version-aware-tx-builder-validation/consensus/enclave/impl/src/lib.rs#L367-L371 into a scoped block and cloning the stuff you need outside of the scope (just the minimum fee and block version)
Should be an easy test and seems like thats a good change regardless.

garbageslam:
I just made a quick patch like this:

https://github.com/garbageslam/mobilecoin/pull/4

and pushed it to build, let's see what happens

jcape (approved):
This generally looks good to me, the only real question is what BlockVersion::default() should actually default to---it feels to me like it should be `ONE` (or `MIN`, or whatever), not `MAX`, but I'm also not sure where it's actually used...

ledger/db/src/lib.rs
```diff
@@ -743,13 +743,18 @@ mod ledger_db_test {
...
+        compute_block_id, membership_proofs::compute_implied_merkle_root, BlockVersion,
+    };
     use mc_util_from_random::FromRandom;
     use rand::{rngs::StdRng, SeedableRng};
     use rand_core::RngCore;
     use tempdir::TempDir;
     use test::Bencher;
 
+    // TODO: Should these tests run over several block versions?
+    const BLOCK_VERSION: BlockVersion = BlockVersion::ONE;
```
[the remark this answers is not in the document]
garbageslam: In my search I see a lot of uses of `BLOCK_VERSION` in this file and no uses of `BlockVersion::ONE`. If you prefer we can mass replace them

consensus/enclave/api/src/config.rs
```diff
@@ -0,0 +1,146 @@
...
+    /// The block version that this enclave will be applying rules for and
+    /// publishing
+    pub block_version: BlockVersion,
+}
+
+impl Default for BlockchainConfig {
+    fn default() -> Self {
+        Self {
+            fee_map: FeeMap::default(),
+            block_version: BlockVersion::MAX,
```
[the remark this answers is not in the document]
garbageslam: Remoun had suggested `::MAX` in earlier review.
I think the default doesn't actually matter here, the actual default in the enclave is `None` basically, and the struct_opt default I think is `1` for backwards compat

consensus/enclave/api/src/fee_map.rs
```diff
@@ -2,34 +2,28 @@
...
-use mc_common::ResponderId;
-use mc_crypto_digestible::{DigestTranscript, Digestible, MerlinTranscript};
+use mc_crypto_digestible::Digestible;
 use mc_transaction_core::{tokens::Mob, Token, TokenId};
 use serde::{Deserialize, Serialize};
 
 /// A thread-safe object that contains a map of fee value by token id.
-#[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq)]
+#[derive(Clone, Debug, Deserialize, Digestible, Eq, Hash, PartialEq, Serialize)]
 pub struct FeeMap {
```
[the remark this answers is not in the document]
garbageslam: Probably but lets do it in a subsequent PR, it's kind of orthogonal to this PR, and there's a lot more token config stuff coming down the pipeline

garbageslam:
I ran it through deploy again -- I think it's possible that when I tested the once-cell change i didn't deploy the commit I thought I was deploying. These numbers look better (after merging the once-cell part to the PR)

Screenshot at 2022-02-19 12-06-14
Screenshot at 2022-02-19 12-06-27
Screenshot at 2022-02-19 12-06-37

PR #1471, opened 2022-02-10
Add jitter to most retry iterators

remoun:
### Motivation

When we have multiple instances of a program that includes retries, there is a risk of having many retries synchronize, creating a kind of thundering herd problem when attempting to recover from transient errors. 

### In this PR
* Add a `.map(delay::jitter)` call to all call sites that create a delay iterator for `retry` calls with a remote connection.

### DIFF
--- a/connection/src/sync.rs
+++ b/connection/src/sync.rs
@@ -148,7 +148,9 @@ macro_rules! impl_sync_connection_retry {
             stringify!($func),
             stringify!($iter)
         );
-        $crate::_retry::retry($iter, || $crate::_retry_wrapper!($obj.$func()))
+        $crate::_retry::retry($iter.map($crate::_retry::delay::jitter), || {
+            $crate::_retry_wrapper!($obj.$func())
+        })
     }};
     ($obj:expr, $logger:expr, $func:ident, $iter:expr, $arg1:expr) => {{
         $crate::_trace_time!(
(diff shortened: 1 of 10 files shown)


eranrund (approved):

jcape (commented):
Am I reading this correctly? Does this mean it could be `secs * (0..f64::MAX)`?
```
/// Apply full random jitter to a duration. (need `random` feature)
pub fn jitter(duration: Duration) -> Duration {
    let jitter = random::<f64>();
    let secs = ((duration.as_secs() as f64) * jitter).ceil() as u64;
    let nanos = ((f64::from(duration.subsec_nanos())) * jitter).ceil() as u32;
    Duration::new(secs, nanos)
}
```

remoun:
> Am I reading this correctly? Does this mean it could be `secs * (0..f64::MAX)`?

Fair question! The[`random` docs](https://docs.rs/rand/latest/rand/fn.random.html) state "Floating point types (f32 and f64): Uniformly distributed in the half-open range [0, 1)"

garbageslam:
@james I think the docu says, `random::<f64>()` is in the range (0,1): https://rust-random.github.io/rand/rand/fn.random.html

jcape (approved):

jcape:
LGTM, then :-)

Issue #1379, opened 2022-01-27
Fog enclave connections should have a retry iterator, rather than hard-coding milliseconds


jcape:
This review comment suggested to surface the retry parameters to the clients that use the fog key image, merkle proof, and fog view connection object: https://github.com/mobilecoinfoundation/mobilecoin/pull/1163#discussion_r746132110

This is a good comment and we should definitely do this, it's just a fair amount of plumbing and having some form of retries in there makes more tests more reliable and makes the test client more reliable for diagnosing deployment issues, so I wanted to fix it in a later PR.

This will be deprecated by #1378, but it's a good cleanup to have in the interim.

PR #1479, opened 2022-02-12
Use configurable retries on all fog connections, errors carry uris

garbageslam:
This addresses connectivity issues we had in fog during testnet
deploy today -- the fog ledger "untrusted txo" endpoint didn't
actually have retries, and it was flaky.

It addresses requests from jason for all these connectivity errors
to carry URIs with them for ease of debugging.

It addresses requests from the earlier PR about fog retries, to
make all the retries configurable at runtime.

(The main reason I didn't in that PR was that I thought I would have to
pipe the retry config through the view connection trait. But I realized
I can pass it to the constructor of the connection object which is simpler.)


eranrund (approved):

garbageslam:
This will resolve #1379

util/grpc/src/retry_config.rs
```diff
@@ -0,0 +1,38 @@
...
+        Self {
+            grpc_retry_count: 3,
+            grpc_retry_millis: 20,
+        }
+    }
+}
+
+impl GrpcRetryConfig {
+    /// Get a duration iterator for use with retry crate based on this config
+    pub fn get_retry_iterator(&self) -> Box<dyn Iterator<Item = Duration>> {
```
remoun, line 31: IIUC this method can return `impl Iterator<Item = Duration>` instead of a `Box`

fog/ledger/connection/src/key_image.rs
```diff
@@ -59,9 +61,10 @@ impl FogKeyImageGrpcClient {
         };
 
         let response: CheckKeyImagesResponse =
-            retry(Fixed::from_millis(100).take(5).map(jitter), || {
+            retry(self.grpc_retry_config.get_retry_iterator(), || {
```
remoun, line 64: How about we add a helper that forwards the call to `retry`?

Cals would then look like:
```rust
self.grpc_retry_config.retry(|| { ... })
```

util/grpc/src/retry_config.rs
```diff
@@ -0,0 +1,38 @@
...
+        Box::new(
+            delay::Fixed::from_millis(self.grpc_retry_millis)
+                .take(self.grpc_retry_count)
+                .map(delay::jitter),
+        )
+    }
+}
```
remoun, line 38: How about we add a helper that forwards the call to `retry`?

Cals would then look like:
```rust
self.grpc_retry_config.retry(|| { ... })
```


```suggestion
  pub fn retry<O, R, E, OR>(operation: O) -> Result<R, retry::Error<E>> 
  where
      O: FnMut() -> OR,
      OR: Into<retry::OperationResult<R, E>>, {
        retry::retry(self.get_retry_iterator, operation)
      }
}
```

util/grpc/src/retry_config.rs
garbageslam: I don't think so, I think that works on arguments to functions but not return values from functions.

We could make struct that implements `Iterator<Duration>` and return that to avoid the Box, that's what they do in the standard library for functions that return iterators. I just felt like the optimization is unnecessary.

fog/ledger/connection/src/key_image.rs
garbageslam: That's a good idea, will do

util/grpc/src/retry_config.rs
remoun: I'm pretty sure impl in return types is stabilized: https://rust-lang.github.io/impl-trait-initiative/explainer/rpit.html

I recommend you try it. Either way, I'm not attached to doing this, as the Box overhead is likely negligible compared to the actual RPCs/calls.

fog/view/connection/src/lib.rs
```diff
@@ -70,10 +95,35 @@ impl FogViewConnection for FogViewGrpcClient {
...
+            let retry_config = self.grpc_retry_config;
+            retry_config
+                .retry(|| {
+                    self.conn
+                        .retriable_encrypted_enclave_request(&req, &aad_bytes)
+                })
+                .map_err(|err| Error {
+                    uri: self.uri.clone(),
+                    error: err,
+                })
```
remoun, line 107: Optional/Equivalent: Rename the closure param for succinctness.
```suggestion
                .map_err(|error| Error {
                    uri: self.uri.clone(),
                    error,
                })
```

fog/view/connection/src/lib.rs
```diff
@@ -1,5 +1,10 @@
 // Copyright (c) 2018-2021 The MobileCoin Foundation
 
+//! A client object that creates an attested connection with a fog view enclave
+//! mediated over grpc. Can be used to make high-level requests.
+
+#![deny(missing_docs)]
```
remoun, line 6: Nice!

util/grpc/src/retry_config.rs
```diff
@@ -0,0 +1,45 @@
+use retry::delay;
+use serde::Serialize;
+use std::time::Duration;
+use structopt::StructOpt;
+
+/// An object which represents a retry policy for retriable errors for a grpc
+/// connection
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Serialize, StructOpt)]
```
remoun, line 8: Basic question: What would we gain from supporting Serialize but not Deserialize?

remoun (approved):

PR #1493, opened 2022-02-15
Remove stale benchmarks

remoun:
It doesn't look like `add_user` ever worked. We can redevelop `add_etxout` as needed.


garbageslam (approved):

eranrund (approved):

garbageslam:
I guess that there is considerable overlap between this benchmark and the fog ingest load testing target (which is a bin target and not a bench) so we may have just forgotten about the benchmarks

PR #1521, opened 2022-02-19
bump repo version to v1.2.0-pre1

garbageslam:
`-pre0` was several months ago and we are going to restart the 1.2 release cycle with a new branch.

i thought it made sense to bump in master first and then branch, since we earlier bumped master to `-pre0` it seems

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2275,7 +2275,7 @@ checksum = "7fc7aa29613bd6a620df431842069224d8bc9011086b1db4c0e0cd47fa03ec9a"
 
 [[package]]
 name = "libmobilecoin"
-version = "1.2.0-pre0"
+version = "1.2.0-pre1"
 dependencies = [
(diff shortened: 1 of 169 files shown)


garbageslam:
@jcape regarding the branching strategy here:

I guess we already have a `release-1.2.0` branch, and I feel like it's not a good idea to delete it. Maybe it's okay to tag the tip of it with `v1.2.0-pre0` or something, and then remake the branch pointing at this commit?

We could also just decide that, the version we are starting now is `1.2.1`, and call the new branch `release-1.2.1`

Lmk what you think

garbageslam:
Maybe in master we should make it like `1.2.0-pre1+dev` or something

jcape (approved):
LGTM. In terms of versioning, I think branching after this bump and then bumping master to 1.3.0-pre0 would clearly distinguish them.

PR #1541, opened 2022-02-25
Replace structopt with Clap v3, and add support for env overrides for all flags

remoun:


consensus/service/src/config/mod.rs
```diff
@@ -26,96 +28,83 @@ pub struct Config {
...
     pub client_auth_token_max_lifetime: Duration,
 
     /// The location for the network.toml/json configuration file.
-    #[structopt(long = "tokens", parse(from_os_str))]
+    #[clap(long = "tokens", env, parse(from_os_str))]
     pub tokens_path: Option<PathBuf>,
 
     /// The configured block version
-    #[structopt(long, env = "MC_BLOCK_VERSION", default_value = "1")]
+    #[clap(long, env, default_value = "1", parse(try_from_str = parse_block_version))]
```
remoun, line 104: This changes the env variable from `MC_BLOCK_VERSION` to `BLOCK_VERSION`

FYI @jgreat

garbageslam:
Can we also port this to 1.2.0 branch? It might simplify deploymebt if there isn't config churn?

jcape:
Is there an an actual reason to do that? I'd prefer if we didn't change the block version env rename (IMO we should prefix everything with `MC_`)

jcape:
> Can we also port this to 1.2.0 branch? It might simplify deployment if there isn't config churn?


Is there an an actual reason to do that? I'd prefer if we didn't change the block version env rename (IMO we should prefix everything with MC_)

consensus/service/src/config/mod.rs
jcape: Could we leave this as-is?

jcape: Could we flip this and make everything that's newly `env`'d `MC_`?

remoun: `clap` does not support this, so we'd have to spell out the env var for each flag. I filed https://github.com/clap-rs/clap/issues/3517 requesting support for a prefix for env vars.

jcape: Sure, but I'm OK defining them as MC_FOO everywhere, just because they don't exist now, and it will be painful to change it later.

jgreat:
This is awesome and will reduce complexity in the long run, but I don't think we need to backport this into 1.2.  We'll need some time to refactor helm charts and deployments to take advantage of not needing to map configs to command line args.

consensus/service/src/config/mod.rs
remoun: Switched to manually specifying MC_FOO everywhere.

jcape (changes requested):
Looks good, but I'd prefer if the test vectors string-as-errors thing was reverted for reasons described in situ.

crypto/x509/test-vectors/src/main.rs
```diff
@@ -26,29 +15,29 @@ enum PathKind {
 }
 
 impl FromStr for PathKind {
-    type Err = PathKindParseError;
+    type Err = String;
```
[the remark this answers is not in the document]
remoun: Good catch! Fixed with `arg_enum`, which provides `from_str` for us.

FWIW I started this change when I was grappling with a lot of compile errors, and didn't finish fixing this one.

jcape (approved):

common/src/responder_id.rs
```diff
@@ -1,4 +1,5 @@
-// Copyright (c) 2018-2021 The MobileCoin Foundation
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+#![deny(missing_docs)]
```
garbageslam, line 2: nit: this usually goes in `lib.rs`

consensus/service/src/config/mod.rs
```diff
@@ -1,22 +1,24 @@
 // Copyright (c) 2018-2022 The MobileCoin Foundation
+#![deny(missing_docs)]
```
garbageslam, line 2: nit: this usually goes in `lib.rs`, but it's a lot of work to do that for all of consensus service, so i see if you want to punt on that

util/uri/src/uri.rs
```diff
@@ -1,4 +1,5 @@
-// Copyright (c) 2018-2021 The MobileCoin Foundation
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+#![deny(missing_docs)]
```
garbageslam, line 2: nit: this could go in `lib.rs` without much work probably

garbageslam (approved):
will re-approve after rebase

jgreat (approved):
This is good work.  I like that all the ENV values are prefixed with `MC`, looks like that was a pain.

The place I'd like to see follow up on is having comments/annotations/help for some of the config options.  It would be really helpful to have examples that shows how to format configuration for things that take complex values (lists or maps).  For example there are lots of places that take a list of `--peer` options, but I don't think there are any examples on how to specify that as a cli option or an ENV var.

consensus/service/src/config/mod.rs
```diff
@@ -26,96 +28,91 @@ pub struct Config {
...
     pub client_auth_token_secret: Option<[u8; 32]>,
 
     /// Maximal client authentication token lifetime, in seconds (only relevant
     /// when --client-auth-token-secret is used. Defaults to 86400 - 24
     /// hours).
-    #[structopt(long, default_value = "86400", parse(try_from_str=parse_duration_in_seconds))]
+    #[clap(long, default_value = "86400", parse(try_from_str = parse_duration_in_seconds), env = "MC_CLIENT_AUTH_TOKEN_MAX_LIFETIME")]
     pub client_auth_token_max_lifetime: Duration,
 
     /// The location for the network.toml/json configuration file.
```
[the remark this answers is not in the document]
jgreat: line 97/98 --tokens is the question

fog/distribution/src/config.rs
```diff
@@ -10,67 +12,70 @@ use mc_connection::{
...
     /// URLs to use for transaction data.
     ///
     /// For example: https://s3-us-west-1.amazonaws.com/mobilecoin.chain/node1.master.mobilecoin.com/
-    #[structopt(long, default_value = "fog_keys")]
+    #[clap(long, default_value = "fog_keys", env = "MC_FOG_KEYS_SUBDIR")]
     pub fog_keys_subdir: String,
 
-    #[structopt(flatten)]
+    /// Peers.
+    #[clap(flatten)]
```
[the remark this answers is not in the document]
remoun: It's a nested structure that is inlined into this config. 

[That struct defines one flag, `peers`, which is renamed to `--peer`](https://github.com/mobilecoinfoundation/mobilecoin/blob/fbf5cdcd25e0203a7b0c93ed72d9fe14463125bb/mobilecoind/src/config.rs#L266). It can be specified multiple times, or passed a CSV of URIs

fog/overseer/server/src/config.rs
```diff
@@ -1,29 +1,35 @@
...
     pub overseer_listen_port: u16,
 
+    /// TODO: Make this an environment variable that can be dynamically
+    /// refreshed. This will allow ops to have one Fog Overseer instance that
+    /// can look at the new Fog Ingest cluster during the blue / green
+    /// deployment.
+    ///
     /// gRPC listening URIs for client requests.
-    #[structopt(long, env, use_delimiter = true)]
+    #[clap(long, use_value_delimiter = true, env = "MC_INGEST_CLUSTER_URIS")]
```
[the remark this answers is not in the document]
remoun: It can be specified multiple times on the command line, or as CSV via env or command-line args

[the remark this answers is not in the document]
remoun: Relevant Clap docs: https://docs.rs/clap/3.1.6/clap/struct.Arg.html#method.use_value_delimiter

common/src/responder_id.rs
remoun: I tried that, but this is a large package, which I don't want to tackle in this already large PR.

fog/distribution/src/config.rs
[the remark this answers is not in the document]
remoun: I actually ended up inlining this particular reference, since the definition in `mobilecoind` referenced another flag that is unique to that binary.

fog/ingest/server/src/config.rs
```diff
@@ -10,47 +12,46 @@ use mc_util_parse::parse_duration_in_seconds;
...
     pub client_listen_uri: FogIngestUri,
 
     /// gRPC listening URI for peer requests.
-    #[structopt(long)]
+    #[clap(long, env = "MC_PEER_LISTEN_URI")]
     pub peer_listen_uri: IngestPeerUri,
 
     /// List of all peers in this cluster
-    #[structopt(long, use_delimiter = true)]
+    #[clap(long, use_value_delimiter = true, env = "MC_PEERS")]
```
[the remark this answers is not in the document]
remoun: I added sample usages.

fog/overseer/server/src/config.rs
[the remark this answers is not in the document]
remoun: I added sample usages, too.

garbageslam (dismissed):

jcape (approved):

common/src/node_id.rs
```diff
@@ -51,7 +51,9 @@ impl From<KeyError> for NodeIDError {
 /// key
 #[derive(Clone, Serialize, Deserialize, Digestible)]
 pub struct NodeID {
+    /// The Responder ID for this node
     pub responder_id: ResponderId,
+    /// The public key for this node
```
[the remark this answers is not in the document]
jcape: Each node has multiple public keys, so it's good to disambiguate: one X25519 key for key exchange, one Ristretto key for block signing, and one Ed25519 key for SCP-message-signing---this is the third one.

fog/ingest/server/src/config.rs
```diff
@@ -10,47 +12,51 @@ use mc_util_parse::parse_duration_in_seconds;
...
+    #[clap(long, env = "MC_PEER_LISTEN_URI")]
     pub peer_listen_uri: IngestPeerUri,
 
     /// List of all peers in this cluster
-    #[structopt(long, use_delimiter = true)]
+    /// Sample usages:
+    ///     --peers mc://foo:123 --peers mc://bar:456
+    ///     --peers mc://foo:123,mc://bar:456
+    ///     env MC_PEERS=mc://foo:123,mc://bar:456
+
```
remoun, line 48:
```suggestion
```

fog/overseer/server/src/lib.rs
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) 2018-2022 The MobileCoin Foundation
 
 #![feature(proc_macro_hygiene, decl_macro)]
-#[deny(missing_docs)]
+
```
remoun, line 4: Accidental removal
```suggestion
#[deny(missing_docs)]
```

jcape (approved):

garbageslam (approved):

jcape (approved):

garbageslam (approved):

PR #1543, opened 2022-02-25
try to improve circle ci times by splitting fog ingest to its own job

garbageslam:
The run mc-fog-tests job currently takes 40 minutes

This does two things:
* Move fog ingest tests to their own job in circle ci, and make
  the fog tests job exclude fog ingest
* Use `--release` when running the fog ingest tests. This is because
  I think it will make the integration tests that run little fog
  ingest clusters a lot faster, since it will make the ORAM and
  all that much faster. Those integration tests probably make more
  sense to run in release mode anyways since then it will more
  closely resemble what we actually deploy.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -392,7 +392,23 @@ commands:
           command: |
             # tell the operating system to remove the file size limit on core dump files
             ulimit -c unlimited
-            cargo test --package "mc-fog-*" -j 4 --frozen --no-fail-fast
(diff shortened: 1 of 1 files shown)


garbageslam:
I wonder if we should run the consensus tests in release mode too, the slowest part of those are setting up the mesh networks etc.

iamalwaysuncomfortable:
I think it would be worth trying the consensus tests in release mode. Also, would there room for us to run parallel executors for some of these jobs?

iamalwaysuncomfortable (approved):
Looks good, maybe we can put them on separate executors for further speedup?

wjuan-mob (approved):

Issue #1482, opened 2022-02-14
Fog Overseer CD


jcape:
This is a tracking issue for getting Fog Overseer into our dev network and CD deployments.

samdealy:
(This issue was prematurely closed by GitHub)

samdealy:
Fog Overseer has been deployed to CD. There are some things that need to be addressed (e.g. start overseer as disableed before fog ingest is running and then enable it to ensure that it doesn't initiate automatic failover), but those will be part of a new issue.

PR #1545, opened 2022-02-25
Keep Fog Overseer Server alive when started via binary

samdealy:
### Motivation
Currently, the fog overseer shuts down soon after being started because the server's `rocket` instance is not kept in scope. This issue was surfaced locally using a local network and on the CD deployment. Fixing this should make overseer run successfully on CD.

Contributes to #1482.

### In this PR
* Keeps the `rocket` instance in scope and calls launch.

### DIFF
--- a/fog/overseer/server/src/bin/main.rs
+++ b/fog/overseer/server/src/bin/main.rs
@@ -41,5 +41,6 @@ fn main() {
             .port(config.overseer_listen_port)
             .unwrap();
 
-    server::initialize_rocket_server(rocket_config, overseer_state);
+    let rocket = server::initialize_rocket_server(rocket_config, overseer_state);
+    rocket.launch();
 }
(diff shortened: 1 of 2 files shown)


fog/overseer/server/src/bin/main.rs
```diff
@@ -41,5 +41,6 @@ fn main() {
             .port(config.overseer_listen_port)
             .unwrap();
 
-    server::initialize_rocket_server(rocket_config, overseer_state);
+    let rocket = server::initialize_rocket_server(rocket_config, overseer_state);
```
remoun, line 44: Consider annotating this helper method with `must_use`, e.g.

```rust
#[must_use = "Use with a Client or call launch"]
fn initialize_rocket_server(...)  {}
```

remoun (approved):

wjuan-mob (approved):

fog/overseer/server/src/bin/main.rs
samdealy: Done.

jcape (approved):

PR #1556, opened 2022-02-28
Replace datatest

### LINKED ISSUE #1369: Remove datatest
Our use of this (effectively unmaintained) crate has caused rust updates to be very painful, we should look to remove it.

One gotcha is that the "test vectors" it's creating/using are (were) being used to verify things in the iOS SDK, so there will be some coordination with the SDK team required.

remoun:
… with a custom `test_with_data` attribute macro.

`datatest` has repeatedly given us trouble when upgrading rust versions, due to its dependency on internal libraries.

Fixes #1369.


util/test-with-data/src/lib.rs
```diff
@@ -0,0 +1,99 @@
...
+///     F: Fn(T) -> (),
+/// {
+///     for case in source {
+///         test(case.into());
+///     }
+/// }
+/// ```
+#[proc_macro_attribute]
+pub fn test_with_data(attr: TokenStream, item: TokenStream) -> TokenStream {
+    let function = syn::parse_macro_input!(item as ItemFn);
```
garbageslam, line 49: This is cool, didn't know about this

garbageslam (approved):

eranrund (approved):

account-keys/src/account_keys.rs
```diff
@@ -462,12 +462,13 @@ impl AccountKey {
...
 mod account_key_tests {
     use super::*;
-    use alloc::boxed::Box;
     use core::convert::TryFrom;
-    use datatest::data;
     use mc_crypto_keys::RistrettoSignature;
-    use mc_test_vectors_account_keys::*;
+    use mc_test_vectors_account_keys::{
+        DefaultSubaddrKeysFromAcctPrivKeys, SubaddrKeysFromAcctPrivKeys,
+    };
```
jcape, line 469: 🙏🏻

PR #1575, opened 2022-03-04
Add Fog Local Network and Test Client to CI

### LINKED ISSUE #1572: Add Fog Local Network and Test Client to CI

samdealy:
### Motivation
As Chris wrote in the previous PR (#1173): 
This adds fog local network to CI, with fog distro using very few transactions. The point is to get test coverage that ensures that test client is able to spend change outputs.

Fixes #1572 

## In this PR
- Fixes previous issue with Fog Test Client. The local_network.py script used the old network fee of 10,000,000,000. The test client gets the fee from consensus, but when it failed to retrieve the fee it defaulted to the new network fee of 40,000,000.
- This manifested as an incorrect balance because the amount used for the expected balance used the new fee while the actual transaction used the old fee. This occurred because the fee was getting retrieved in two places. I fixed this by reusing the fee from the transaction in the balance check.
- I also had to update the old network fee to the new network fee because the test client would use the new fee if it couldn't get the fee from consensus, Then consensus would reject the transaction because the fee (40,000,000) was too low (< 10,000,000,000).

### Future Work
- Modify the Fog Local network README to describe how to run the network using  `./mob prompt`.

### DIFF
--- a/fog/sample-paykit/src/client.rs
+++ b/fog/sample-paykit/src/client.rs
@@ -324,7 +324,7 @@ impl Client {
             .tx_data
             .get_transaction_inputs(amount + Mob::MINIMUM_FEE, TARGET_NUM_INPUTS)?;
         let inputs: Vec<(OwnedTxOut, TxOutMembershipProof)> = self.get_proofs(&inputs)?;
-        let rings: Vec<Vec<(TxOut, TxOutMembershipProof)>> = self.get_rings(inputs.len(), rng)?;
+        let rings: Vec<Vec<(TxOut, TxOutMembershipProof)>> = self.get_rings(&inputs, rng)?;
 
         let tombstone_block = self.compute_tombstone_block()?;
 
@@ -444,26 +444,35 @@ impl Client {
     /// self.ring_size elements.
     fn get_rings<T: RngCore + CryptoRng>(
         &mut self,
-        num_rings: usize,
+        true_inputs: &[(OwnedTxOut, TxOutMembershipProof)],
         rng: &mut T,
     ) -> Result<Vec<Vec<(TxOut, TxOutMembershipProof)>>> {
         mc_common::trace_time!(self.logger, "MobileCoinClient.get_rings");
 
+        let true_input_indices: HashSet<u64> = true_inputs
+            .iter()
+            .map(|input| input.0.global_index)
+            .collect();
+
+        let num_rings = true_inputs.len();
         let num_requested = num_rings * self.ring_size;
         let sample_limit = self.tx_data.get_global_txo_count() as usize;
 
-        // Randomly sample `num_requested` TxOuts, without replacement.
-        if sample_limit < num_requested {
+        // Randomly sample `num_requested` TxOuts, without replacement, not using
+        // true_inputs
+        if sample_limit < num_requested + true_inputs.len() {
             return Err(Error::InsufficientTxOutsInBlockchain(
                 sample_limit,
-                num_requested,
+                num_requested + true_inputs.len(),
             ));
         }
 
         let mut sampled_indices: HashSet<u64> = HashSet::default();
         while sampled_indices.len() < num_requested {
-            let index = rng.gen_range(0..sample_limit);
-            sampled_indices.insert(index as u64);
+            let index = rng.gen_range(0..sample_limit) as u64;
+            if !true_input_indices.contains(&index) {
+                sampled_indices.insert(index);
+            }
         }
         let indices: Vec<u64> = sampled_indices.iter().cloned().collect();
         // FIXME: We are not sure whether this is a necessary parameter under ORAM.
--- a/fog/test-client/src/test_client.rs
+++ b/fog/test-client/src/test_client.rs
@@ -224,7 +224,7 @@ impl TestClient {
         &self,
         source_client: &mut Client,
         target_client: &mut Client,
-    ) -> Result<(Tx, u64), TestClientError> {
+    ) -> Result<(Tx, u64, u64), TestClientError> {
         self.tx_info.clear();
         let target_address = target_client.get_account_key().default_subaddress();
         log::debug!(
@@ -269,7 +269,7 @@ impl TestClient {
             block_count
         };
         self.tx_info.set_tx_propose_block_count(block_count);
-        Ok((transaction, block_count))
+        Ok((transaction, block_count, fee))
     }
 
     /// Waits for a transaction to be accepted by the network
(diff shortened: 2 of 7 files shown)


fog/test-client/src/test_client.rs
```diff
@@ -224,7 +224,7 @@ impl TestClient {
         &self,
         source_client: &mut Client,
         target_client: &mut Client,
-    ) -> Result<(Tx, u64), TestClientError> {
+    ) -> Result<(Tx, u64, u64), TestClientError> {
```
garbageslam, line 227: Can we add some docu explaining what this return value is?

fog/test-client/src/test_client.rs
```diff
@@ -623,6 +622,10 @@ impl TestClient {
 
         let start_time = Instant::now();
         for ti in 0..num_transactions {
+            // FIXME: Should not be needed
+            log::debug!(self.logger, "sleeping");
+            std::thread::sleep(Duration::from_millis(5000));
+
```
garbageslam, line 628: is this actually still needed?

tools/local-network/local_network.py
```diff
@@ -152,7 +152,7 @@ def __init__(self, name, node_num, client_port, peer_port, admin_port, admin_htt
         self.admin_http_gateway_port = admin_http_gateway_port
         self.peers = peers
         self.quorum_set = quorum_set
-        self.minimum_fee = 10_000_000_000
+        self.minimum_fee = 400_000_000
```
garbageslam, line 155: thank you

i wonder if there is a more maintainable way to do this

is the issue that sample paykit is not detecting the fee dynamically? maybe that is a different bug?

garbageslam: because like, we use the test client in prod to detect outages, and if it will break if the fee changes then that is bad

garbageslam: nm, reading the code again, i see that this is working i guess

garbageslam (changes requested):
only small comments:

* can we add docu for the extra return value
* are the sleeps actually still needed or was that just during investigation

fog/test-client/src/test_client.rs
samdealy: No, removed.

fog/test-client/src/test_client.rs
samdealy: Done.

garbageslam (approved):
Thank you!

tools/local-network/local_network.py
jcape: I opened https://github.com/mobilecoinfoundation/mobilecoin/issues/1582 for this

jcape (changes requested):

jcape:
Also: this *explodes* CI tiem to 3h

fog/sample-paykit/src/client.rs
```diff
@@ -444,26 +444,35 @@ impl Client {
...
+
+        let num_rings = true_inputs.len();
         let num_requested = num_rings * self.ring_size;
         let sample_limit = self.tx_data.get_global_txo_count() as usize;
 
-        // Randomly sample `num_requested` TxOuts, without replacement.
-        if sample_limit < num_requested {
+        // Randomly sample `num_requested` TxOuts, without replacement, not using
+        // true_inputs
+        if sample_limit < num_requested + true_inputs.len() {
```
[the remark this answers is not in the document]
samdealy: This is a hold over form @garbageslam's original PR. Chris, wwhat are your thoughts?

garbageslam:
> Also: this explodes CI time to 3h

Maybe we should be building with `--release`?

garbageslam:
never mind, we are using `--release`...

I think maybe the 10 seconds of sleeping was contributing to the CI time?

fog/sample-paykit/src/client.rs
```diff
@@ -444,26 +444,35 @@ impl Client {
...
+                num_requested + true_inputs.len(),
             ));
         }
 
         let mut sampled_indices: HashSet<u64> = HashSet::default();
         while sampled_indices.len() < num_requested {
-            let index = rng.gen_range(0..sample_limit);
-            sampled_indices.insert(index as u64);
+            let index = rng.gen_range(0..sample_limit) as u64;
+            if !true_input_indices.contains(&index) {
```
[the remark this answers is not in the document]
garbageslam: so, the reason I changed this code this way was, when running the test with a relatively small bootstrap, (which makes the test faster), I was getting problems where, my transactions were failing because the real input was being chosen as one of the real input's mixins. So the most obvious fix was to exclude all the real inputs from the sample. If we don't do something to prevent that, then the client can build an invalid transaction and the test can fail.

I think you may be right that a more subtle fix is better, where when sampling a ring for `x` we only exclude `x` as a mixin.

Since this is just used by the test client, I feel like it's not security critical. But to the extent that it's a model for real SDKs to crib from, I think your proposed change would be an improvement.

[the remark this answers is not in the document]
jcape: Cool, I think we're on the same page.

jcape:
The CI issue was due to some weird errors coming back when running this (tx is successfully submitted, but we also get an attestation failure caused by gRPC throwing ENANCE_YOUR_CALM/"too many pings", which is bonkers).

After the attestation failure, it gets itself into an infinite loop trying to re-send the same transaction which contains a spent key image.

samdealy:
> > Also: this explodes CI time to 3h
> 
> Maybe we should be building with `--release`?

The 3 hours is misleading- the CI was continuing to run because the tx submit kept failing and looping infinitely. After this fix I just pushed, this infinite loop should not happen. When the CI succeeds, the new step that this PR adds doesn't add that much time.

fog/sample-paykit/src/client.rs
```diff
@@ -444,26 +444,35 @@ impl Client {
...
 
-        // Randomly sample `num_requested` TxOuts, without replacement.
-        if sample_limit < num_requested {
+        // Randomly sample `num_requested` TxOuts, without replacement, not using
+        // true_inputs
+        if sample_limit < num_requested + true_inputs.len() {
             return Err(Error::InsufficientTxOutsInBlockchain(
                 sample_limit,
-                num_requested,
+                num_requested + true_inputs.len(),
```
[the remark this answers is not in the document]
samdealy: See chris's comment.

garbageslam (approved):

fog/test-client/src/test_client.rs
```diff
@@ -220,11 +220,16 @@ impl TestClient {
...
+    /// Returns:
+    ///  * Tx: the transaction that represents the transfer.
+    ///  * u64: the block count at which the transaction was submitted.
+    ///  * u64: the fee associated with the transaction.
     fn transfer(
         &self,
         source_client: &mut Client,
         target_client: &mut Client,
-    ) -> Result<(Tx, u64), TestClientError> {
+    ) -> Result<(Tx, u64, u64), TestClientError> {
```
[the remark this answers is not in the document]
samdealy: Done.

garbageslam (approved):

fog/sample-paykit/src/client.rs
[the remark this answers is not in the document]
jcape: @samdealy I think the upshot here is that this code should get modified to test for real inputs. I'm fine if that happens in a follow-up PR, but please create an issue for this.

jcape (approved):

PR #1583, opened 2022-03-08
Make Fog Overseer disabled on start up

samdealy:
### Motivation

Previously, Fog Overseer started up in the "enabled" state, which means that it monitors the Fog Ingest cluster for incorrect state. This was an issue because Overseer was started right when Fog Ingest started, and the ingest nodes didn't have time to activate. In this state, where no active nodes exist, Overseer initiated automatic failover, which is not desirable because ingest node was in the process of activating.

### In this PR
* Starts Fog Overseer in the disabled state. The new deployment flow will involve activating Overseer after the Fog Ingest cluster has reached the active state.

### DIFF
--- a/fog/overseer/server/src/service.rs
+++ b/fog/overseer/server/src/service.rs
@@ -61,7 +61,7 @@ where
             logger,
             overseer_worker: None,
             recovery_db,
-            is_enabled: Arc::new(AtomicBool::new(true)),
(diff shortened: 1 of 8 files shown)


garbageslam (approved):
LGTM

joekottke (approved):

Issue #1573, opened 2022-03-04
Streaming gRPC wrappers for Block Streaming


remoun:
Implement the API defined in #1568, and provide async wrappers that provide a Stream of blocks.

PR #1584, opened 2022-03-08
Add GrpcBlockSource

remoun:
### Motivation

#1573

### In this PR
* Add `GrpcBlockSource`, which wraps the gRPC API introduced in https://github.com/mobilecoinfoundation/mobilecoin/pull/1568 with the `BlockSource` trait also introduced in that PR.

### Future Work
* Implement server streaming API.
* Other sources and sinks.


ledger/streaming/api/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+displaydoc = { version = "0.2", default-features = false }
+futures = { version = "0.3.21" }
+grpcio = { version = "0.9.0" }
+protobuf = "2.22.1"
+
+[build-dependencies]
+mc-util-build-grpc = { path = "../../../util/build/grpc" }
+mc-util-build-script = { path = "../../../util/build/script" }
+
+cargo-emit = "0.2.1"
```
iamalwaysuncomfortable, line 25: Should we merge the cargo.toml so that there's only 1 for streaming?

ledger/streaming/client/src/source.rs
```diff
@@ -0,0 +1,120 @@
...
+    /// A logger object
+    logger: Logger,
+}
+
+impl GrpcBlockSource {
+    pub fn new(channel: grpcio::Channel, logger: Logger) -> Self {
+        let client = LedgerUpdatesClient::new(channel);
+        Self { client, logger }
+    }
+}
```
iamalwaysuncomfortable, line 27: Perhaps we should have some kind of sense of uri stored in GRPC block source. At least another field specifying it?  This seems important for enabling logging and record keeping, I am also not %100 certain on this, but I think the ResponderIDs are included in the Merlin Transcripts.

What might be the options for this?

ledger/streaming/api/Cargo.toml
remoun: I've opted to keep them separate since we expect the server(s) and clients to live in different binaries, with different deps.

ledger/streaming/api/proto/streaming_blocks.proto
```diff
@@ -0,0 +1,34 @@
...
+message BlockWithQuorumSet {
+    /// The block data.
+    blockchain.ArchiveBlock block = 1;
+    /// QuorumSet when the block was externalized.
+    quorum_set.QuorumSet quorum_set = 2;
+    /// Attestation Verification report.
+    external.VerificationReport report = 3;
+}
+
+message SubscribeResponse {
```
iamalwaysuncomfortable, line 27: On second thought this should probably include the NodeID so that a proper accounting of quorum can be done when doing block verification. The sig is representative, but it makes it potentially difficult to get an accounting of what is going wrong if there are ever issues

ledger/streaming/api/src/traits.rs
```diff
@@ -0,0 +1,8 @@
+use crate::Result;
+use mc_transaction_core::BlockData;
+
+pub trait BlockSource {
+    type BlockStream: futures::Stream<Item = Result<BlockData>>;
```
iamalwaysuncomfortable, line 5: We potentially want to update this to the following: 

```Rust
type BlockStream: futures::Stream<Item = Result<(NodeID, BlockData)>>
```

iamalwaysuncomfortable (commented):
I know it was originally thought that NodeID might not be needed, but I think it makes handling the data on a consumer side a lot easier to account with. Especially in the case that there's a potentially bad actor in the network.

ledger/streaming/client/src/source.rs
remoun: Added a URI parameter. I opted for an opaque impl as the parameter, since we haven't converged on where the server will live.

ledger/streaming/api/proto/streaming_blocks.proto
remoun: Why NodeID and not the whole `BlockWithQuorumSet`?

remoun:
I had to rebase on top of #1592 for the tests to pass; this PR will be smaller after that PR is merged.

ledger/streaming/api/proto/streaming_blocks.proto
iamalwaysuncomfortable: Yes, that'll include the members correct, but is it easily identifiable which member the node is in that set in the stream being received without the node id?

Issue #1576, opened 2022-03-06
Maybe there is a bug in fog/ocall_oram_storage/trusted/src/lib.rs


AmbitionXiang:
Hi, there.

At line 331, `if idx == first_treetop_index` seems to be  `if idx == first_treetop_index - 1`? Otherwise, the check against trusted roots is never encountered.

garbageslam:
I am still investigating and started working on this -- I think you are right that there is an off by one mistake that makes the merkle root check not happen. I will finish writing a patch and tests soon.

Thanks for reporting this issue!

AmbitionXiang:
You're welcome!

PR #1585, opened 2022-03-08
add tests for untrusted tampering with data and metadata

garbageslam:
This is a test-driven approach to investigating and fixing the issue
https://github.com/mobilecoinfoundation/mobilecoin/issues/1576

These tests confirm that the trusted side indeed does not always
detect when untrusted tampers with the memory, at this revision
I have these results:

```
running 5 tests
test helpers::exercise_oram_storage_clear_data ... FAILED
test helpers::exercise_oram_storage_hammer_data ... FAILED
test helpers::exercise_oram_storage_hammer_metadata ... ok
test helpers::exercise_oram_storage_clear_metadata ... FAILED
test helpers::exercise_oram_storage_shims ... ok
```

So the shims work normally when the memory is not tampered with,
and the code detects when the untrusted metadata is hammered.
But it does not detect when it is cleared, or when the data is
tampered with.

This is likely because of the issue point out in #1576, in the
next commits we will fix it and get all these tests to pass.

### DIFF
--- a/fog/ocall_oram_storage/trusted/src/lib.rs
+++ b/fog/ocall_oram_storage/trusted/src/lib.rs
@@ -84,15 +84,13 @@ lazy_static! {
     // Yogseh Swami "Intel SGX Remote Attestation is not sufficient" discusses such issues in detail:
     // https://www.blackhat.com/docs/us-17/thursday/us-17-Swami-SGX-Remote-Attestation-Is-Not-Sufficient-wp.pdf
     //
-    // A simliar measure is in place for the trusted side fog-recovery-db-iface.
-    //
     // It is not clear that this is actually needed to prevent problems,
     // we just haven't done a detailed anaysis.
     //
     // If you do such an analysis and conclude it's safe, you could remove this and leave a code comment.
     // However, the cost of this is likely very low since the mutex is uncontended -- our real enclaves
     // only have one OMAP object, and its API is &mut, so the caller must wrap it in a mutex anyways,
-    // and this is unlikely to change.
+    // and this is unlikely to change anytime soon. (Maybe we will have parallel ORAM someday???)
     static ref OCALL_REENTRANCY_MUTEX: Mutex<()> = Mutex::new(());
 }
 
@@ -166,14 +164,11 @@ where
         } else {
             // eprintln!("count = {} > TREETOP_MAX_COUNT = {}, we must allocate in
             // untrusted", count, treetop_max_count);
-            unsafe {
-                allocate_oram_storage(
-                    count - treetop_max_count,
-                    DataSize::U64,
-                    MetaSize::U64 + ExtraMetaSize::U64,
-                    &mut allocation_id,
-                );
-            }
+            allocation_id = helpers::allocate_ocall(
+                count - treetop_max_count,
+                DataSize::U64,
+                MetaSize::U64 + ExtraMetaSize::U64,
+            );
             if allocation_id == 0 {
                 panic!("Untrusted could not allocate storage! count = {}, data_size = {}, meta_size + extra_meta_size = {}",
                        count - treetop_max_count,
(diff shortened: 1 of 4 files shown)


fog/ocall_oram_storage/trusted/src/lib.rs
```diff
@@ -571,3 +597,775 @@ extern "C" {
...
+        TREETOP_CACHING_THRESHOLD_LOG2.store(10, Ordering::SeqCst);
+
+        let mut st = StorageType::new(131072, &mut rng);
+
+        let mut data_scratch = vec![A64Bytes::<U1024>::default(); 17];
+        let mut meta_scratch = vec![A8Bytes::<U16>::default(); 17];
+
+        // Write 1's along branch at 131072 - 1
+        {
+            st.checkout(131072 - 1, &mut data_scratch, &mut meta_scratch);
```
remoun, line 749: This `131072` repeats a lot; let's make it a constant and document its significance -- why 128 KiB?

remoun (commented):
Thanks for adding these tests!

fog/ocall_oram_storage/trusted/src/lib.rs
garbageslam: i think i just needed a number

fog/ocall_oram_storage/trusted/src/lib.rs
```diff
@@ -351,6 +333,16 @@ where
...
+
+            // Check the last hash with the trusted merkle root storage
+            // This unwrap is valid because if first_treetop_index is zero, then we didn't
+            // enter this `if`
+            let (last_idx, last_hash) = last_hash.expect("should not be empty at this point");
+            assert!(
+                bool::from(self.trusted_merkle_roots[last_idx as usize].ct_eq(&last_hash)),
+                "authentication failed, trusted merkle root mismatch at {}",
+                last_idx
+            );
```
garbageslam, line 345: @wjuan-mob this is what i thought to do to fix it at first, but the tests are still failing, so this can't be the whole story

lets think about it some more on monday

wjuan-mob: Still working through this. So far my suspicion is that this fix is correct, but that the test is wrong somewhere. I went through and printed out what the merkle root is after "clear" and I got the following which seems to suggest we may have cleared some hashes over zealously: 
```
cargo:warning=First treetop index 16
cargo:warning=trusted merkle root: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0], lash_hash: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
```

fog/ocall_oram_storage/trusted/src/lib.rs
```diff
@@ -571,3 +596,775 @@ extern "C" {
...
+            assert_eq!(meta_scratch[16], a8_bytes(3));
+
+            st.checkin(131072 - 1, &mut data_scratch, &mut meta_scratch);
+        }
+    }
+
+    // Test what happens when we start screwing with the untrusted memory,
+    // by hammering it with junk value
+    #[test]
+    fn exercise_oram_storage_hammer_data() {
```
wjuan-mob, line 894: It's really odd to me that this isn't failing on line 314. I must be missing something about the way we calculate our hashes. If we go through and hammer the data, shouldn't the newly calculated hash be different throughout the tree and not just at the merkle root?

wjuan-mob: On further investigation. This is because for some reason, hammer meta data is getting that the self.meta_scratch_buffer[idx] == Default::default() is false, whereas hammer data is having this condition return true, which seems to be what's causing the last hash to be coming out as 0s.

PR #1614, opened 2022-03-12
Block Streaming: Ledger Sink Component

iamalwaysuncomfortable:
## Motivation
As part of the Block Streaming mcip we are composing a bunch of simple streams which each take on a specific domain of functionality. This particular PR represents the final link in this chain and assumes a validated stream of blocks from a consensus node.

It's expected that ahead of this node, there's a validation streaming component that is collating streams from several peers and doing all the necessary validations (SCP, Block Content Verification, etc.)

## In this PR
Addittions/Changes:

* Adds the beginning of a config file to launch the client
* Adds a Block Sink that takes a single stream of validated blocks and pushes them to a sync thread for ingest into ledgerdb
* Suggests some changes to Block streaming API namings (notably Error & Result as those override keywords in the Rust 
prologue)

Ticket Status: In Progress
* Needs fuller test coverage
* Needs more error handling

## Future Work
* The sink should be self-healing (i.e. try to re-establish connection) rather than crashing if a stream dies, but this is somewhat dependent on the upstream component.
* Validation stream needs to be added directly in front of this
* This chose tokio, but generic executors may be able to be used
* Blockstreaming client components need to be decided

### DIFF
--- a/ledger/streaming/api/src/error.rs
+++ b/ledger/streaming/api/src/error.rs
@@ -20,6 +20,9 @@ pub enum Error {
 
     /// Invalid signature: {0}
     Signature(String),
+
+    /// Ledger DB Access Error
+    DBError,
 }
 
 impl std::error::Error for Error {}
--- a/ledger/streaming/api/src/error.rs
+++ b/ledger/streaming/api/src/error.rs
@@ -22,7 +22,7 @@ pub enum Error {
     Signature(String),
 
     /// Ledger DB Access Error
-    DBError,
+    DBAccess,
 }
 
 impl std::error::Error for Error {}
--- a/ledger/streaming/client/Cargo.toml
+++ b/ledger/streaming/client/Cargo.toml
@@ -13,6 +13,7 @@ test_utils = []
 [dependencies]
 mc-common = { path = "../../../common" }
 mc-crypto-keys = { path = "../../../crypto/keys" }
+mc-ledger-db = { path = "../../../ledger/db" }
 mc-ledger-streaming-api = { path = "../api" }
 mc-transaction-core = { path = "../../../transaction/core" }
 mc-util-grpc = { path = "../../../util/grpc" }
@@ -21,6 +22,7 @@ mc-util-uri = { path = "../../../util/uri" }
 displaydoc = "0.2"
 futures = "0.3"
 grpcio = "0.10"
+pin-project = "1.0.10"
 
 [build-dependencies]
 mc-util-build-grpc = { path = "../../../util/build/grpc" }
@@ -30,6 +32,7 @@ cargo-emit = "0.2"
 
 [dev-dependencies]
 mc-common = { path = "../../../common", features = ["loggers"] }
+mc-ledger-db = { path = "../../../ledger/db", features = ["test_utils"] }
 mc-ledger-streaming-api = { path = "../api", features = ["test_utils"] }
 mc-util-from-random = { path = "../../../util/from-random" }
 
--- a/ledger/streaming/client/src/error.rs
+++ b/ledger/streaming/client/src/error.rs
@@ -1,0 +1,66 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
+//! MobileCoin stream errors to manage client future & stream operation
+
+use displaydoc::Display;
+use mc_ledger_db::{Error as LedgerDBError, Ledger};
+use mc_transaction_core::BlockIndex;
+use std::sync::{RwLockReadGuard, RwLockWriteGuard, TryLockError};
+
+/// Errors specific to client
+#[derive(Debug, Eq, PartialEq, Clone, Display)]
+pub enum Error {
+    /// LedgerDBError,
+    LedgerDBNotFound,
+
+    /// Failed to access DB
+    CantAccessDB,
+
+    /// Other Ledger DB Error
+    LedgerDB(LedgerDBError),
+
+    /// Block index {0} is too far ahead current index {1}, wait
+    BlockIndexTooFar(BlockIndex, BlockIndex),
+
+    /// Mutex Lock Error {0}
+    Locked(LockReason),
+}
+
+/// A non-generic helper enum to avoid having to pass in generic type params
+#[derive(Debug, Eq, PartialEq, Clone, Display)]
+pub enum LockReason {
+    /// Acquiring the lock would block
+    WouldBlock,
+
+    /// Lock poisoned
+    Poisoned,
+}
+
+// Return specific errors for checking
+impl From<LedgerDBError> for Error {
+    fn from(ledgerdb_error: LedgerDBError) -> Self {
+        match ledgerdb_error {
+            LedgerDBError::NotFound => Error::LedgerDBNotFound,
+            LedgerDBError::BadRslot => Error::CantAccessDB,
+            err => Error::LedgerDB(err),
+        }
+    }
+}
+
+impl<L: Ledger> From<TryLockError<RwLockReadGuard<'_, L>>> for Error {
+    fn from(lock_error: TryLockError<RwLockReadGuard<'_, L>>) -> Self {
+        match lock_error {
+            TryLockError::WouldBlock => Error::Locked(LockReason::WouldBlock),
+            TryLockError::Poisoned(..) => Error::Locked(LockReason::Poisoned),
+        }
+    }
+}
+
+impl<L: Ledger> From<TryLockError<RwLockWriteGuard<'_, L>>> for Error {
+    fn from(lock_error: TryLockError<RwLockWriteGuard<'_, L>>) -> Self {
+        match lock_error {
+            TryLockError::WouldBlock => Error::Locked(LockReason::WouldBlock),
+            TryLockError::Poisoned(..) => Error::Locked(LockReason::Poisoned),
+        }
+    }
+}
--- a/ledger/streaming/client/src/lib.rs
+++ b/ledger/streaming/client/src/lib.rs
@@ -7,6 +7,10 @@
 
 mod grpc;
 
+pub mod error;
+pub mod ledger_sink;
+pub mod streaming_futures;
+
 #[cfg(any(test, feature = "test_utils"))]
 pub mod test_utils;
 
--- a/ledger/streaming/client/src/streaming_futures.rs
+++ b/ledger/streaming/client/src/streaming_futures.rs
@@ -84,10 +84,10 @@ impl<L: Ledger> Future for WriteBlock<L> {
                     if reason == LockReason::WouldBlock {
                         Poll::Pending
                     } else {
-                        Poll::Ready(Err(StreamError::DBError))
+                        Poll::Ready(Err(StreamError::DBAccess))
                     }
                 }
-                _ => Poll::Ready(Err(StreamError::DBError)),
+                _ => Poll::Ready(Err(StreamError::DBAccess)),
             },
         }
     }
@@ -145,10 +145,10 @@ impl<'a, L: Ledger> Future for ReadLedger<'a, L> {
                     if reason == LockReason::WouldBlock {
                         Poll::Pending
                     } else {
-                        Poll::Ready(Err(StreamError::DBError))
+                        Poll::Ready(Err(StreamError::DBAccess))
                     }
                 }
-                _ => Poll::Ready(Err(StreamError::DBError)),
+                _ => Poll::Ready(Err(StreamError::DBAccess)),
             },
         }
     }
--- a/ledger/streaming/client/src/streaming_futures.rs
+++ b/ledger/streaming/client/src/streaming_futures.rs
@@ -1,0 +1,231 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
+//! Custom Future implementations for the block streaming client
+
+use crate::error::{Error as ClientError, LockReason};
+use futures::{
+    task::{Context, Poll},
+    Future, Stream,
+};
+use mc_ledger_db::Ledger;
+use mc_ledger_streaming_api::{
+    BlockStreamComponents, Error as StreamError, Result as StreamResult,
+};
+use mc_transaction_core::{ring_signature::KeyImage, BlockIndex};
+use pin_project::pin_project;
+use std::{
+    pin::Pin,
+    sync::{Arc, RwLock},
+};
+
+/// Enum to allow monadic results other than Some & None for
+/// stream types were None would terminate the stream. Useful for
+/// stream combinations like Scan --> Filter_Map
+#[allow(dead_code)]
+pub enum Ready<T> {
+    /// Value with data container to indicate a value is ready
+    Ready(T),
+
+    /// Indicates to subsequent future/stream upstream is not ready
+    NotReady,
+}
+
+/// Future for writing blocks to LedgerDB
+pub struct WriteBlock<L: Ledger> {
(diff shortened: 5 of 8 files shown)


ledger/streaming/api/src/error.rs
```diff
@@ -5,24 +5,27 @@ use grpcio::Error as GrpcError;
 use mc_api::ConversionError;
 
 #[derive(Debug, Display)]
-pub enum Error {
+pub enum LedgerStreamingError {
```
remoun, line 8: I disagree with this rename. Callers can alias this when importing.

FWIW I modeled this after examples from our repo as well as other libs:
https://docs.rs/grpcio/latest/grpcio/enum.Error.html
https://github.com/mobilecoinfoundation/mobilecoin/blob/93e41f75df3f6fc202eaaa3099033bc845a92abe/ledger/db/src/error.rs#L9
https://github.com/mobilecoinfoundation/mobilecoin/search?q=enum.Error

ledger/streaming/client/Cargo.toml
```diff
@@ -5,22 +5,28 @@ authors = ["MobileCoin"]
 license = "MIT OR Apache-2.0"
 edition = "2018"
 
-[lib]
```
remoun, line 8: AFAIK we need this for Cargo to emit a library

ledger/streaming/client/Cargo.toml
```diff
@@ -5,22 +5,28 @@ authors = ["MobileCoin"]
...
-
 [features]
 test_utils = []
 
 [dependencies]
+
```
remoun, line 12: Nit: unnecessary blank line.

ledger/streaming/client/Cargo.toml
```diff
@@ -5,22 +5,28 @@ authors = ["MobileCoin"]
...
 
 displaydoc = { version = "0.2", default-features = false }
 futures = { version = "0.3.21" }
 grpcio = { version = "0.10.0" }
-protobuf = "2.22.1"
+protobuf = { version = "2.22.1" }
+structopt = { version = "0.3" }
+serde_json = "1.0"
+tokio = { version = "1.17.0" }
+pin-project = {version = "1.0.10" }
```
remoun, line 29: Nit: Sort alphabetically

ledger/streaming/client/src/ledgerdb_sink.rs
```diff
@@ -0,0 +1,115 @@
...
+            ledger
+                .append_block(
+                    block_data.block(),
+                    block_data.contents(),
+                    block_data.signature().clone(),
+                )
+                .unwrap();
+            log::trace!(logger, "Block {} written", index)
+        } else {
+            log::warn!(logger, "Receiver shut down!");
```
remoun, line 70: Is that the only reason we might get an error? It seems better to capture and log the error here.

remoun (dismissed):

jcape (changes requested):
So some of the things here should be changed, but I think this suffers from the same issue as the block publisher PR: like, ideally `LedgerDbSink` is itself a `Stream` implementation that can be polled to turn the crank on itself and all upstream objects.

The only thing special about a sink is that instead of `Result<Data, Error>` it returns `Result<(), Error>` (even that isn't really necessary).

ledger/streaming/client/src/ledgerdb_sink.rs
```diff
@@ -0,0 +1,115 @@
...
+
+        for i in 1..200 {
+            test_block_data.push(Ok(source_ledger.get_block_data(i).unwrap()));
+        }
+        let stream = futures::stream::iter(test_block_data);
+        let mut bs = BlockSink::new(stream, logger.clone(), tx);
+
+        let dest_ledger = Arc::new(Mutex::new(get_mock_ledger(1)));
+        let ledger_ref = dest_ledger.clone();
+        std::thread::spawn(move || sink(ledger_ref, rcv, logger.clone()));
```
remoun, line 97: How about we have BlockSink own spawning a thread calling `sink()`, and the associated `mpsc::channel`? It would then track a handle to that thread, cleaning it up in `stop()`, and call `stop()` on `drop()`.

ledger/streaming/client/src/ledgerdb_sink.rs
```diff
@@ -0,0 +1,115 @@
...
+
+        let rt = Builder::new_current_thread().enable_all().build().unwrap();
+
+        rt.block_on(async move {
+            let mut counter = 0usize;
+            loop {
+                counter += 1;
+                bs.broadcast().await;
+                if counter < 200 {
```
remoun, line 106: Did you mean `>` here? As written, this breaks after the first iteration.

ledger/streaming/client/Cargo.toml
iamalwaysuncomfortable: Indeed!

ledger/streaming/api/src/error.rs
iamalwaysuncomfortable: Okay cool, thanks for the references, I'll use it the original

ledger/streaming/client/Cargo.toml
iamalwaysuncomfortable: Will do, is there linter automation that can do this?

ledger/streaming/client/src/ledgerdb_sink.rs
iamalwaysuncomfortable: Yeah this definitely a todo

ledger/streaming/client/src/source.rs
```diff
@@ -1,12 +1,13 @@
 // Copyright (c) 2018-2022 The MobileCoin Foundation
+#![allow(dead_code)]
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: Only temporary while the PR is WIP status, will be removed.

ledger/streaming/client/src/lib.rs
```diff
@@ -1,10 +1,9 @@
...
 #![feature(min_type_alias_impl_trait)]
-
 mod source;
 
+pub mod config;
+mod ledgerdb_sink;
 #[cfg(any(test, feature = "test_utils"))]
 pub mod test_utils;
-
-pub use self::source::GrpcBlockSource;
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: Thank you for the ref

ledger/streaming/api/src/error.rs
```diff
@@ -5,24 +5,27 @@ use grpcio::Error as GrpcError;
...
 
-impl From<ConversionError> for Error {
+impl From<ConversionError> for LedgerStreamingError {
     fn from(src: ConversionError) -> Self {
         Self::Conversion(src)
     }
 }
 
-pub type Result<T> = core::result::Result<T, Error>;
+pub type StreamResult<T> = core::result::Result<T, LedgerStreamingError>;
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: Agreed

ledger/streaming/client/src/config.rs
```diff
@@ -0,0 +1,96 @@
+#![allow(dead_code)]
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: Agreed!

ledger/streaming/client/src/ledgerdb_sink.rs
iamalwaysuncomfortable: Indeed I did!

ledger/streaming/client/src/ledgerdb_sink.rs
```diff
@@ -0,0 +1,115 @@
...
+use std::sync::{Arc, Mutex};
+use tokio::sync::mpsc;
+/// A block sink that takes blocks from a passed stream and puts them into
+/// ledger db. This sink should live downstream from a verification source that
+/// has already done block content, scp, and avr validation and thus is
+/// considered a trusted stream
+#[pin_project]
+pub struct BlockSink<BS: Stream<Item = StreamResult<BlockData>> + std::marker::Unpin> {
+    /// Stream of block data
+    stream: BS,
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: Agreed. Yeah I think this should be fairly simple to create a Stream implementation that can be cranked. The main question is "where' the crank" and I believe that this crank is should be something that lives in a file like stream_service.rs which aligns all the streams and does the crank turning. On all client-side streams. I think this receiving a stream is the right choice - and the validator (potentially) will be the one that takes the GrpcBlockSource upstream component and a BlockService struct turns the crank.

ledger/streaming/client/src/ledgerdb_sink.rs
iamalwaysuncomfortable: Indeed I Did!

ledger/streaming/client/src/ledgerdb_sink.rs
iamalwaysuncomfortable: This was a detail I'm sort of wondering about in the entire chain. As James points out the sink should be a stream chain, so I'm guessing that this is the final stream component sinking into the a final component like BlockStreamService that owns the spawning thread (and turns the crank) & manages the sink thread and the BlockSink perhaps just does the mpsc sending. 

Also, I think we all need to mind meld and think about what components "client" has. I think it's probably your GrpcSource component (remote stream origin) -> validation -> sink -> BlockStreamClient (the crank turner).

What are your thoughts?

ledger/streaming/client/Cargo.toml
remoun: Not that I know of

remoun: Looks like https://crates.io/crates/cargo-sort fits the bill

ledger/streaming/client/src/ledgerdb_sink.rs
remoun: Then let's add a TODO ;)

ledger/streaming/client/src/ledgerdb_sink.rs
remoun: Yeah, let's discuss tomorrow. In the meantime, I'm thinking of this thread as an implementation detail of using LMDB with async Rust, so the thread management is the responsibility of the BlockSink, not its callers, and I'd argue they shouldn't know about the thread. This also simplifies the code, as we can then trust the sender side to be our sink, and not some random code.

iamalwaysuncomfortable: While it does seem seem natural to have the sink be the final component, there probably needs to be a stream driver at the end mainly because there's a good amount of diversity of node types in the network which would lead to a wide diversity in configuration & synchronization needs

#### Examples
The following have very different ways in which they sync their ledgers with the existing services: 
* full_service: https://github.com/mobilecoinofficial/full-service/blob/main/full-service/src/validator_ledger_sync.rs 
vs: 
* consensus: https://github.com/mobilecoinfoundation/mobilecoin/blob/master/consensus/service/src/byzantine_ledger/worker.rs

What I think could probably happen is that the blocksink could have an option to spin up a sinking thread which an end node could easily spin up in their configuration or choose the option to just consume the stream that it can decide how to execute.

Right now this is setup to only target the ledger trait and not specifically ledger db (so that the underlying implementation can be different) so it's general enough to be used by nodes not using ledgedb.

ledger/streaming/client/src/error.rs
```diff
@@ -0,0 +1,41 @@
...
+use mc_transaction_core::BlockIndex;
+
+/// Errors specific to client
+#[derive(Debug, Eq, PartialEq, Clone, Display)]
+pub enum Error {
+    /// LedgerDBError,
+    LedgerDBNotFound,
+
+    /// Failed to access DB
+    CantAccessDB,
```
remoun, line 17: These seems redundant with the `LedgerDb` variant

ledger/streaming/client/src/error.rs
```diff
@@ -0,0 +1,41 @@
...
+
+    /// Other Ledger DB Error
+    LedgerDB(LedgerDBError),
```
remoun, line 20: Nit: Use snake case
```suggestion
    LedgerDb(LedgerDbError),
```

ledger/streaming/client/src/ledgerdb_sink.rs
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    /// Execute storing the blocks in an opinionated way
+    pub async fn store_blocks(&mut self, limit: Option<usize>) {
+        let mut counter = 0usize;
+        while let Some(index) = self.block_sink.next() {
+            log::trace!(self.logger, "successfully wrote block {}", index);
+            if let Some(limit) = limit {
+                counter += 1;
+                if counter > limit  { break }
```
remoun, line 71: IIUC we want >= here.
```suggestion
                if counter >= limit  { break }
```

ledger/streaming/client/src/lib.rs
```diff
@@ -1,10 +1,16 @@
 // Copyright (c) 2018-2022 The MobileCoin Foundation
 
+//! MobileCoin Ledger Streaming Client
+
 #![feature(min_type_alias_impl_trait)]
 
+pub mod error;
+pub mod config;
+pub mod ledgerdb_sink;
+pub mod streaming_futures;
```
remoun, line 10: Let's keep these `mod`s private and re-export what we need to.

ledger/streaming/client/src/streaming_futures.rs
```diff
@@ -0,0 +1,129 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
+//! Custom Future implementations for the block streaming client
+
+use std::pin::Pin;
+use std::sync::{Arc, RwLock, TryLockError};
+
```
remoun, line 7: Nit: Remove blank lines and sort imports

ledger/streaming/client/src/streaming_futures.rs
```diff
@@ -0,0 +1,129 @@
...
+    type Output = BlockIndex;
+
+    ///Attempt to write block and return the block index written
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        match self.write_block() {
+            Ok(block_index) => {
+                Poll::Ready(block_index)
+            }
+            _ => {
+                cx.waker().wake_by_ref();
```
remoun, line 79: Waking the task then returning Pending doesn't seem right. It seems easier and safer to reuse the StreamExt/FutureExt helpers than trying to implement `Future` ourselves.

ledger/streaming/client/src/streaming_futures.rs
```diff
@@ -0,0 +1,129 @@
...
+            Ok(ledger_lock) => {
+                block_index = self.block_data.block().index;
+                if ledger_lock.num_blocks()? + 1 < block_index {
+                    return Err(ClientError::BlockIndexTooFar(block_index));
+                }
+            },
+            Err(err) => {
+                match err {
+                    TryLockError::WouldBlock => return Err(ClientError::Locked),
+                    _ => return Err(ClientError::LockPoisoned),
```
remoun, line 45: How is the caller expected to handle these?

iamalwaysuncomfortable:
Feature rebased on feature/block-streaming - ready for re-review

iamalwaysuncomfortable:
Superseded by PR #1831

PR #561, opened 2020-11-20
[MCC-1994] Adds validate_outputs_are_sorted

mfaulk:
### Motivation

Transaction outputs should be deterministically ordered. A required ordering removes an unneeded "degree of freedom" in the transaction structure, and prevents a possible source of information leakage (e.g. client software that always orders transaction outputs from largest value to smallest value).

### In this PR
* Adds `validate_outputs_are_sorted` to transaction validation.

PR #1615, opened 2022-03-14
feat: added a check for outputs to be sorted for version 3 of the blo…

integral-llc:
…ck chain

### Motivation

Reverted back the validated output check to bes orted

### In this PR

### Future Work
* as of no transactions are already sorted, might need a negative unit test to see how it behaves with the new functionality


integral-llc:
this is the reversal of 561[https://github.com/mobilecoinfoundation/mobilecoin/pull/561] PR

transaction/core/src/validation/validate.rs
```diff
@@ -57,6 +57,8 @@ pub fn validate<R: RngCore + CryptoRng>(
 
     validate_inputs_are_sorted(&tx.prefix)?;
 
+    validate_outputs_are_sorted(&tx.prefix)?;
```
remoun, line 60: Remove this unconditional call.

transaction/core/src/validation/validate.rs
```diff
@@ -920,6 +945,46 @@ mod tests {
         }
     }
 
+    #[test]
+    /// Should reject a transaction with unsorted outputs.
+    fn test_validate_outputs_are_sorted() {
+        for block_version in BlockVersion::iterator() {
```
remoun, line 951: Let's extend this test (or add another) to verify that `validate` rejects transactions with unsorted outputs if `BlockVersion.validate_transaction_outputs_sorted`, and accepts them otherwise.

remoun (changes requested):

consensus/api/src/conversions.rs
```diff
@@ -94,6 +95,7 @@ impl TryInto<Error> for ProposeTxResult {
             Self::TxFeeError => Ok(Error::TxFeeError),
             Self::KeyError => Ok(Error::KeyError),
             Self::UnsortedInputs => Ok(Error::UnsortedInputs),
+            Self::UnsortedOutputs => Ok(Error::UnsortedOutputs),
```
jcape, line 98: This should be at the end of this statement.

consensus/api/src/conversions.rs
```diff
@@ -46,6 +46,7 @@ impl From<Error> for ProposeTxResult {
             Error::TxFeeError => Self::TxFeeError,
             Error::KeyError => Self::KeyError,
             Error::UnsortedInputs => Self::UnsortedInputs,
+            Error::UnsortedOutputs => Self::UnsortedOutputs,
```
jcape, line 49: This should be at the end of this statement.

jcape (changes requested):

transaction/core/src/validation/validate.rs
integral-llc: can you help me with this. I checked the other test we have and could not find a direct call to `validate` itself. Mainly I don't know how to correctly create all the needed parameters besides `tx and block_version`. Thx!

remoun: Ah, I see there are no other tests that call the top-level `validate()`. This can be a follow-up, then.

remoun (approved):

jcape (approved):

Issue #1295, opened 2022-01-07
Watcher should publish metrics on nodes, their block heights, and their blocks synced


jcape:
Watcher should provide a table via prometheus metrics containing a node per-row with the following details:

1. Node being synchronized
2. The block height of the node
3. The block height that has been synchronized so-far

PR #1633, opened 2022-03-15
Prometheus Metrics for Watcher

iamalwaysuncomfortable:
### Motivation

This PR addresses issue #1295 which identifies the desire to track watcher blocks synced vs ledger height. Watcher's main purpose is to collect AVRs, but it also serves as a proxy for node health because if a node stops syncing, a Prometheus alert that compares watcher blocks sync vs. ledger height can throw an alert that the node is unhealthy before it starts to affect the network.

### In this PR
* A basic Prometheus metric service is implemented
* Ledger height and block_height metrics are tracked after every sync iteration
* Copyright annotations for watcher are updated

Ticket Status:
Addresses issue #1295


watcher/src/watcher.rs
```diff
@@ -389,6 +401,11 @@ impl WatcherSyncThread {
 
             // Maybe sync, maybe wait and check again.
             if is_behind {
+                watcher
+                    .metrics
+                    .ledger_block_height
+                    .set(ledger_num_blocks as i64);
+
```
iamalwaysuncomfortable, line 408: This is accidental, to be removed

PR #1643, opened 2022-03-16
Make `TxOut::new` take `Amount` as parameter

garbageslam:
This is consistent with `MaskedAmount::new`, and we generally
want to pass `value: u64` around with its token id to avoid confusing
token types.

This is a minor fixup after confidential token ids PR

### DIFF
--- a/transaction/core/src/tx.rs
+++ b/transaction/core/src/tx.rs
@@ -24,7 +24,7 @@ use crate::{
     memo::{EncryptedMemo, MemoPayload},
     onetime_keys::{create_shared_secret, create_tx_out_public_key, create_tx_out_target_key},
     ring_signature::{KeyImage, SignatureRctBulletproofs},
-    CompressedCommitment, NewMemoError, NewTxError, TokenId,
+    CompressedCommitment, NewMemoError, NewTxError,
 };
 
 /// Transaction hash length, in bytes.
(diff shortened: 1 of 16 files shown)


sugargoat (approved):

transaction/std/src/transaction_builder.rs
```diff
@@ -463,8 +464,8 @@ fn create_output_with_fog_hint<RNG: CryptoRng + RngCore>(
     rng: &mut RNG,
 ) -> Result<(TxOut, RistrettoPublic), TxBuilderError> {
     let private_key = RistrettoPrivate::from_random(rng);
-    let mut tx_out =
-        TxOut::new_with_memo(value, token_id, recipient, &private_key, fog_hint, memo_fn)?;
+    let amount = Amount { value, token_id };
```
remoun, line 467: How about we make the `create_output*` functions also take an `Amount`?

remoun (approved):

transaction/std/src/transaction_builder.rs
garbageslam: so i think the thing here was, the token id is a config parameter to the tx builder, and so it is implied. so that was why it was passing `value: u64` around. We could still make it take `Amount` and then double check the token ids I guess.

transaction/core/src/tx.rs
```diff
@@ -291,18 +291,17 @@ impl TxOut {
     /// This uses a defaulted (all zeroes) MemoPayload.
     ///
     /// # Arguments
-    /// * `value` - Value of the output.
+    /// * `amount` - Amount contained within the TxOut
```
mfaulk, line 294: This does look nicer :)

eranrund (dismissed):

iamalwaysuncomfortable (approved):

remoun (dismissed):

mfaulk (dismissed):

eranrund (approved):

remoun (approved):

Marc27100 (approved):

Issue #1552, opened 2022-02-28
Overseer Productionization/Testing


jcape:
- [x] Prometheus/Sentry wiring
 - [x] Ensuring startup logic is correct
 - [x] Wargaming / testing
 - [x] Production helm / deployment

samdealy:
All of the work that eng is responsible for has been completed. Will assist ops as is needed in the future. Closing ticket.

PR #1646, opened 2022-03-17
Fog Overseer improvements

samdealy:
### Motivation

Contains minor fixes for things that came up during Fog overseer testing. 

Contributes to #1552 

### In this PR
- Logs when an inactive outstanding key isn't matched + when a previously unresponsive node becomes responsive again.
- Adds a prometheus counter for number of unresponsive nodes. 
- Logs node URI rather than it's "index."


fog/overseer/server/src/worker.rs
```diff
@@ -235,12 +246,23 @@ where
     /// Returns the latest round of ingest summaries for each
     /// FogIngestGrpcClient that communicates with a node that is online.
     fn retrieve_ingest_summary_node_mappings(
-        &self,
+        &mut self,
```
iamalwaysuncomfortable, line 249: Could this potentially lead to situation where this function fails multiple mutable references exist to the parent object?

iamalwaysuncomfortable:
Potentially enumerate the peer nodes for the metrics + double check the mutable references

fog/overseer/server/src/metrics/counters.rs
```diff
@@ -20,4 +20,7 @@ lazy_static::lazy_static! {
 
     /// Number of idle Fog Ingest nodes.
     pub static ref IDLE_NODE_COUNT: IntGauge = OP_COUNTERS.gauge("idle_node_count");
+
+    /// Number of idle Fog Ingest nodes.
+    pub static ref UNRESPONSIVE_NODE_COUNT: IntCounter = OP_COUNTERS.counter("unresponsive_node_count");
```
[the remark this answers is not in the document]
samdealy: What methods are you referring to?

fog/overseer/server/src/worker.rs
samdealy: I don't understand. Can you explain more?

fog/overseer/server/src/metrics/utils.rs
```diff
@@ -6,6 +6,11 @@ use mc_common::logger::{log, Logger};
 use mc_fog_api::ingest_common::{IngestControllerMode, IngestSummary};
 use std::{collections::HashSet, convert::TryInto};
 
+pub fn increment_unresponsive_node_count(logger: &Logger) {
+    log::trace!(logger, "Setting unresponsive node metric.");
+    counters::UNRESPONSIVE_NODE_COUNT.inc();
+}
+
```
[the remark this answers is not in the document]
samdealy: This is being logged in the worker.

garbageslam (dismissed):
LGTM

fog/overseer/server/src/worker.rs
```diff
@@ -257,6 +279,8 @@ where
                         ingest_client.get_uri(),
                         err
                     );
+                    self.unresponsive_node_urls
+                        .insert(ingest_client.get_uri().clone());
```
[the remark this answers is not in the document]
samdealy: That would be redundant- we log when a node doesn't return a response. Ops would be able to use this info to track down when it started not responding.

fog/overseer/server/src/worker.rs
```diff
@@ -235,12 +246,23 @@ where
...
     ) -> Result<Vec<IngestSummaryNodeMapping>, OverseerError> {
         let mut ingest_summary_node_mappings: Vec<IngestSummaryNodeMapping> = Vec::new();
         for (ingest_client_index, ingest_client) in self.ingest_clients.iter().enumerate() {
             match ingest_client.get_status() {
                 Ok(ingest_summary) => {
+                    if self
+                        .unresponsive_node_urls
+                        .contains(ingest_client.get_uri())
+                    {
+                        self.unresponsive_node_urls.remove(ingest_client.get_uri());
```
remoun, line 259: Optional/Equivalent: Use the return value of `remove()`
```suggestion
                    if self
                        .unresponsive_node_urls
                        .remove(ingest_client.get_uri())
                    {
```

fog/overseer/server/src/worker.rs
remoun: FWIW the single-mutable-borrow constraint is enforced at compile time.

fog/overseer/server/src/worker.rs
```diff
@@ -235,12 +246,23 @@ where
...
+                        log::info!(
+                            self.logger,
+                            "Node {} was previously unresponsive, but just successfully responded!",
```
remoun, line 262: Would it make sense to decrement `unresponsive_node_count` here?

fog/overseer/server/src/metrics/counters.rs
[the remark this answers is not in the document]
iamalwaysuncomfortable: These: https://github.com/mobilecoinfoundation/mobilecoin/blob/master/util/metrics/src/op_counters.rs#L78-L102

But also a matter of choice.

fog/overseer/server/src/metrics/utils.rs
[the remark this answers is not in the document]
iamalwaysuncomfortable: I'm sorry, I didn't mean log, I meant the prometheus metric capturing - the unresponsive node could potentially annotated with the prometheus peer_counter so that they show up on the metrics dashboard (separate from the logstash logs) with the node name.

https://github.com/mobilecoinfoundation/mobilecoin/blob/master/util/metrics/src/op_counters.rs#L89-L97

iamalwaysuncomfortable (approved):
Main feedback is consider using peer_gauge/peer_counter when collecting metrics on remote peers, otherwise LGTM

fog/overseer/server/src/metrics/utils.rs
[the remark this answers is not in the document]
samdealy: Ah I see. I think this is something we can consider implementing in a follow-up PR if ops wants this info to be captured.

fog/overseer/server/src/worker.rs
samdealy: I think we want to keep a running total of how many unresponsive nodes there have been. If we wanted a snapshot of unresponsive node count, then we'd use a prometheus gauge.

fog/overseer/server/src/metrics/counters.rs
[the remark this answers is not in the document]
samdealy: I'll keep it as is, but this can be a follow up PR if we so choose.

iamalwaysuncomfortable (approved):
LGTM

garbageslam (approved):
Lgtm

remoun (approved):

fog/overseer/server/src/worker.rs
```diff
@@ -472,8 +497,8 @@ where
                             let number_of_remaining_tries =
                                 Self::NUMBER_OF_TRIES - current_try as usize;
                             let error_message = match number_of_remaining_tries {
-                                0 => format!("Did not succeed in setting a new key on node at index {}. Underlying error: {}", i, err),
-                                _ => format!("New keys were not successfully set on the ingest node at index {}. Will try {} more times. Underlying error: {}", i, number_of_remaining_tries, err),
+                                0 => format!("Did not succeed in setting a new key on node {}. Underlying error: {}", ingest_client.get_uri(), err),
+                                _ => format!("New keys were not successfully set on the ingest node {}. Will try {} more times. Underlying error: {}", ingest_client.get_uri(), number_of_remaining_tries, err),
```
remoun, line 501: Nit: This variant seems strictly more informative. How about we replace this `match` with the one `format!` call?

fog/overseer/server/src/worker.rs
```diff
@@ -511,13 +537,14 @@ where
                             Self::NUMBER_OF_TRIES - current_try as usize;
                         let error_message = match number_of_remaining_tries {
                             0 => format!(
-                                "Did not succeed in setting a new key on node at index {}. Underlying error: {}",
-                                activated_node_index,
+                                "Did not succeed in setting a new key on node {}. Underlying error: {}",
```
remoun, line 540: This message is rather different from the other match arm. Consider consolidating these messages.

remoun (changes requested):

fog/overseer/server/src/worker.rs
samdealy: Done.

fog/overseer/server/src/worker.rs
samdealy: Done

remoun: This one still has differing messages.

samdealy: Done

fog/overseer/server/src/worker.rs
```diff
@@ -496,30 +515,21 @@ where
...
-                                "Did not succeed in setting a new key on node at index {}. Underlying error: {}",
-                                activated_node_index,
-                                err
-                            ),
-                            _ => format!(
-                                "Node at index {} not activated. Will try {} more times. Underlying error: {}",
-                                activated_node_index, number_of_remaining_tries, err
-                            ),
-                        };
+                        let error_message = format!("Node at index {} not activated. Will try {} more times. Underlying error: {}",ingest_client.get_uri(), number_of_remaining_tries, err);
```
remoun, line 532:
```suggestion
                        let error_message = format!("Node {} not activated. Will try {} more times. Underlying error: {}",ingest_client.get_uri(), number_of_remaining_tries, err);
```

remoun (dismissed):

fog/overseer/server/src/worker.rs
```diff
@@ -496,30 +515,26 @@ where
...
-                                activated_node_index, number_of_remaining_tries, err
-                            ),
-                        };
+                        let error_message = format!(
+                            "Node {} not activated. Will try {} more times. Underlying error: {}",
+                            ingest_client.get_uri(),
+                            number_of_remaining_tries,
+                            err
+                        );
                         OperationResult::Retry(OverseerError::ActivateNode(error_message))
```
remoun, line 538: ... On further thought, do we still want to return Retry when num_retries = 0? If not, then we might as well change back to using a `match`, and returning a different message and result when num_retries=0.

samdealy: That was the original reason why I kept the messages as they were, but I think it is fine as is. And IIUC, I think that it does not matter if we return a Retry when num_tries is 0 because the logic on line 497 will ensure that it is not retried after the num_tries is 0.

remoun (approved):

iamalwaysuncomfortable (dismissed):
LGTM

remoun (approved):

iamalwaysuncomfortable (approved):

PR #1647, opened 2022-03-17
circle ci --exclude fix attempt

garbageslam:
try to remove the globs from `--exclude` in cargo invocations,
since ehuss says it's not implemented in some github comment,
and instead the implementation uses `starts_with()`, so maybe this
will have the effect we were seeking

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -333,7 +333,7 @@ commands:
     parameters:
       test_command:
         type: string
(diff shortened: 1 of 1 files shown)


garbageslam:
I think this isn't working, I'm seeing an error in CI log:

```
warning: excluded package(s) `mc-fog-ingest-` not found in workspace `/root/project`
```

garbageslam:
this actually caused `run-mc-tests` to fail due to not being able to find enclave files

garbageslam:
Closing because this didn't work

Issue #1370, opened 2022-01-27
More prometheus metrics for fog services


jcape:
Per @jgreat

We need Prometheus "counters" for "total", "successful" and "error" (by type) transactions in the various fog services.  Something like the 2xx, 4xx, 5xx counters you commonly see when monitoring metrics for a http web server.

I think consensus has "consensus_service_tx_validation_errors" for this function.

**Definition of Done:**

Fog Operators are able to monitor and set up alerts for error rates so we can be proactive in heading off network attacks.

**Next Steps:**

Identify what error conditions should be tracked for each service, and define those metrics.

PR #1648, opened 2022-03-17
Add Prometheus int counters that track peers

iamalwaysuncomfortable:
### Motivation

Several situations have arisen where Prometheus counters that track (mostly misbehaving) peers would be very useful, this small PR addresses that need.

### In this PR
* Addition of peer counters
* Addition of helper methods for peer counters

Part of issue #1370


iamalwaysuncomfortable:
@samdealy would love your weigh in on this

Issue #287, opened 2020-07-09
ConsensusEnclave::enclave_init panicks on unparsable sealed key


brson:
If I pass a bogus file to `mc-consensus-service --sealed-block-signing-key` the enclave will panick when calling `IntelSealed::try_from` in `ConsensusEnclave::enclave_init`.

This is harmless, but not the right error handling mode for this error - the enclave probably shouldn't be panicking on bad input. Instead of unwrapping errors in `ConsensusEnclave::enclave_init` they should probably be propagated with `?`.

PR #1658, opened 2022-03-17
Feature/no panic on unparsable sealed key

nick-mobilecoin:
Previously `ConsensusEnclave::enclave_init` would panic if there was a failure parsing or sealing the key.  Now an an ` mc_attest_core::IntelSealingError` is provided back.

### In this PR
fixes: #287

### DIFF
--- a/fog/ingest/enclave/impl/src/lib.rs
+++ b/fog/ingest/enclave/impl/src/lib.rs
@@ -20,7 +20,8 @@ use aligned_cmov::{typenum::U32, A8Bytes, Aligned, GenericArray};
 use alloc::vec::Vec;
 use core::convert::TryFrom;
 use mc_attest_core::{
-    IasNonce, IntelSealed, IntelSealingError, Quote, QuoteNonce, Report, TargetInfo, VerificationReport,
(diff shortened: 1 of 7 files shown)


fog/ingest/enclave/impl/src/lib.rs
```diff
@@ -391,7 +392,7 @@ fn seal_private_key(src: &RistrettoPrivate) -> Result<SealedIngestKey> {
 // Helper for converting error type living only in mc_attest_trusted to the
 // error type in enclave_api (The attest_trusted crate will not compile in
 // non-sgx environment.)
-fn map_sealing_error(src: mc_attest_trusted::IntelSealingError) -> Error {
+fn map_sealing_error(src: IntelSealingError) -> Error {
```
garbageslam, line 395: we might be able to redo this function, this was a workaround from the time when consensus and fog were in separate repos and this error was not in attest core yet

garbageslam: like we could replace this with a `From<IntelSealingError> for Error` and then drop the `map_err(map_sealing_error)` part

nick-mobilecoin: I think b4c1879 should address this

PR #1659, opened 2022-03-18
Fix linter

### LINKED ISSUE #1304: Lint requires sgx

wjuan-mob:
Soundtrack of this PR: [link to song that really fits the mood of this PR](https://www.youtube.com/watch?v=cFbn7vEvT3w)

### Motivation

The linter should not require sgx to run

### In this PR
Fixes a typo in BUILD.md
Fixes #1304


consensus/enclave/build.rs
```diff
@@ -29,6 +30,14 @@ fn main() {
         SGX_LIBS
     };
 
+    // This is a hack for lint to run, this wouldn't be necessary after we move the
+    // enclave to its own crate
+    rerun_if_env_changed!("CONSENSUS_ENCLAVE_CSS");
+    if let Ok(value) = var("CONSENSUS_ENCLAVE_CSS") {
+        warning!("Found consensus enclave css: {}", value);
+        return;
+    }
```
iamalwaysuncomfortable, line 39: Are there conditions in which this could introduce side effects?

iamalwaysuncomfortable (commented):
Seems good, but I wonder if this cold introduce side effects for other envar changes? I'm fuzzy on the details of this. Also, how was the speedup of CI with this change?

consensus/enclave/build.rs
jcape: Unfortunately this will break builds of consensus with provided CSS files, because the generated code that lives in this file will not be created (which is required).

jcape (changes requested):
This approach will break how we're doing production/testnet builds, because we actually do need the SGX SDK to build `mc-consensus-enclave`---specifically, it includes codegen that comes from SGX SDK's `edger8r` utility.

PR #1667, opened 2022-03-18
Prefactoring for block streaming

remoun:
* Add `BlockStreamComponents` and conversion helpers.
* Make `BlockStreamComponents` implement `Digestible`
* Add `BlockFetcher` trait with a mock implementation.
* Add a few other test utilities, including a mock `BlockStream`.
* Add `deny(missing_docs)` and necessary comments.
* Make `mc_ledger_streaming_api::Error` implement `Clone`.
* Reorganize files to make upcoming additions simpler.


ledger/streaming/client/src/grpc.rs
```diff
@@ -0,0 +1,139 @@
...
+    streaming_blocks_grpc::LedgerUpdatesClient,
+    BlockStream, BlockStreamComponents, Result,
+};
+use mc_util_grpc::ConnectionUriGrpcioChannel;
+use mc_util_uri::ConnectionUri;
+use std::{sync::Arc, time::Duration};
+
+/// A BlockStream that streams blocks using the `LedgerUpdates` gRPC API.
+#[derive(Display)]
+pub struct GrpcBlockSource {
```
remoun, line 19: This is a rename of `source.rs` with minor modifications, enough to throw off git's rename detection.

ledger/streaming/api/src/convert/components.rs
```diff
@@ -0,0 +1,73 @@
...
+        }
+        if let Some(ref report) = data.verification_report {
+            proto.set_report(report.into());
+        }
+        proto
+    }
+}
+
+impl TryFrom<&BlockWithQuorumSet> for BlockStreamComponents {
+    type Error = ConversionError;
```
iamalwaysuncomfortable, line 26: Should we merge our error set? Or does an error set for each crate make sense? I think putting some thought into separating by ontology of errors + their handling is a worthy discussion.

ledger/streaming/api/src/response.rs
```diff
@@ -0,0 +1,44 @@
...
+
+use crate::{streaming_blocks::SubscribeResponse, BlockStreamComponents, Result};
+use mc_crypto_keys::{Ed25519Pair, Ed25519Public, Ed25519Signature, Signer, Verifier};
+use protobuf::Message;
+use std::convert::TryFrom;
+
+/// Creates a signed SubscribeResponse from the given components and signer.
+pub fn make_subscribe_response(
+    data: &BlockStreamComponents,
+    signer: &Ed25519Pair,
```
iamalwaysuncomfortable, line 11: I think ultimately we want to find out how often this changes, I've heard slightly differing things. What would happen currently if the node sig changed?

iamalwaysuncomfortable (approved):
This is good, let's merge this

ledger/streaming/api/src/response.rs
remoun: I'm not sure what you mean by node sig. This key will be for signing and verifying the `SubscribeResponse`s

jcape: This is one possible reason to do generics now, to have a "semantic signing" element.

iamalwaysuncomfortable: I thought it was the node signature that was being passed along, I see about signing the subscribe responses. As far as the purpose of collecting blocks, ultimately client-side SCP would take care ensuring peer streams are indeed a consensus. If there's a hijacked stream in the set of peers, the SCP validation would catch it.

ledger/streaming/api/src/components.rs
```diff
@@ -0,0 +1,18 @@
...
+pub struct BlockStreamComponents {
+    /// The block data.
+    pub block_data: BlockData,
+    /// The SCP quorum set for this block.
+    /// Currently optional, will become required with a future BlockVersion.
+    pub quorum_set: Option<QuorumSet>,
+    /// The AVR for this block.
+    /// Currently optional, will become required with a future BlockVersion.
+    pub verification_report: Option<VerificationReport>,
+}
```
iamalwaysuncomfortable, line 18: Seems like this could possibly be merged with components.rs?

remoun: I don't understand your comment. This is `components.rs`...

ledger/streaming/api/src/convert/components.rs
remoun: We do have an `mc_ledger_streaming_api::Error`, but `mc_api::ConversionError` seems sufficient here, and matches the other conversion code in `mc_api::convert`. `ConversionError` is already an enum, so I expect Chris' work won't affect this.

remoun: Also note that `mc_ledger_streaming_api::Error` has an enum variant wrapping `ConversionError`

ledger/streaming/api/src/response.rs
remoun: After discussing offline with James, we agreed to have the consensus node sign the `SubscribeResponse`s, so I removed the `make_subscribe_response` calls from `GrpcServerSink` (in #1670 ), and switched these methods to use a digest of the `BlockStreamComponents`, rather than proto encoding.

ledger/streaming/api/src/components.rs
iamalwaysuncomfortable: I suspect consumers are eventually going to want a NodeID in the component, but I understand the motivation behind not having it. I'm not entirely sure that the overhead in memory/speed would be enough of a con against including it in the component object for the convenience of tracking, error handling, and business logic. Having to run an SCP discovery predicate/patch it into a stream every time you want to know the node you're dealing with sounds bad from a maintainability perspective.

It would be interesting to clock the pipeline speed with NodeID to see if makes any appreciable difference.

Other than that, this is good.

ledger/streaming/api/src/components.rs
```diff
@@ -2,10 +2,11 @@
 
 use mc_attest_core::VerificationReport;
 use mc_consensus_scp::QuorumSet;
+use mc_crypto_digestible::Digestible;
 use mc_transaction_core::BlockData;
 
 /// Wrapper for the components needed for block streaming.
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, Digestible, PartialEq, Eq)]
```
iamalwaysuncomfortable, line 9: This doesn't really need to be digestible IMO. Digesting is more for when dealing with things that are an immutable record that need to be verified later - these component objects are highly ephemeral and unlikely to be hashed/stored anywhere. With the blockstream, the hash will be changing with each new block/avr and the blocks/avrs already have digests consumers can compare downstream. Mishbehaving streaming nodes lying about quorum set are more easily detected by scanning those directly than inspecting a hash.

ledger/streaming/api/src/response.rs
```diff
@@ -13,8 +13,8 @@ pub fn make_subscribe_response(
     let mut proto = SubscribeResponse::new();
     proto.set_result(data.into());
 
-    let result_bytes = proto.get_result().write_to_bytes()?;
-    let signature = signer.try_sign(&result_bytes)?;
+    let digest = get_digest(data);
+    let signature = signer.try_sign(&digest)?;
```
iamalwaysuncomfortable, line 17: We shouldn't really need to sign these unless protobuf requires it? This is a fairly heavy operation that doesn't add much security benefit given downstream components necessarily need to do verification of blocks they receive from peers.

iamalwaysuncomfortable (commented):
I think this is good modulo questions about the need signing/hashing. I'd be happy to get this merged and base off it!

ledger/streaming/api/src/components.rs
jcape: @remoun and I discussed some changes to how we're storing data in S3 (and potentially the ledger) which should help this situation. There's a couple MCIPs that I need to write which will generate future PRs to update this code.

ledger/streaming/api/src/convert/components.rs
jcape: It sounds like there's no reason *not* to use `mc_ledger_streaming_api::Error`, then.

I'm fine to leave this here now, but only because we need to refactor *all* our proto stuff to make it easier for non-rust clients to use, and that's a good time to split up our monolithic API crates into smaller/more manageable.

ledger/streaming/client/src/grpc.rs
```diff
@@ -0,0 +1,141 @@
...
+use mc_util_uri::ConnectionUri;
+use std::{sync::Arc, time::Duration};
+
+/// A [BlockStream] that streams blocks using the `LedgerUpdates` gRPC API.
+#[derive(Display)]
+pub struct GrpcBlockSource {
+    /// The gRPC client
+    client: LedgerUpdatesClient,
+    /// The public key of the consensus node we're streaming from.
+    public_key: Arc<Ed25519Public>,
```
jcape, line 23: This component shouldn't be checking signatures on blocks it receives, there should be a downstream object which is doing that. The only thing this object should be doing is handling the gRPC to get the blocks.

ledger/streaming/api/src/response.rs
jcape: My point was that the validator should be the one checking signatures, not the gRPC components, so all this code should be removed.

jcape (approved):
So, I'm fine for this to get merged, with the knowledge that a lot of it will need to be reworked to support generic traits and remove unnecessary operations (e.g. gRPC streaming components shouldn't be checking signatures or even doing business logic).

ledger/streaming/client/src/grpc.rs
iamalwaysuncomfortable: Currently the validator does check the sigs & the SCP component checks quorums/combines based SCP rules

remoun: I see what you're getting at, but there's currently no way to propagate the `SubscribeResponse.result_signature`, so we basically have to verify when we convert from `SubscribeResponse` to `BlockStreamComponents`. I'm up for splitting this sink into an element that basically maps gRPC->local and another that converts `SubscribeResponse` to `BlockStreamComponents`, in a follow-up PR.

PR #1670, opened 2022-03-18
Add GrpcServerSink

### LINKED ISSUE #1573: Streaming gRPC wrappers for Block Streaming

remoun:
Resolves #1573

### DIFF
--- a/ledger/streaming/publisher/src/grpc.rs
+++ b/ledger/streaming/publisher/src/grpc.rs
@@ -1,0 +1,235 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
+//! A sink that consumes a [Stream] of [SubscribeResponse]s and publishes the
+//! results over gRPC [LedgerUpdates].
+
+use flo_stream::{ExpiringPublisher, MessagePublisher, Subscriber};
+use futures::{lock::Mutex, FutureExt, Stream, StreamExt, TryStreamExt};
+use grpcio::ServerStreamingSink;
+use mc_common::logger::Logger;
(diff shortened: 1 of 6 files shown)


iamalwaysuncomfortable (approved):
LGTM

iamalwaysuncomfortable:
I think Remoun was waiting on a review from @jcape, but this sits good from my perspective  Do we need prefactoring merged before this can be merged? If so, let's resolve the questions there and merge. I think we're about ready to start stichting components together.

ledger/streaming/publisher/src/grpc.rs
```diff
@@ -0,0 +1,230 @@
...
+                let block_index = resp
+                    .get_result()
+                    .get_block()
+                    .get_v1()
+                    .get_block()
+                    .get_index();
+                futures::future::ready(block_index < starting_height)
+            })
+            .map(|resp| Ok((resp, grpcio::WriteFlags::default())));
+        let fut = stream.forward(sink).map(|_| ());
```
iamalwaysuncomfortable, line 109: Would protobuf be able to send an actual stream object across or must it send futures?

iamalwaysuncomfortable (approved):
LGTM - only question is about the source and if we can send a stream object (perhaps as an option)

ledger/streaming/publisher/src/grpc.rs
remoun: I'm not sure what you mean

jcape (approved):

PR #1672, opened 2022-03-18
Add BackfillingStream

### LINKED ISSUE #1604: Write a block stream multiplexer
Implements `BlockStream`, takes a `BlockStream` and can backfill from S3.

remoun:
Resolves #1604

### DIFF
--- a/ledger/streaming/client/src/backfill.rs
+++ b/ledger/streaming/client/src/backfill.rs
@@ -1,0 +1,178 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
+//! A BlockStream that backfills missing indices from a given BlockFetcher.
+
+use displaydoc::Display;
+use futures::{FutureExt, Stream, StreamExt};
+use mc_common::logger::{log, Logger};
+use mc_ledger_streaming_api::{BlockFetcher, BlockStream, BlockStreamComponents, Result};
+use mc_transaction_core::BlockIndex;
+use std::{pin::Pin, sync::Arc};
+
+/// A [BlockStream] that backfills another [BlockStream] using a [BlockFetcher].
+#[derive(Debug, Display)]
+pub struct BackfillingStream<S: BlockStream + 'static, F: BlockFetcher + 'static> {
+    upstream: S,
+    fetcher: Arc<F>,
+    logger: Logger,
+}
+
+impl<S: BlockStream + 'static, F: BlockFetcher + 'static> BackfillingStream<S, F> {
+    /// Instantiate a [BackfillingStream].
+    pub fn new(upstream: S, fetcher: F, logger: Logger) -> Self {
+        Self {
+            upstream,
+            fetcher: Arc::new(fetcher),
+            logger,
+        }
+    }
+}
+
+impl<S: BlockStream + 'static, F: BlockFetcher + 'static> BlockStream for BackfillingStream<S, F> {
+    type Stream = impl Stream<Item = Result<BlockStreamComponents>>;
+
+    fn get_block_stream(&self, starting_height: u64) -> Result<Self::Stream> {
+        self.upstream
+            .get_block_stream(starting_height)
+            .map(|upstream| {
+                backfill_stream(
+                    upstream,
+                    starting_height,
+                    self.fetcher.clone(),
+                    self.logger.clone(),
+                )
+            })
+    }
+}
+
+fn backfill_stream<
+    S: Stream<Item = Result<BlockStreamComponents>> + 'static,
+    F: BlockFetcher + 'static,
+>(
+    upstream: S,
+    starting_height: u64,
+    fetcher: Arc<F>,
+    logger: Logger,
+) -> impl Stream<Item = Result<BlockStreamComponents>> {
+    use futures::stream::{empty, once};
+
+    let mut prev_index: Option<BlockIndex> = None;
+    upstream.flat_map(
+        move |result| -> Pin<Box<dyn Stream<Item = Result<BlockStreamComponents>>>> {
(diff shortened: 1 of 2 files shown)


ledger/streaming/client/src/backfill.rs
```diff
@@ -0,0 +1,178 @@
...
+                            index,
+                            prev_index.unwrap()
+                        );
+                        return Box::pin(empty());
+                    }
+
+                    let item_stream = once(async { Ok(components) });
+                    if index == next_index {
+                        prev_index = Some(index);
+                        Box::pin(item_stream)
```
iamalwaysuncomfortable, line 79: We could potentially use Pin-project here as it is meant to generating pinning in optimized ways

ledger/streaming/client/src/backfill.rs
```diff
@@ -0,0 +1,178 @@
...
+                        }
+                    }
+                }
+                Err(upstream_error) => {
+                    // If we get an error, fetch one item.
+                    match fetcher.fetch_single(next_index) {
+                        Ok(future) => {
+                            prev_index = Some(next_index);
+                            Box::pin(future.into_stream())
+                        }
```
iamalwaysuncomfortable, line 104: I wonder how this would cascade, what is the cost of a single backfill? Might this get expensive if a node is having a longer term failure + what should we do in case of a longer term failure? I think at some point we'll have to do more fine grained error handling based upon exactly what error gets passed down the line.

ledger/streaming/client/src/backfill.rs
```diff
@@ -0,0 +1,178 @@
...
+    upstream: S,
+    starting_height: u64,
+    fetcher: Arc<F>,
+    logger: Logger,
+) -> impl Stream<Item = Result<BlockStreamComponents>> {
+    use futures::stream::{empty, once};
+
+    let mut prev_index: Option<BlockIndex> = None;
+    upstream.flat_map(
+        move |result| -> Pin<Box<dyn Stream<Item = Result<BlockStreamComponents>>>> {
```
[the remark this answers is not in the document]
remoun: No, I had to box here because of the different stream types across the match arms. This boxed Stream is consumed by the `flat_map` call above.

I agree we should profile these pipelines when we have all the pieces.

ledger/streaming/client/src/backfill.rs
remoun: I don't think so, since the Box here is to use different types of streams (via `dyn`), not operating on members of a struct.

ledger/streaming/client/src/backfill.rs
remoun: Yeah, I expect we'll evolve the error handling with usage.

ledger/streaming/client/src/backfill.rs
```diff
@@ -0,0 +1,178 @@
...
+
+        let result_fut = source
+            .get_block_stream(0)
+            .expect("Failed to start upstream")
+            .map(|resp| resp.expect("expected no errors").block_data.block().index)
+            .collect::<Vec<_>>();
+
+        let result = block_on(result_fut);
+        assert_eq!(result, vec![0, 1, 2, 3, 4])
+    }
```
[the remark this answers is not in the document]
remoun: What would this test exercise? I suspect exercising the error handling would end up looking much like the `backfills_on_error` test below.

ledger/streaming/client/src/backfill.rs
```diff
@@ -0,0 +1,178 @@
...
+    fetcher: Arc<F>,
+    logger: Logger,
+}
+
+impl<S: BlockStream + 'static, F: BlockFetcher + 'static> BackfillingStream<S, F> {
+    /// Instantiate a [BackfillingStream].
+    pub fn new(upstream: S, fetcher: F, logger: Logger) -> Self {
+        Self {
+            upstream,
+            fetcher: Arc::new(fetcher),
```
iamalwaysuncomfortable, line 25: Perhaps, move the Arc creation to the stream object
```suggestion
            fetcher,
```

ledger/streaming/client/src/backfill.rs
```diff
@@ -0,0 +1,178 @@
...
+    type Stream = impl Stream<Item = Result<BlockStreamComponents>>;
+
+    fn get_block_stream(&self, starting_height: u64) -> Result<Self::Stream> {
+        self.upstream
+            .get_block_stream(starting_height)
+            .map(|upstream| {
+                backfill_stream(
+                    upstream,
+                    starting_height,
+                    self.fetcher.clone(),
```
iamalwaysuncomfortable, line 41: This is a clone, but it's an Arc clone where the actual arg is attached to the BackfillingStream object. This could potentially cause lifetime issues. Likely better to make the arc within the stream.

```suggestion
                    Arc::new(self.fetcher.clone()),
```

iamalwaysuncomfortable (changes requested):
Small nit about Arc creation location

ledger/streaming/client/src/backfill.rs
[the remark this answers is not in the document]
iamalwaysuncomfortable: Let's discuss Monday when we try to pull together all the components!

iamalwaysuncomfortable:
LGTM minus the one nit about arc, excited to get this merged

ledger/streaming/client/src/backfill.rs
remoun: This is intentional -- if the `BackfillingStream` somehow goes out of scope while one of its returned streams is running, those streams will retain a copy of the fetcher.

What you're suggesting would clone the fetcher, which is not what I want here.

iamalwaysuncomfortable (approved):

integral-llc (approved):

PR #1683, opened 2022-03-22
Create Test Vectors for SDK Engineers

### LINKED ISSUE #1561: Make More Fog Test Vectors
Many times, SDK team needs to bind to code in this repo and then they want to be able to unit test their bindings.

Getting test data is difficult because usually the rust tests fabricate random data in their unit tests. So this would be captured somehow and then hard-coded into the SDK test code.

To make this more maintainable, Kyle added the "data test" functionality, which allows that:
* test case data is generated by rust (and written to a json file)
* rust unit tests can load the data from the json file and test it
* the json file is committed and swift and java unit tests can also just load the test cases and run them

However, this was problematic for a while because of the architecture of data test. Remoun recently replaced the data test dependency with something simpler: https://github.com/mobilecoinfoundation/mobilecoin/pull/1556

We should make more extensive use of this to help the SDKs:
* test vectors for the different memo types, with expected pass and expected fail validation: https://github.com/mobilecoinfoundation/mobilecoin/blob/453756e0ac44872a7938be042bb04667cb1f201b/transaction/std/src/memo/mod.rs#L166
* test vectors for `TxOutRecord`, which has become more complicated after we started omitting the compressed commitment: https://github.com/mobilecoinfoundation/mobilecoin/blob/b7bfe32d697430b162a177a07206947a0471f972/fog/api/proto/view.proto#L219 ideally we would have test cases relevant to this functionality: https://github.com/mobilecoinfoundation/mobilecoin/blob/b7bfe32d697430b162a177a07206947a0471f972/fog/types/src/view.rs#L365

samdealy:
### Motivation
As described in #1561, SDK engineers often need to test their code using test data that is only able to be generated by Rust code that lives in mobilecoin.git. Recently, they needed test data related to memos and TxOutRecords, and they had to manually run unit tests and capture log outputs to create the needed test data. This is cumbersome and error-prone.

### In this PR
* Defines memo test vectors and generates correct and incorrect memos for the three types of memos.
* Defines TxOutRecord test vectors and generates correct and incorrect TxOutRecords.
* Uses the TxOutRecords to test the FogTxOut#try_recover_tx_out_method

### Notes
I chose to use hex-encoded bytes for the test vectors. In my experience, it's a lot easier to copy and paste a hex string than an array of bytes. Moreover, I use a naming convention for the fields that hex encode bytes to distinguish between proto bytes and raw bytes. See the updated README for more info.

Fixes #1561[


test-vectors/tx-out-records/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "mc-test-vectors-tx-out-records"
+version = "1.3.0-pre0"
+authors = ["MobileCoin"]
+edition = "2018"
+
+[dependencies]
+hex = "0.4"
+mc-test-vectors-definitions = { path = "../definitions" }
```
remoun, line 9: Nit: Separate MC and external deps, throughout this file.

test-vectors/tx-out-records/build.rs
```diff
@@ -0,0 +1,161 @@
...
+                recipient_view_private_key_hex_proto_bytes: hex::encode(mc_util_serial::encode(
+                    &tx_out_record_data.recipient_view_private_key,
+                )),
+                tx_out_record_hex_proto_bytes: hex::encode(mc_util_serial::encode(tx_out_record)),
+            };
+
+            correct_tx_out_record_data_set.push(correct_tx_out_record_data);
+        }
+
+        correct_tx_out_record_data_set
```
remoun, line 40: Optional/Equivalent: Use iterator chaining, e.g.
```suggestion
        let tx_out_record_data = generate_tx_out_record_data();
        let private_key_hex_proto_bytes = hex::encode(mc_util_serial::encode(
            &tx_out_record_data.recipient_view_private_key,
        ));
        tx_out_record_data
            .tx_out_records 
            .iter()
            .map(|tx_out_record| {
                CorrectTxOutRecordData {
                    recipient_view_private_key_hex_proto_bytes: private_key_hex_proto_bytes.clone(),
                    tx_out_record_hex_proto_bytes: hex::encode(mc_util_serial::encode(tx_out_record)),
                }
            })
            .collect()
```

test-vectors/tx-out-records/Cargo.toml
samdealy: I had it that way before, but I think it was giving me issues with the linter's "sorted dependency" requirement. Regardless, I prefer it to be this way because when I want to find a dependency, it's easier if the dependencies exist in one sorted block. If there are multiple blocks, then I sometimes have trouble quickly locating the dependency, and have to resort to ctrl - F.

test-vectors/tx-out-records/build.rs
samdealy: I think this is a good suggestion, and I will implement this in a follow up PR.

remoun (approved):

dolanbernard (approved):
Looks great!

garbageslam:
Thank you Sam!

Issue #1618, opened 2022-03-14
Block Stream Verification 


iamalwaysuncomfortable:
As part of block streaming framework (#1433), the blocks being sent by each stream need to be properly validated.

Validations that need to be done include:
* Block Hash validation (i.e. block contents need to be hashed matched with the Block Id)
* Parent block ID validation (to ensure it's part of an ordered stream)
* Attestation Verification Report (AVR) validation

Streams will be grabbed from an upstream source. When streams arrive, validation should be done on each stream's blocks. They should then go through SCP validation, and passing that, then combined into a single stream that the block sink component can take to ingest.

PR #1690, opened 2022-03-23
Block data validation for block streaming

iamalwaysuncomfortable:
### Motivation

Blocks need to be validated before being sent on to other block streaming components like server or database Block Sinks. This component implements block validation within the streaming framework via.

### In this PR
* A stream combinator struct which takes an upstream, does validation computations, and passes either the block data or a validation error along the chain of streams
* Addition of small useful errors for block streaming
* Addition of a small test utility

Currently WIP: the following validations need to be added in
* Validation of parent block id
* Optional validation of key images (based on whether the stream creator has access to past key images)
* Optional attestation verification (based on whether stream creator has access to AVRs)

Addresses #1618 

### Future Work
* This is one of the most CPU heavy pieces of the block streaming framework and offloading computations to something like an associated thread pool in an async aware way might be desired.
* SCP validation of blocks needs to combine several of these streams into a single, consensus stream


ledger/streaming/client/src/block_validator.rs
```diff
@@ -0,0 +1,216 @@
...
+        //let logger = self.logger.clone();
+        let prev_block_id = self.start_block.id.clone();
+        let ledger = self.ledger.clone();
+        let additional_key_images: HashSet<KeyImage> = HashSet::default();
+
+        log::info!(self.logger, "Creating block validation stream");
+        let stream = self.upstream.get_block_stream(starting_height)?;
+
+        Ok(stream.scan(
+            (ledger, prev_block_id, additional_key_images),
```
remoun, line 51: Consider using a dedicated struct for this. I expect it would make the state tracking easier to follow

ledger/streaming/client/src/block_validator.rs
```diff
@@ -0,0 +1,216 @@
...
+    BlockStream, BlockStreamComponents, Error as StreamError, Result as StreamResult,
+};
+use mc_transaction_core::{compute_block_id, ring_signature::KeyImage, Block};
+
+/// Create stream factory for validating individual blocks within a stream.
+/// Valid blocks will passed on, blocks that don't pass will pass an error.
+pub struct BlockValidator<US: BlockStream + 'static, L: Ledger + 'static> {
+    upstream: US,
+    start_block: Block,
+    ledger: Option<L>,
```
remoun, line 21: Why is the ledger optional here? IIUC this validation doesn't work without a Ledger

remoun (commented):
I'm not sure how this fits in with #1744, especially with the duplicated code across these 2 PRs.

ledger/streaming/client/src/block_validator.rs
iamalwaysuncomfortable: Agreed

ledger/streaming/client/src/block_validator.rs
iamalwaysuncomfortable: It's because. If there is no ledger (i.e. if it's a pass through component) and someone wants to validate the blocks coming from a peer, they can. The ledger step is required if you're somehow writing into a ledger.

iamalwaysuncomfortable:
> I'm not sure how this fits in with #1744, especially with the duplicated code across these 2 PRs.

Via Discussion with James that they should be decoupled. 

The validation components differ in the following way:
1. The above PR is validation of individual block correctness (i.e. the blocks they contain no spent key images, the block id matches the computed proof, and it has the correct parent). 
2. SCP Validation is validating that the peers you've chosen have come to consensus/are saying the same things

It was stated that these two validations are different enough that they should be decoupled. Coupling them would force a node to have to do both in a certain designed order. 

For instance if you're a repeater node you might have
1 Consensus Node Peer --> BlockData --> Validation --> gRPC broadcast

A particular consumer might have
gRPC Client from 8 peers --> SCP Consensus --> Individual Block Validation --> DBSink

PR #1694, opened 2022-03-24
Test transaction automation script for mobilecoind release testing

iamalwaysuncomfortable:
### Motivation

Working out mobilecoind based release issues has historically been very much a pain point and done manually. This script which works via mobilecoin-d json attempts to allow transaction testing via mobilecoind in a one command invocation. This alternatively can be used with the test network.

### In this PR
* A script which uses mobilecoind & mobilecoind-json to create payment monitors, import addresses, and start a payment loop (requires funded accounts)

### Future Work
* Allow for slightly more configuration possibly (pending outside opinions)

### DIFF
--- a/tools/local-network/send-mobilecoind-txs.sh
+++ b/tools/local-network/send-mobilecoind-txs.sh
@@ -3,21 +3,36 @@
 
 ## .json file with account keys in form {"view_private_key":"..", "spend_private_key":".."}
 
+## Amount to send
+AMOUNT=0.001
+## Location of keyfile with view_private_key and spend_private_key
 KEYFILE=""
+## Length to sleep between sending transactions
+SLEEP=1.5
+## Curl encoding
 enc="Content-Type: application/json"
 
 while [ "$1" != "" ]; do
-	case $1 in
+	case "$1" in
+		-a | --amount)
+		  shift
+		  AMOUNT="$1";;
 		-k | --keyfile)
-		  KEYFILE=$1
+		  shift
+		  KEYFILE="$1";;
+		-s | --sleep)
+		  shift
+		  SLEEP="$1";;
 		-h)
 			echo "Imports key file for an account starts sending test transactions via mobilecoind-json"
 			echo "Note: both mobilecoind & mobilecoin-json must both be launched prior to running this script"
 			echo "Usage: $(basename $0) [-k | --keyfile]"
 			echo ""
 			echo "Args:"
-			echo "[-k |--keyfile] location of keyfile with view & spend private keys in format:"
+			echo "[-a | --amount] amount to send per transaction"
+			echo "[-k | --keyfile] location of keyfile with view & spend private keys in format: "
 			echo '{"view_private_key":"..", "spend_private_key":".."}'
+			echo "[-s | --sleep] period to sleep between sending transactions"
 			exit 0
 			;;
 		*)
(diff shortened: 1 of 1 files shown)


tools/local-network/send-mobilecoind-txs.sh
```diff
@@ -0,0 +1,60 @@
...
+do
+	for var in 1
+	do
+		m="mon_id${var}"
+		p="payrpc${var}"
+		declare -n mon="$m"
+		declare -n pay="$p"
+		sleep 1.5
+		r=$(curl localhost:9090/monitors/$mon/subaddresses/0/pay-address-code -d "$pay" -X POST -H "$enc")
+		echo $r
```
remoun, line 58: If we're just echoing the output, we might as well call the command directly
```suggestion
		curl localhost:9090/monitors/$mon/subaddresses/0/pay-address-code -d "$pay" -X POST -H "$enc"
```

tools/local-network/send-mobilecoind-txs.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
```
remoun, line 1: I wonder if this would be better as a Python script. We'd get better structs/encapsulation and cleaner arrays/looping, as well as easier JSON de/serialization

tools/local-network/send-mobilecoind-txs.sh
```diff
@@ -0,0 +1,60 @@
...
+#Setup monitor and get it's id and get the account address associated with the keyfile
+monitor1=$(curl localhost:9090/monitors -d "$mon_data1" -X POST -H "$enc")
+mon_id1=$(echo $monitor1 | jq -r ".monitor_id")
+add1=$(curl localhost:9090/monitors/$mon_id1/subaddresses/0/public-address | jq -r ".b58_address_code")
+
+#Setup payment invocations
+payrpc1="{\"receiver_b58_address_code\": \"$add1\",\"value\": \"0.001\"}"
+r1=$(curl localhost:9090/monitors/$mon_id1/subaddresses/0/pay-address-code -d "$payrpc1" -X POST -H "$enc")
+
+#Run payments to self every 1.5 seconds. More addresses can be added above if desired
```
remoun, line 47: It seems useful to support CLI args.

jcape (commented):
LGTM, will approve after comments resolved.

tools/local-network/send-mobilecoind-txs.sh
iamalwaysuncomfortable: I'd like to keep the echo in here actually as this prints the success/failure of the transaction

tools/local-network/send-mobilecoind-txs.sh
iamalwaysuncomfortable: Potentially, I think it would add more bloat than is necessary at this point

tools/local-network/send-mobilecoind-txs.sh
iamalwaysuncomfortable: Yeah amount config would be useful as a cli arg certainly, I'll add those

iamalwaysuncomfortable: Done!

tools/local-network/send-mobilecoind-txs.sh
```diff
@@ -0,0 +1,72 @@
...
+add1=$(curl localhost:9090/monitors/$mon_id1/subaddresses/0/public-address | jq -r ".b58_address_code")
+
+#Setup payment invocations
+payrpc1="{\"receiver_b58_address_code\": \"$add1\",\"value\": \"$AMOUNT\"}"
+r1=$(curl localhost:9090/monitors/$mon_id1/subaddresses/0/pay-address-code -d "$payrpc1" -X POST -H "$enc")
+
+#Run payments to self every 1.5 seconds. More addresses can be added above if desired
+while [ true ]
+do
+	for var in 1
```
remoun, line 62: Do you expect we'll use other vars? If not, how about we omit this loop?

tools/local-network/send-mobilecoind-txs.sh
```diff
@@ -0,0 +1,72 @@
...
+	do
+		m="mon_id${var}"
+		p="payrpc${var}"
+		declare -n mon="$m"
+		declare -n pay="$p"
+		sleep $SLEEP
+		r=$(curl localhost:9090/monitors/$mon/subaddresses/0/pay-address-code -d "$pay" -X POST -H "$enc")
+		echo $r
+	done
+done
```
remoun, line 72: Nit: Add newline
```suggestion
done

```

tools/local-network/send-mobilecoind-txs.sh
```diff
@@ -0,0 +1,72 @@
...
+enc="Content-Type: application/json"
+
+while [ "$1" != "" ]; do
+	case $1 in
+		-a | --amount)
+		  AMOUNT=$1
+		-k | --keyfile)
+		  KEYFILE=$1
+		-s | --sleep)
+		  SLEEP=$1
```
remoun, line 22: IIUC these should consume the next arg
```suggestion
		-a | --amount)
		  shift
		  AMOUNT=$1
		-k | --keyfile)
		  shift
		  KEYFILE=$1
		-s | --sleep)
		  shift
		  SLEEP=$1
```

tools/local-network/send-mobilecoind-txs.sh
iamalwaysuncomfortable: Yeah originally I had 4 against the test network, so we likely will, but I wanted to keep it simple so it doesn't require more than one to be run if not needed as getting account keys is a painful process

tools/local-network/send-mobilecoind-txs.sh
iamalwaysuncomfortable: I'd prefer to keep the indent, I think it makes more legible

tools/local-network/send-mobilecoind-txs.sh
iamalwaysuncomfortable: Done

remoun: What you have now will set `AMOUNT="-a"` and so on -- the loop already checks `$1`, so we need to consume it. You could `shift` after assigning `X=$2` (not `$1`)

iamalwaysuncomfortable: Correct! Fix'd, tested against the test network

wjuan-mob (approved):

jcape (approved):

PR #1698, opened 2022-03-25
decrement block version everywhere, bump to -pre2

garbageslam:
we have to decrement block version because of a compatibility issue
it turns out that v1.0 always published block version = 0, and not
one as the source code indicated after a bunch of development.

### DIFF
--- a/api/proto/blockchain.proto
+++ b/api/proto/blockchain.proto
@@ -65,6 +65,7 @@ message BlockSignature {
 }
 
 // Version 1 of an archived block.
+// Note: The block.version field within the block may or may not be equal to 1.
 message ArchiveBlockV1 {
     // Block
     Block block = 1;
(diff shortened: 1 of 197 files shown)


transaction/core/src/blockchain/block_version.rs
```diff
@@ -69,13 +71,13 @@ impl BlockVersion {
     /// Iterator over block versions from one up to max, inclusive. For use in
     /// tests.
     pub fn iterator() -> BlockVersionIterator {
-        BlockVersionIterator(1)
+        BlockVersionIterator(0)
     }
 
     /// The encrypted memos [MCIP #3](https://github.com/mobilecoinfoundation/mcips/pull/3)
-    /// feature is introduced in block version 2.
+    /// feature is introduced in block version 1.
```
garbageslam, line 78: will write a new MCIP soon

api/proto/blockchain.proto
```diff
@@ -65,6 +65,7 @@ message BlockSignature {
 }
 
 // Version 1 of an archived block.
+// Note: The block.version field within the block may or may not be equal to 1.
```
sugargoat, line 68: Thank you for this clarifying comment!

sugargoat (approved):

remoun (approved):

PR #1700, opened 2022-03-25
decrement block version everywhere (master branch)

garbageslam:
we have to decrement block version because of a compatibility issue
it turns out that v1.0 always published block version = 0, and not
one as the source code indicated after a bunch of development.

this was made by cherry-picking the fix from release 1.2 branch, then removing `BlockVerision::THREE` for now and fixing up a bunch of things in minting and ring signature logic and tests.

### DIFF
--- a/ledger/db/src/lib.rs
+++ b/ledger/db/src/lib.rs
@@ -2528,7 +2528,7 @@ mod ledger_db_test {
             ..Default::default()
         };
         let block = Block::new_with_parent(
-            BlockVersion::ZERO,
+            BLOCK_VERSION,
             &origin_block,
             &Default::default(),
             &block_contents,
@@ -2570,12 +2570,8 @@ mod ledger_db_test {
             ..Default::default()
         };
         let parent = ledger_db.get_block(n_blocks - 1).unwrap();
-        let block = Block::new_with_parent(
-            BlockVersion::ZERO,
-            &parent,
-            &Default::default(),
-            &block_contents,
-        );
+        let block =
+            Block::new_with_parent(BLOCK_VERSION, &parent, &Default::default(), &block_contents);
 
         ledger_db
             .append_block(&block, &block_contents, None)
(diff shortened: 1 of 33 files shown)


ledger/db/src/lib.rs
```diff
@@ -2528,7 +2528,7 @@ mod ledger_db_test {
             ..Default::default()
         };
         let block = Block::new_with_parent(
-            BlockVersion::ONE,
+            BlockVersion::ZERO,
```
remoun, line 2531: How about we use the `BLOCK_VERSION` const throughout this file?

remoun (dismissed):

ledger/db/src/lib.rs
garbageslam: Will fix

eranrund (dismissed):

remoun (approved):

iamalwaysuncomfortable (approved):

transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -394,7 +394,7 @@ fn compute_extended_message_either_version(
     pseudo_output_commitments: &[CompressedCommitment],
     range_proof_bytes: &[u8],
 ) -> Vec<u8> {
-    if block_version >= BlockVersion::THREE {
+    if block_version >= BlockVersion::TWO {
```
jcape, line 397: Shouldn't this use the `BlockVersion::has_feature()` API?

garbageslam: it's not exactly the same as the token id feature

this is basically, there was a lack of proper framing and domain separation when creating the "message" that is signed by the MLSAG, and I wanted to fix it.

we could add a new feature like "fix extended message framing" and make it contingent on that I guess. that seems fine to me. lmk what you want

garbageslam: I think any fix should be in a follow up PR, this is already too big

Issue #1639, opened 2022-03-16
Improve transaction output sorting requirement unit tests


jcape:
Per [PR comments](https://github.com/mobilecoinfoundation/mobilecoin/pull/1615#discussion_r826145270) while making transaction output sorting mandatory, we should improve the unit tests of transaction validation to also test the validate function for the positive/negative cases of version-dependent validation.

PR #1705, opened 2022-03-28
added a mechanism to generate unsorted outputs to be able to execute …

integral-llc:
…the global transaction validate method

### Motivation

we didn't have a way to generate a wrongly sorted transaction outputs.  

### In this PR
added a mechanism to sort outputs using a custom comparator

#1639

### DIFF
--- a/transaction/core/src/validation/validate.rs
+++ b/transaction/core/src/validation/validate.rs
@@ -510,9 +510,11 @@ mod tests {
     use mc_crypto_keys::{CompressedRistrettoPublic, ReprBytes};
     use mc_ledger_db::{Ledger, LedgerDB};
     use mc_transaction_core_test_utils::{
-        create_ledger, create_transaction, create_transaction_with_amount, initialize_ledger,
-        AccountKey, INITIALIZE_LEDGER_AMOUNT,
+        create_ledger, create_transaction, create_transaction_with_amount_and_comparer,
+        initialize_ledger, outputs_comparer_inverse, outputs_comparer_normal, AccountKey,
+        INITIALIZE_LEDGER_AMOUNT,
     };
+    use mc_transaction_std::TxOutputsComparerOption;
     use rand::{rngs::StdRng, SeedableRng};
     use serde::{de::DeserializeOwned, ser::Serialize};
 
@@ -567,6 +569,20 @@ mod tests {
         amount: u64,
         fee: u64,
     ) -> (Tx, LedgerDB) {
+        create_test_tx_with_amount_and_comparer(
+            block_version,
+            amount,
+            fee,
+            Some(outputs_comparer_normal),
+        )
+    }
+
+    fn create_test_tx_with_amount_and_comparer(
+        block_version: BlockVersion,
+        amount: u64,
+        fee: u64,
+        outputs_comparer_option: TxOutputsComparerOption,
+    ) -> (Tx, LedgerDB) {
         let mut rng: StdRng = SeedableRng::from_seed([1u8; 32]);
         let sender = AccountKey::random(&mut rng);
         let mut ledger = create_ledger();
@@ -584,7 +600,7 @@ mod tests {
         let tx_out = block_contents.outputs[0].clone();
 
         let recipient = AccountKey::random(&mut rng);
-        let tx = create_transaction_with_amount(
+        let tx = create_transaction_with_amount_and_comparer(
             adapt_hack(&block_version),
             &mut ledger,
             &tx_out,
@@ -594,6 +610,7 @@ mod tests {
             fee,
             n_blocks + 1,
             &mut rng,
+            outputs_comparer_option,
         );
 
         (adapt_hack(&tx), ledger)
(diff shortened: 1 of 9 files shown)


transaction/std/Cargo.toml
```diff
@@ -4,8 +4,12 @@ version = "1.3.0-pre0"
 authors = ["MobileCoin"]
 edition = "2018"
 
+[features]
+test-only=[]
```
jcape, line 8:
```suggestion
test-only = []
```

jcape (changes requested):
Why can't this use a functor-esque mechanism with a comparison trait that contains a single function, and is compiled statically?

```
pub trait TxOutputSorter {
    fn sort(a: &RistrettoPublic, b: &RistrettoPublic) -> Ordering;
}

struct DefaultTxOutputSorter;

impl TxOutputSorter for DefaultTxOutputSorter {
    /* actual impl here */
}

impl TransactionBuilder { 
    /* ... */
    pub fn build<RNG: CryptoRng + RngCore, S: TxOutputSorter = DefaultTxOutputSorter>(/* .. */) {
        /* .. */
    }
}
```

transaction/std/src/transaction_builder.rs
```diff
@@ -23,6 +23,12 @@ use mc_transaction_core::{
 use mc_util_from_random::FromRandom;
 use rand_core::{CryptoRng, RngCore};
 
+/// This is a datatype for the comparer of the transaction outputs
+pub type TxComparerFn = fn(&TxOut, &TxOut) -> core::cmp::Ordering;
+
+/// This is a datatype helper used for transaction outputs sorting mechanism
+pub type TxOutputsComparerOption = Option<TxComparerFn>;
+
```
jcape, line 31: When doing stuff like this you should just use the Option in-situ.

transaction/core/src/validation/validate.rs
```diff
@@ -510,9 +510,11 @@ mod tests {
     use mc_crypto_keys::{CompressedRistrettoPublic, ReprBytes};
     use mc_ledger_db::{Ledger, LedgerDB};
     use mc_transaction_core_test_utils::{
-        create_ledger, create_transaction, create_transaction_with_amount, initialize_ledger,
-        AccountKey, INITIALIZE_LEDGER_AMOUNT,
+        create_ledger, create_transaction, create_transaction_with_amount_and_comparer,
+        initialize_ledger, outputs_comparer_inverse, outputs_comparer_normal, AccountKey,
```
iamalwaysuncomfortable, line 514: 2 name change suggestions: 
1. outputs_comparer_inverse --> inverse_outputs_comparer 
2. create_transaction_with_amount could probably stay the same given it has an optional parameter

transaction/core/test-utils/src/lib.rs
```diff
@@ -38,6 +40,14 @@ pub fn create_ledger() -> LedgerDB {
     LedgerDB::open(path).unwrap()
 }
 
+pub fn outputs_comparer_normal(a: &TxOut, b: &TxOut) -> core::cmp::Ordering {
+    a.public_key.cmp(&b.public_key)
+}
+
+pub fn outputs_comparer_inverse(a: &TxOut, b: &TxOut) -> core::cmp::Ordering {
+    b.public_key.cmp(&a.public_key)
+}
```
iamalwaysuncomfortable, line 49: perhaps forward_sort_txouts and inverse_sort_txouts to keep in convention of naming functions as verbs

transaction/core/test-utils/src/lib.rs
```diff
@@ -96,6 +106,42 @@ pub fn create_transaction_with_amount<L: Ledger, R: RngCore + CryptoRng>(
     fee: u64,
     tombstone_block: BlockIndex,
     rng: &mut R,
+) -> Tx {
+    create_transaction_with_amount_and_comparer(
```
iamalwaysuncomfortable, line 110: What's the tradeoff between creating a new function for this and just including an option in the existing fn "create_transaction_with_amount"?

iamalwaysuncomfortable (commented):
In the review above:
- Potential renames for clarity
- Asked some questions about the tradeoff of just condensing the test function into existing one with extra parameter

transaction/std/Cargo.toml
integral-llc: fixed

transaction/core/test-utils/src/lib.rs
integral-llc: changed to traits now, fixed

transaction/core/src/validation/validate.rs
integral-llc: changed to traits, naming changed

transaction/std/src/transaction_builder.rs
integral-llc: not using anymore, fixed

transaction/std/src/transaction_builder.rs
```diff
@@ -22,12 +22,22 @@ use mc_transaction_core::{
...
 use mc_util_from_random::FromRandom;
 use rand_core::{CryptoRng, RngCore};
+use std::cmp::Ordering;
 
-/// This is a datatype for the comparer of the transaction outputs
-pub type TxComparerFn = fn(&TxOut, &TxOut) -> core::cmp::Ordering;
+/// A trait used to compare the transaction outputs
+pub trait TxOutputsOrdering {
+    /// comparer method
+    fn cmp(&self, a: &CompressedRistrettoPublic, b: &CompressedRistrettoPublic) -> Ordering;
```
jcape, line 30: This probably doesn't need self, it can be a static method.

integral-llc: fixed

transaction/std/src/transaction_builder.rs
```diff
@@ -22,12 +22,22 @@ use mc_transaction_core::{
...
 use rand_core::{CryptoRng, RngCore};
+use std::cmp::Ordering;
 
-/// This is a datatype for the comparer of the transaction outputs
-pub type TxComparerFn = fn(&TxOut, &TxOut) -> core::cmp::Ordering;
+/// A trait used to compare the transaction outputs
+pub trait TxOutputsOrdering {
+    /// comparer method
+    fn cmp(a: &CompressedRistrettoPublic, b: &CompressedRistrettoPublic) -> Ordering;
+}
```
iamalwaysuncomfortable, line 31: A trait seems like maybe too much work to achieve this? Would this be useful elsewhere, if so that could justify this approach.

iamalwaysuncomfortable (approved):
I'd be okay to approve this modulo given someone else giving their thoughts on the trait based approach

transaction/std/src/transaction_builder.rs
integral-llc: what other way would you recomend?
cc: @jcape

jcape: This was my suggestion, because I don't think we should be afraid of generics, traits, and static dispatch.

In a normal build, there's no reason to ever change this sorting, so we shouldn't allow access to change it, and with a trait/functor, the compiler should be able to generate identical assembly for this vs. the original.

jcape (approved):

PR #1729, opened 2022-03-29
Use `verify_strict` instead of `verify` in crypto-multisig

garbageslam:
This is expected to mitigate malleability issues in ed25519
signatures.

Per documentation:

https://doc.dalek.rs/ed25519_dalek/struct.PublicKey.html#on-the-multiple-sources-of-malleability-in-ed25519-signatures

There are two sources of malleability in RFC 8032 ed25519 signatures:

* Scalar Malleability
* Point malleability

This method performs both of the above signature malleability checks.


garbageslam:
thank you, I didn't realize that

PR #1730, opened 2022-03-29
add block-version field to propose tx response (#1720) (release-1.2.0)

### LINKED ISSUE #1718: Add a `block_version` field to `ProposeTxResponse`

garbageslam:
Fixes #1718 (in release 1.2.0 branch)

This is a cherry-pick of https://github.com/mobilecoinfoundation/mobilecoin/pull/1720

### DIFF
--- a/consensus/service/src/api/client_api_service.rs
+++ b/consensus/service/src/api/client_api_service.rs
@@ -153,7 +153,6 @@ mod client_api_tests {
         counters,
         tx_manager::{MockTxManager, TxManagerError},
     };
-    use clap::Parser;
     use grpcio::{
         ChannelBuilder, Environment, Error as GrpcError, RpcStatusCode, Server, ServerBuilder,
     };
@@ -177,6 +176,7 @@ mod client_api_tests {
     use mc_util_grpc::{AnonymousAuthenticator, TokenAuthenticator};
     use serial_test_derive::serial;
     use std::{sync::Arc, time::Duration};
+    use structopt::StructOpt;
 
     /// Starts the service on localhost and connects a client to it.
     fn get_client_server(instance: ClientApiService) -> (ConsensusClientApiClient, Server) {
@@ -196,7 +196,7 @@ mod client_api_tests {
 
     /// Get a dummy config object
     fn get_config() -> Config {
-        Config::try_parse_from(&[
+        Config::from_iter(&[
             "foo",
             "--peer-responder-id=localhost:8081",
             "--client-responder-id=localhost:3223",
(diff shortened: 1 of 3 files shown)


consensus/service/src/api/client_api_service.rs
```diff
@@ -187,6 +194,25 @@ mod client_api_tests {
...
+            "--client-responder-id=localhost:3223",
+            "--msg-signer-key=MC4CAQAwBQYDK2VwBCIEIC50QXQll2Y9qxztvmsUgcBBIxkmk7EQjxzQTa926bKo",
+            "--network=network.toml",
+            "--peer-listen-uri=insecure-mcp://0.0.0.0:8081/",
+            "--client-listen-uri=insecure-mc://0.0.0.0:3223/",
+            "--admin-listen-uri=insecure-mca://0.0.0.0:9090/",
+            "--sealed-block-signing-key=/tmp/key",
+            "--ledger-path=/tmp/ledger",
+            "--ias-spid=22222222222222222222222222222222",
+            "--ias-api-key=asdf",
```
integral-llc, line 211: should not we have here an entry for `block_version` which is used above on line 138? or that's a default value coming form somewhere?

eranrund (approved):

consensus/service/src/api/client_api_service.rs
garbageslam: so what i'm doing here is actually passing an array of string literals which the `clap` code parses the same way it would parse `argc, argv`. So anything that is defaulted in the struct will be defaulted here also, I only had to list the required parameters to start the server

garbageslam: i did it that way because it was less complicated than doing `Config { ... } ` and actually initializing everything

integral-llc (approved):

consensus/service/src/api/client_api_service.rs
garbageslam: and yeah i think block-version is defaulting to zero for backwards compat

the-real-adammork (approved):

PR #1761, opened 2022-04-01
Make GetLastBlockInfo return the configured block version, and use it

garbageslam:
We realized that there is a chicken and egg problem for many clients around
the block version:

* For mobilecoind and for fog clients, they were getting block version
  from their local copy of the ledger, or fog-ledger's copy of the ledger
* However, when we upgrade consensus and increase block version, the
  first transaction takes place when the ledger has the old block version
  but the consensus config is the new block version.
* There has to be at least one client capable of determining the configured
  block version and making a transaction according to those rules.

To fix this, James suggested that we use the `GetLastBlock` call, where
most clients are getting the minimum fees right now. We added a field
on this response that returns the configured block version, so this is
a good source of truth for the mentioned situation. Then we made mobilecoind,
fog-distro, and fog sample paykit all track the responded value and incorporate
it into their caches, so they are tracking the max of the ledger and this
response value.

This will also ease CD work because fog distro doesn't need to be told
the block version any more, and none of these other clients will have
problems after we deploy a new version.

We may want to backport this to 1.2.0 also.

### DIFF
--- a/fog/distribution/src/config.rs
+++ b/fog/distribution/src/config.rs
@@ -73,10 +73,6 @@ pub struct Config {
     #[clap(long, default_value = "0", env = "MC_ADD_TX_DELAY_MS")]
     pub add_tx_delay_ms: u64,
 
-    /// Block version to use (otherwise fall back to ledger)
-    #[clap(long, env = "MC_BLOCK_VERSION")]
-    pub block_version: Option<u32>,
-
     /// Destination keys subdirectory. Defaults to `fog_keys`
     #[clap(long, default_value = "fog_keys", env = "MC_FOG_KEYS_SUBDIR")]
     pub fog_keys_subdir: String,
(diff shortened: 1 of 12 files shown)


eranrund (approved):

the-real-adammork (approved):

fog/distribution/src/config.rs
```diff
@@ -73,10 +73,6 @@ pub struct Config {
     #[clap(long, default_value = "0", env = "MC_ADD_TX_DELAY_MS")]
     pub add_tx_delay_ms: u64,
 
-    /// Block version to use (otherwise fall back to ledger)
-    #[clap(long, env = "MC_BLOCK_VERSION")]
-    pub block_version: Option<u32>,
```
remoun, line 78: This still seems useful for trying older block versions

remoun (approved):

fog/distribution/src/config.rs
garbageslam: maybe, but this isn't a very convenient tool for that kind of testing

PR #1767, opened 2022-04-01
mobilecoind python test logging and dependency update. 

jgreat:
### Motivation

The current output for this scrip is very verbose.  Set up basic logging and sort output into info/warn/error/debug. Default output is now `info`. 

The dependencies for this script were pretty outdated.  Update dependencies to support Ubuntu 20.04 image without compiling grpc. 

### In this PR

* update requirements.txt 
* replace print statements with logging.


mobilecoind/strategies/accounts.py
```diff
@@ -12,6 +14,9 @@
 from random import randint
 from google.protobuf.empty_pb2 import Empty
 
+logging.basicConfig(stream = sys.stdout, level = logging.INFO)
```
iamalwaysuncomfortable, line 17: Is stdout where we want, or stderr or somewhere else too?

In general we should have a python logging module which sets config for all files and log with that instead of printing across all python files.

For now:
```suggestion
logging.basicConfig(stream = sys.stdout, level = logging.INFO, format='"{levelname} {module}:{lineno}: {message}", style="{"')
```

mobilecoind/strategies/accounts.py
```diff
@@ -119,20 +124,20 @@ def poll_mitosis(starting_balance, account_data, tx_stats, stub):
...
+    logging.debug(tx_stats)
     return tx_stats
 
 
 def poll(monitor_id, tx_stats, stub):
     complete = {t: False for t in tx_stats.keys()}
     receipts = {t: tx_stats[t] for t in tx_stats.keys()}
     pending = complete.keys()
-    while not all(complete.values()):
+    while not all(complete.values()):
```
iamalwaysuncomfortable, line 140:
```suggestion
    while not all(complete.values()):
```

mobilecoind/strategies/requirements.txt
```diff
@@ -1,22 +1,30 @@
 appnope==0.1.0
 backcall==0.1.0
-boto3==1.9
-botocore==1.4.93
+boto3==1.21.13
+botocore==1.24.13
+certifi==2021.10.8
+chardet==3.0.4
```
iamalwaysuncomfortable, line 6: Was the dependency update intentional?

mobilecoind/strategies/accounts.py
```diff
@@ -73,25 +78,25 @@ def register_random_key(stub, outdir) -> AccountData:
 
 
 def wait_for_accounts_sync(stub, accounts, wait_secs):
-    print("accounts = ", accounts[0])
+    logging.debug("accounts = %s", accounts[0])
     block_count = stub.GetLedgerInfo(Empty()).block_count
     synced_ids = {a: False for a in accounts}
     while not all(synced_ids.values()):
-        print(".", end="", flush=True)
+        logging.info(".")
```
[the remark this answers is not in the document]
jgreat: Actually this is useful output (or well maybe could be more useful with "waiting for sync" or something). One of the common failure is I've seen is a the script just hanging on sync, its good to know that its still trying.

mobilecoind/strategies/accounts.py
jgreat: We collect both stdout/stderr streams on the container, but in general all logging output should go to stdout by default.

I don't disagree that we could have a logging module, but that's way beyond the simple cleaning up the output I want to achieve here.

mobilecoind/strategies/requirements.txt
jgreat: Short answer is yes. The main goal was getting grpcio up to date so I could use a pre-compiled pip package on Ubuntu 20.04. This of course ended up requiring a chain of updates and some new packages added.

mobilecoind/strategies/test_client.py
```diff
@@ -11,16 +11,17 @@
 ```
 """
 import argparse
-import concurrent.futures
+import logging
+import sys
 import glob
-import grpc
```
jgreat, line 16: Clean up unused imports

mobilecoind/strategies/accounts.py
jgreat: Also the main output here is being run from a `kubectl exec` command via CI, so stdout is where we want the info so the CI system can display it.

jgreat:
> Few nits and also question about the dependencies. Otherwise looks good.
> 
> In general we need a python logging module that's configured properly for shipping to cover this. All prints should be turned into log statements. What are your thoughts on that?

The reason want to switch the output to a logger is so I can classify the output.  The script was outputting 100K+ lines of protobuf and structured responses. Made it really difficult to track if the tests were passing/failing.   I didn't want to remove that functionality, just hide that output for normal use.

mobilecoind/strategies/test_client.py
```diff
@@ -11,16 +11,17 @@
 ```
 """
 import argparse
-import concurrent.futures
+import logging
+import sys
```
garbageslam, line 15: Might be better to keep this sorted

garbageslam (approved):
Lgtm -- we should map these updates onto drain- accounts.py which started as a copy of this file and is now used in jenkins CD

sugargoat (approved):
Awesome 🔥 Thank you for tackling this!

PR #1778, opened 2022-04-07
Add ability to count individual GRPC status codes

iamalwaysuncomfortable:
### Motivation

Fog & Consensus outages often happen with not enough early warning nor enough context as to what's happening. Issue #1370 was created to address this. HTTP servers often track HTTP codes (2xx/3xx/4xx) but given FOG uses a lot of GRPC, OPS specified the desire to have GRPC response codes tracked in a similar manner. This PR slightly modifies existing GRPC tracking code to enable tracking response codes.

Upon approval, this tracker will be instrumented to track response codes of FOG services (ingest/server/etc.).

### In this PR
* Cleanup of 
* Cleanup of GRPC metrics documentation

### Future Work
* Track response codes of FOG GRPC methods

### DIFF
--- a/util/metrics/src/service_metrics.rs
+++ b/util/metrics/src/service_metrics.rs
@@ -10,6 +10,7 @@ metrics about a gRPC server.
 For each method, the counters that are captured are:
 - num_req: number of requests
 - num_error: number of errors (can be used to calculate error rate)
+- num_status_code: number of GRPC status code (for more fine grained error rates/diagnostics)
 - duration: duration (in units determined by the exporter) the request took, bucketed
 
 Example use:
@@ -26,7 +27,7 @@ fn sample_service_method(ctx: RpcContext, params: Params) {
 }
 */
 
-use grpcio::RpcContext;
+use grpcio::{RpcContext, RpcStatusCode};
 use mc_common::logger::global_log;
 use prometheus::{
     core::{Collector, Desc},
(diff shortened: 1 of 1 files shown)


util/metrics/src/service_metrics.rs
```diff
@@ -170,9 +197,6 @@ fn path_from_ctx(ctx: &RpcContext) -> Option<String> {
 /// `/{package}.{service_name}/{method}`
 /// and converts it into a dot-delimited string, dropping the 1st `/`
 fn path_from_byte_slice(bytes: &[u8]) -> Option<String> {
-    // The content of method() is expected to be:
-    // `/{package}.{service}/{method}`
-    // 47 ascii = '/'
```
garbageslam, line 175: nit: is there a good reason to remove this comment?

iamalwaysuncomfortable: De-duping. `/{package}.{service}/{method}` is duplicated in the method docstring above. The // 47 ascii specification could be moved up to the docstring or left where it is. Thoughts?

util/metrics/src/service_metrics.rs
```diff
@@ -100,17 +119,25 @@ impl ServiceMetrics {
         )
     }
 
+    /// Count number of errors by method
```
samdealy, line 122: I'm guessing this method is called when a gRPC response is received, and isn't only called to count number of errors by method? If so, maybe add a sentence before this sentence to ensure that the reader understands the full context.

util/metrics/src/service_metrics.rs
```diff
@@ -100,17 +119,25 @@ impl ServiceMetrics {
...
-        // number of responses to number of requests
         if let Some(name) = path_from_ctx(ctx) {
             self.num_error
                 .with_label_values(&[name.as_str()])
                 .inc_by(if success { 0 } else { 1 });
         }
     }
 
+    /// Count number of response codes by method
+    pub fn status_code(&self, ctx: &RpcContext, response_code: RpcStatusCode) {
```
samdealy, line 132: Should this method be renamed to something like `count_number_of_response_codes_per_method`? The name `status_code` is somewhat vague and doesn't describe what's going on (unless I'm misunderstanding and this is like some trait or gRPC construct).

samdealy (approved):

util/metrics/src/service_metrics.rs
iamalwaysuncomfortable: That might be too verbose, when modifying libraries like this I try to keep the semantics close to what the other methods have. I could do "count_status_codes"

util/metrics/src/service_metrics.rs
iamalwaysuncomfortable: I could expand on the comment

util/metrics/src/service_metrics.rs
garbageslam: The 47 comment seems good

I think we didn't actually write this code, we lifted it from some other company. Don't remember exactly

util/metrics/src/service_metrics.rs
```diff
@@ -10,6 +10,7 @@ metrics about a gRPC server.
 For each method, the counters that are captured are:
 - num_req: number of requests
 - num_error: number of errors (can be used to calculate error rate)
+- num_status_code: number of GRPC status code (for more fine grained error rates/diagnostics)
```
wjuan-mob, line 13:
```suggestion
- num_status_code: number of GRPC status codes (for more fine grained error rates/diagnostics)
```

util/metrics/src/service_metrics.rs
```diff
@@ -100,17 +119,25 @@ impl ServiceMetrics {
...
+    /// Count number of response codes by method
+    pub fn status_code(&self, ctx: &RpcContext, response_code: RpcStatusCode) {
+        if let Some(name) = path_from_ctx(ctx) {
+            self.num_status_code
+                .with_label_values(&[name.as_str(), response_code.to_string().as_str()])
+                .inc();
+        }
+    }
+
+    /// Track GRPC message size
```
wjuan-mob, line 140: Similar comment as on the resp() call. I think this should indicate what the method is intended for so you know how to use it or extend it with additional metrics. Is this for counting metrics on messages that are sent or received?

iamalwaysuncomfortable: Yeah I actually agree, however, these methods were named as such prior to this and are used in a lot of different places which could cause larger breakage. If I change these method names, I'd want to do so as part of a larger refactor, which could be submitted as a separate PR and reviewed. At the moment though, I'm trying to prevent that type of scope creep by introducing a single method.

wjuan-mob (approved):

util/metrics/src/service_metrics.rs
```diff
@@ -75,16 +95,15 @@ impl ServiceMetrics {
         }
     }
 
+    /// Register service
     pub fn new_and_registered() -> ServiceMetrics {
         let svc = ServiceMetrics::default();
-        //TODO: should be OK to panic here
         let _res = prometheus::register(Box::new(svc.clone()));
```
remoun, line 101: It seems better to log a warning than just swallow this result.

util/metrics/src/service_metrics.rs
```diff
@@ -37,19 +38,34 @@ use prometheus::{
...
+
+    /// Duration of method call
+    duration: HistogramVec,
+
+    /// Histogram of message sizes
+    message_size: HistogramVec,
 }
 
 impl ServiceMetrics {
+    /// Create a default constructor that initializes all metrics
```
remoun, line 68: Nit: s/constructor/instance/

PR #1783, opened 2022-04-08
Monitor gRPC status codes with prometheus

iamalwaysuncomfortable:
### Motivation

As part of #1370 operations has expressed the desire to be able to profile different gRPC status (and error) codes in a similar way to how HTTP codes are often profiled as 2XX/4XX/5XX, etc. so that errors in FOG & Consensus can be profiled and monitored for early alerts when FOG services become unhealthy. This PR enables Prometheus metrics to be shipped for most gRPC calls in Consensus & FOG.

### In this PR
* Implementation status code tracking in the "send_result" method in the mc-util-grpc lib that most FOG & Consensus grpc calls go through
* gRPC metrics service used by Mobilecoin was ported from a 3rd party with comments not up to MC doc standards. This PR improves them.

### DIFF
--- a/util/grpc/src/lib.rs
+++ b/util/grpc/src/lib.rs
@@ -67,6 +67,7 @@ pub fn send_result<T>(
 ) {
     let logger = logger.clone();
     let success = resp.is_ok();
+    let mut code = RpcStatusCode::from(0);
 
     match resp {
         Ok(ok) => ctx.spawn(
@@ -74,14 +75,18 @@ pub fn send_result<T>(
                 .map_err(move |err| log::error!(logger, "failed to reply: {}", err))
                 .map(|_| ()),
         ),
-        Err(e) => ctx.spawn(
-            sink.fail(e)
-                .map_err(move |err| log::error!(logger, "failed to reply: {}", err))
-                .map(|_| ()),
-        ),
+        Err(e) => {
+            code = e.code();
+            ctx.spawn(
+                sink.fail(e)
+                    .map_err(move |err| log::error!(logger, "failed to reply: {}", err))
+                    .map(|_| ()),
+            )
+        }
     }
 
     SVC_COUNTERS.resp(&ctx, success);
+    SVC_COUNTERS.status_code(&ctx, code);
 }
 
 macro_rules! report_err_with_code(
--- a/util/grpc/src/lib.rs
+++ b/util/grpc/src/lib.rs
@@ -67,7 +67,7 @@ pub fn send_result<T>(
 ) {
     let logger = logger.clone();
     let success = resp.is_ok();
-    let mut code = RpcStatusCode::from(0);
+    let mut code = RpcStatusCode::OK;
 
     match resp {
         Ok(ok) => ctx.spawn(
--- a/util/grpc/src/lib.rs
+++ b/util/grpc/src/lib.rs
@@ -67,7 +67,10 @@ pub fn send_result<T>(
 ) {
     let logger = logger.clone();
     let success = resp.is_ok();
-    let mut code = RpcStatusCode::OK;
+    let code = match &resp {
+        Ok(_) => RpcStatusCode::OK,
+        Err(e) => e.code(),
+    };
 
     match resp {
         Ok(ok) => ctx.spawn(
@@ -75,14 +78,11 @@ pub fn send_result<T>(
                 .map_err(move |err| log::error!(logger, "failed to reply: {}", err))
                 .map(|_| ()),
         ),
-        Err(e) => {
-            code = e.code();
-            ctx.spawn(
-                sink.fail(e)
-                    .map_err(move |err| log::error!(logger, "failed to reply: {}", err))
-                    .map(|_| ()),
-            )
-        }
+        Err(e) => ctx.spawn(
+            sink.fail(e)
+                .map_err(move |err| log::error!(logger, "failed to reply: {}", err))
+                .map(|_| ()),
+        ),
     }
 
     SVC_COUNTERS.resp(&ctx, success);
--- a/util/metrics/src/service_metrics.rs
+++ b/util/metrics/src/service_metrics.rs
@@ -49,10 +49,10 @@ use std::str;
 /// e.g., calc_service.duration_sum{method="add"} = 6
 #[derive(Clone)]
 pub struct ServiceMetrics {
-    /// Number of requests made by each gRPC method tracked
+    /// Count of requests made by each gRPC method tracked
     num_req: IntCounterVec,
 
-    /// Number of error responses for each gRPC method tracked
+    /// Count of error responses for each gRPC method tracked
     num_error: IntCounterVec,
 
     /// Count of gRPC status codes for each gRPC method tracked
@@ -193,8 +193,8 @@ impl Collector for ServiceMetrics {
     }
 }
 
-/// This method reads the full URI from gRpcContext (looks like:
-/// `/{package}.{service_name}/{method}`
+/// This method reads the full URI from gRpcContext
+/// which looks like `/{package}.{service_name}/{method}`
 /// ('/' equates to ascii code 47)
 /// and converts it into a dot-delimited string, dropping the 1st `/`
 fn path_from_ctx(ctx: &RpcContext) -> Option<String> {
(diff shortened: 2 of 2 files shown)


util/grpc/src/lib.rs
```diff
@@ -67,21 +67,26 @@ pub fn send_result<T>(
 ) {
     let logger = logger.clone();
     let success = resp.is_ok();
+    let mut code = RpcStatusCode::from(0);
```
nick-mobilecoin, line 70:
```suggestion
    let mut code = RpcStatusCode::OK;
```

util/metrics/src/service_metrics.rs
```diff
@@ -48,19 +49,19 @@ use std::str;
...
-    /// Number of requests made by methods
+    /// Number of requests made by each gRPC method tracked
     num_req: IntCounterVec,
 
-    /// Number of error responses for methods
+    /// Number of error responses for each gRPC method tracked
     num_error: IntCounterVec,
 
-    /// Number of GRPC status codes for methods
+    /// Count of gRPC status codes for each gRPC method tracked
```
nick-mobilecoin, line 58: nit: why did this get changed to `Count`? Seems inconsistent with the variable name of `num` and the other member docs using `Number`.

as an aside I generally lean toward "count" over "number". A "number" can mean; how many (i.e. a counter), or the number 3 (an id).  While "count" seems to be less ambiguous, unless you're dealing with count chocula.

util/metrics/src/service_metrics.rs
```diff
@@ -95,14 +96,16 @@ impl ServiceMetrics {
...
     pub fn new_and_registered() -> ServiceMetrics {
         let svc = ServiceMetrics::default();
         let _res = prometheus::register(Box::new(svc.clone()));
         svc
     }
 
-    /// Track number of requests and durations
+    /// Takes the RpcContext used during a gRPC method call to get the method
+    /// name and increments counters tracking the number of calls to and
+    /// durations of that method
```
nick-mobilecoin, line 108: nit: This feels like it's describing the inner workings of the function and not necessarily what a user wants/needs to know.

Perhaps:
```suggestion
    /// Increments the number of requests for the active method in `ctx`,
    /// returning a [Option<HistogramTimer>] to track the duration.
    /// The returned [Option<HistogramTimer>] is meant to be used as an RAII
    /// guard in order to accurately track the method duration.
```
Not sure on the format to correctly document the `HistogramTimer` perhaps the `Option` should be omitted from it in the docs.

I'm hoping someone knows a better term. RAII always felt like an unintuitive term, but I ran across this, https://rust-unofficial.github.io/patterns/patterns/behavioural/RAII.html

util/metrics/src/service_metrics.rs
```diff
@@ -137,7 +144,7 @@ impl ServiceMetrics {
         }
     }
 
-    /// Track GRPC message size
+    /// Tracks gRPC message name and corresponding message size into a histogram
```
nick-mobilecoin, line 147: It's quite possible I'm missing it in the inner workings but the "into a histogram" part doesn't seem like it is happening here.  This seems like an _input_ and eventually when one wants the output that might be a _histogram_.

util/metrics/src/service_metrics.rs
```diff
@@ -187,6 +194,7 @@ impl Collector for ServiceMetrics {
 
 /// This method reads the full URI from gRpcContext (looks like:
 /// `/{package}.{service_name}/{method}`
+/// ('/' equates to ascii code 47)
 /// and converts it into a dot-delimited string, dropping the 1st `/`
```
nick-mobilecoin, line 198: nit: it feels like the notation about the ascii code interrupts the thought.

```suggestion
/// and converts it into a dot-delimited string, dropping the 1st `/`
/// ('/' equates to ascii code 47)
```

util/metrics/src/service_metrics.rs
```diff
@@ -195,6 +203,7 @@ fn path_from_ctx(ctx: &RpcContext) -> Option<String> {
...
```
nick-mobilecoin, line 207: nit: it feels like the notation about the ascii code interrupts the thought.
```suggestion
/// and converts it into a dot-delimited string, dropping the 1st `/`
/// ('/' equates to ascii code 47)
```

nick-mobilecoin (approved):

util/metrics/src/service_metrics.rs
iamalwaysuncomfortable: Changed to count it creates a Prometheus Counter

util/metrics/src/service_metrics.rs
iamalwaysuncomfortable: I think just "returning a histogram timer to track the duration" I don't want get overly verbose.

util/metrics/src/service_metrics.rs
iamalwaysuncomfortable: This call message call tracks the message size and ships it to a Prometheus histogram

util/grpc/src/lib.rs
iamalwaysuncomfortable: Good suggestion!

util/grpc/src/lib.rs
```diff
@@ -67,21 +67,26 @@ pub fn send_result<T>(
...
-        ),
+        Err(e) => {
+            code = e.code();
+            ctx.spawn(
+                sink.fail(e)
+                    .map_err(move |err| log::error!(logger, "failed to reply: {}", err))
+                    .map(|_| ()),
+            )
+        }
     }
```
jcape, line 86: This should be `let code = match { .. }`

util/metrics/src/service_metrics.rs
jcape: The other comments that are changed here should be changed to "Count" as well, then ;-)

util/metrics/src/service_metrics.rs
jcape: Was this suggestion meant to be committed, or ignored?

util/metrics/src/service_metrics.rs
jcape: Same ?

jcape (changes requested):
One nitpick, some questions about resolved conversations.

util/metrics/src/service_metrics.rs
iamalwaysuncomfortable: Can do

util/grpc/src/lib.rs
iamalwaysuncomfortable: I don't think that's the case. The issue referenced specified that a list of gRPC status codes for FOG services should be tracked. All the above code does is record it so that it can be tracked by a single gRPC function below. I'm not sure what the match would achieve.

The surrounding function `send_result` is called by most FOG gRPC methods to do the actual work of spawning a task into the gRPC async context. Changing the match in `send_result` to match on error codes vs. the `resp` Result would significantly changing the way FOG services currently interact with gRPC.

remoun: We can avoid `mut` with a second match
```suggestion
    let code = match &resp {
        Ok(_) => RpcStatusCode::OK,
        Err(e) => e.code(),
    };
    match resp {
        Ok(ok) => ctx.spawn(
            sink.success(ok)
                .map_err(move |err| log::error!(logger, "failed to reply: {}", err))
                .map(|_| ()),
        ),
        Err(e) => ctx.spawn(
            sink.fail(e)
                .map_err(move |err| log::error!(logger, "failed to reply: {}", err))
                .map(|_| ()),
        ),
    }
```

iamalwaysuncomfortable: Ah yes, I see the thought

util/metrics/src/service_metrics.rs
```diff
@@ -10,7 +10,8 @@ metrics about a gRPC server.
 For each method, the counters that are captured are:
 - num_req: number of requests
 - num_error: number of errors (can be used to calculate error rate)
-- num_status_code: number of GRPC status code (for more fine grained error rates/diagnostics)
+- num_status_code: number of gRPC status codes (to establish statistics on
+gRPC status codes, similar to how HTTP 2XX/4XX/5XX codes are profiled)
```
remoun, line 14: Nit: indent
```suggestion
  gRPC status codes, similar to how HTTP 2XX/4XX/5XX codes are profiled)
```

remoun (approved):

util/metrics/src/service_metrics.rs
iamalwaysuncomfortable: I feel like indents in a list like that leaves odd whitespace

PR #1792, opened 2022-04-11
Android SDK RTH Release 1.2.0

dolanbernard:
### Motivation

This PR contains Android Bindings changes for the 1.2 release

### In this PR
* Recoverable Transaction History
* Parameterize transaction builder bindings for block version
* Fix one-time private key generation to use correct subaddress


garbageslam:
It looks to me like this is merging master into release 1.2 -- not sure, there are a lot of dependant commits in here though

I think maybe we should try rebasing your commits onto 1.2.0. Or else, try merging this version into master and makinf a new PR for 1.2 that is rebased or cherry picked

PR #1793, opened 2022-04-11
Update iOS Bindings Header File Definition - `release-1.2.0`

the-real-adammork:
Soundtrack of this PR: [Orlann Divo - Onde Anda O Meu Amor](https://youtu.be/fe-PXWEbCGg)

### Motivation

Function parameter types do not match definition.

### In this PR
* update to be correct.


the-real-adammork:
Im merging this now because its a small change and a blocker for CICD on the Swift SDK PR

Issue #1582, opened 2022-03-07
Look into intermittent failures in GetLastBlockInfo gRPC


jcape:
This is a follow-on to issue #1575. The `local-network.py` script was setting the fee to 10mMOB, but the `fog-test-client` binary was using the current default fee of 400uMOB when it could not contact the consensus service, which happens both in CI environments *and* on local desktop use. This would tend to indicate the consensus service has relatively poor reliability with the `GetLastBlockInfo` gRPC call, at least at load induced by the fog test client (and potentially others).

To resolve this issue, the reason for gRPC call failures when both client and server are running on localhost should be found and fixed.

jcape:
This could be caused by the server and clients being created with identical keepalive settings, meaning client and server settings leave no room for error.

PR #1800, opened 2022-04-12
added extra configuration consensus service serve side.

integral-llc:
### Motivation

There were intermittent issues when clients would lose connection to the server. 

### In this PR
* added a default config for server

#1582

### DIFF
--- a/consensus/service/src/consensus_service.rs
+++ b/consensus/service/src/consensus_service.rs
@@ -14,7 +14,7 @@ use crate::{
 use base64::{encode_config, URL_SAFE};
 use displaydoc::Display;
 use futures::executor::block_on;
-use grpcio::{ChannelBuilder, EnvBuilder, Environment, Server, ServerBuilder};
+use grpcio::{EnvBuilder, Environment, Server, ServerBuilder};
 use mc_attest_api::attest_grpc::create_attested_api;
 use mc_attest_enclave_api::{ClientSession, PeerSession};
 use mc_attest_net::RaClient;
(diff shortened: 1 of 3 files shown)


iamalwaysuncomfortable (commented):
Small questions about keepalive time/testing client disconnection.

util/grpc/src/grpcio_extensions.rs
```diff
@@ -61,6 +61,15 @@ pub trait ConnectionUriGrpcioServer {
     /// hot-reloading certificates when TLS is used.
     #[must_use]
     fn bind_using_uri(self, uri: &impl ConnectionUri, logger: Logger) -> Self;
+
```
remoun, line 64: Optional: IIUC we can add a helper method that calls `channel_args`, e.g.

```suggestion

/// Set the channel args to our defaults.
fn set_default_channel_args(self, env: Arc<Environment>) -> Self {
  self.channel_args(Self::default_channel_builder(env).build_args())
}
```

integral-llc: good idea, added.

consensus/service/src/consensus_service.rs
```diff
@@ -382,12 +382,16 @@ impl<
...
+        let server_builder = ServerBuilder::new(env.clone())
             .register_service(client_service)
             .register_service(blockchain_service)
             .register_service(health_service)
             .register_service(attested_service)
             .register_service(build_info_service)
+            .channel_args(
+                <ServerBuilder as ConnectionUriGrpcioServer>::default_channel_builder(env)
+                    .build_args(),
+            )
```
remoun, line 394: This can now be `set_default_channel_args`
```suggestion
            .set_default_channel_args(env)
```

jcape (approved):

jcape:
(approved pending inclusion of changes)

iamalwaysuncomfortable (approved):

remoun (approved):

PR #1816, opened 2022-04-18
change token id from u32 to u64

garbageslam:
This is done because it is conceivable that we will have more than
4 billion tokens on our chain, particularly if this mechanism is
used to create NFTs or something someday.

There is very little reason from an engineering point of view to
try to save space by using a u32 instead of a u64 -- u32's do
sometimes run out in practice, and a u64 will not here, and 4
bytes usually doesn't matter.

The size of an `ETxOutRecord` changes from `233` bytes to `237` bytes
in this commit.
Per MCIP #3, the calculation is `2048 / (16 + 1 + 237) = 8.062`, so we still
have a little room before we hit the threshold for 8, so this will not negatively
impact fog.


eranrund (approved):

transaction/core/src/mint/tx.rs
```diff
@@ -17,8 +17,8 @@ use serde::{Deserialize, Serialize};
 )]
 pub struct MintTxPrefix {
     /// Token ID we are minting.
-    #[prost(uint32, tag = "1")]
-    pub token_id: u32,
+    #[prost(uint64, tag = "1")]
+    pub token_id: u64,
```
remoun, line 21: How about we make this a `TokenId`? (with the Prost tag specifying uint64)

```suggestion
    pub token_id: TokenId,
```

consensus/mint-client/src/config.rs
```diff
@@ -137,7 +137,7 @@ pub struct MintTxPrefixParams {
 
     /// The token id we are minting.
     #[clap(long, env = "MC_MINTING_TOKEN_ID")]
-    pub token_id: u32,
+    pub token_id: u64,
```
remoun, line 140:
```suggestion
    pub token_id: TokenId,
```

transaction/core/src/amount/mod.rs
```diff
@@ -91,7 +91,7 @@ impl MaskedAmount {
 
         // The token_id is XORed with the first 4 bytes of the mask.
```
nick-mobilecoin, line 92:
```suggestion
        // The token_id is XORed with the first 8 bytes of the mask.
```

transaction/core/src/amount/mod.rs
```diff
@@ -232,11 +232,11 @@ fn get_value_mask(shared_secret: &RistrettoPublic) -> u64 {
 ///
 /// # Arguments
 /// * `shared_secret` - The shared secret, e.g. `rB`.
-fn get_token_id_mask(shared_secret: &RistrettoPublic) -> u32 {
+fn get_token_id_mask(shared_secret: &RistrettoPublic) -> u64 {
```
nick-mobilecoin, line 235: The comment should be updated as this is now 8 bytes and a u64
```
/// Computes `Blake2B(token_id_mask | shared_secret)`,
/// then interprets the first 8 canonical bytes as a u64 number in
/// little-endian representation.
```

transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -70,7 +70,7 @@ impl SignatureRctBulletproofs {
         input_secrets: &[(RistrettoPrivate, u64, Scalar)],
         output_values_and_blindings: &[(u64, Scalar)],
         fee: u64,
-        token_id: u32,
+        token_id: u64,
```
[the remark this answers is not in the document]
garbageslam: I don't this is correct, have you tested that this compiles?

transaction/core/src/mint/tx.rs
garbageslam: Have you tested that this compiles?

transaction/core/src/ring_signature/rct_bulletproofs.rs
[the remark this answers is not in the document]
garbageslam: the custom wrappers like `TokenId` are not compatible with `prop_test` so I would rather not do this here

nick-mobilecoin (approved):

Issue #1682, opened 2022-03-21
Include QuorumSet and AVR in ArchiveBlock


remoun:
To make [`ArchiveBlock`](https://github.com/mobilecoinfoundation/mobilecoin/blob/7077b418fab65e05a05e4fca1d272e3dddd2e392/api/proto/blockchain.proto#L80) fully interchangeable with [`BlockWithQuorumSet`](https://github.com/mobilecoinfoundation/mobilecoin/blob/5a4581804a56b227b084bb202f0d2b215fc62888/ledger/streaming/api/proto/streaming_blocks.proto#L18).

This will require a block version bump (to enforce that the AVR and QS are required for new transactions), and potentially a new version for `ArchiveBlock` (TBD, since this change is additive w.r.t. protobufs).

PR #1823, opened 2022-04-19
Refactor BlockData+ArchiveBlock to include the signed block metadata 

remoun:
### Motivation
Add the Quorum Set and AVR to `ArchiveBlock` (#1682)

### In this PR
* Restructure where we define and include the Quorum Set and AVR, moving them into `mc_transaction_core`.
* Replace `BlockStreamComponents` holder struct with `BlockData`, now that the latter has all the necessary fields.
* Replace `SubscribeResponse` with `ArchiveBlock`, now that the latter has all the necessary fields.

### Future Work
* Update all callers of `BlockData::new` to include block metadata.
* Make `quorum_set` and `verification_report` mandatory, with a `BlockVersion` bump.
* Potentially re-add `derive(Message)` for `BlockData`. This would require adding `prost` annotations to `QuorumSet` and `NodeID`.


fog/ingest/enclave/trusted/Cargo.lock
```diff
@@ -1279,6 +1279,7 @@ dependencies = [
  "hkdf",
  "lazy_static",
  "mc-account-keys",
+ "mc-attest-core",
```
jcape, line 1282: Oof, this is pretty undesirable.

remoun: OK, I'll duplicate `VerificationReport` and remove this dep

transaction/core/src/blockchain/block_metadata.rs
```diff
@@ -0,0 +1,122 @@
...
+
+    /// Get the signing key.
+    pub fn node_key(&self) -> &Ed25519Public {
+        &self.node_key
+    }
+
+    /// Get the signature.
+    pub fn signature(&self) -> &Ed25519Signature {
+        &self.signature
+    }
```
iamalwaysuncomfortable, line 121: We need a cryptography review of this

iamalwaysuncomfortable (commented):
Few questions, nits, and statement that we need a cryptography review of the signing.

api/src/convert/archive_block.rs
```diff
@@ -43,8 +48,19 @@ impl TryFrom<&blockchain::ArchiveBlock> for BlockData {
...
+            .transpose()?;
+
         if block.contents_hash == block_contents.hash() {
-            Ok(BlockData::new(block, block_contents, signature))
+            Ok(BlockData::new_with_metadata(
+                block,
+                block_contents,
+                signature,
+                metadata,
+            ))
```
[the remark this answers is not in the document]
remoun: We generally convert to `ArchiveBlock` when storing blocks, e.g. on chain, and vice versa when parsing the blockchain. Another case to note is that the `LedgerUpdates` gRPC API uses `ArchiveBlock`, so we'll convert from `BlockData` to `ArchiveBlock` in `GrpcServerSink`, and vice versa in `GrpcBlockSource`

attest/verifier/types/src/verification.rs
```diff
@@ -0,0 +1,124 @@
...
+    /// Attestation Report Signing Certificate Chain, as an array of
+    /// DER-formatted bytes, from the X-IASReport-Signing-Certificate HTTP
+    /// header.
+    #[prost(bytes, repeated)]
+    pub chain: Vec<Vec<u8>>,
+    /// The raw report body JSON, as a byte sequence
+    #[prost(string, required)]
+    #[digestible(never_omit)]
+    pub http_body: String,
+}
```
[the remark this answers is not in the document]
remoun: Added to prefactoring PR.

iamalwaysuncomfortable (approved):

isis-mc (approved):
Great work! I'm happy to see the AVRs will be stored onchain. For the block signing parts specifically, I do not think we need to worry about the malleability issues of ed25519 signatures, as a duplicate block with a "different" signature would get rejected anyway, and additionally the signer can just produce a new different signature by varying the nonce.

Issue #1564, opened 2022-03-02
Use time-memory tradeoff speedups from Dalek crate


garbageslam:
(TODO flesh this out)

Issue #1522, opened 2022-02-19
Should mobilecoind support recoverable transaction history memos in 1.2 release?


garbageslam:
And if so, how exactly?

Specifically, the question is if mobilecoind should continue using `EmptyMemoBuilder`, or if it should actually use the `RTHMemoBuilder` and write memos.

`mobilecoind` is primarily used by exchanges in my understanding, and the ideal user experience for mobile users is likely:
* The public address where I can deposit funds (this is a subaddress of the exchange) is known to my mobile wallet, so that I can deposit funds to the exchange easily. (And this content will then also appear in my change memos)
* The exchange uses this same public address when creating RTH memos when I do a withdrawal. Then my app automatically associates payments coming from the exchange to payments going to the exchange, into one interaction, based on the RTH memos.
* This was described in MCIP number 4 (https://github.com/mobilecoinfoundation/mcips/pull/4)

How we do this for mobilecoind may also inform how full-service should work. We might also decide it's more important to implement RTH in full-service than it is for mobilecoind.

PR #1827, opened 2022-04-19
allow transactions with "mixed" token types

garbageslam:
This allows transactions to involve inputs and outputs of different types, as long as they are balanced with respect to all of the types.

This is described in MCIP 31

TODO:
- [x] full transaction builder support
- [x] more testing

Future work:
* Make RTH memos that allow for mixed transactions (or at least, fee paid in another token?)
* Avoid regenerating pseudo commitments?
* Make a ticket to change asserts to debug asserts before branching for release?

### DIFF
--- a/android-bindings/src/bindings.rs
+++ b/android-bindings/src/bindings.rs
@@ -48,7 +48,7 @@ use mc_transaction_core::{
     ring_signature::KeyImage,
     tokens::Mob,
     tx::{Tx, TxOut, TxOutConfirmationNumber, TxOutMembershipProof},
-    BlockVersion, CompressedCommitment, MaskedAmount, Token,
+    Amount, BlockVersion, CompressedCommitment, MaskedAmount, Token,
 };
 
 use mc_transaction_std::{
@@ -1692,12 +1692,19 @@ pub unsafe extern "C" fn Java_com_mobilecoin_lib_TransactionBuilder_add_1output(
 
             let value = jni_big_int_to_u64(env, value)?;
 
+            // TODO: If you want to support mixed transactions, use something other
+            // than fee_token_id here.
+            let amount = Amount {
+                value: value as u64,
+                token_id: tx_builder.get_fee_token_id(),
+            };
+
             let recipient: MutexGuard<PublicAddress> =
                 env.get_rust_field(recipient, RUST_OBJ_FIELD)?;
 
             let mut rng = McRng::default();
             let (tx_out, confirmation_number) =
-                tx_builder.add_output(value as u64, &recipient, &mut rng)?;
+                tx_builder.add_output(amount, &recipient, &mut rng)?;
             if !confirmation_number_out.is_null() {
                 let len = env.get_array_length(confirmation_number_out)?;
                 if len as usize >= confirmation_number.to_vec().len() {
--- a/android-bindings/src/bindings.rs
+++ b/android-bindings/src/bindings.rs
@@ -1693,7 +1693,7 @@ pub unsafe extern "C" fn Java_com_mobilecoin_lib_TransactionBuilder_add_1output(
 
             let value = jni_big_int_to_u64(env, value)?;
 
-            // TODO: If you want to support mixed transactions, use something other
+            // TODO: If you want to do mixed transactions, use something other
             // than fee_token_id here.
             let amount = Amount {
                 value: value as u64,
@@ -1751,8 +1751,15 @@ pub unsafe extern "C" fn Java_com_mobilecoin_lib_TransactionBuilder_add_1change_
                 ChangeDestination::from(&*source_account_key);
             let mut rng = McRng::default();
 
+            // TODO: If you want to do mixed transactions, use something other
+            // than fee_token_id here.
+            let amount = Amount {
+                value: value as u64,
+                token_id: tx_builder.get_fee_token_id(),
+            };
+
             let (tx_out, confirmation_number) =
-                tx_builder.add_change_output(value, &change_destination, &mut rng)?;
+                tx_builder.add_change_output(amount, &change_destination, &mut rng)?;
             if !confirmation_number_out.is_null() {
                 let len = env.get_array_length(confirmation_number_out)?;
                 if len as usize >= confirmation_number.to_vec().len() {
--- a/android-bindings/src/bindings.rs
+++ b/android-bindings/src/bindings.rs
@@ -1693,7 +1693,7 @@ pub unsafe extern "C" fn Java_com_mobilecoin_lib_TransactionBuilder_add_1output(
 
             let value = jni_big_int_to_u64(env, value)?;
 
-            // TODO: If you want to do mixed transactions, use something other
+            // TODO (GH #1867): If you want to do mixed transactions, use something other
             // than fee_token_id here.
             let amount = Amount {
                 value: value as u64,
@@ -1751,7 +1751,7 @@ pub unsafe extern "C" fn Java_com_mobilecoin_lib_TransactionBuilder_add_1change_
                 ChangeDestination::from(&*source_account_key);
             let mut rng = McRng::default();
 
-            // TODO: If you want to do mixed transactions, use something other
+            // TODO (GH #1867): If you want to do mixed transactions, use something other
             // than fee_token_id here.
             let amount = Amount {
                 value: value as u64,
--- a/android-bindings/src/bindings.rs
+++ b/android-bindings/src/bindings.rs
@@ -1603,12 +1603,13 @@ pub unsafe extern "C" fn Java_com_mobilecoin_lib_TransactionBuilder_init_1jni(
         // FIXME #1595: The token id should be a parameter and not hard coded to Mob
         // here
         let token_id = Mob::ID;
+        let fee_amount = Amount::new(Mob::MINIMUM_FEE, token_id);
         let tx_builder = TransactionBuilder::new_with_box(
             block_version,
-            token_id,
+            fee_amount,
             fog_resolver.clone(),
             memo_builder_box,
-        );
+        )?;
 
         Ok(env.set_rust_field(obj, RUST_OBJ_FIELD, tx_builder)?)
     })
--- a/api/proto/external.proto
+++ b/api/proto/external.proto
@@ -242,8 +242,8 @@ message TxPrefix {
     // The block index at which this transaction is no longer valid.
     uint64 tombstone_block = 4;
 
-    // Token id for this transaction
-    fixed64 token_id = 5;
+    // Token id for the fee of this transaction
+    fixed64 fee_token_id = 5;
 }
 
 message RingMLSAG {
@@ -255,7 +255,10 @@ message RingMLSAG {
 message SignatureRctBulletproofs {
     repeated RingMLSAG ring_signatures = 1;
     repeated CompressedRistretto pseudo_output_commitments = 2;
-    bytes range_proofs = 3;
+    bytes range_proof_bytes = 3;
+    repeated bytes range_proofs = 4;
+    repeated fixed64 pseudo_output_token_ids = 5;
+    repeated fixed64 output_token_ids = 6;
 }
 
 message Tx {
--- a/api/proto/external.proto
+++ b/api/proto/external.proto
@@ -246,18 +246,44 @@ message TxPrefix {
     fixed64 fee_token_id = 5;
 }
 
+// A ring mlsag is a group-ring signature conferring spending authority of one TxOut
+// which is part of a TxIn.
 message RingMLSAG {
+    // The initial challenge value for the ring signature
     CurveScalar c_zero = 1;
+    // The "responses", one for each input which is signed
     repeated CurveScalar responses = 2;
+    // The key image is a hash unique to the "true" spent input. This cannot
+    // be linked back to determine the true spent input, but the input cannot be
+    // spent again without producing the same key image value, so this is used to
+    // prevent double-spends.
     KeyImage key_image = 3;
 }
 
 message SignatureRctBulletproofs {
+    // A ring-signature, one for each TxIn, producing one pseudo-output and key image.
     repeated RingMLSAG ring_signatures = 1;
+    // The amount commitments for each pseudo-output.
+    // There must be one of these for each TxIn.
     repeated CompressedRistretto pseudo_output_commitments = 2;
+    // Before mixed transactions feature, there is one range proof for all pseudo-output
+    // and output commitments, whose serialized bytes appear here.
+    // After mixed transactions feature, this field is empty.
     bytes range_proof_bytes = 3;
+    // Before mixed transactions feature, this field is empty.
+    // After mixed transactions feature, this field contains one range proof for each
+    // token id which appears in the transaction, in sorted order of token ids.
+    // It range-proofs the pseudo-outputs and outputs with that token id, in the order
+    // that they appear in the transaction.
     repeated bytes range_proofs = 4;
+    // The token ids of each pseudo ouptut. There must be one of these for each TxIn.
+    // Before mixed transactions feature, this field is empty, and the token ids of
+    // all pseudo-outputs are inferred from the tx.prefix.fee_token_id.
     repeated fixed64 pseudo_output_token_ids = 5;
+    // The token ids of each output. There must be one of these for each output of the Tx.
+    // (tx.prefix.outputs).
+    // Before mixed transactions feature, this field is empty, and the token ids of
+    // all outputs are inferred from the tx.prefix.fee_token_id.
     repeated fixed64 output_token_ids = 6;
 }
 
--- a/consensus/enclave/impl/src/lib.rs
+++ b/consensus/enclave/impl/src/lib.rs
@@ -236,7 +236,7 @@ impl SgxConsensusEnclave {
         // We need to make sure all transactions are valid. We also ensure they all
         // point at the same root membership element.
         for (tx, proofs) in transactions_with_proofs.iter() {
-            let token_id = TokenId::from(tx.prefix.token_id);
+            let token_id = TokenId::from(tx.prefix.fee_token_id);
 
             let minimum_fee = ct_min_fees
                 .get(&token_id)
@@ -685,7 +685,7 @@ impl ConsensusEnclave for SgxConsensusEnclave {
             .decrypt_bytes(locally_encrypted_tx.0)?;
         let tx: Tx = mc_util_serial::decode(&decrypted_bytes)?;
 
-        let token_id = TokenId::from(tx.prefix.token_id);
+        let token_id = TokenId::from(tx.prefix.fee_token_id);
 
         // Validate.
         let mut csprng = McRng::default();
@@ -782,7 +782,7 @@ impl ConsensusEnclave for SgxConsensusEnclave {
         // Compute the total fees for each known token id, for tx's in this block.
         let mut total_fees: CtTokenMap<u128> = ct_min_fee_map.keys().cloned().collect();
         for tx in transactions.iter() {
-            let token_id = TokenId::from(tx.prefix.token_id);
+            let token_id = TokenId::from(tx.prefix.fee_token_id);
             total_fees.add(&token_id, tx.prefix.fee as u128);
         }
 
--- a/consensus/enclave/impl/src/lib.rs
+++ b/consensus/enclave/impl/src/lib.rs
@@ -236,10 +236,10 @@ impl SgxConsensusEnclave {
         // We need to make sure all transactions are valid. We also ensure they all
         // point at the same root membership element.
         for (tx, proofs) in transactions_with_proofs.iter() {
-            let token_id = TokenId::from(tx.prefix.fee_token_id);
+            let fee_token_id = TokenId::from(tx.prefix.fee_token_id);
 
             let minimum_fee = ct_min_fees
-                .get(&token_id)
+                .get(&fee_token_id)
                 .ok_or(TransactionValidationError::TokenNotYetConfigured)?;
 
             mc_transaction_core::validation::validate(
@@ -685,12 +685,12 @@ impl ConsensusEnclave for SgxConsensusEnclave {
             .decrypt_bytes(locally_encrypted_tx.0)?;
         let tx: Tx = mc_util_serial::decode(&decrypted_bytes)?;
 
-        let token_id = TokenId::from(tx.prefix.fee_token_id);
+        let fee_token_id = TokenId::from(tx.prefix.fee_token_id);
 
         // Validate.
         let mut csprng = McRng::default();
         let minimum_fee = ct_min_fee_map
-            .get(&token_id)
+            .get(&fee_token_id)
             .ok_or(TransactionValidationError::TokenNotYetConfigured)?;
         mc_transaction_core::validation::validate(
             &tx,
@@ -782,8 +782,8 @@ impl ConsensusEnclave for SgxConsensusEnclave {
         // Compute the total fees for each known token id, for tx's in this block.
         let mut total_fees: CtTokenMap<u128> = ct_min_fee_map.keys().cloned().collect();
         for tx in transactions.iter() {
-            let token_id = TokenId::from(tx.prefix.fee_token_id);
-            total_fees.add(&token_id, tx.prefix.fee as u128);
+            let fee_token_id = TokenId::from(tx.prefix.fee_token_id);
+            total_fees.add(&fee_token_id, tx.prefix.fee as u128);
         }
 
         // Sort the token ids which did not appear to the end.
--- a/consensus/service/src/validators.rs
+++ b/consensus/service/src/validators.rs
@@ -575,10 +575,11 @@ mod combine_tests {
 
             let mut transaction_builder = TransactionBuilder::new(
                 block_version,
-                Mob::ID,
+                Amount::new(Mob::MINIMUM_FEE, Mob::ID),
                 MockFogResolver::default(),
                 EmptyMemoBuilder::default(),
-            );
+            )
+            .unwrap();
             transaction_builder.add_input(input_credentials);
             transaction_builder.set_fee(0).unwrap();
             transaction_builder
@@ -634,10 +635,11 @@ mod combine_tests {
 
                     let mut transaction_builder = TransactionBuilder::new(
                         block_version,
-                        Mob::ID,
+                        Amount::new(Mob::MINIMUM_FEE, Mob::ID),
                         MockFogResolver::default(),
                         EmptyMemoBuilder::default(),
-                    );
+                    )
+                    .unwrap();
 
                     // Create InputCredentials to spend the TxOut.
                     let onetime_private_key = recover_onetime_private_key(
@@ -739,10 +741,11 @@ mod combine_tests {
 
                 let mut transaction_builder = TransactionBuilder::new(
                     block_version,
-                    Mob::ID,
+                    Amount::new(Mob::MINIMUM_FEE, Mob::ID),
                     MockFogResolver::default(),
                     EmptyMemoBuilder::default(),
-                );
+                )
+                .unwrap();
                 transaction_builder.add_input(input_credentials);
                 transaction_builder.set_fee(0).unwrap();
                 transaction_builder
@@ -780,10 +783,11 @@ mod combine_tests {
 
                 let mut transaction_builder = TransactionBuilder::new(
                     block_version,
-                    Mob::ID,
+                    Amount::new(Mob::MINIMUM_FEE, Mob::ID),
                     MockFogResolver::default(),
                     EmptyMemoBuilder::default(),
-                );
+                )
+                .unwrap();
                 transaction_builder.add_input(input_credentials);
                 transaction_builder.set_fee(0).unwrap();
                 transaction_builder
@@ -844,10 +848,11 @@ mod combine_tests {
 
                 let mut transaction_builder = TransactionBuilder::new(
                     block_version,
-                    Mob::ID,
+                    Amount::new(Mob::MINIMUM_FEE, Mob::ID),
                     MockFogResolver::default(),
                     EmptyMemoBuilder::default(),
-                );
+                )
+                .unwrap();
                 transaction_builder.add_input(input_credentials);
                 transaction_builder.set_fee(0).unwrap();
                 transaction_builder
@@ -935,10 +940,11 @@ mod combine_tests {
 
                 let mut transaction_builder = TransactionBuilder::new(
                     block_version,
-                    Mob::ID,
+                    Amount::new(Mob::MINIMUM_FEE, Mob::ID),
                     MockFogResolver::default(),
                     EmptyMemoBuilder::default(),
-                );
+                )
+                .unwrap();
                 transaction_builder.add_input(input_credentials);
                 transaction_builder.set_fee(0).unwrap();
                 transaction_builder
@@ -977,10 +983,11 @@ mod combine_tests {
 
                 let mut transaction_builder = TransactionBuilder::new(
                     block_version,
-                    Mob::ID,
+                    Amount::new(Mob::MINIMUM_FEE, Mob::ID),
                     MockFogResolver::default(),
                     EmptyMemoBuilder::default(),
-                );
+                )
+                .unwrap();
                 transaction_builder.add_input(input_credentials);
                 transaction_builder.set_fee(0).unwrap();
                 transaction_builder
@@ -1042,10 +1049,11 @@ mod combine_tests {
 
                 let mut transaction_builder = TransactionBuilder::new(
                     block_version,
-                    Mob::ID,
+                    Amount::new(Mob::MINIMUM_FEE, Mob::ID),
                     MockFogResolver::default(),
                     EmptyMemoBuilder::default(),
-                );
+                )
+                .unwrap();
                 transaction_builder.add_input(input_credentials);
                 transaction_builder.set_fee(0).unwrap();
                 transaction_builder
--- a/consensus/service/src/validators.rs
+++ b/consensus/service/src/validators.rs
@@ -585,7 +585,14 @@ mod combine_tests {
             transaction_builder.add_input(input_credentials);
             transaction_builder.set_fee(0).unwrap();
             transaction_builder
-                .add_output(123, &bob.default_subaddress(), &mut rng)
+                .add_output(
+                    Amount {
+                        value: 123,
+                        token_id: Mob::ID,
+                    },
+                    &bob.default_subaddress(),
+                    &mut rng,
+                )
                 .unwrap();
 
             let tx = transaction_builder.build(&mut rng).unwrap();
@@ -669,7 +676,14 @@ mod combine_tests {
                     transaction_builder.add_input(input_credentials);
                     transaction_builder.set_fee(0).unwrap();
                     transaction_builder
-                        .add_output(88, &bob.default_subaddress(), &mut rng)
+                        .add_output(
+                            Amount {
+                                value: 88,
+                                token_id: Mob::ID,
+                            },
+                            &bob.default_subaddress(),
+                            &mut rng,
+                        )
                         .unwrap();
 
                     let tx = transaction_builder.build(&mut rng).unwrap();
@@ -744,7 +758,14 @@ mod combine_tests {
                 transaction_builder.add_input(input_credentials);
                 transaction_builder.set_fee(0).unwrap();
                 transaction_builder
-                    .add_output(123, &bob.default_subaddress(), &mut rng)
+                    .add_output(
+                        Amount {
+                            value: 123,
+                            token_id: Mob::ID,
+                        },
+                        &bob.default_subaddress(),
+                        &mut rng,
+                    )
                     .unwrap();
 
                 let tx = transaction_builder.build(&mut rng).unwrap();
@@ -781,7 +802,14 @@ mod combine_tests {
                 transaction_builder.add_input(input_credentials);
                 transaction_builder.set_fee(0).unwrap();
                 transaction_builder
-                    .add_output(123, &recipient_account.default_subaddress(), &mut rng)
+                    .add_output(
+                        Amount {
+                            value: 123,
+                            token_id: Mob::ID,
+                        },
+                        &recipient_account.default_subaddress(),
+                        &mut rng,
+                    )
                     .unwrap();
 
                 let tx = transaction_builder.build(&mut rng).unwrap();
@@ -844,7 +872,14 @@ mod combine_tests {
                 transaction_builder.add_input(input_credentials);
                 transaction_builder.set_fee(0).unwrap();
                 transaction_builder
-                    .add_output(123, &recipient_account.default_subaddress(), &mut rng)
+                    .add_output(
+                        Amount {
+                            value: 123,
+                            token_id: Mob::ID,
+                        },
+                        &recipient_account.default_subaddress(),
+                        &mut rng,
+                    )
                     .unwrap();
 
                 let tx = transaction_builder.build(&mut rng).unwrap();
@@ -937,7 +972,14 @@ mod combine_tests {
                 transaction_builder.add_input(input_credentials);
                 transaction_builder.set_fee(0).unwrap();
                 transaction_builder
-                    .add_output(123, &bob.default_subaddress(), &mut rng)
+                    .add_output(
+                        Amount {
+                            value: 123,
+                            token_id: Mob::ID,
+                        },
+                        &bob.default_subaddress(),
+                        &mut rng,
+                    )
                     .unwrap();
 
                 let tx = transaction_builder.build(&mut rng).unwrap();
@@ -975,7 +1017,14 @@ mod combine_tests {
                 transaction_builder.add_input(input_credentials);
                 transaction_builder.set_fee(0).unwrap();
                 transaction_builder
-                    .add_output(123, &recipient_account.default_subaddress(), &mut rng)
+                    .add_output(
+                        Amount {
+                            value: 123,
+                            token_id: Mob::ID,
+                        },
+                        &recipient_account.default_subaddress(),
+                        &mut rng,
+                    )
                     .unwrap();
 
                 let mut tx = transaction_builder.build(&mut rng).unwrap();
@@ -1039,7 +1088,14 @@ mod combine_tests {
                 transaction_builder.add_input(input_credentials);
                 transaction_builder.set_fee(0).unwrap();
                 transaction_builder
-                    .add_output(123, &recipient_account.default_subaddress(), &mut rng)
+                    .add_output(
+                        Amount {
+                            value: 123,
+                            token_id: Mob::ID,
+                        },
+                        &recipient_account.default_subaddress(),
+                        &mut rng,
+                    )
                     .unwrap();
 
                 let tx = transaction_builder.build(&mut rng).unwrap();
--- a/consensus/service/src/validators.rs
+++ b/consensus/service/src/validators.rs
@@ -625,10 +625,7 @@ mod combine_tests {
                     let tx_secret_key_for_txo = RistrettoPrivate::from_random(&mut rng);
 
                     let tx_out = TxOut::new(
-                        Amount {
-                            value: 88,
-                            token_id: Mob::ID,
-                        },
+                        Amount::new(88, Mob::ID),
                         &alice.default_subaddress(),
                         &tx_secret_key_for_txo,
                         Default::default(),
@@ -677,10 +674,7 @@ mod combine_tests {
                     transaction_builder.set_fee(0).unwrap();
                     transaction_builder
                         .add_output(
-                            Amount {
-                                value: 88,
-                                token_id: Mob::ID,
-                            },
+                            Amount::new(88, Mob::ID),
                             &bob.default_subaddress(),
                             &mut rng,
                         )
@@ -759,10 +753,7 @@ mod combine_tests {
                 transaction_builder.set_fee(0).unwrap();
                 transaction_builder
                     .add_output(
-                        Amount {
-                            value: 123,
-                            token_id: Mob::ID,
-                        },
+                        Amount::new(123, Mob::ID),
                         &bob.default_subaddress(),
                         &mut rng,
                     )
@@ -803,10 +794,7 @@ mod combine_tests {
                 transaction_builder.set_fee(0).unwrap();
                 transaction_builder
                     .add_output(
-                        Amount {
-                            value: 123,
-                            token_id: Mob::ID,
-                        },
+                        Amount::new(123, Mob::ID),
                         &recipient_account.default_subaddress(),
                         &mut rng,
                     )
@@ -824,10 +812,7 @@ mod combine_tests {
                 // The transaction keys.
                 let tx_secret_key_for_txo = RistrettoPrivate::from_random(&mut rng);
                 let tx_out = TxOut::new(
-                    Amount {
-                        value: 123,
-                        token_id: Mob::ID,
-                    },
+                    Amount::new(123, Mob::ID),
                     &alice.default_subaddress(),
                     &tx_secret_key_for_txo,
                     Default::default(),
@@ -873,10 +858,7 @@ mod combine_tests {
                 transaction_builder.set_fee(0).unwrap();
                 transaction_builder
                     .add_output(
-                        Amount {
-                            value: 123,
-                            token_id: Mob::ID,
-                        },
+                        Amount::new(123, Mob::ID),
                         &recipient_account.default_subaddress(),
                         &mut rng,
                     )
@@ -909,10 +891,7 @@ mod combine_tests {
 
             // Create two TxOuts that were sent to Alice.
             let tx_out1 = TxOut::new(
-                Amount {
-                    value: 123,
-                    token_id: Mob::ID,
-                },
+                Amount::new(123, Mob::ID),
                 &alice.default_subaddress(),
                 &RistrettoPrivate::from_random(&mut rng),
                 Default::default(),
@@ -920,10 +899,7 @@ mod combine_tests {
             .unwrap();
 
             let tx_out2 = TxOut::new(
-                Amount {
-                    value: 123,
-                    token_id: Mob::ID,
-                },
+                Amount::new(123, Mob::ID),
                 &alice.default_subaddress(),
                 &RistrettoPrivate::from_random(&mut rng),
                 Default::default(),
@@ -973,10 +949,7 @@ mod combine_tests {
                 transaction_builder.set_fee(0).unwrap();
                 transaction_builder
                     .add_output(
-                        Amount {
-                            value: 123,
-                            token_id: Mob::ID,
-                        },
+                        Amount::new(123, Mob::ID),
                         &bob.default_subaddress(),
                         &mut rng,
                     )
@@ -1018,10 +991,7 @@ mod combine_tests {
                 transaction_builder.set_fee(0).unwrap();
                 transaction_builder
                     .add_output(
-                        Amount {
-                            value: 123,
-                            token_id: Mob::ID,
-                        },
+                        Amount::new(123, Mob::ID),
                         &recipient_account.default_subaddress(),
                         &mut rng,
                     )
@@ -1040,10 +1010,7 @@ mod combine_tests {
                 // The transaction keys.
                 let tx_secret_key_for_txo = RistrettoPrivate::from_random(&mut rng);
                 let tx_out = TxOut::new(
-                    Amount {
-                        value: 123,
-                        token_id: Mob::ID,
-                    },
+                    Amount::new(123, Mob::ID),
                     &alice.default_subaddress(),
                     &tx_secret_key_for_txo,
                     Default::default(),
@@ -1089,10 +1056,7 @@ mod combine_tests {
                 transaction_builder.set_fee(0).unwrap();
                 transaction_builder
                     .add_output(
-                        Amount {
-                            value: 123,
-                            token_id: Mob::ID,
-                        },
+                        Amount::new(123, Mob::ID),
                         &recipient_account.default_subaddress(),
                         &mut rng,
                     )
--- a/consensus/service/src/validators.rs
+++ b/consensus/service/src/validators.rs
@@ -583,10 +583,7 @@ mod combine_tests {
             transaction_builder.set_fee(0).unwrap();
             transaction_builder
                 .add_output(
-                    Amount {
-                        value: 123,
-                        token_id: Mob::ID,
-                    },
+                    Amount::new(123, Mob::ID),
                     &bob.default_subaddress(),
                     &mut rng,
                 )
--- a/consensus/service/src/validators.rs
+++ b/consensus/service/src/validators.rs
@@ -536,10 +536,7 @@ mod combine_tests {
             let tx_secret_key_for_txo = RistrettoPrivate::from_random(&mut rng);
 
             let tx_out = TxOut::new(
-                Amount {
-                    value: 123,
-                    token_id: Mob::ID,
-                },
+                Amount::new(123, Mob::ID),
                 &alice.default_subaddress(),
                 &tx_secret_key_for_txo,
                 Default::default(),
--- a/libmobilecoin/src/transaction.rs
+++ b/libmobilecoin/src/transaction.rs
@@ -378,14 +378,12 @@ pub extern "C" fn mc_transaction_builder_create(
 
         let mut transaction_builder = TransactionBuilder::new_with_box(
             block_version,
-            token_id,
+            Amount::new(fee, token_id),
             fog_resolver,
             memo_builder_box,
-        );
+        )
+        .expect("Could not create transaction builder");
 
-        transaction_builder
-            .set_fee(fee)
-            .expect("failure not expected");
         transaction_builder.set_tombstone_block(tombstone_block);
         Some(transaction_builder)
     })
--- a/libmobilecoin/src/transaction.rs
+++ b/libmobilecoin/src/transaction.rs
@@ -19,7 +19,7 @@ use mc_transaction_core::{
     ring_signature::KeyImage,
     tokens::Mob,
     tx::{TxOut, TxOutConfirmationNumber, TxOutMembershipProof},
-    BlockVersion, CompressedCommitment, EncryptedMemo, MaskedAmount, Token,
+    Amount, BlockVersion, CompressedCommitment, EncryptedMemo, MaskedAmount, Token,
 };
 
 use mc_transaction_std::{
@@ -498,6 +498,13 @@ pub extern "C" fn mc_transaction_builder_add_output(
             .as_slice_mut_of_len(TxOutConfirmationNumber::size())
             .expect("out_tx_out_confirmation_number length is insufficient");
 
+        // TODO: If you want to support mixed transactions, use something other
+        // than fee_token_id here.
+        let amount = Amount {
+            value: amount,
+            token_id: transaction_builder.get_fee_token_id(),
+        };
+
         let (tx_out, confirmation) =
             transaction_builder.add_output(amount, &recipient_address, &mut rng)?;
 
@@ -540,6 +547,13 @@ pub extern "C" fn mc_transaction_builder_add_change_output(
             .into_mut()
             .as_slice_mut_of_len(TxOutConfirmationNumber::size())
             .expect("out_tx_out_confirmation_number length is insufficient");
+
+        // TODO: If you want to support mixed transactions, use something other
+        // than fee_token_id here.
+        let amount = Amount {
+            value: amount,
+            token_id: transaction_builder.get_fee_token_id(),
+        };
 
         let (tx_out, confirmation) =
             transaction_builder.add_change_output(amount, &change_destination, &mut rng)?;
--- a/libmobilecoin/src/transaction.rs
+++ b/libmobilecoin/src/transaction.rs
@@ -496,8 +496,8 @@ pub extern "C" fn mc_transaction_builder_add_output(
             .as_slice_mut_of_len(TxOutConfirmationNumber::size())
             .expect("out_tx_out_confirmation_number length is insufficient");
 
-        // TODO: If you want to support mixed transactions, use something other
-        // than fee_token_id here.
+        // TODO (GH #1867): If you want to support mixed transactions, use something
+        // other than fee_token_id here.
         let amount = Amount {
             value: amount,
             token_id: transaction_builder.get_fee_token_id(),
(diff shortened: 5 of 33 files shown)


mobilecoind-json/src/data_types.rs
```diff
@@ -841,7 +841,10 @@ impl From<&RingMLSAG> for JsonRingMLSAG {
 pub struct JsonSignatureRctBulletproofs {
     pub ring_signatures: Vec<JsonRingMLSAG>,
     pub pseudo_output_commitments: Vec<String>,
-    pub range_proofs: String,
+    pub range_proof_bytes: String,
+    pub range_proofs: Vec<String>,
```
remoun, line 845: Optional/Equivalent: Add a `_hex` suffix to these new members
```suggestion
    pub range_proof_hex: String,
    pub range_proofs_hex: Vec<String>,
```

mobilecoind-json/src/data_types.rs
```diff
@@ -910,9 +920,17 @@ impl TryFrom<&JsonSignatureRctBulletproofs> for SignatureRctBulletproofs {
         let mut signature = SignatureRctBulletproofs::new();
         signature.set_ring_signatures(RepeatedField::from_vec(ring_sigs));
         signature.set_pseudo_output_commitments(RepeatedField::from_vec(commitments));
-        let proofs_bytes = hex::decode(&src.range_proofs)
+        let range_proof_bytes = hex::decode(&src.range_proof_bytes)
             .map_err(|err| format!("Could not decode from hex: {}", err))?;
```
remoun, line 924: Let's specify which item we couldn't decode.
```suggestion
        let range_proof_bytes = hex::decode(&src.range_proof_bytes)
            .map_err(|err| format!("Could not decode top-level range proof from hex '{}': {}", &src.range_proof_bytes, err))?;
```

transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -170,8 +251,52 @@ impl SignatureRctBulletproofs {
...
+                assert!(
+                    !commitments.is_empty(),
+                    "logic error when accumulating commitments for token id"
+                );
+
+                let range_proof = RangeProof::from_bytes(range_proof)
+                    .map_err(|_e| Error::RangeProofDeserializationError)?;
+
+                check_range_proofs(&range_proof, &commitments, generator, rng)
+                    .map_err(|_e| Error::RangeProofError)?;
```
remoun, line 298: Can we include some info about the underlying error in the returned error?

transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -295,10 +437,7 @@ fn sign_with_balance_check<CSPRNG: RngCore + CryptoRng>(
     //
     // Note: This implicit fee output is not the same as the accumulated fee output
     // produced by the enclave -- the blinding of that output is not zero.
-    let sum_of_output_blindings: Scalar = output_values_and_blindings
-        .iter()
-        .map(|(_, blinding)| blinding)
-        .sum();
+    let sum_of_output_blindings: Scalar = output_secrets.iter().map(|secret| secret.blinding).sum();
```
remoun, line 440: Do we need to worry about overflowing here?

garbageslam: Overflowing definitely happens, and it's supposed to wrap around mod p -- the dalek Scalar takes care of that

mobilecoind-json/src/data_types.rs
garbageslam: I think these names impact the Json API, but I'm okay with your suggestions

transaction/core/src/ring_signature/rct_bulletproofs.rs
garbageslam: Yea let's do it

mobilecoind-json/src/data_types.rs
garbageslam: In the rest of this API, we map the protobuf schema onto json names exactly. But actually in this specific spot we had "range_proof" vs "range_proof_bytes"

remoun: These are strings while the proto fields are bytes, so adding/changing the suffix seems OK

remoun: With that said, I don't feel strongly about renaming this.

transaction/core/src/ring_signature/rct_bulletproofs.rs
garbageslam: I had to stringify the range proof error because it doesn't implement traits that `TxValidationError` requires, and I don't want to break everything

mobilecoind-json/src/data_types.rs
```diff
@@ -841,7 +841,10 @@ impl From<&RingMLSAG> for JsonRingMLSAG {
 pub struct JsonSignatureRctBulletproofs {
     pub ring_signatures: Vec<JsonRingMLSAG>,
     pub pseudo_output_commitments: Vec<String>,
-    pub range_proofs: String,
+    pub range_proof_bytes: String,
+    pub range_proofs: Vec<String>,
+    pub pseudo_output_token_ids: Vec<u64>,
```
eranrund, line 846: I think `u64` is a problem in some JSON parsers, which is why in a bunch of places it is encoded as a string ;(

transaction/core/src/ring_signature/error.rs
```diff
@@ -49,15 +49,48 @@ pub enum Error {
      */
     ValueNotConserved,
 
-    /// Invalid RangeProof
-    RangeProofError,
+    /// Invalid RangeProof: {0}
+    RangeProofError(String),
```
eranrund, line 53: Nit: `Error` suffix is redundant

transaction/core/src/ring_signature/error.rs
```diff
@@ -49,15 +49,48 @@ pub enum Error {
...
 
 impl From<mc_util_repr_bytes::LengthMismatch> for Error {
     fn from(src: mc_util_repr_bytes::LengthMismatch) -> Self {
         Error::LengthMismatch(src.found, src.expected)
     }
 }
+
+impl From<RangeProofError> for Error {
+    fn from(src: RangeProofError) -> Self {
+        Error::RangeProofError(src.to_string())
```
eranrund, line 94: Is it not possible to embed the actual `RangeProofError`? I'm guessing some traits missing?

transaction/core/src/ring_signature/generator_cache.rs
```diff
@@ -0,0 +1,25 @@
+//! A simple generator cache
```
eranrund, line 1: Nit: copyright

transaction/std/src/memo_builder/rth_memo_builder.rs
```diff
@@ -195,11 +212,36 @@ impl MemoBuilder for RTHMemoBuilder {
...
+        } else {
+            // If no outlays occurred yet, this should be the token id for the whole tx.
+            self.outlay_token_id = Some(amount.token_id);
+        }
+
+        // If the fee is not the same token id as the outlay, then
+        // total_outlay.checked_add(self.fee.value) is wrong.
+        // We need to specify token-id aware RTH memos
+        if self.fee.token_id != amount.token_id {
+            panic!(
```
eranrund, line 229: This doesn't seem like it belongs here

transaction/std/src/transaction_builder.rs
```diff
@@ -2595,4 +2779,333 @@ pub mod transaction_builder_tests {
...
+
+            assert!(tx_out1.view_key_match(&sender.view_private_key()).is_err());
+
+            assert!(tx_out2.view_key_match(&sender.view_private_key()).is_err());
+        }
+    }
+
+    #[test]
+    // Test mixed transactions expected failure case 1
+    fn test_mixed_transactions_expected_failure1() {
```
eranrund, line 2922: What is the expected failure?

mobilecoind-json/src/data_types.rs
garbageslam: i see -- thank you

transaction/core/src/ring_signature/error.rs
garbageslam: yeah i tried to pass the whole error, but it doesn't support `Serialize, Deserialize`, and `TransactionValidationError` wants that, presumably to pass across enclave boundary.

we could do upstream work in bulletproofs crate maybe, we already forked them anyways

transaction/std/src/memo_builder/rth_memo_builder.rs
garbageslam: sorry, that was debugging, will fix

transaction/std/src/transaction_builder.rs
garbageslam: i subtracted 1 from one of the outputs, so that it isn't balanced

garbageslam: will document this a little better

transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -308,35 +446,110 @@ fn sign_with_balance_check<CSPRNG: RngCore + CryptoRng>(
...
+                "logic error when accumulating commitments for token id"
+            );
+
+            let (range_proof, _commitments) =
+                generate_range_proofs(&values, &blindings, generator, rng)?;
+
+            range_proofs.push(range_proof.to_bytes());
+        }
+
+        (Default::default(), range_proofs)
```
eranrund, line 524: Nit: Above you used `vec![]` instead of `Default::default()`, I think that is easier to read/understand.

transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -308,35 +446,110 @@ fn sign_with_balance_check<CSPRNG: RngCore + CryptoRng>(
...
+            .cloned()
+            .chain(
+                output_secrets
+                    .iter()
+                    .map(|secret| (secret.value, secret.blinding)),
+            )
             .unzip();
-        generate_range_proofs(&values, &blindings, &generator, rng)
-            .map_err(|_e| Error::RangeProofError)?
+        let (range_proof, _commitments) =
```
eranrund, line 470: I'm wondering about discarding the commitments here and want to check my understanding. You are simply regenerating them below? Is there a cost to doing that?

transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -347,46 +560,70 @@ fn sign_with_balance_check<CSPRNG: RngCore + CryptoRng>(
         }
     }
 
-    let pseudo_output_commitments: Vec<CompressedCommitment> = commitments
-        .iter()
-        .take(num_inputs)
-        .map(CompressedCommitment::from)
+    // The actual pseudo output commitments use the blindings from
+    // `pseudo_output_blinding` and not the original true input.
+    let pseudo_output_commitments: Vec<CompressedCommitment> = pseudo_output_commitments
```
eranrund, line 565: Repurposing the `pseudo_output_commitments` here seems a bit janky, any reason not to `CompressedCommitment::from` at the block that generates the original `pseudo_output_commitments` from `input_secrets`/`pseudo_output_blindings`?

transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -347,46 +560,70 @@ fn sign_with_balance_check<CSPRNG: RngCore + CryptoRng>(
...
+        .collect();
+
+    if !block_version.mixed_transactions_are_supported() {
+        pseudo_output_token_ids.clear();
+        output_token_ids.clear();
+        assert!(!range_proof.is_empty());
+        assert!(range_proofs.is_empty());
+    } else {
+        assert!(range_proof.is_empty());
+        assert!(!range_proofs.is_empty());
```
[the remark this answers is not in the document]
eranrund: My take is that this code runs on clients, I think it is acceptable that it panics instead of produces an unexpected result.

transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -804,11 +1111,60 @@ mod rct_bulletproofs_tests {
...
+                fee,
+                params.fee_token_id,
+                &mut rng,
+            );
+
+            match result {
+                Err(Error::RangeProofError(_)) => {},
+                Err(err) => { panic!("Expected: RangeProofError, found {}", err) },
+                Ok(()) => { panic!("Expected: RangeProofError, found Ok") },
+            };
```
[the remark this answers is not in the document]
eranrund: IIRC that requires an unstable feature, `assert!(matches!(...))` doesn't.

transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -396,10 +633,17 @@ fn compute_extended_message_either_version(
     message: &[u8],
     pseudo_output_commitments: &[CompressedCommitment],
     range_proof_bytes: &[u8],
+    range_proofs: &[Vec<u8>],
 ) -> Vec<u8> {
     if block_version >= BlockVersion::TWO {
```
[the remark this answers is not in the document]
garbageslam: Yea we can do that -- I will do that in master and rebase on it I think

transaction/core/src/ring_signature/rct_bulletproofs.rs
garbageslam: yeah i'm regenerating them below

the thing is, in the old way, there was only one range proof, and it proves each of the commitments in order of pseudo_outputs, followed by outputs.

in the new world, there have to be multiple range proofs, one for each pedersen generator.
so I have to like, sort them and pick out the ones that match each token id.

if we try to take the commitments returned by `generate_range_proofs`, we have to "unsort" them back into the right order to be useful later, because the order of the pseudo output commtiments has to match the order of the TxIn in the prefix, which is sorted by tx public key.

i felt like it was simpler to just regenerate them, at least when trying to get the code to pass tests.
we can try to go back and do the "unsorting" thing, it will probably save some elliptic curve operations if we don't regenerate the commitments.

transaction/core/src/ring_signature/rct_bulletproofs.rs
garbageslam: we need to add them at line 550, and they can't be compressed to add them

we could change the name of this variable if you think that's better

eranrund: ah I see. I don't feel strongly about it.

transaction/core/src/ring_signature/rct_bulletproofs.rs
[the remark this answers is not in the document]
garbageslam: trying with this crate: https://crates.io/crates/assert_matches

i think the benefit of this over `assert!(matches!` is that the panic message shows you what didn't match. if they stabilize `assert_matches` in `std` then we can drop this dependency

mobilecoind-json/src/data_types.rs
```diff
@@ -837,11 +837,33 @@ impl From<&RingMLSAG> for JsonRingMLSAG {
     }
 }
 
+// Representa TokenId (u64) using string, when serializing to Json
+// This does not rely on the serde-json arbitrary precision feature, which
+// (we fear) might break other things
+#[derive(Deserialize, Serialize, Default, Debug, Clone)]
+#[serde(transparent)]
+pub struct JsonTokenId(#[serde(with = "serde_with::rust::display_fromstr")] pub u64);
```
garbageslam, line 845: @eranrund @remoun @jcape lmk if you guys are okay with this fix

transaction/core/src/ring_signature/rct_bulletproofs.rs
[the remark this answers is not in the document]
garbageslam: I am okay with either way

In the future I hope that we will drop support for older block versions that are no longer relevant, and then we can delete the branches having to do with `range_proof`, so ultimately there is a path to get rid of this.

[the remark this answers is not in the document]
garbageslam: @remoun maybe we can leave it as assert for now, so that when SDKs are testing with it, it's a hard assert, but we can make a ticket so that before we branch for release we change it to debug assert?

transaction/core/src/ring_signature/rct_bulletproofs.rs
garbageslam: If you don't mind, I would like to try to develop this unsorting in a subsequent commit, because it's going to make the signed contingent input commit more complicated

garbageslam: i think the unsorting thing only impacts signing a transaction, so any slowdown is only client side and won't affect the performance of the network

consensus/service/src/validators.rs
```diff
@@ -585,7 +582,14 @@ mod combine_tests {
             transaction_builder.add_input(input_credentials);
             transaction_builder.set_fee(0).unwrap();
             transaction_builder
-                .add_output(123, &bob.default_subaddress(), &mut rng)
+                .add_output(
+                    Amount {
+                        value: 123,
+                        token_id: Mob::ID,
+                    },
```
remoun, line 589:
```suggestion
                    Amount::new(123, Mob::ID),
```

transaction/core/src/ring_signature/generator_cache.rs
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) 2022 The MobileCoin Foundation
```
remoun, line 1:
```suggestion
// Copyright (c) 2018-2022 The MobileCoin Foundation
```

transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -42,9 +65,29 @@ pub struct SignatureRctBulletproofs {
     /// This contains range_proof.to_bytes(). It is stored this way so that this
     /// struct may derive Default, which is a requirement for serializing
     /// with Prost.
+    ///
+    /// Note: This is EMPTY if mixed transactions is enabled
```
remoun, line 69:
```suggestion
    /// Note: This is EMPTY if mixed transactions are enabled
```

transaction/core/src/ring_signature/rct_bulletproofs.rs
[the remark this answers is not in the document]
remoun: SGTM

transaction/std/src/memo_builder/rth_memo_builder.rs
```diff
@@ -79,8 +83,12 @@ impl Default for RTHMemoBuilder {
             wrote_destination_memo: false,
             last_recipient: Default::default(),
             total_outlay: 0,
+            outlay_token_id: None,
             num_recipients: 0,
-            fee: Mob::MINIMUM_FEE,
+            fee: Amount {
+                value: Mob::MINIMUM_FEE,
+                token_id: Mob::ID,
+            },
```
remoun, line 91:
```suggestion
            fee: Amount::new(Mob::MINIMUM_FEE, Mob::ID),
```

transaction/std/src/transaction_builder.rs
```diff
@@ -1303,14 +1342,21 @@ pub mod transaction_builder_tests {
...
                     )
                     .unwrap();
 
                 transaction_builder
-                    .add_change_output(change_value, &sender_change_dest, &mut rng)
+                    .add_change_output(
+                        Amount {
+                            value: change_value,
+                            token_id,
+                        },
```
remoun, line 1356:
```suggestion
                        Amount::new(change_value, token_id),
```

transaction/std/src/transaction_builder.rs
```diff
@@ -1460,14 +1506,21 @@ pub mod transaction_builder_tests {
...
```
remoun, line 1520:
```suggestion
                        Amount::new(change_value, token_id),
```

transaction/std/src/transaction_builder.rs
```diff
@@ -2089,14 +2154,21 @@ pub mod transaction_builder_tests {
...
                     )
                     .unwrap();
 
                 transaction_builder
-                    .add_change_output(change_value, &alice_change_dest, &mut rng)
+                    .add_change_output(
+                        Amount {
+                            value: change_value,
+                            token_id,
+                        },
```
remoun, line 2168:
```suggestion
                        Amount::new(change_value, token_id),
```

transaction/std/src/transaction_builder.rs
```diff
@@ -2536,14 +2624,21 @@ pub mod transaction_builder_tests {
...
                 )
                 .unwrap();
 
             transaction_builder
-                .add_change_output(change_value, &sender_change_dest, &mut rng)
+                .add_change_output(
+                    Amount {
+                        value: change_value,
+                        token_id,
+                    },
```
remoun, line 2638:
```suggestion
                        Amount::new(change_value, token_id),
```

transaction/std/src/transaction_builder.rs
```diff
@@ -2590,4 +2685,219 @@ pub mod transaction_builder_tests {
...
+            token_id: Mob::ID,
+        };
+        let change_amount = Amount {
+            value: 128 * MILLIMOB_TO_PICOMOB,
+            token_id: Mob::ID,
+        };
+        let amount2 = Amount {
+            value: 999999,
+            token_id: 2.into(),
+        };
```
remoun, line 2711:
```suggestion
        let amount1 = Amount::new(1475 * MILLIMOB_TO_PICOMOB, Mob::ID);
        let change_amount = Amount::new(128 * MILLIMOB_TO_PICOMOB, Mob::ID);
        let amount2 = Amount::new(999999, 2.into())'
```

transaction/std/src/transaction_builder.rs
```diff
@@ -2590,4 +2685,219 @@ pub mod transaction_builder_tests {
...
```
remoun, line 2846:
```suggestion
        let amount1 = Amount::new(1475 * MILLIMOB_TO_PICOMOB, Mob::ID);
        let change_amount = Amount::new(128 * MILLIMOB_TO_PICOMOB, Mob::ID);
        let amount2 = Amount::new(999999, 2.into())'
```

remoun (approved):
LGTM

transaction/core/Cargo.toml
```diff
@@ -52,6 +52,7 @@ features = ["default-code-coverage"]
 curve25519-dalek = { version = "4.0.0-pre.2", default-features = false, features = ["nightly", "u64_backend"] }
 
 [dev-dependencies]
+assert_matches = "1.5"
```
eranrund, line 55: Why do you need this? `assert!(matches!(` worked in other places of the codebase and didn't require another dependency.

mobilecoind-json/src/data_types.rs
garbageslam: I think this is basically the same as "just using strings" in the json, it's just less typing

We could change this to `JsonU64` I guess and use it for other u64s

transaction/core/src/ring_signature/generator_cache.rs
garbageslam: see i feel like this is wierd, this file was written in 2022 so how can we claim it is copyrighted since 2018?

in another open source project i worked on the files all had different starting years and the same ending year for copyright claims. i'm happy to change it, i just want to make sure I understand why we are doing this

transaction/core/Cargo.toml
garbageslam: i think `assert_matches` is nicer because `assert!(matches!(`, when it fails, just tells you "matches! was false",
but `assert_matches!` will be able to log the expression that failed to match which helps you debug it

transaction/std/src/transaction_builder.rs
garbageslam: will swap `'` for `;` in next commit

transaction/std/src/transaction_builder.rs
garbageslam: will swap `'` for `;` in next commit

transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -168,10 +247,52 @@ impl SignatureRctBulletproofs {
...
+                        } else {
+                            None
+                        }
+                    })
+                    .collect();
+
+                assert!(
+                    !commitments.is_empty(),
+                    "logic error when accumulating commitments for token id"
+                );
```
garbageslam, line 289: change this to an error

transaction/std/src/transaction_builder.rs
```diff
@@ -103,23 +102,34 @@ impl<FPR: FogPubkeyResolver> TransactionBuilder<FPR> {
...
+    /// * `block_version` - The block version to use when building a transaction
+    /// * `fee_token_id` - The token id of the fee output
     /// * `fog_resolver` - Source of validated fog keys to use with this
     ///   transaction
     /// * `memo_builder` - An object which creates memos for the TxOuts in this
     ///   transaction
     pub fn new_with_box(
         block_version: BlockVersion,
-        token_id: TokenId,
+        fee_token_id: TokenId,
```
garbageslam, line 113: TODO: change this to an amount

api/proto/external.proto
```diff
@@ -255,7 +255,10 @@ message RingMLSAG {
 message SignatureRctBulletproofs {
     repeated RingMLSAG ring_signatures = 1;
     repeated CompressedRistretto pseudo_output_commitments = 2;
-    bytes range_proofs = 3;
+    bytes range_proof_bytes = 3;
```
wjuan-mob, line 258: This should maybe be a oneof (range_proof_bytes/range_proofs) since you only want to be able to have either the range_proof_bytes or range_proofs.

consensus/enclave/impl/src/lib.rs
```diff
@@ -236,7 +236,7 @@ impl SgxConsensusEnclave {
         // We need to make sure all transactions are valid. We also ensure they all
         // point at the same root membership element.
         for (tx, proofs) in transactions_with_proofs.iter() {
-            let token_id = TokenId::from(tx.prefix.token_id);
+            let token_id = TokenId::from(tx.prefix.fee_token_id);
```
wjuan-mob, line 239: I think it'd be helpful for clarity to change the variable name to also be fee_token_id here and below.

transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -42,9 +65,29 @@ pub struct SignatureRctBulletproofs {
...
     #[prost(bytes, tag = "3")]
     #[digestible(never_omit)]
     pub range_proof_bytes: Vec<u8>,
+
+    /// A range proof, one for each token id that is used in the transaction.
+    ///
+    /// The range proofs correspond to the sorted order of token ids used.
+    ///
+    /// Note: This is EMPTY if mixed transactions is not enabled
+    #[prost(bytes, repeated, tag = "4")]
```
wjuan-mob, line 79: Similarly to the oneof, I think we could use an enum type: RangeProof {
SingleProof: Vec<u8>
MultiProof: Vec<Vec<u8>>
}

garbageslam: So the way I think the schema evolution is likely to work is:

Initially: Only `Vec<u8>`

Transition: Both a `Vec<u8>` field and a `Vec<Vec<u8>>` field

End: Remove the (no longer used) `Vec<u8>` field, and only the `Vec<Vec<u8>>` field remains.

This is normally the forwards-and-backwards compatibilie method of protobuf schema evolution. If we already had a OneOf here then we could use that method instead, but we can't just change the `Vec<u8>` to a `OneOf` now without a breaking change.

We could make the next thing `OneOf<Vec<Vec<u8>>>` instead. I'm not sure if that makes sense. If Isis is right then we may be able to aggregate all these range proofs. But even in that case `Vec<Vec<u8>>` is fine.

api/proto/external.proto
wjuan-mob: Per discussion, adding a oneof in this case is likely backwards compatible in protobuf, however we would likely encounter difficulties due to potential conflicts in the hash of old tx structure in the mc-crypt-digestible crate. With additive changes we can avoid this conflict because new fields simply hash to 0.

garbageslam: >  simply hash to 0.

simply don't contribute to the hash when they are `None`

yeah...

we could add a OneOf with one option, but I think what we actually want to do here is like, make `SignatureRctBulletproofsv2` or something and just start over with this structure, and do schema evolution in `Tx` instead

iamalwaysuncomfortable (approved):
Review #1 - questions about a few core things

transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -155,11 +213,32 @@ impl SignatureRctBulletproofs {
...
+                token_ids.insert(token_id.into());
+            }
+            for token_id in &self.pseudo_output_token_ids {
+                token_ids.insert(token_id.into());
+            }
+            token_ids
+        };
+
+        // Get a generator cache
+        let mut generator_cache = GeneratorCache::default();
```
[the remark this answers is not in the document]
garbageslam: 1. Yeah, in the current design for MCIP #25, we are going with generator per token id, and so this object is caching those.
2. Yeah, the generators are computed here: https://github.com/mobilecoinfoundation/mobilecoin/blob/93647188e543f5a9eaaf6e2a49587ebfd828a512/transaction/core/src/ring_signature/mod.rs#L46 This formula has to be fixed pretty much for all time or older tokens will not be recognized. The formula is a bit funky, mainly because, the way we computed the initial generator for MOB was a bit funky, and we have to be backwards compatible to that forever.
3. The idea here was that this cache is built for every transaction. The thing is, the cache as it currently exists is not constant time. The idea I had here was like, we're not trying to avoid revealing how many token ids you have in your Tx (which would be basically infeasible because we have different range proofs per token id), or, which TxOut's have the same token id as which other TxOut's or pseudo outputs. (We could maybe do that). We don't want to reveal when a token id used by Bob is the same as the token id used by Alice. So, we trash the cache when we look at a new transaction, to try to kill off cache timing attacks.

This means that e.g. if your transaction has 5 inputs and outputs, and two token ids, you do two elliptic curve operations to get generators, instead of 5 before the cache.

We could make this better from a privacy point of view if we had a constant-time map for ristretto points, but currently that is impossible because ristretto point doesn't implement `subtle::ConditionalAssign`. But also the constant-time map we have in the enclave now is pretty dumb and just does linear scanning, so if that thing can live forever and get very big i might get worried about performance. But it wouldn't get big unless there are a lot of token ids. But someday that may happen.

I feel like doing a lot of work to avoid 2 elliptic curve operations when validating a Tx may be misguided -- I was able to eliminate two elliptic curve operations anyways during validation by doing `RistrettoPoint::identity()` instead of `generator.commit(Zero, Zero)`. I think there is a lot of low hanging optimization fruit right now. So I feel like engineering a massive, fast, constant-time global generator cache might not be the easiest way to make it faster.

We might be able to do something like, hard-code a table of the first 16 token ids or something, and constant-time scan across that, then fall back to the btree map? That might be worth doing?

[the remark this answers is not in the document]
iamalwaysuncomfortable: Dumping the cache makes a lot sense. It might make sense for us to take a quick look at how much work putting `subtle::ConditionalAssign` into Dalek Ristretto might be. If it's low hanging fruit it might worth it! I can take a peek at what that would entail. 

I agree with you that there's potentially other easier optimizations to do in the meantime - although at this very moment, I'm not sure what they are. But I don't even think they're needed for this PR. It's already big. But subsequent 'quickening' PRs might be a good idea.

I guess importantly, we're not currently taking a measure of the speed. It sounds like a wise thing to do to potentially put benchmarks gauge this?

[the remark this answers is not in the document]
garbageslam: I think we used to have some benchmarks targets, but they weren't in CI and eventually got stale and then got deleted

I remember joey used to run benchmarks of transaction construction when we were deciding how big rings should be

We should make some more cargo benchmarks. It would be great if we actually tracked those numbers over time, we could have something that watches for perf regressions when a PR goes in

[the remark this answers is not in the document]
garbageslam: > f it's low hanging fruit it might worth it! I can take a peek at what that would entail.

:+1: 

> potentially other easier optimizations to do in the meantime - although at this very moment, I'm not sure what they are.

I tried to make some tickets.. let me see if i can track them down...

I opened this PR a while ago: https://github.com/mobilecoinfoundation/mobilecoin/pull/1423/files

I think this "if check_value_is_preserved" is a client-side operation that is done when building a transaction, to confirm that consensus will agree that the tx is valid, before we submit it. But the way it is done today, it entails a bunch of elliptic curve operations on the phone. I think we might be able to do it better if we just check that that, sum of values (u64's) is balanced, and sum of blindings (dalek::Scalar) is balanced.

I suspect that the "scalar mul precomputation table" that Isis talked about (I made a ticket here: https://github.com/mobilecoinfoundation/mobilecoin/issues/1564) may have significantly more impact, but it's also going to entail either:
1. Sticking that table in a lazy static, which would work for existing targets (and which we currently do for bulletproofs generators), but is possibly a tech debt
2. Passing the scalar mul table around everywhere (a lot of work, but more flexibility for embedded devices)

android-bindings/src/bindings.rs
```diff
@@ -1692,12 +1693,19 @@ pub unsafe extern "C" fn Java_com_mobilecoin_lib_TransactionBuilder_add_1output(
 
             let value = jni_big_int_to_u64(env, value)?;
 
+            // TODO: If you want to do mixed transactions, use something other
+            // than fee_token_id here.
```
sugargoat, line 1697: Is this TODO a note for the consumers of the android-bindings? Maybe providing a little more clarity on where to look to be able to do mixed transactions, e.g.

```suggestion
            // NOTE: This transaction builder does not support mixed transactions at this time. To 
            // implement mixed transactions, see the implementation in TransactionBuilder
            // TODO: Provide mixed-transaction functionality in these bindings.
```

api/proto/external.proto
sugargoat, line 258: Could we add documentation here about the new fields to help external consumers of this API? (that can be a follow-on PR if needed)

libmobilecoin/src/transaction.rs
```diff
@@ -541,6 +546,13 @@ pub extern "C" fn mc_transaction_builder_add_change_output(
             .as_slice_mut_of_len(TxOutConfirmationNumber::size())
             .expect("out_tx_out_confirmation_number length is insufficient");
 
+        // TODO: If you want to support mixed transactions, use something other
```
sugargoat, line 549: Refer to https://github.com/mobilecoinfoundation/mobilecoin/pull/1827/files#r859868566

mobilecoind-json/src/bin/main.rs
```diff
@@ -295,12 +295,8 @@ fn create_request_code(
     // Generate b58 code
```
sugargoat, line 295: Confirming: the changes in this file are all improving syntax/style, but not material to mixed token types?

garbageslam: Yes, there was a lot of discussion about u64s in Json earlier, that ended with, the JsonU64 thing is fine if we use in other places

mobilecoind-json/src/data_types.rs
```diff
@@ -1446,10 +1462,25 @@ mod test {
...
+            proto_proposal
+                .get_tx()
+                .get_signature()
+                .pseudo_output_token_ids,
+            proto2.get_tx().get_signature().pseudo_output_token_ids,
+        );
+        assert_eq!(
+            proto_proposal.get_tx().get_signature().output_token_ids,
+            proto2.get_tx().get_signature().output_token_ids,
+        );
```
sugargoat, line 1484: Is there a follow-on ticket to implement mixed transactions tests for mobilecoind-json? I don't think it needs to be blocking, but maybe there's a ticket for "ensure mixed token types test coverage in all clients" that can have a prioritized checklist.

android-bindings/src/bindings.rs
garbageslam: The note is for Android bindings maintainers to have guidance --

The way things work now, libmobilecoin and android have their own headers and cross language linkage matching to this code, so I can't change the API without breaking the build. When I add functionality to Tx builder, I configure it to just work the way it was working before so as not to break their stuff, but then leave a note about where that configuration is happening.

So to me this comment is confusing : "This transaction builder does not support mixed transactions at this time " because the transaction builder does support that as of this revision. It's just that the bindings still expose only the previous functionality. So the note is to the binding maintainer about how they could expose that if desired

mobilecoind/src/payments.rs
```diff
@@ -963,8 +962,14 @@ impl<T: BlockchainConnection + UserTxConnection + 'static, FPR: FogPubkeyResolve
         let mut tx_out_to_outlay_index = HashMap::default();
         let mut outlay_confirmation_numbers = Vec::default();
         for (i, outlay) in destinations.iter().enumerate() {
+            // TODO: If you want to support mixed transactions, use outlay-specific token id
+            // here
```
sugargoat, line 966: Similar comment as: https://github.com/mobilecoinfoundation/mobilecoin/pull/1827/files#r859858717

Also maybe a similar comment as https://github.com/mobilecoinfoundation/mobilecoin/pull/1827/files#r859876530, to have a prioritized checklist of implementing this functionality in clients.

api/proto/external.proto
garbageslam: 👍

mobilecoind/src/payments.rs
```diff
@@ -982,15 +987,22 @@ impl<T: BlockchainConnection + UserTxConnection + 'static, FPR: FogPubkeyResolve
         }
         let change = input_value - total_value - tx_builder.get_fee();
 
-        // If we do, add an output for that as well.
+        // If we do have nonzero change, add an output for that as well.
         // TODO: Should the exchange write destination memos?
```
sugargoat, line 991: Not in your PR - but wondering if we have this TODO captured as a ticket anywhere?

transaction/core/src/blockchain/block_version.rs
```diff
@@ -68,6 +68,9 @@ impl BlockVersion {
     /// Constant for block version two
     pub const TWO: Self = Self(2);
 
+    /// Constant for block version two
```
sugargoat, line 71:
```suggestion
    /// Constant for block version three
```

mobilecoind-json/src/data_types.rs
garbageslam: I am not sure we're planning even to support multiple token types txs in the mobilecoind-json bindings -- let's sync with Eran, maybe I'm out of the loop

mobilecoind/src/payments.rs
garbageslam: Yes there is a GitHub issue

transaction/core/src/ring_signature/error.rs
```diff
@@ -2,13 +2,13 @@
...
+use alloc::string::{String, ToString};
 use displaydoc::Display;
 use serde::{Deserialize, Serialize};
 
 /// An error which can occur in connection to a ring signature
-#[derive(
-    Clone, Copy, Debug, Display, Eq, Hash, Ord, PartialEq, PartialOrd, Serialize, Deserialize,
-)]
+#[derive(Clone, Debug, Deserialize, Display, Eq, Hash, Ord, PartialEq, PartialOrd, Serialize)]
 pub enum Error {
```
sugargoat, line 12: Wow nice - love the error enumeration expansion to clearly capture the failure modes!

mobilecoind/src/payments.rs
garbageslam: #1522

transaction/core/src/ring_signature/generator_cache.rs
```diff
@@ -0,0 +1,27 @@
...
+//! A simple generator cache
+
+use super::{generators, PedersenGens};
+use crate::TokenId;
+use alloc::collections::BTreeMap;
+
+/// GeneratorCache is a simple object which caches computations of
+/// generator: TokenId -> PedersenGens
+///
+/// This is intended just to be used in the scope of a single transaction,
```
sugargoat, line 12: In the scope of constructing or verifying a single transaction?

transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -23,10 +26,30 @@ use crate::{
...
     range_proofs::{check_range_proofs, generate_range_proofs},
-    ring_signature::{generators, mlsag::RingMLSAG, Error, KeyImage, Scalar},
-    BlockVersion, Commitment, CompressedCommitment,
+    ring_signature::{mlsag::RingMLSAG, Error, GeneratorCache, KeyImage, Scalar},
+    Amount, BlockVersion, Commitment, CompressedCommitment,
 };
 
+/// The secrets corresponding to an input needed to create a signature
+#[derive(Clone, Debug)]
+pub struct InputSecret {
```
sugargoat, line 35: Love these two structs 👏

api/proto/external.proto
sugargoat: e.g. from rct_bulletproofs for `range_proof_bytes` `This field is empty if mixed transactions are enabled`

transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -96,19 +137,19 @@ impl SignatureRctBulletproofs {
     ///   commitments.
     /// * `output_commitments` - Output amount commitments.
     /// * `fee` - Value of the implicit fee output.
-    /// * `token id` - This determines the pedersen generator for commitments
+    /// * `fee_token id` - This determines the pedersen generator for fee
```
sugargoat, line 140:
```suggestion
    /// * `fee_token_id` - This determines the pedersen generator for fee
```

transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -128,6 +169,23 @@ impl SignatureRctBulletproofs {
...
 
+        // pseudo output token ids must be provided if mixed transactions is enabled
+        if block_version.mixed_transactions_are_supported() {
+            if self.pseudo_output_commitments.len() != self.pseudo_output_token_ids.len() {
+                return Err(Error::MissingPseudoOutputTokenIds);
+            }
+            if output_commitments.len() != self.output_token_ids.len() {
+                return Err(Error::MissingOutputTokenIds);
+            }
+        } else {
```
sugargoat, line 180: also check if `range_proofs` is empty?

transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -155,11 +213,32 @@ impl SignatureRctBulletproofs {
             decompressed_pseudo_output_commitments.push(commitment);
         }
 
-        // Compute the pedersen generators for this token_id
-        let generator = generators(token_id);
+        // Collect list of of unique token ids
+        let token_ids = {
```
sugargoat, line 217: Confirming my understanding: 1) Do we care about constant-timeness in `verify` (I think the answer is yes, for certain operations that would leak material information about sender, receiver, amount, or token type), 2) is it correct to say, "This operation does not need to be constant time because it does not leak any information about which `token_ids` are being used, just that multiple `token_ids` are being used.

transaction/core/src/ring_signature/generator_cache.rs
garbageslam: Yea that's how it's used in this commit

transaction/core/src/ring_signature/rct_bulletproofs.rs
garbageslam: yeah that happens later right now, when validating the range proofs, but we could do all those checks here instead

transaction/core/src/ring_signature/rct_bulletproofs.rs
garbageslam: thanks

transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -155,11 +213,32 @@ impl SignatureRctBulletproofs {
...
+            let mut token_ids = BTreeSet::default();
+            token_ids.insert(fee.token_id);
+            for token_id in &self.output_token_ids {
+                token_ids.insert(token_id.into());
+            }
+            for token_id in &self.pseudo_output_token_ids {
+                token_ids.insert(token_id.into());
```
sugargoat, line 224: Confirming my understanding of transaction data: 

`pseudo_output_commitments` length is 1:1 with `inputs`, and the `token_id` set for `pseudo_output_commitments` needs to match `inputs` exactly, or else they will not properly construct commitments to zero (this is verified below when checking `Ristretto::identity` equality).

Therefore, `pseudo_output_token_ids` should have the same set of `token_ids` as `output_token_ids`, because there is no way (yet) to introduce a new `token_id` without spending an `input` from that `token_id`.

Is that correct?

transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -168,31 +247,111 @@ impl SignatureRctBulletproofs {
...
+
+                check_range_proofs(&range_proof, &commitments, generator, rng)?
+            }
+        }
 
+        // Transaction must be balanced (not create or destroy value).
+        //
         // Output commitments - pseudo_outputs must be zero.
+        //
+        // Note: Why does this imply that the transaction is balanced?
```
sugargoat, line 301: Awesome - thank you for this write up

transaction/core/src/ring_signature/error.rs
garbageslam: yep, dont want to make it any harder to debug all this than it has to be :)

transaction/core/src/tx_error.rs
```diff
@@ -65,6 +65,8 @@ pub enum NewMemoError {
     OutputsAfterChange,
     /// Changing the fee after the change output is not supported
     FeeAfterChange,
+    /// Mixed Token Ids is not supported in these memos
```
sugargoat, line 68:
```suggestion
    /// Mixed Token Ids are not supported in these memos
```

transaction/core/src/validation/validate.rs
```diff
@@ -1330,22 +1334,25 @@ mod tests {
         }
     }
 
+    // FIXME: This test needs to involve a Tx with more than one output to make
+    // sense
```
sugargoat, line 1338: 👍

sugargoat: Maybe worth capturing in a testing fast-follow-on ticket?

sugargoat: Maybe we should also consider what tests should be blocking for moving forward with the release?

transaction/std/src/memo_builder/rth_memo_builder.rs
```diff
@@ -143,16 +148,25 @@ impl MemoBuilder for RTHMemoBuilder {
...
-        value: u64,
+        amount: Amount,
         recipient: &PublicAddress,
         memo_context: MemoContext,
     ) -> Result<MemoPayload, NewMemoError> {
         if self.wrote_destination_memo {
             return Err(NewMemoError::OutputsAfterChange);
         }
+        // Check if the outlay is mixing token ids
+        if let Some(prev_token_id) = self.outlay_token_id {
```
sugargoat, line 159: I wonder if this should be checked in another method? It seems to be mixing the responsibility for `make_memo_for_output` with making sure that the outlay is well-formed?

transaction/std/src/transaction_builder.rs
```diff
@@ -81,49 +80,56 @@ impl<FPR: FogPubkeyResolver> TransactionBuilder<FPR> {
...
 
     /// Initializes a new TransactionBuilder, using a Box<dyn MemoBuilder>
     /// instead of statically typed
     ///
     /// # Arguments
+    /// * `block_version` - The block version rules to use when building this
+    ///   transaction
+    /// * `fee` - The fee (and token id) to use for this transaction. Note: The
+    ///   fee token id cannot be changed later, and before mixed transactions
+    ///   feature, every input and output must have this token id.
```
sugargoat, line 109:
```suggestion
    ///   feature, every input and output must have the same token id as the fee.
```

transaction/std/src/transaction_builder.rs
```diff
@@ -81,49 +80,56 @@ impl<FPR: FogPubkeyResolver> TransactionBuilder<FPR> {
...
+        fee: Amount,
         fog_resolver: FPR,
-        memo_builder: Box<dyn MemoBuilder + Send + Sync>,
-    ) -> Self {
-        TransactionBuilder {
+        mut memo_builder: Box<dyn MemoBuilder + Send + Sync>,
+    ) -> Result<Self, TxBuilderError> {
+        // make sure that the memo builder
+        // is initialized to the same fee as the transaction builder
+        memo_builder.set_fee(fee)?;
```
sugargoat, line 122: Should this be `self.set_fee(fee)?` - or rather, why does `self.set_fee` exist if the fee cannot be changed later, and we are calling `set_fee` directly from the `memo_builder` here?

(Still reading - so I may answer this question for myself)

transaction/std/src/memo_builder/mod.rs
```diff
@@ -30,20 +30,20 @@ pub trait MemoBuilder: Debug {
     /// and gets a chance to report an error, if the fee is too large, or if it
     /// is being changed too late
     /// in the process, and memos that are already written would be invalid.
-    fn set_fee(&mut self, value: u64) -> Result<(), NewMemoError>;
+    fn set_fee(&mut self, amount: Amount) -> Result<(), NewMemoError>;
```
sugargoat, line 33: Maybe this is also part of my question below (https://github.com/mobilecoinfoundation/mobilecoin/pull/1827/files#r859983201), why the `memo_builder` also has responsibility for the fee management. That may be out of scope for this PR, but I'm trying to understand the ways that the `TransactionBuilder` can fail, and the introduction of the `memo_builder`'s fee management for the `token_id`-specified fee has introduced a way for the `TransactionBuilder` to fail on `new`, which I'm curious about.

transaction/std/src/transaction_builder.rs
```diff
@@ -2590,4 +2693,211 @@ pub mod transaction_builder_tests {
...
+                fog_resolver.clone(),
+                memo_builder,
+            )
+            .unwrap();
+
+            let input_credentials =
+                get_input_credentials(block_version, amount1, &sender, &fog_resolver, &mut rng);
+            transaction_builder.add_input(input_credentials);
+
+            let input_credentials =
```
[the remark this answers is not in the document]
sugargoat: Oh, I see - it's fine to overwrite because you're adding a new input

transaction/std/src/transaction_builder.rs
```diff
@@ -2590,4 +2693,211 @@ pub mod transaction_builder_tests {
                 .is_err());
         }
     }
+
+    #[test]
+    // Test that sending mixed transactions works
```
sugargoat, line 2698:
```suggestion
    // Test that sending mixed transactions works
    // 
    // This test uses inputs of two different token IDs, paying the fee and creating change with
    // TokenID1, and "passing through" the second token ID with its full amount as output.
```

transaction/std/src/transaction_builder.rs
```diff
@@ -2590,4 +2693,211 @@ pub mod transaction_builder_tests {
...
+            );
+            assert!(
+                !subaddress_matches_tx_out(&sender, CHANGE_SUBADDRESS_INDEX, &tx_out2).unwrap()
+            );
+            assert!(
+                subaddress_matches_tx_out(&sender, CHANGE_SUBADDRESS_INDEX, &change_tx_out)
+                    .unwrap()
+            );
+
+            // Test that recipients's default subaddress owns the change, and not the other
```
sugargoat, line 2781:
```suggestion
            // Test that recipients's default subaddress owns the correct output, and not the other
```

mobilecoind-json/src/data_types.rs
sugargoat: I don't think you're out of the loop - just trying to figure out what our bookkeeping process is if we did decide to support it

mobilecoind/src/payments.rs
sugargoat: Ty!!

transaction/std/src/transaction_builder.rs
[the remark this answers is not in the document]
garbageslam: yeah the previous version is "moved" when we call `add_input`, and the previous binding no longer exists at this line

transaction/core/src/ring_signature/rct_bulletproofs.rs
sugargoat: SGTM - ty!

sugargoat (commented):
Awesome - this is looking great. I'm going to take one more pass on the `rct_bulletproofs` code and the `transaction_builder` changes, but very close to approval from my perspective!

transaction/std/src/memo_builder/mod.rs
garbageslam: the memo builder has to know the fee, because it has to get written into the destination memo. if the memo builder has the wrong fee, then the RTH data will be wrong.

So like, i've been thinking about this more, and I think that what happened here is, there is a tension:

* The way the transaction builder is currently architected, we actually build all `TxOut`'s at the time of calling `add_output`.
* The simplest way to add the encrypted memo stuff was, make memo encryption happen as part of `TxOut::new`. So that means each memo is finalized at time of `add_output` call, and not at the later `build` call.
* The memo builder is responsible for writing those memos. There are some "bad things" that can happen like, you could write the change output, and then try to change the fee. The way it works now is, the memo builder flags an error that gets returned by the `set_fee` call, and prevents the fee from changing and tells the user "you can't do that".

We could try to change it all so that, `TxOut::new` is not actually called until `.build()` is called, and instead we just store the data for `TxOut::new`. We could then give the memo builder all the final data at once -- all the inputs and `TxOut` amounts, the final fee value, etc. etc., and it could write all the memos for everything, which would simplify the memo builder.

I didn't do it that way before because it wasn't the simplest thing that could work, it would have been a bigger change to the `TransactionBuilder`. But we might be justified in doing a refactor like that, especially if the memo stuff gets more complicated, because that would probably remove the need for the memo builder to be stateful and update its state as outputs are written.

transaction/std/src/transaction_builder.rs
garbageslam: the fee value can be changed, because that was how Matt originally set it up, and that API is exposed and used, but the fee token id cannot be changed

transaction/core/src/validation/validate.rs
garbageslam: Yea i chatted with James and Eugene about this offline, they are planning a fix to land after this PR

IMO we should change the `transaction_core_test_utils::get_transaction` thing so that the test transaction is creates has a change output as well, since it's better test coverage, all of the real life transactions have a change output i think

transaction/std/src/memo_builder/rth_memo_builder.rs
garbageslam: So the idea here is, anything having to do with memos / memo policy is the job of the memo builder.

There is no reason we can't specify RTH to allow some types of mixed transactions -- in fact I think we plan to, and then this wouldn't be an error. But I think it's logically correct to say "today, the RTH scheme is not compatible with mixed transactions, and so it falls to the RTH Memo builder (and not the transaction builder, which is generic over the memo builder) to report that as an error. And then for a future version of RTH memo builder, this may not always be an error.

transaction/core/src/validation/validate.rs
garbageslam: this is now #1866

transaction/core/src/ring_signature/rct_bulletproofs.rs
garbageslam: (Note: if we did things Koe's way, so that token ids are secret even from the enclave, then there would be no way to prevent this (creating zero valued outputs for nonexisting token ids). So I'm not sure it's actually worth trying to prevent it.

I guess the only reason to try to prevent it is, if people start using e.g. u64::MAX, u64::MAX - 1, etc. as token ids, then we are on the hook for actually computing those generators, which costs elliptic curve operations in prod. It would be nicer if we could say "only the first N token ids are actually enabled right now, and we cache those generators and never compute generators dynamically" as a perf optimization.

api/proto/external.proto
garbageslam: I think Matt omitted all the doc strings on these initially, because no one is really using these structs directly, they only handle `Tx` usually, and these are just constituents inside that.

But maybe there are people who are making their own transaction builders or something who will use these structs directly.

transaction/core/src/ring_signature/rct_bulletproofs.rs
garbageslam: I guess the safest thing to do is make `consensus-enclave-impl` also check these extra token ids and prohibit them if they aren't known. If we decide we don't care about this, we can remove that check later. There's no legitimate use-case that could be broken by this I think.

eranrund: This happens implicitly, since for a token have TxOuts at all, it needs to be mintable - which would imply it is configured in `tokens.toml`.

transaction/std/src/memo_builder/mod.rs
eranrund: Some random thoughts:
1) Completely agree on the tension between when memos are created and when TxOuts are created. I think there is no clear winner.
2) It is nice that `add_output` returns the `TxOut` - for example you get to store the public key, in case you want to reference it later in association with whatever you did that made you call `add_output`. This could potentially be resolved by returning some `TxOutContext` or something like that.
3) I think the way we are handling the fee right now is just incorrect - every transaction built must have a fee, and defaulting to `Mob::MINIMUM_FEE` is a hack. Now that `TransactionBuilder::new` accepts a fee, why not remove `set_fee()`. That way the fee cannot change after initialization, which I think is perfectly fine.

eranrund (approved):
LGTM once all of @sugargoat 's comments are addressed.

transaction/core/src/ring_signature/rct_bulletproofs.rs
garbageslam: I think that's not actually right at this revision --

I think if I have a mixed transaction, and token id = u64::MAX, with no existing TxOut's of that token id on the chain, it would work for me to add an output with zero value and this token id to the transaction builder, to any other normal transaction, and the transaction would be balanced and well-formed (since the value of that TxOut is zero). So it's not exactly minting but the TxOut would exist in the chain, and could be added as an input (even though it has zero value).

transaction/std/src/memo_builder/mod.rs
garbageslam: (2) is a very good point, and I think in the future we will want even more stuff returned, for instance, there's this stuff about having a key that you can use to sign in a way that proves that you are the creator of a TxOut, that's in MCIP #31. And if we did the refactor I'm proposing then that would all get harder.
(3) The only use I can see for this is like, currently to get the minimum fees, clients make a network call to consensus. It might be helpful to the phones if they can make this call in parallel to building the transaction, since it takes like a second or so to actually build the transaction? and then they could update the fee only at the very end right before the change memo. Not sure that actually makes sense in real life, I think `.add_output` is really fast and only the `.build` step is slow.

We could make it so that, `TxOut::new` is called at `.add_output`, and we still return the confirmation number (and even the public key and stuff?), but in `.build`, the TransactionBuilder rewrites all the memos after everything about the transaction is known. But it's kind of messy.

Maybe the simplest thing is what you say, to just get rid of the `set_fee` option.

eranrund: Replacing the memos after we return a `TxOut` feels questionable to me. If someone later compares or hashes the entire `TxOut` they are going to have a bad time...

My vote is to remove the `set_fee` for now. As to the point of parallelizing, I think you could get the fee at the same time you are getting rings. Like you said, `add_output` is fast so waiting with the `add_output` calls doesn't seem problematic to me. And you can't add inputs until you have fetched rings.

garbageslam: > and the introduction of the memo_builder's fee management for the token_id-specified fee has introduced a way for the TransactionBuilder to fail on new, which I'm curious about.

@sugargoat the reason that set-fee is allowed to fail is, when we specified RTH memos, we decided that i was okay to only reserve 6 bytes for the fee amount, since it seems wildly unlikely that higher fees than that would ever be paid in real life. So the memo builder gets to say "Sorry, this fee is too large to be represented. You can still use the transaction builder if you want but you have to turn off RTH or use a different memo builder or something."

If set-fee is now moving into the `new` call, then this failure gets propagated from there.

garbageslam:
@jcape I added a follow-on issue #1869 with some thoughts about some of the issues you raised, lmk what you think

mobilecoind-json/src/data_types.rs
garbageslam: this is now #1867

transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -168,31 +247,111 @@ impl SignatureRctBulletproofs {
...
+        //
+        // Therefore, if someone submits a transaction where this sum is zero, and they
+        // cannot not know any "nontrivial" linear combination of G, H_1, ...
+        // then this must be a trivial linear combination.
+        // A trivial linear combination is one where all the coefficients are zero.
+        // This implies that w_j = 0 for all j, which implies that value was not created
+        // or destroyed in any token id, as required.
+        //
+        // So we don't need to do a separate loop here once per token id, we can just
+        // add everything together and check for zero.
```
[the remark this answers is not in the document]
isis-mc: Yep! Taking a look now.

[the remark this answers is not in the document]
isis-mc: This seems correct to me.

isis-mc (approved):

transaction/std/src/memo_builder/mod.rs
garbageslam: > My vote is to remove the set_fee for now. As to the point of parallelizing, I think you could get the fee at the same time you are getting rings. Like you said, add_output is fast so waiting with the add_output calls doesn't seem problematic to me. And you can't add inputs until you have fetched rings.

Okay, let's make a ticket to explore removing the set fee API. I'm not sure if any client is using it, but we can find out

sugargoat (approved):

PR #1833, opened 2022-04-21
Fix Clippy warnings across the repo

remoun:
… when running with `--all-targets`, so that we have a baseline without warnings for #1693

### DIFF
--- a/util/keyfile/tests/b58-length.rs
+++ b/util/keyfile/tests/b58-length.rs
@@ -15,12 +15,7 @@ fn test_b58pub_length<T: RngCore + CryptoRng>(
     num_trials: usize,
     rng: &mut T,
 ) -> bool {
-    let url = format!(
-        "fog://{}",
-        std::iter::repeat("a")
-            .take(domain_length)
-            .collect::<String>()
-    );
+    let url = format!("fog://{}", "a".repeat(domain_length));
     for _ in 0..num_trials {
         let root_id = RootIdentity::random_with_fog(
             rng,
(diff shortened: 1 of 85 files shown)


util/keyfile/tests/b58-length.rs
```diff
@@ -15,12 +15,7 @@ fn test_b58pub_length<T: RngCore + CryptoRng>(
     num_trials: usize,
     rng: &mut T,
 ) -> bool {
-    let url = format!(
-        "fog://{}",
-        std::iter::repeat("a")
-            .take(domain_length)
-            .collect::<String>()
-    );
+    let url = format!("fog://{}", "a".repeat(domain_length));
```
garbageslam, line 18: wow clippy's gotten pretty good

garbageslam (approved):

remoun:
> Nice. What changed that clippy is suddenly finding all of those? I thought it already runs as part of `./tools/lint.sh`?

The difference is passing `--all-targets` to clippy, which I noticed we're missing while working on CI stuff. I'll update tools/lint.sh to keep that going forward.

remoun:
> > Nice. What changed that clippy is suddenly finding all of those? I thought it already runs as part of `./tools/lint.sh`?
> 
> The difference is passing `--all-targets` to clippy, which I noticed we're missing while working on CI stuff. I'll update tools/lint.sh to keep that going forward.

The enclave workspaces error out with `--all-targets` (`duplicate lang item in crate ``std`` (which ``test`` depends on)`), and it seems to be working OK in https://github.com/mobilecoinfoundation/mobilecoin/pull/1693, so I'll skip that for now. Nick's SGX work will likely get us away from needing multiple workspaces.

PR #1841, opened 2022-04-22
add feature extended message digest

### LINKED ISSUE #1706: Should the extended-message toggle use a `BlockVersion::has_feature()` API?
Shouldn't this use the `BlockVersion::has_feature()` API?

_Originally posted by @jcape in https://github.com/mobilecoinfoundation/mobilecoin/pull/1700#discussion_r836646544_

In confidential token ids PR, we added this code:

```
/// Toggles between old-style and new-style extended message
fn compute_extended_message_either_version(
    block_version: BlockVersion,
    message: &[u8],
    range_proof_bytes: &[u8],
) -> Vec<u8> {
    if block_version >= BlockVersion::THREE {
        // New-style extended message using merlin
        digest_extended_message(message, pseudo_output_commitments, range_proof_bytes).to_vec()
    } else {
        // Old-style extended message
        extend_message(message, pseudo_output_commitments, range_proof_bytes)
    }
}
```

This is actually not directly related to confidential tokens implementation.

We wanted to change the way the extended message was being computed to follow best practices like domain separation and framing when signing a message.

Reviewers agreed that this is an improvement.

To avoid toggling explicitly on block version values, we could treat this as a feature `use_merlin_extended_message_digest` or something, which can happen at the same time as the token id change. We need to describe this in an MCIP, either MCIP 25 or another MCIP.

garbageslam:
this was requested in other PR reviews

closes #1706

### DIFF
--- a/transaction/core/src/blockchain/block_version.rs
+++ b/transaction/core/src/blockchain/block_version.rs
@@ -96,6 +96,12 @@ impl BlockVersion {
     pub fn mint_transactions_are_supported(&self) -> bool {
         self.0 >= 2
     }
+
+    /// The extended message digest is used when signing MLSAGs
+    /// in block version 2 and higher. This is described in MCIP #25.
+    pub fn mlsags_sign_extended_message_digest(&self) -> bool {
+        self.0 >= 2
+    }
 }
 
 impl Deref for BlockVersion {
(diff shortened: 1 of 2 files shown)


eranrund (approved):

remoun (approved):
LGTM. Thank you!

PR #1848, opened 2022-04-22
Development Env and Release CI/CD - Part 3

jgreat:
## Motivation
Breaking this work up into multiple parts to hopefully make this easier to review.

- Part 1: Mobilecoind python tests (should match what's in master now) Docker and helper scripts.
- Part 2: Helm charts
- Part 3: Github Actions and test wrappers.

The main goal for refactoring the release workflow was to enable block_version 0 -> block_version 1 testing.

1. Build rust and go binaries.
2. Build/Publish Docker images.
3. Build/Publish Helm Charts
4. Deploy "previous" release (v1.1.3)
5. Run integration tests against current release.
6. Upgrade to current release block_version=0.
7. Run integrations tests.
8. Upgrade consensus to block_version=1
9. Run integration tests.
 
CI/CD "improvements"

- Updated OS and utilities for build and runtime.
- Slimmed down build and runtime images.
- Using new build image will provide a static, versioned, verifiable and repeatable build process.
- pre-install of rust/cargo targets.
- Versioned helm charts published to S3 repo for each release tied to the versioned docker images.
- More consistent runtime environment setup through helm configuration sub-charts.
- GitHub Actions with private runners.
- Dynamically built dev environments for feature/* branches.
- Ability to retry failed steps.
- Skip build/ci with head commit messages.
- Build -dev release on release/*
- reduce chance of "shared" secret leaks, generate unique dev env secrets and keys.
- Manual ad-hoc dev environment actions: deploy, reset, delete, test

## In this PR

Add Github actions workflows to do block version upgrade testing.

### .github/workflows

**Naming**

Since we have multiple projects in the monorepo, I've done a simple namespacing of the workflows to try and keep things organized. Names follow this pattern. 

- `mobilecoin` - these are workflows for testing core mobilecoin funtions.
- `dev` - these workflows interact with CD for the dynamic "dev" environments used for integration testing. 
- `workflow|dispatch` - type of workflow.
- `<description>` 

The workflows here take advantage of workflow_call (reusable workflows) and workflow_dispatch (manual workflows) to keep things DRYish and allow us to run steps manually, like reset and redeploy to a dev namespace. In general the main workflows triggered by pushes call the "dispatch" workflows that and they use the "workflow" events.

Base workflows triggered by github events.

- `.github/workflows/mobilecoin-dev-cd.yaml` - Main workflow triggered on pushes, drives build and publishing of artifacts.
- `.github/workflows/mobilecoin-dev-delete.yaml` - Workflow run on branch delete events to clean up old dev environments.

Dispatch (manual) events.

- `.github/workflows/mobilecoin-dispatch-dev-delete.yaml` - Delete a dev environment (cluster namespace)
- `.github/workflows/mobilecoin-dispatch-dev-deploy.yaml` - Deploy a dev environment (apps to the cluster namespace)
- `.github/workflows/mobilecoin-dispatch-dev-reset.yaml` - Reset an existing dev environment (delete all data in a namespace)
- `.github/workflows/mobilecoin-dispatch-dev-test.yaml` - Run Integration tests (fog-distribution, fog-test-client, mobilecoind python)
- `.github/workflows/mobilecoin-dispatch-dev-update-consensus.yaml` - Update consensus config and restart pods.

Reusable Workflows.

- `.github/workflows/mobilecoin-workflow-dev-delete.yaml` - Delete a dev environment (cluster namespace)
- `.github/workflows/mobilecoin-workflow-dev-deploy.yaml` - Deploy a dev environment (apps to the cluster namespace)
- `.github/workflows/mobilecoin-workflow-dev-reset.yaml` - Reset an existing dev environment (delete all data in a namespace)
- `.github/workflows/mobilecoin-workflow-dev-test.yaml` - Run Integration tests (fog-distribution, fog-test-client, mobilecoind python)
- `.github/workflows/mobilecoin-workflow-dev-update-block-version.yaml` - Update consensus config and restart pods.

### Build image

The build is using a dockerhub based image that includes rust and all the usual targets pre-installed.  We reference this with the sha256 identifier so we can guarantee the image and build environment is consistent. I worked to create a minimal image needed to build our apps, but I'm sure we will need more tools as we integrate other CI actions.

This image is automatically built and published when we update https://github.com/mobilecoinofficial/docker-rust-sgx-base

### Caching

This build takes advantage of cargo and binary caching for rebuilds.  It generates the cache keys based on hashes of .proto, Cargo., .rs or .go files as appropriate. Cache should be shared with the default branch.

### Custom Actions

Most of the Kubernetes interactions are powered by a custom action here: https://github.com/mobilecoinofficial/gha-k8s-toolbox

### Artifacts

- Docker images will be published to the `mobilecoin` org.
- Helm charts will be published to the S3 repository.
- Binaries and measurements will be published as "artifacts" attached to the specific build.
- Environment access information can be found in the `👾 Environment Info 👾` job/step in the main CD workflow.

### Caveat :)

I _think_ I have all the required repo config set up, but there's a pretty good chance this will need a follow up or two to make everything functional. It was tested and working on the fork, but its really hard to test the CI/CD system without committing code.

## Future Work

- manual dispatch workflow to build artifacts for TestNet and MainNet deployments.
- refactor of entrypoint scripts to align internal deployment configuration with partner deployments.
- generate fog-report signing keys, instead of using shared key.


.github/workflows/mobilecoin-workflow-dev-delete.yaml
```diff
@@ -0,0 +1,62 @@
+# Copyright (c) 2018-2022 The MobileCoin Foundation
+#
+# MobileCoin Core projects - Reusable Workflow - Delete development namespace and associated components.
+
+name: mobilecoin-workflow-dev-delete
```
remoun, line 5: Basic question: Why do we need/have a `workflow_dispatch` and `workflow_call` variant of each workflow?

.github/workflows/mobilecoin-dispatch-dev-delete.yaml
```diff
@@ -0,0 +1,23 @@
...
+        description: "Target Namespace"
+        type: string
+        required: true
+
+jobs:
+  delete:
+    uses: ./.github/workflows/mobilecoin-workflow-dev-delete.yaml
+    with:
+      namespace: ${{ github.event.inputs.namespace }}
+    secrets:
```
MCrank, line 20: Do we not need to pass in the two `LEDGER_AWS_*` secrets to the mobilecoin-workflow-dev-delete.yaml?  They are marked as required.  I thought you could not access the repo secrets from a workflow_call unless they were passed in.

.github/workflows/mobilecoin-workflow-dev-update-block-version.yaml
```diff
@@ -0,0 +1,118 @@
...
+    inputs:
+      namespace:
+        description: "Target Namespace"
+        type: string
+        required: true
+      block_version:
+        description: "block_version"
+        type: string
+        required: true
+        default: '0'
```
MCrank, line 18: Do we want to set a default here?  Thought is that once we move off  version zero is there a potential that this does not get bumped and we deploy back to version 0?  Is that even possible or would it error out because the block version is lower than what is currently deployed?

MCrank (commented):
Looks  awesome.  A lot of moving parts in here but made sense to me.  Just a couple of minor questions.

.github/workflows/mobilecoin-dev-cd.yaml
```diff
@@ -0,0 +1,563 @@
...
+  build-cargo-hardware:
+    runs-on: [self-hosted, Linux, large]
+    container:
+      image: mobilecoin/rust-sgx-base@sha256:565e0b31ba40379098de222d63f13fee9fc3f7b72318743794f103f79e85fbca
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+
+    - name: Cache rust build binaries
+      if: "! contains(github.event.head_commit.message, '[skip build]')"
```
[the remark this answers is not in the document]
jgreat: If you skip a job, you then need to do a bunch of complicated if statements on each downstream job to see if the `needs` job skipped or failed.  And even then its super messy because you need to check the status and add needs for all of the upstream jobs not just the ones the current job is dependent on. 

Believe it or not I think this is way cleaner.

.github/workflows/mobilecoin-dispatch-dev-delete.yaml
jgreat: Yeah good catch.  The delete workflows have not been thoroughly tested yet.

.github/workflows/mobilecoin-workflow-dev-update-block-version.yaml
jgreat: Deploying back to a previous version would probably make a big mess. You're probably right, removing the version default is a good idea,

MCrank (approved):
LGTM

joekottke (approved):
Some minor naming nits, but nothing that I would require holding up the review.

.github/workflows/mobilecoin-dev-cd.yaml
```diff
@@ -0,0 +1,563 @@
...
+        apt install -y protobuf-compiler zstd
+
+    - name: Cache go build binaries
+      if: "! contains(github.event.head_commit.message, '[skip build]')"
+      id: go_artifact_cache
+      uses: actions/cache@v3
+      with:
+        path: |
+          go_build_artifacts
+        # Key is a hash of all the .rs, .proto and Cargo.toml files.
```
nick-mobilecoin, line 196: nit: copy pasted comment from the rust version
```suggestion
        # Key is a hash of all the .go, go.*, and .proto files.
```

.github/workflows/mobilecoin-dev-cd.yaml
```diff
@@ -0,0 +1,563 @@
...
+        steps.artifact_cache.outputs.cache-hit != 'true' &&
+        ! contains(github.event.head_commit.message, '[skip build]')
+      shell: bash
+      run: |
+        cd build_artifacts
+        for i in *.signed.so
+        do
+          css=$(echo -n "${i}" | sed -r 's/(.*)\.signed\.so/\1/')
+          sgx_sign dump -enclave "${i}" -dumpfile /dev/null -cssfile ${css}.css
+        done
```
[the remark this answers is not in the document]
jgreat: Yes, since the tools are installed we might as well create the .css files.  Just so the measurement files are readily available as an artifact for use in clients that might need them.

nick-mobilecoin (commented):
only reviewed up to (and including) `dev-reset` in `.github/workflows/mobilecoin-dev-cd-yaml`

.github/workflows/mobilecoin-dev-cd.yaml
```diff
@@ -0,0 +1,563 @@
+# Copyright (c) 2018-2022 The MobileCoin Foundation
```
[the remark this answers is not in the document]
jgreat: I see what you're saying here, but I think for convenience for future updates we should keep these headers the same as all the other files in the repo?

joekottke (approved):

.github/workflows/mobilecoin-dispatch-dev-test.yaml
```diff
@@ -0,0 +1,48 @@
...
+        required: false
+        default: true
+
+jobs:
+  test:
+    uses: ./.github/workflows/mobilecoin-workflow-dev-test.yaml
+    with:
+      namespace: ${{ github.event.inputs.namespace }}
+      ingest_color: ${{ github.event.inputs.ingest_color }}
+      fog_distribution: ${{ github.event.inputs.fog_distribution == 'true' && 'true' || 'false' }}
```
nick-mobilecoin, line 41: More a question than anything.
I found https://github.com/actions/runner/issues/1483
I'm guessing this approach is an alternative to quoting the entire expression like in, https://github.com/actions/runner/issues/1483#issuecomment-994986996

I get the `|| 'false'` output, but I'm not quite understanding how a `true && 'true'` will result in the string `'true'`, I would think it would still result in `true`

.github/workflows/mobilecoin-dispatch-dev-deploy.yaml
```diff
@@ -0,0 +1,47 @@
...
+        type: string
+        required: true
+      version:
+        description: "Chart Version"
+        type: string
+        required: true
+      docker_image_org:
+        description: "Docker Image Org"
+        type: string
+        required: true
```
nick-mobilecoin, line 21: I was going to ask why have `required` and `default`, but it looks like `required` is just a _communication_ mechanism between authors and users,  https://github.com/actions/runner/issues/1070

.github/workflows/mobilecoin-dispatch-dev-test.yaml
jgreat: I really wish the types worked correctly here, but I'm pretty sure things are just being compared inconstantly as strings in the background. 

Something gets lost in the translation in workflow_call if I just use `true`  instead of `'true'`.   Lots and lots of trail and error here.

.github/workflows/mobilecoin-dispatch-dev-deploy.yaml
jgreat: Well I thought that "required" did something, but maybe not :shrug:   

Although we set a default, we still want to make sure a user puts _something_ in this value. Having an empty docker image org value would break the workflow.

nick-mobilecoin (approved):

PR #1860, opened 2022-04-26
Block streaming prefactoring

remoun:
Prefactoring for #1823: Restructure crates to minimize the footprint of `mc-transaction-core` including IAS `VerificationReport` and SCP `QuorumSet`

While the enclave already depends on `VerificationReport`, we want to minimize the number of crates other clients have to pull in for `mc-transaction-core`, especially with the addition of `QuorumSet`


api/src/convert/verification_report.rs
```diff
@@ -3,15 +3,14 @@
 //! Convert to/from external::VerificationReport
 
 use crate::external;
-use mc_attest_core::{VerificationReport, VerificationSignature};
-use protobuf::RepeatedField;
+use mc_attest_verifier_types::{VerificationReport, VerificationSignature};
```
iamalwaysuncomfortable, line 6: To sanity check - the mc_attest_verifier_types crate is being created to make it easier for to access the AVR data right? What's the benefit of moving it?

common/src/hash.rs
```diff
@@ -0,0 +1,31 @@
...
+
+/// A HashMap that replaces the default hasher with an implementation that
+/// relies on mcrand for randomess.
+/// See [hashbrown::HashMap] and [HasherBuilder]
+pub type HashMap<K, V> = hashbrown::HashMap<K, V, HasherBuilder>;
+
+/// A HashSet that replaces the default hasher with an implementation that
+/// relies on mcrand for randomess.
+/// See [hashbrown::HashSet] and [HasherBuilder]
+pub type HashSet<K> = hashbrown::HashSet<K, HasherBuilder>;
```
iamalwaysuncomfortable, line 15: Eventually the the HasherBuilder should be re-reviewed. I believe it's using some hashing primitives which were made to be fast back in the day but there's been some evolution in the technology since then. I'll bring this us for re-evaluation.

common/src/hash.rs
```diff
@@ -0,0 +1,31 @@
...
+/// of the blocks in the blockchain, where you should be specific about what
+/// hash you are using.
+pub fn fast_hash(data: &[u8]) -> Hash {
+    use sha3::Digest;
+
+    let hash = sha3::Sha3_256::digest(data);
+    let mut output = [0u8; 32];
+    output.copy_from_slice(hash.as_slice());
+    output
+}
```
iamalwaysuncomfortable, line 31: @jcape here's where the hash function used in the U256 priority calculations in SCP is defined!

consensus/scp/core/src/core_types.rs
```diff
@@ -125,7 +125,7 @@ impl<V: Value> PartialOrd for Ballot<V> {
 // This makes debugging easier when looking at large ballots.
 impl<V: Value> fmt::Display for Ballot<V> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        let mut hasher = DefaultHasher::new();
+        let mut hasher = HasherBuilder::default().build_hasher();
```
iamalwaysuncomfortable, line 128: There's not specifically good reason to use HasherBuilder anywhere we can. It's uses specific qualities which are useful in the enclave but it's not necessarily the best default choice everywhere.

consensus/scp/core/src/msg.rs
```diff
@@ -629,9 +633,10 @@ impl<V: Value, ID: GenericNodeId> fmt::Display for Msg<V, ID> {
         };
 
         // Returns "<set.len, hash(set)>".
+        let hasher_builder = HasherBuilder::default();
         let format_b_tree_set = |b_tree_set: &BTreeSet<V>| {
             let hash = {
-                let mut hasher = DefaultHasher::new();
+                let mut hasher = hasher_builder.build_hasher();
```
iamalwaysuncomfortable, line 639: This needs a sanity check - there are an exceptional volume of msgs that get created - if we're changing this, there should be mindfulness towards the impact.

consensus/scp/core/src/quorum_set.rs
```diff
@@ -430,9 +429,26 @@ impl<ID: GenericNodeId + AsRef<ResponderId>> From<&QuorumSet<ID>> for QuorumSet<
 mod quorum_set_tests {
     use super::*;
     use crate::{core_types::*, msg::*, predicates::*, test_utils::test_node_id};
-    use mc_common::ResponderId;
-    use std::collections::hash_map::DefaultHasher;
+    use core::hash::{BuildHasher, Hash, Hasher};
+    use mc_common::HasherBuilder;
```
iamalwaysuncomfortable, line 433: Again nit - why is this being used?

iamalwaysuncomfortable (commented):
Asked a few questions about why the AVRs were moved to a separate crate - I'm familiar with the reasoning as to why but don't think I have the knowledge as to why.

Also, hasherbuilder shouldn't be blanket subsituted across the entire project - we should get some further review as to why it's being used/where it's most suitable. Siphash13 works well for small/medium sized messages, but there are better choices for different situations. I'd like to get an outside weigh in on what'd go best in these situations.

api/src/convert/verification_report.rs
remoun: James pointed out that `mc-attest-core` has a large footprint, and while it's already included in the enclave, it would also get included in all clients that use `mc-transaction-core` and by extension `mc-api`.

consensus/scp/core/src/core_types.rs
remoun: The main reason here is that `DefaultHasher` is not `no_std` compatible, and `HasherBuilder` is a compatible impl within reach. This is in `Display` so we don't have to maintain backwards compatibility.

consensus/scp/core/src/quorum_set.rs
remoun: Because `mc-consensus-scp-core` is `no_std` (because `mc-transaction-core` is `no_std`), and this is a compatible hasher builder within easy reach.

consensus/scp/core/src/msg.rs
remoun: Fair point. This is in `Display`, so it's fairly safe to change; if we're logging all messages we're in trouble anyway.

common/src/hash.rs
remoun: Sounds reasonable. Wanna file a tracking issue?

samdealy (approved):
The PR looks good. Overall, it would be helpful if you could add a PR description that provides (a) background and (b) a picture of how this PR fits into the string of PRs that follow. What does this prefactoring establish, etc.?

common/src/hash.rs
iamalwaysuncomfortable: Here we are! https://github.com/mobilecoinfoundation/mobilecoin/issues/1937

api/src/convert/verification_report.rs
iamalwaysuncomfortable: Should we perhaps also create a separate PR that puts this into main as well?

iamalwaysuncomfortable (approved):

api/src/convert/verification_report.rs
remoun: It seems OK to defer that till we merge this feature branch. LMK if you disagree.

PR #1863, opened 2022-04-26
Add Clippy check to lint script

remoun:
Add `clippy --all-targets`. Supports #1693

### DIFF
--- a/tools/lint.sh
+++ b/tools/lint.sh
@@ -8,6 +8,10 @@ if [[ ! -z "$1" ]]; then
     cd "$1"
 fi
 
+# We want to check with --all-targets since it checks test code, but that flag
+# leads to build errors in enclave workspaces, so check it here.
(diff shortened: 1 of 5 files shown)


tools/lint.sh
```diff
@@ -8,6 +8,9 @@ if [[ ! -z "$1" ]]; then
     cd "$1"
 fi
 
+# Do an initial pass with --all-targets
+cargo clippy --all --all-features --all-targets
```
nick-mobilecoin, line 12: This feels like a narration comment re-iterating _what_ but not addressing _why_
Why is there an initial pass with `--all-targets` instead of including `--all-targets` in the loop for each workspace?

nick-mobilecoin (changes requested):

nick-mobilecoin (approved):

PR #1871, opened 2022-04-27
Feature/signed contingent inputs

garbageslam:
This is the second step of MCIP 31

This is a pretty large diff, says +3500 lines, but more than half of it is test code, which seems good.

This:
* Adds a `SignedContingentInput` object as envisioned in the MCIP
* Adds a verification function to check the signature in the signed contingent input (and tests of this)
* Adds a `SignedContingentInputBuilder` object similar to the `TransactionBuilder`, which takes input credentials and builds a signed contingent input.
* Makes the transaction builder able to take `SignedContingentInput`'s as well as normal `InputCredentials`, and build Tx's using these.
* Add lots of testing around:
  * What happens if the person who builds the SCI is a fog user
  * What happens if there are multiple SCI's in the transaction
  * What happens if the inputs aren't balanced
  * What happens if someone doesn't follow the input rules
  * What happens if someone tries to remove the input rules
  * What if happens if someone tries to change the input rules
  * Testing of the tx core validation methods
  * Testing of round tripping of the new protobuf payloads


transaction/std/src/input_materials.rs
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    /// Get the amount for whichever type of input this is
+    pub fn amount(&self) -> Amount {
+        match self {
+            InputMaterials::Signable(cred) => cred.input_secret.amount,
+            InputMaterials::Presigned(input) => Amount {
+                value: input.pseudo_output_amount.value,
+                token_id: input.pseudo_output_amount.token_id.into(),
+            },
```
remoun, line 54: Would it be useful if we `impl From<UnmaskedAmount> for Amount`?

transaction/std/src/signed_contingent_input_builder.rs
```diff
@@ -0,0 +1,2272 @@
...
+    /// # Arguments
+    /// * `block_version` - The block version rules to use when signing the
+    ///   input
+    /// * `input_credentials` - Credentials for the input we are signing
+    /// * `tx_out_global_indices` - Global indices for the tx out's in the ring
+    /// * `fog_resolver` - Source of validated fog keys to use with this
+    ///   transaction
+    /// * `memo_builder` - An object which creates memos for the TxOuts in this
+    ///   transaction
+    pub fn new<MB: MemoBuilder + 'static + Send + Sync>(
```
remoun, line 66: OOC, why is `'static` required here but not in `new_with_box`?

transaction/std/src/signed_contingent_input_builder.rs
```diff
@@ -0,0 +1,2272 @@
...
+        Ok((tx_out, confirmation))
+    }
+
+    /// Sets the tombstone block, clamping to smallest pubkey expiry value.
+    ///
+    /// # Arguments
+    /// * `tombstone_block` - Tombstone block number.
+    pub fn set_tombstone_block(&mut self, tombstone_block: u64) -> u64 {
+        self.tombstone_block = min(tombstone_block, self.fog_tombstone_block_limit);
+        self.tombstone_block
```
remoun, line 274: IIUC we can just return the result of the assignment
```suggestion
        self.tombstone_block = min(tombstone_block, self.fog_tombstone_block_limit)
```

transaction/std/src/signed_contingent_input_builder.rs
```diff
@@ -0,0 +1,2272 @@
...
+                        .default_subaddress()
+                        .fog_report_url()
+                        .unwrap()
+                        .to_string()
+                =>
+                    FullyValidatedFogPubkey {
+                        pubkey: RistrettoPublic::from(&ingest_private_key2),
+                        pubkey_expiry: 1500,
+                    },
+            });
```
remoun, line 540: This pattern seems more readable with `from_iter`

```suggestion
            let fog_resolver = MockFogResolver(BTreeMap::from_iter([
                (
                    sender
                        .default_subaddress()
                        .fog_report_url()
                        .unwrap()
                        .to_string(),
                    FullyValidatedFogPubkey {
                        pubkey: RistrettoPublic::from(&ingest_private_key1),
                        pubkey_expiry: 1000,
                    },
                ),
                (
                    recipient
                        .default_subaddress()
                        .fog_report_url()
                        .unwrap()
                        .to_string(),
                    FullyValidatedFogPubkey {
                        pubkey: RistrettoPublic::from(&ingest_private_key2),
                        pubkey_expiry: 1500,
                    },
                ),
            ]));
```

transaction/std/src/transaction_builder.rs
```diff
@@ -50,10 +53,10 @@ impl TxOutputsOrdering for DefaultTxOutputsOrdering {
 pub struct TransactionBuilder<FPR: FogPubkeyResolver> {
     /// The block version that we are targeting for this transaction
     block_version: BlockVersion,
-    /// The input credentials used to form the transaction
-    input_credentials: Vec<InputCredentials>,
-    /// The outputs created by the transaction, and associated shared secrets
-    outputs_and_shared_secrets: Vec<(TxOut, RistrettoPublic)>,
+    /// The input material used to form the transaction
```
remoun, line 56:
```suggestion
    /// The input materials used to form the transaction.
```

transaction/std/src/transaction_builder.rs
```diff
@@ -50,10 +53,10 @@ impl TxOutputsOrdering for DefaultTxOutputsOrdering {
...
+    input_materials: Vec<InputMaterials>,
+    /// The outputs created by the transaction, and associated output secret
```
remoun, line 58:
```suggestion
    /// The outputs created by the transaction, and associated output secrets.
```

transaction/std/src/transaction_builder.rs
```diff
@@ -139,7 +142,72 @@ impl<FPR: FogPubkeyResolver> TransactionBuilder<FPR> {
...
+    ///
+    /// Note: Before adding a signed_contingent_input, you probably want to:
+    /// * validate it (call .validate())
+    /// * check if key image appreared already (call .key_image())
+    /// * provide merkle proofs of membership for each ring member (see
+    ///   .tx_out_global_indices)
+    ///
+    /// # Arguments
+    /// * `signed_contingent_input` - The pre-signed input we are adding
+    pub fn add_presigned_input_raw(&mut self, sci: SignedContingentInput) {
```
remoun, line 209: Does this need to be `pub`?

transaction/std/src/input_materials.rs
```diff
@@ -0,0 +1,87 @@
...
+        match self {
+            InputMaterials::Signable(cred) => cred.ring.len(),
+            InputMaterials::Presigned(input) => input.tx_in.ring.len(),
+        }
+    }
+}
+
+// Helper which converts from InputMaterials (TransactionBuilder type) to
+// InputRing (rct_bulletproofs type)
+impl From<InputMaterials> for InputRing {
```
remoun, line 69: Would it make sense to make this `From<&InputMaterials>` (by ref) to match the impl below?

remoun (commented):
LGTM overall, just some nitpicks

transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -367,9 +429,23 @@ impl SignatureRctBulletproofs {
...
         // Each MLSAG must be valid.
         for (i, ring) in rings.iter().enumerate() {
+            let rules_digest = ring
+                .input_rules
+                .as_ref()
+                .map(|rules| rules.digest())
+                .unwrap_or_default();
+
+            // If there are input rules, then the signature is over the rules digest
+            // If not, then it is the entire extended message digest
```
eranrund, line 439: Can you please elaborate on why this is the case?

eranrund: I am guessing because thats how the ring MLSAG was generated when it was presigned, but I am wondering if it shouldn't include more than just the rules?

transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -390,21 +466,18 @@ impl SignatureRctBulletproofs {
 /// # Arguments
 /// * `block_version` - This may influence details of the signature
 /// * `message` - The messages to be signed, e.g. Hash(TxPrefix).
-/// * `rings` - One or more rings of one-time addresses and amount commitments.
-/// * `real_input_indices` - The index of the real input in each ring.
-/// * `input_secrets` - Input secret for each real input.
-/// * `output_values_and_blindings` - Output secret for each output amount
-///   commitment.
+/// * `rings` - One or more rings of one-time addresses and amount commitments,
```
eranrund, line 469: Nit
```suggestion
/// * `rings` - One or more rings of one-time addresses and amount commitments.
```

transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -390,21 +466,18 @@ impl SignatureRctBulletproofs {
...
+///   with secrets for the real input
+/// * `presigned_rings` - Zero or more pre-signed input rings
```
eranrund, line 471: This doesn't actually exist

transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -431,57 +504,67 @@ fn sign_with_balance_check<CSPRNG: RngCore + CryptoRng>(
...
+    }
+
     if rings.is_empty() {
         return Err(Error::NoInputs);
     }
-    let num_inputs = rings.len();
-    let ring_size = rings[0].len();
+    let ring_size = rings
+        .iter()
+        .filter_map(|ring| {
```
eranrund, line 529: Nit: `.filter_map().next()` is `.find_map()`.

transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -1202,16 +1382,15 @@ mod rct_bulletproofs_tests {
 
             // Duplicate one of the rings.
             params.rings[2] = params.rings[3].clone();
-            params.output_secrets[2] = params.output_secrets[3].clone();
-            params.input_secrets[2] = params.input_secrets[3].clone();
-            params.real_input_indices[2] = params.real_input_indices[3];
+            // Duplicate corresponding the output also, so we don't get "value not conserved" error
```
eranrund, line 1385:
```suggestion
            // Duplicate the corresponding output also, so we don't get "value not conserved" error
```

transaction/std/src/signed_contingent_input_builder.rs
```diff
@@ -0,0 +1,2272 @@
...
+/// and attaching fog hint and memos as appropriate.
+///
+/// This is generic over FogPubkeyResolver because there are several reasonable
+/// implementations of that.
+///
+/// This is generic over MemoBuilder to allow injecting a policy for how to
+/// use the memos in the TxOuts.
+#[derive(Debug)]
+pub struct SignedContingentInputBuilder<FPR: FogPubkeyResolver> {
+    /// The block version that we are targeting for this transaction
```
eranrund, line 28:
```suggestion
    /// The block version that we are targeting for this input
```

This is making me wonder about whether a contingent input should include a block version in it...
Might be useful to have that information in case there are some contingent inputs floating around, and we switch to a newer block version that makes the old ones incompatible. Right now I speculate that this would manifest as weird signature errors or such, depending on where the incompatibility comes from.

transaction/std/src/signed_contingent_input_builder.rs
```diff
@@ -0,0 +1,2272 @@
...
+    block_version: BlockVersion,
+    /// The input material used to form the transaction
+    input_credentials: InputCredentials,
+    /// Global indices for the tx out's in the ring
+    tx_out_global_indices: Vec<u64>,
+    /// The outputs created by the transaction, and associated secrets
+    outputs_and_secrets: Vec<(TxOut, OutputSecret)>,
+    /// The tombstone_block value, a block index in which the transaction
```
eranrund, line 36: Nit: I think its more correct that the contingent input is no longer usable?

transaction/std/src/signed_contingent_input_builder.rs
```diff
@@ -0,0 +1,2272 @@
...
+    /// expires, and can no longer be added to the blockchain
+    tombstone_block: u64,
+    /// The source of validated fog pubkeys used for this transaction
+    fog_resolver: FPR,
+    /// The limit on the tombstone block value imposed pubkey_expiry values in
+    /// fog pubkeys used so far
+    fog_tombstone_block_limit: u64,
+    /// An policy object implementing MemoBuilder which constructs memos for
```
eranrund, line 44:
```suggestion
    /// A policy object implementing MemoBuilder which constructs memos for
```

transaction/std/src/signed_contingent_input_builder.rs
eranrund: I believe you need `'static` here so that you could `Box::new` the object. What I think `'static` implies when used as a trait bound is that the object does not contain any fields with references that have non-`'static ` lifetimes, or to put it differently - it is safe to move the object around knowing that it does not point at anything that could potentially go away.
From reading https://doc.rust-lang.org/reference/lifetime-elision.html#default-trait-object-lifetimes I am  getting that `'static` is implied in a `Box<dyn ...>` so it is not explicitly required to be specified, due to `dyn` implying `'static` when an explicit lifetime is not specified.

transaction/std/src/signed_contingent_input_builder.rs
```diff
@@ -0,0 +1,2272 @@
...
+            } else {
+                self.tombstone_block
+            },
+        };
+
+        // Now we can create the mlsag
+        let mut tx_in = TxIn::from(&self.input_credentials);
+        tx_in.input_rules = Some(input_rules.clone());
+        let mut ring = SignableInputRing::from(self.input_credentials);
+        ring.input_rules = Some(input_rules.clone());
```
eranrund, line 326: Does `ring` (and by association `ring.input_rules`) actually get used anywhere here? It looks like the code is only reading specific fields of `ring`.

transaction/std/src/signed_contingent_input_builder.rs
```diff
@@ -0,0 +1,2272 @@
...
+    /// * `input_credentials` - Credentials for the input we are signing
+    /// * `tx_out_global_indices` - Global indices for the tx out's in the ring
+    /// * `fog_resolver` - Source of validated fog keys to use with this signed
+    ///   contingent input
+    /// * `memo_builder` - An object which creates memos for the TxOuts in this
+    ///   signed contingent input
+    pub fn new_with_box(
+        block_version: BlockVersion,
+        input_credentials: InputCredentials,
+        tx_out_global_indices: Vec<u64>,
```
eranrund, line 96: Might be worth validating that `tx_out_global_indices.len() == input_credentials.ring.len()`.

eranrund (commented):
I went over it and left a few comments/questions. Mostly looks great to me, and really nice test coverage for the `SignedContingentInputBuilder`.
I do wonder about the similarities it has with `TransactionBuilder`, mostly having to duplicate a bunch of stuff from it.

Since they are in the same crate perhaps `SignedContingentInputBuilder` could internally use `TransactionBuilder` (it seems like it is mostly a `TransactionBuilder` with a single input and a different `build()` method).

This also makes me wonder about the fee in the destination memo. It appears nothing is going to call `set_fee` on the `memo_builder`, which I think is going to result in `add_change_output` misbehaving if `RTHMemoBuilder` is used for example.

garbageslam:
More possible action items:
- [ ]: Think about if signed contingent input should sign more than the input rules
- [ ]: Enforce in the verifier that at least one ring is not presigned
- [ ]: Share code better between signed contingent input builder and transaction builder

api/proto/external.proto
```diff
@@ -226,6 +226,20 @@ message TxIn {
...
+}
+
+// Rules enforced on a transaction by a signed input within it (MCIP #31)
+message InputRules {
+    // Outputs required to appear in the TxPrefix for the Tx to be valid
+    repeated TxOut required_outputs = 1;
+
+    // A maximum value which the tombstone block for the Tx cannot exceed
+    //
+    // A value of zero here means no limit is enforced
```
samdealy, line 241: Could this make use of proto's `optional` field? https://developers.google.com/protocol-buffers/docs/proto#optional. 

Might be clearer than using `0` to signify "not set".

api/src/convert/tx.rs
```diff
@@ -51,69 +50,144 @@ mod tests {
...
-                    (charlie.default_subaddress(), 65536),
-                ];
-
-                mc_transaction_core_test_utils::get_outputs(
-                    block_version,
-                    &recipient_and_amounts,
-                    &mut rng,
-                )
-            };
+            let fpr = MockFogResolver::default();
```
samdealy, line 54: nit: Can you use full words here to increase readability.? I'm not sure what this acronym stands for,

transaction/core/src/input_rules.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    /// An upper bound on the tombstone block which must be respected for the
+    /// transaction to be valid
+    #[prost(fixed64, tag = "2")]
+    pub max_tombstone_block: u64,
+}
+
+impl InputRules {
+    /// Verify that a Tx conforms to the rules.
+    pub fn verify(&self, _block_version: BlockVersion, tx: &Tx) -> Result<(), InputRuleError> {
```
samdealy, line 41: optional: In the future, we may have many InputRules and this method could get bloated. We may consider having an `InputRule` trait that has a `verify` method, and then the InputRules struct just contains a Vec of these and goes through each input rule and calls `verify`. 

This can be implemented in the future if we find it necessary!

transaction/core/src/input_rules.rs
```diff
@@ -0,0 +1,68 @@
...
+    #[prost(message, repeated, tag = "1")]
+    pub required_outputs: Vec<TxOut>,
+
+    /// An upper bound on the tombstone block which must be respected for the
+    /// transaction to be valid
+    #[prost(fixed64, tag = "2")]
+    pub max_tombstone_block: u64,
+}
+
+impl InputRules {
```
samdealy, line 39: Are there tests for these input rules? Would be good to have simple unit tests verifying failure cases if not.

transaction/core/src/validation/validate.rs
```diff
@@ -86,6 +85,12 @@ pub fn validate<R: RngCore + CryptoRng>(
         validate_outputs_are_sorted(&tx.prefix)?;
     }
 
+    if block_version.signed_input_rules_are_supported() {
+        validate_all_input_rules(block_version, tx)?;
+    } else {
+        validate_no_input_rules_exist(tx)?;
```
samdealy, line 91: Consider renaming this method. I read it as "validate zero input rules" with an "exist" tacked on at the end. Maybe something like "validate_when_no_input_rules_exist" OR "validate_input_rules_not_supported"

transaction/std/src/input_materials.rs
```diff
@@ -0,0 +1,87 @@
+// Copyright (c) 2022 The MobileCoin Foundation
+
+//! Input Materials is a helper struct for the transaction builder.
+//! The transaction builder can get inputs either from input credentials,
+//! or signed contingent inputs. In one case the input is being signed by us,
```
samdealy, line 5: nit: specify who "us" refers to. "Us" == the user building the transaction?

transaction/std/src/transaction_builder.rs
```diff
@@ -139,7 +142,72 @@ impl<FPR: FogPubkeyResolver> TransactionBuilder<FPR> {
...
+    ///
+    /// Note: Before adding a signed_contingent_input, you probably want to:
+    /// * validate it (call .validate())
+    /// * check if key image appreared already (call .key_image())
+    /// * provide merkle proofs of membership for each ring member (see
+    ///   .tx_out_global_indices)
+    ///
+    /// # Arguments
+    /// * `signed_contingent_input` - The pre-signed input we are adding
+    pub fn add_presigned_input(
```
samdealy, line 161: Tying into my comment above re: creating an InputRule trait for reach individual rule- it could perhaps be used here as well. This would make adding a new input rule easier for another engineer. Right now, you'd have to know to add to the InputRules verify method then add the verification logic here as well.

samdealy (approved):

api/proto/external.proto
eranrund: We are using `proto3` where the `optional` keyword has been deprecated. All fields in proto land are optional.

transaction/core/src/signed_contingent_input.rs
```diff
@@ -0,0 +1,171 @@
...
+/// motivation.
+#[derive(Clone, Eq, Message, PartialEq)]
+pub struct SignedContingentInput {
+    /// The tx_in which was signed over
+    #[prost(message, required, tag = "1")]
+    pub tx_in: TxIn,
+
+    /// The Ring MLSAG signature, conferring spending authority
+    #[prost(message, required, tag = "2")]
+    pub mlsag: RingMLSAG,
```
iamalwaysuncomfortable, line 44: When a signed contingent input is sent to the consensus service - how would consensus treat it? Does it get included into the actual block structure which can then be scanned?

iamalwaysuncomfortable (commented):
Left a few discussion comments - my main questions are:
- What were the cryptographic concerns here, what are we potentially afraid could be a vulnerability here?
- When this gets submitted to the consensus service, does this go into a block? Like how people on a dex search for this? Would this create a situation where there's a lot of crawlers on our chain?
- Least important: floated ideas about how the plaintext orders might be blinded, or situations where they might be optionally blinded.

api/proto/external.proto
```diff
@@ -416,3 +430,34 @@ message ValidatedMintConfigTx {
...
+// The amount and blinding factor of a TxOut
+message UnmaskedAmount {
+    // The value of the amount commitment
+    fixed64 value = 1;
+
+    // The token_id of the amount commitment
+    fixed64 token_id = 2;
+
+    // The blinding factor of the amount commitment
+    CurveScalar blinding = 3;
```
wjuan-mob, line 443: Do we even need a blinding factor here? If these are going to be unblinded amounts anyway, could we not just have a convention that for unmasked amounts we intentionally choose a blinding factor of 0?

api/proto/external.proto
```diff
@@ -226,6 +226,20 @@ message TxIn {
...
     repeated TxOutMembershipProof proofs = 2;
+
+    // Any rules specified by the signed input
+    InputRules input_rules = 3;
+}
+
+// Rules enforced on a transaction by a signed input within it (MCIP #31)
+message InputRules {
+    // Outputs required to appear in the TxPrefix for the Tx to be valid
+    repeated TxOut required_outputs = 1;
```
wjuan-mob, line 237: I am wondering because we reveal so much information about these particular TxOuts, whether that will have an impact on the network, or if these txouts will be penalized for spending because of their traceability.

api/src/convert/tx.rs
```diff
@@ -51,69 +50,144 @@ mod tests {
...
+                .add_output(
+                    Amount::new(1000 * MILLIMOB_TO_PICOMOB, Mob::ID),
+                    &charlie.default_subaddress(),
+                    &mut rng,
+                )
+                .unwrap();
+
+            let sci = sci_builder.build(&mut rng).unwrap();
+
+            // Alice sends this token2 amount to Bob from Charlie, paying Charlie 1000 MOB
```
wjuan-mob, line 156: Nit: Alice pays Charlie only 1.475 Mob and getting change for .475 Mob. Alice is paying Bob 1000 Token2.

api/src/convert/tx_in.rs
```diff
@@ -40,7 +44,53 @@ impl TryFrom<&external::TxIn> for tx::TxIn {
             proofs.push(tx_proof);
         }
 
-        let tx_in = tx::TxIn { ring, proofs };
+        let input_rules = source
+            .input_rules
+            .as_ref()
+            .map(InputRules::try_from)
+            .transpose()?;
```
wjuan-mob, line 51: Why do we need the transpose here?

api/tests/prost.rs
```diff
@@ -36,6 +43,99 @@ fn root_identity_examples<T: RngCore + CryptoRng>(rng: &mut T) -> Vec<RootIdenti
...
+        vec![0u64, 1u64],
+        fpr.clone(),
+        EmptyMemoBuilder::default(),
+    );
+    builder
+        .add_output(Amount::new(400, 0.into()), &recipient, rng)
+        .unwrap();
+    result.push(builder.build(rng).unwrap());
+
+    let input_credentials = get_input_credentials(
```
wjuan-mob, line 89: Nit: This seems like a duplicate of line 70. Do we need to define this again? Since builder is mut I think we should just be able to set it?

transaction/core/src/signed_contingent_input.rs
```diff
@@ -0,0 +1,171 @@
...
+    #[prost(fixed64, tag = "1")]
+    pub value: u64,
+
+    /// The token id of the amount commitment
+    #[prost(fixed64, tag = "2")]
+    pub token_id: u64,
+
+    /// The blinding factor of the amount commitment
+    #[prost(message, required, tag = "3")]
+    pub blinding: CurveScalar,
```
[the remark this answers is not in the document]
wjuan-mob: It's used in line 101. When the code goes through and validates the blinding aligns with the commitments in input_rules. I am not sure why we need a blinding factor though, since if we're publishing these unmasking fields, wouldn't it be equivalently unblinded?

transaction/std/src/signed_contingent_input_builder.rs
wjuan-mob: I feel like the contingent input should try to be block version agnostic. It seems like an undesireable behaviour for a contract to be invalidated by a backend change to the block version.

transaction/std/src/signed_contingent_input_builder.rs
```diff
@@ -0,0 +1,2272 @@
...
+        // Put the memo builder back
+        self.memo_builder = Some(mb);
+        result
+    }
+
+    /// Add an output to the transaction, using `fog_hint_address` to construct
+    /// the fog hint.
+    ///
+    /// This is a private implementation detail, and generally, fog users expect
+    /// that the transactions that they recieve from fog belong to the account
```
wjuan-mob, line 218:
```suggestion
    /// that the transactions that they receive from fog belong to the account
```

transaction/std/src/signed_contingent_input_builder.rs
```diff
@@ -0,0 +1,2272 @@
...
+    }
+
+    #[test]
+    // Test that if you add a signed contingent input, but don't add any of your own
+    // input credentials, it fails with "AllRingsPresigned".
+    // (This is expected because (1) there would be no reason for someone to make
+    // signed inputs like this, and (2) there needs to be at least one
+    // pseudo-output where the transaction builder can choose a blinding factor
+    // for it.)
+    fn test_contingent_input_rules_no_input_credentials_doesnt_work() {
```
wjuan-mob, line 1386: Theoretically, couldn't this happen naturally by the market maker? E.g. Alice and Bob both make signed contingent inputs that match each other. The market maker just pairs them together and all the rings are presigned and everyone is happy?

transaction/std/src/signed_contingent_input_builder.rs
```diff
@@ -0,0 +1,2272 @@
...
+                validate_all_input_rules(block_version, &tx),
+                Err(TransactionValidationError::InputRule(
+                    InputRuleError::MissingRequiredOutput
+                ))
+            );
+        }
+    }
+
+    #[test]
+    // Test that if you delete the required ouptut from a signed contingent input,
```
wjuan-mob, line 1796:
```suggestion
    // Test that if you delete the required output from a signed contingent input,
```

transaction/core/src/signed_contingent_input.rs
garbageslam: So the idea here is, you can't send this directly to consensus, you can only send a full Transaction to consensus, and we aren't actually changing the consensus service here. So the only thing you can do with this is use it to build a full transaction and then submit that.

transaction/std/src/signed_contingent_input_builder.rs
wjuan-mob: Maybe we should expect it to work. It might actually be a sensible use case that we could create a transaction using only presigned outputs if they do overlap and in that case the blinding factor seems unnecessary. I guess there is probably a concern about who is paying the fee in that situation.

api/src/convert/tx_in.rs
garbageslam: transpose is a thing that turns `Option<Result<T>>` into `Result<Option<T>>`, if I do `.transpose()?` it has the effect of pulling an error case out of `Option<Result<T>>` and returning it in a fairly succinct way.

wjuan-mob: Oh that's a neat trick. Thank you. I should use that.

transaction/core/src/signed_contingent_input.rs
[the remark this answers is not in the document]
wjuan-mob: Don't we publish the unmasking amount fields though? If the unmasking amount contains the blinding factor, don't we have to assume that the information is public?

transaction/std/src/signed_contingent_input_builder.rs
garbageslam: that is a really good point, thank you

api/src/convert/tx.rs
garbageslam: yeah sure -- in a lot of test code right now `fpr` is `FogPubkeyResolver`, that's the trait that the `TransactionBuilder` requires when you want to make a new transaction

transaction/core/src/signed_contingent_input.rs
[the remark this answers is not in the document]
iamalwaysuncomfortable: I think Chris is saying since the blinding factor is used only to blind the amount, and we're revealing the amount, there's no harm in revealing that blinding factor.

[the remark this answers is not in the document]
wjuan-mob: Right, my point is that if we are revealing the blinding factor, we could just have a convention for these that the blinding factor is 0 and exclude it from calculations.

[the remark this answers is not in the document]
iamalwaysuncomfortable: I'm not all the way through the mlsag & bulletproof math, but when you go through those, making the curve scalar zero would make any arbitrary counterparty amount work as valid when it's run through the proofs.

transaction/std/src/signed_contingent_input_builder.rs
garbageslam: So the only way I can see to permit this is, stop requiring that the sum of blinding factors is zero, instead the Tx contains a scalar which is the sum of blinding factors. Then the balance equation is changed to subtract that. Then this could work.

But it's not clear if it should work, because then we can no longer say that at least one MLSAG signs the whole tx prefix. So in this case the tx would be malleable I think which seems bad.

transaction/std/src/transaction_builder.rs
garbageslam: So, potentially in this revision it could just be `pub(crate)`.

But in the future it may have to be `pub`: for instance, one feature that was requested was to have rules that could be satisfied in more than one way, for instance like "I will trade my MOB for 100 of this or 200 of that". In that case, the `add_presigned_input` approach won't really be able to work because it won't know which way you want to satisfy the rule. Right now I think the simplest thing to do for that would be to say that, you can add it using the `_raw` version and then have like an `add_output_raw` so that you add the outputs that satisfy it the way you want to.

Brian was separately requesting an `add_output_raw` for use in the hardware wallet project. I'm not sure if we still need that but it's at least conceivable that it's useful in cases like that.

transaction/std/src/signed_contingent_input_builder.rs
garbageslam: interesting -- yeah maybe there's no good reason to use `maplit`.

We are using `maplit` in many places in the test code now, maybe we can replace all uses of that as you suggest. But let's do it in a separate ticket.

transaction/core/src/validation/validate.rs
```diff
@@ -1375,4 +1393,50 @@ mod tests {
             )
         }
     }
+
+    // Test that input rules validation is working
+    #[test]
+    fn test_input_rules_validation() {
+        let block_version = BlockVersion::THREE;
+
+        for _ in 0..3 {
```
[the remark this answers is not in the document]
garbageslam: yeah that's wierd -- there should be some randomness in here somewhere that would make this more reasonable. but it seems that there isn't. i'll remove the loop i guess

transaction/std/src/input_materials.rs
garbageslam: So like, I am somewhat torn about this. We only need the `From<InputMaterials>` in real life, and coding it that way avoids some unnecessary copies.

There are other places in our code where we choose to implement `From<&_>` instead. But I wonder if it's not more idiomatic to implement `From<_>` and then clone before converting only if necessary. We consistently do `From<_>` conversions whenever error objects are being converted, because otherwise the `?` operator doesn't work.

I'm inclined to only implement the version I actually need, and also, to make it possible for developers to avoid unnecessary copies / dynamic memory allocations, as a rule of thumb.

transaction/core/src/validation/validate.rs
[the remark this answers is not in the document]
garbageslam: This is a pretty good question.

It is obviously not very good to reveal the amount of a TxOut given that our general goal is not to reveal that.

However:
* In any sort of "proposal to trade", the counterparty who accepts the trade necessarily has some insight into what they traded, and also, what outputs were created by the Tx that they submitted. So as a practical matter I don't know how we can avoid this.
* Assuming that is unavoidable, this scheme still doesn't reveal the public address of who is receiving the amount.
* This scheme also doesn't reveal who submitted the offer to trade (the originator) to the counter-party, or at least, it doesn't necessarily do so.
* The TxOut would only be "penalized for spending" if it was obvious when someone was spending it. But by design our network makes it very difficult to figure out if any TxOut has been spent. No one can link the key image back to the TxOut unless they have the spend private key.

So I think, the "penalized for spending" thing is not real, and this is likely the best we can do for now. If we had MobZero then we could also avoid revealing the input rings of the true input -- that's the most feasible way that I can see that we can improve this.

You might imagine that there's a system where like, instead of sending your trade proposal to a network of counterparties, you could propose it to some kind of zero-knowledge smart contract. But right now I'm not sure I believe that that is real. I don't believe right now that e.g. you can have a DEX on chain that doesn't expose enough state to be able to infer what it's treasury is, and therefore also infer individual changes in its treasury value. I don't think an off-chain SGX thing works either because SGX doesn't really allow for durable state, all the state in an enclave has to be designed to be discardable. But the DEX private keys are obviously not discardable.

So right now the only thing that makes sense to me is like, having a network which makes it difficult to infer who submitted a particular message, and accepting that messages will reveal the amounts of some TxOut's. Happy to be wrong.

api/proto/external.proto
garbageslam: This is a pretty good question.

It is obviously not very good to reveal the amount of a TxOut given that our general goal is not to reveal that.

However:
* In any sort of "proposal to trade", the counterparty who accepts the trade necessarily has some insight into what they traded, and also, what outputs were created by the Tx that they submitted. So as a practical matter I don't know how we can avoid this.
* Assuming that is unavoidable, this scheme still doesn't reveal the public address of who is receiving the amount.
* This scheme also doesn't reveal who submitted the offer to trade (the originator) to the counter-party, or at least, it doesn't necessarily do so.
* The TxOut would only be "penalized for spending" if it was obvious when someone was spending it. But by design our network makes it very difficult to figure out if any TxOut has been spent. No one can link the key image back to the TxOut unless they have the spend private key.

So I think, the "penalized for spending" thing is not real, and this is likely the best we can do for now. If we had MobZero then we could also avoid revealing the input rings of the true input -- that's the most feasible way that I can see that we can improve this.

You might imagine that there's a system where like, instead of sending your trade proposal to a network of counterparties, you could propose it to some kind of zero-knowledge smart contract. But right now I'm not sure I believe that that is real. I don't believe right now that e.g. you can have a DEX on chain that doesn't expose enough state to be able to infer what it's treasury is, and therefore also infer individual changes in its treasury value. I don't think an off-chain SGX thing works either because SGX doesn't really allow for durable state, all the state in an enclave has to be designed to be discardable. But the DEX private keys are obviously not discardable.

So right now the only thing that makes sense to me is like, having a network which makes it difficult to infer who submitted a particular message, and accepting that messages will reveal the amounts of some TxOut's. Happy to be wrong.

api/tests/prost.rs
garbageslam: the builder is consumed when we call `.build`: the `.build` function takes `self` and not `&self` or `&mut self`. So the builder is gone at this line and if we try to refer to it it's just a syntax error.

api/proto/external.proto
garbageslam: Yeah, we do need a blinding factor here, because, the recipient of the `TxOut` may not know that it is unblinded -- they need to be able to recover their balance just from the blockchain. Currently, the blinding factor is derived as a hash of the TxOut shared secret, which is a random value.

We could try to adopt a convention that sometimes the blinding factor is zero, and so view key scanning would have to be modified so that it always tries both the derived value and zero. But this seems bad because it would make view key scanning twice as slow. It seems simpler to just keep using the blinding factor as it is working today.

transaction/core/src/signed_contingent_input.rs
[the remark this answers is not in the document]
garbageslam: I think this question is answered adequately here: https://github.com/mobilecoinfoundation/mobilecoin/pull/1871#discussion_r867258634

api/proto/external.proto
garbageslam: But also, for things like integers, "optional" in proto3 means "defaults to zero" and the default cannot be adjusted. This is something we cannot change if we want to use proto3

transaction/std/src/signed_contingent_input_builder.rs
garbageslam: > I feel like the contingent input should try to be block version agnostic. It seems like an undesireable behaviour for a contract to be invalidated by a backend change to the block version.

Yeah for sure, but it may be unavoidable. For instance, in block version 2 vs. block version 3 we changed the hash that was being signed by the MLSAGs. There's no way a contingent input could be compatible across a change like that.

We may be able to make the transaction builder smart though, like, if no changes that would impact singed contignent inputs happen between block version 3 and block version 4, then maybe the transaction builder can still accept signed contingent inputs at block version 3 even when it's building for block version 4.

I feel like it can't hurt anything to put a block version in the signed contingent input payload -- the `Tx` proto still doesn't itself have a block version, and the transaction builder can make a decision if it thinks it will work or not, before dropping the block version value from the input.

transaction/core/src/validation/validate.rs
garbageslam: the other functions in this file are named like `validate_no_memos_exist` etc. but most of these were named by me.

i could change it to like `validate_that_no_memos_exist` ? does that seem better?

transaction/core/src/input_rules.rs
garbageslam: yeah -- i would opt to do that later, just so that we can see how it evolves

but i will definitely add a comment suggesting to maybe split this up if it gets too big in the future

transaction/core/src/ring_signature/rct_bulletproofs.rs
garbageslam: i mean the short answer is that it's by design of the MCIP -- i'd be inclined to just add a code comment like "For more information about the rationale see discussion in the MCIP"

transaction/std/src/input_materials.rs
garbageslam: yeah that seems like a good idea

transaction/std/src/signed_contingent_input_builder.rs
garbageslam: in my understanding, this works in C but not in rust -- the result of assignment is `()`. willing to be wrong

transaction/std/src/input_materials.rs
garbageslam: I think that `TryFrom<&>` makes total sense though, because usually if the conversion is invalid you don't want to consume the argument.

transaction/std/src/signed_contingent_input_builder.rs
garbageslam: yeah, it doesn't build that way. i think this is the reference:

https://doc.rust-lang.org/reference/expressions/operator-expr.html#basic-assignments

I think they got rid of this because they felt that in C and C++ this creates too much cryptic code, like the

```
a = ++b = c++
```
etc.

transaction/std/src/signed_contingent_input_builder.rs
garbageslam: this is #1942

transaction/std/src/transaction_builder.rs
garbageslam: I think it's okay for engineers to have to know that if they add a new feature, they have to adjust multiple places in the code. Right now if you add a new transaction validation rule, you have to modify the validation module, and also the builder to ensure that the builder makes things that satisfy the rule. Additionally, it's possible that not all the rules can be enforced in isolation, they may tie into each other somehow, and then a trait probably wouldn't work.

Also, it's not clear if this API is actually going to work in general, it has been proposed that we should have rules that are like "or" rules that can be satisfied in several possible ways, potentially numerous. Then the builder will not be able to infer which way you have in mind, so you would have to use `add_presigned_input_raw` for that, and then some kind of `add_output_raw` function as well (which doesn't exist yet). Or we would need some other more helpful API here.
So given that we might be headed in that direction I'm not sure if a trait is going to work out.

transaction/core/src/input_rules.rs
garbageslam: Yeah -- there are the validation tests which cover this, but we should also test this API I guess

api/src/convert/tx.rs
garbageslam: i'm going to hold off on this in this PR because a lot of test code currently uses `fpr`. but we can look to improve that in a subsequent PR

transaction/std/src/input_materials.rs
```diff
@@ -0,0 +1,87 @@
...
+        }
+    }
+
+    /// Get the amount for whichever type of input this is
+    pub fn amount(&self) -> Amount {
+        match self {
+            InputMaterials::Signable(cred) => cred.input_secret.amount,
+            InputMaterials::Presigned(input) => Amount {
+                value: input.pseudo_output_amount.value,
+                token_id: input.pseudo_output_amount.token_id.into(),
```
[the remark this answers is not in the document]
garbageslam: > Does it make sense to make a contingent input optionally blinded if we find a useful case for it?

I think, we should figure out the use-case for this before writing code that allows it and does it, because extra possibilities makes the code more complex.

Right now the only thing that a signed contingent input can be used for is to add it to the transaction builder. There is no way that an MLSAG can be added to a transaction which will be valid if we don't know the value, token id, and blinding factor of the pseudo-output that it produces.

Moreover, if the counter party has no insight into the value and token id, then there's no way that they can build a balanced transaction. So it's hard to see how any system could work that conceals this.

If the value and token id are known, then there doesn't seem to be anything lost by also revealing the blinding factor, since concealing those is its purpose.

transaction/core/src/tx.rs
```diff
@@ -237,6 +243,44 @@ pub struct TxIn {
...
+    /// This is the digest of the TxIn which is signed per MCIP #31 if there
+    /// are input rules present.
+    ///
+    /// See MCIP #31 for rationale -- by not signing the whole TxPrefix, we
+    /// allow that someone can create this signature who does not have the
+    /// whole TxPrefix.
+    ///
+    /// The membership proofs are not signed, because it is useful to allow that
+    /// someone later may update those proofs. See MCIP #31 for discussion.
+    pub fn signed_digest(&self) -> Option<[u8; 32]> {
```
garbageslam, line 262: calling this out for reviewers:

in the latest revision, this is the digest that is signed, and the `input_rules.digest()` thing went away.

the idea is, sign the entire `TxIn`, but remove the merkle proofs first. The reason for this is that there are tx-wide rules about merkle proofs having matching trust roots, so it may be necessary for the recipient of a signed input to update these proofs without breaking the signature.

transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -50,6 +101,22 @@ pub struct OutputSecret {
...
+    /// member we have only:
+    /// * The onetime-address (tx_out.target_key)
+    /// * The compressed commitment (tx_out.amount.commitment)
+    pub members: Vec<(CompressedRistrettoPublic, CompressedCommitment)>,
+
+    /// The digest that this signature is supposed to have signed.
+    /// If omitted, then the entire extended message digest should have been
+    /// signed. If the input has rules, then a reduced digest is signed
+    /// instead. (See MCIP #31 for rationale)
+    pub signed_digest: Option<[u8; 32]>,
```
garbageslam, line 117: this was changed since last review from `Option<InputRule>` to `Option<[u8; 32]>`, so, if the input signs something other than the extended message digest, the thing it is supposed to sign instead goes here. That makes it easier to do a digest of the whole `TxIn` without bringing the whole `TxIn` into this scope.

garbageslam:
@eranrund So I'm looking more at how to share code with the transaction builder, and what I'm thinking is that, this object is ultimately going to grow a lot if we add more features to the signed contingent inputs (like partial fills, or the "or rules"). I'm worried that the cost of putting all that interface and internal state into the transaction builder is that it will complicate the transaction builder, which numerous clients are likely to need to use, while this object may be less widely used.

I think maybe there could be like an "output creator" object which is shared by these two objects, which has responsibility for the fog pubkey resolver, the fee, the tombstone block, and the memos. However I'm worried about increasing the scope of this PR significantly. So I'd like to make a ticket to follow up on that. LMK if you agree with that.

eranrund:
> @eranrund So I'm looking more at how to share code with the transaction builder, and what I'm thinking is that, this object is ultimately going to grow a lot if we add more features to the signed contingent inputs (like partial fills, or the "or rules"). I'm worried that the cost of putting all that interface and internal state into the transaction builder is that it will complicate the transaction builder, which numerous clients are likely to need to use, while this object may be less widely used.
> 
> I think maybe there could be like an "output creator" object which is shared by these two objects, which has responsibility for the fog pubkey resolver, the fee, the tombstone block, and the memos. However I'm worried about increasing the scope of this PR significantly. So I'd like to make a ticket to follow up on that. LMK if you agree with that.

I'm okay with this getting addressed at a followup PR in the near future. I remain concerned that it is very likely we'll forget to change stuff in two places when for example memo or change logic change, which is why I think this is important. But yeah, no need to make this particular PR any bigger.

transaction/core/src/validation/validate.rs
```diff
@@ -461,6 +455,29 @@ pub fn validate_tombstone(
...
+) -> TransactionValidationResult<()> {
+    for input in tx.prefix.inputs.iter() {
+        if let Some(rules) = input.input_rules.as_ref() {
+            rules.verify(block_version, tx)?;
+        }
+    }
+    Ok(())
+}
+
+/// Any input rules imposed on the Tx must satisfied
```
eranrund, line 471:
```suggestion
/// Validate that no input have input rules
```

transaction/std/src/input_credentials.rs
```diff
@@ -78,21 +80,51 @@ impl InputCredentials {
...
+            input_rules: None,
+        }
+    }
+}
+
 impl Zeroize for InputCredentials {
     fn zeroize(&mut self) {
-        self.onetime_private_key.zeroize();
-        self.view_private_key.zeroize();
+        self.real_index.zeroize();
```
eranrund, line 127: Shouldn't `input_secret` be zeroized as well?

eranrund (approved):

transaction/std/src/input_credentials.rs
garbageslam: yeah probably all of this should be zeroized, there was some build issue when i was looking at it though, so i skipped it. will look at it again

transaction/core/src/tx.rs
garbageslam: I believe that this is the largest thing here that could possibly be signed, everything else in the signed contingent input is either implied, or is supposed to be secret from the consensus enclave

iamalwaysuncomfortable (approved):
This is great. we should probably spend a round with an outside auditor to check that any weird edge cases can't be exploited here.

transaction/core/src/tx.rs
iamalwaysuncomfortable: Yeah that's a good point that the trust roots will change when the matching contingent inputs are combined.

api/proto/external.proto
```diff
@@ -416,3 +430,37 @@ message ValidatedMintConfigTx {
...
+
+    // The amount and blinding of the pseudo-output of the MLSAG
+    UnmaskedAmount pseudo_output_amount = 4;
+
+    /// The amount and blinding of any TxOut required by the input rules
+    repeated UnmaskedAmount required_output_amounts = 5;
+
+    /// The tx_out global index of each ring member
+    /// This helps the recipient of this payload construct proofs of membership for the ring
+    repeated fixed64 tx_out_global_indices = 6;
```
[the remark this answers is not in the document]
garbageslam: so, the global index refers to the number of the `TxOut` in the blockchain, ignoring all block structure. So the very first `TxOut` is 0, the next is 1, etc. So these numbers are fixed for each `TxOut` and never change.

There are database functions that allow looking these numbers up and searching for TxOut's using them, this is in the LedgerDB. Also, the Merkle tree infrastructure uses these numbers to organize the merkle tree. This is also basically why fog uses these numbers

iamalwaysuncomfortable:
This commit is so awesome btw, that it deserves the "Pyramid Of Preparing For Victory"

```stl
solid cube_corner
  facet normal 0.0 -1.0 0.0
    outer loop
      vertex 0.0 0.0 0.0
      vertex 1.0 0.0 0.0
      vertex 0.0 0.0 1.0
    endloop
  endfacet
  facet normal 0.0 0.0 -1.0
    outer loop
      vertex 0.0 0.0 0.0
      vertex 0.0 1.0 0.0
      vertex 1.0 0.0 0.0
    endloop
  endfacet
  facet normal -1.0 0.0 0.0
    outer loop
      vertex 0.0 0.0 0.0
      vertex 0.0 0.0 1.0
      vertex 0.0 1.0 0.0
    endloop
  endfacet
  facet normal 0.577 0.577 0.577
    outer loop
      vertex 1.0 0.0 0.0
      vertex 0.0 1.0 0.0
      vertex 0.0 0.0 1.0
    endloop
  endfacet
endsolid
```

isis-mc (commented):
:+1:

PR #1906, opened 2022-05-02
Create ResponderId out of host and port in PollingNetworkState

### LINKED ISSUE #848: Update mobilecoind to work with load balanced consensus networks
## Summary

Update or replace [`PollingNetworkState`](https://github.com/mobilecoinfoundation/mobilecoin/blob/9741017957ad28bd82cb22e3e7cb187f0f018ee6/ledger/sync/src/network_state/polling_network_state.rs#L35) in ledger sync and mobilecoind with something that will not dedupe nodes based on `ResponderId`.

## Background Information

The current situation is this: Consensus Service nodes nominally support operating with their client-facing connections operating behind a load balancer, by setting the `--client-responder-id` config option, e.g. `consenus-service --listen-uri=mc://mynode1234.foo.com --client-responder-id=loadbalancer.foo.com`.

As an aside, `ResponderId` is simply a string that is baked into the key exchange process---both the client and server (consensus service) inject this string as the Noise Framework prologue, providing a very minimal amount of hostname checking (this feature is more useful in enclave-to-enclave communications, where `ResponderId` also contains a hash of configuration details---i.e. nodes must agree on critical config for their enclaves to communicate---but we don't diverge this feature in our kex implementation for simplicity sake).

At any rate, this `loadbalancer.foo.com` is intended to support mobile clients who are wishing to have a smaller/more reliable subset of URLs to send transaction proposals to (as server-side load balancing is always going to be superior to the multiple round-trips required to do client-side load balancing). A mobile client in this scenario would simply send their TX proposals to `mc://loadbalancer.foo.com` and it would work, regardless of which actual `mynodeXXXX` host was the real destination.

That said, thick clients such as `mobilecoind` and `full-service` need to replicate the SCP network state when synchronizing a given block, which they do by polling something that looks like our existing mainnet full-mesh quorum set, e.g.:

```
./mobilecoind \
    --peer mc://node1.foo.com \
    --peer mc://node2.foo.com \
    --peer mc://node3.foo.com \
    --peer mc://node4.foo.com \
    --peer mc://node5.foo.com \
    --quorum-set '{ "threshold": 3, "members": [{"args":"node1.foo.com:443","type":"Node"},/* ... */}]}'
```

When run in this way, `mobilecoind` will construct a fake SCP instance, and get results from as many nodes as it can, and then run SCP. If the threshold of "3" (which is, sadly, actually 4 nodes) is met, then the block will be accepted as "read". This all works fine, until you want to stick a load balancer (for mobile clients) in front of these nodes, ala: 

```
./mobilecoind \
    --peer mc://node1.foo.com/?responder-id=loadbalancer.foo.com \
    --peer mc://node2.foo.com/?responder-id=loadbalancer.foo.com \
    --peer mc://node3.foo.com/?responder-id=loadbalancer.foo.com \
    --peer mc://node4.foo.com/?responder-id=loadbalancer.foo.com \
    --peer mc://node5.foo.com/?responder-id=loadbalancer.foo.com \
    --quorum-set '{ "threshold": 3, "members": [{"args":"node1.foo.com:443","type":"Node"},/* ... */}]}'
```

Inside `mobilecoind`, the `PollingNetworkState` uses [`SCPNetworkState`](https://github.com/mobilecoinfoundation/mobilecoin/blob/9741017957ad28bd82cb22e3e7cb187f0f018ee6/ledger/sync/src/network_state/scp_network_state.rs#L18) parameterized by `ResponderId`, which is pulled from the URLs list that's being polled, which means it will de-duplicate "votes" (really, externalized blocks) based on `ResponderId` -- so you end up with five duplicate "yes" votes, which is collapsed to one "yes" vote. Unfortunately, `1 > 4`, so this can never succeed.

## Suggested Fixes

The likely fix, then, is to either make `PollingNetworkState` use `SCPNetworkState<NodeId>` instead, and fill in node ID from the `host:port`. An alternative fix would be to hack around this issue, and make everything use  `host:port` as the `ResponderId`, regardless of any actual responder ID configured.

samdealy:
### Motivation
See #848. 

### In this PR
* Creates and uses `host_and_port_responder_id` method.
* Updates mobilecoind test configuration to test this new feature 

Fixes #848

### DIFF
--- a/connection/src/manager.rs
+++ b/connection/src/manager.rs
@@ -40,12 +40,13 @@ impl<C: Connection> ConnectionManager<C> {
                     .into_iter()
                     .map(|conn| {
                         let name = conn.to_string();
-                        let responder_id = conn.uri().host_and_port_responder_id().unwrap_or_else(|_| {
-                            panic!(
-                                "Could not create responder_id from {:?}",
-                                conn.uri().to_string()
-                            )
-                        });
+                        let responder_id =
+                            conn.uri().host_and_port_responder_id().unwrap_or_else(|_| {
+                                panic!(
+                                    "Could not create responder_id from {:?}",
+                                    conn.uri().to_string()
+                                )
+                            });
                         let sync_conn =
                             SyncConnection::new(conn, logger.new(o!("mc.peers.peer_name" => name)));
                         (responder_id, sync_conn)
--- a/connection/src/manager.rs
+++ b/connection/src/manager.rs
@@ -40,7 +40,7 @@ impl<C: Connection> ConnectionManager<C> {
                     .into_iter()
                     .map(|conn| {
                         let name = conn.to_string();
-                        let responder_id = conn.uri().responder_id().unwrap_or_else(|_| {
+                        let responder_id = conn.uri().host_and_port_responder_id().unwrap_or_else(|_| {
                             panic!(
                                 "Could not create responder_id from {:?}",
                                 conn.uri().to_string()
--- a/connection/src/manager.rs
+++ b/connection/src/manager.rs
@@ -41,12 +41,15 @@ impl<C: Connection> ConnectionManager<C> {
                     .map(|conn| {
                         let name = conn.to_string();
                         let responder_id =
-                            conn.uri().host_and_port_responder_id().unwrap_or_else(|_| {
-                                panic!(
-                                    "Could not create responder_id from {:?}",
-                                    conn.uri().to_string()
-                                )
-                            });
+                            conn.uri()
+                                .host_and_port_responder_id()
+                                .unwrap_or_else(|err| {
+                                    panic!(
+                                        "Could not create responder_id from {:?}: {}",
+                                        conn.uri().to_string(),
+                                        err
+                                    )
+                                });
                         let sync_conn =
                             SyncConnection::new(conn, logger.new(o!("mc.peers.peer_name" => name)));
                         (responder_id, sync_conn)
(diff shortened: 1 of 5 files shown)


connection/src/manager.rs
```diff
@@ -40,12 +40,13 @@ impl<C: Connection> ConnectionManager<C> {
...
-                                conn.uri().to_string()
-                            )
-                        });
+                        let responder_id =
+                            conn.uri().host_and_port_responder_id().unwrap_or_else(|_| {
+                                panic!(
+                                    "Could not create responder_id from {:?}",
+                                    conn.uri().to_string()
+                                )
+                            });
```
remoun, line 49: Let's include the error 
```suggestion
                            conn.uri().host_and_port_responder_id().unwrap_or_else(|err| {
                                panic!(
                                    "Could not create responder_id from {:?}: {}",
                                    conn.uri().to_string(),
                                    err
                                )
                            });
```

samdealy: Done.

iamalwaysuncomfortable (approved):
This looks decent as a bridge. What tech debt is standing in the way of refactoring Consensus to use the NodeID?

samdealy:
> This looks decent as a bridge. What tech debt is standing in the way of refactoring Consensus to use the NodeID?

Lots of code in other directories requires ScpNetworkState to use a ResponderId. Refactoring it to use NodeId would be a huge undertaking.

PR #1909, opened 2022-05-03
port a select subset of the changes from mixed transactions for candidate 1.2

garbageslam:
The changes ported here, from (https://github.com/mobilecoinfoundation/mobilecoin/pull/1827) are:
* Rename `token_id` to `fee_token_id` in `TxPrefix`. This avoids a breaking
  change between this and 1.3 in the hashes.
* Make `TransactionBuilder::new` take a fee amount instead of a fee token id
  This also changes te memo builder trait, so that it will be compatible with
  the burn redemption memo builder which we hope to port next.
  This also makes `TransactionBuilder::new` possibly return an error.
* Adapt all clients and sdks for `TransactionBuilder::new` changes.
* Adapt all tests for this change
* Bring the  `Amount::new` function which makes the test code nicer.

This does NOT port:
* Any other changes to transaction builder or tx protos
* The change that makes `add_output` take an amount instead of a `value: u64`. That is needed for mixed transactions but isn't needed here, and requires more changes to the clients and sdks, so we avoid it in this branch.

### DIFF
--- a/transaction/std/src/transaction_builder.rs
+++ b/transaction/std/src/transaction_builder.rs
@@ -18,7 +18,6 @@ use mc_transaction_core::{
     tokens::Mob,
     tx::{Tx, TxIn, TxOut, TxOutConfirmationNumber, TxPrefix},
     Amount, BlockVersion, CompressedCommitment, MemoContext, MemoPayload, NewMemoError, Token,
-    TokenId,
 };
 use mc_util_from_random::FromRandom;
 use rand_core::{CryptoRng, RngCore};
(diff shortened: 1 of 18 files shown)


transaction/std/src/transaction_builder.rs
```diff
@@ -316,7 +322,12 @@ impl<FPR: FogPubkeyResolver> TransactionBuilder<FPR> {
 
     /// Gets the transaction fee.
     pub fn get_fee(&self) -> u64 {
-        self.fee
+        self.fee.value
+    }
+
+    /// Gets the transaction fee token id.
+    pub fn get_fee_token_id(&self) -> TokenId {
```
remoun, line 329: Can we afford to make `get_fee` return `Amount`? I'm a bit concerned about people not noticing/using this method.

remoun (approved):

transaction/std/src/transaction_builder.rs
garbageslam: yeah we could probably do that

garbageslam: done

eranrund (approved):

PR #1917, opened 2022-05-03
Remove bigint in favor of primitive-types.

jcape:
### Motivation

The `bigint` crate has not been maintained for a long time, we shouldn't use it.

### In this PR
* Update SCP crate to calculate weights and round hashes using primitive-types instead of bigint.


eranrund (approved):

remoun (approved):

consensus/scp/src/slot.rs
```diff
@@ -494,10 +496,11 @@ impl<V: Value, ValidationError: Display> Slot<V, ValidationError> {
...
             let (num, denom) = self.weight(node_id);
-            let mut tmp = bigint::U512::from(bigint::U256::max_value());
-            tmp = tmp.saturating_mul(bigint::U512::from(num));
-            tmp = tmp.overflowing_div(bigint::U512::from(denom)).0;
-            let weight256 = bigint::U256::from(tmp);
+            let mut tmp = U512::from(U256::max_value());
+            tmp = tmp.saturating_mul(U512::from(num));
+            tmp = tmp / U512::from(denom);
+            let weight256 = U256::try_from(tmp)
+                .expect("failure calculating weight (max_u256 * k -> 2^512) / n");
```
iamalwaysuncomfortable, line 503: Good changeout for the crate, but why are does it use such huge numbers here?

iamalwaysuncomfortable (approved):

consensus/scp/src/slot.rs
```diff
@@ -494,10 +496,11 @@ impl<V: Value, ValidationError: Display> Slot<V, ValidationError> {
             // weight256 is the node's weight, scaled to 0..<max uint256>
             // (weight256 = <max uint256> * <num> / <denom>)
             let (num, denom) = self.weight(node_id);
-            let mut tmp = bigint::U512::from(bigint::U256::max_value());
-            tmp = tmp.saturating_mul(bigint::U512::from(num));
-            tmp = tmp.overflowing_div(bigint::U512::from(denom)).0;
-            let weight256 = bigint::U256::from(tmp);
+            let mut tmp = U512::from(U256::max_value());
+            tmp = tmp.saturating_mul(U512::from(num));
+            tmp = tmp / U512::from(denom);
```
iamalwaysuncomfortable, line 501:
```suggestion
            tmp /= U512::from(denom);
```

iamalwaysuncomfortable (approved):

consensus/scp/src/slot.rs
jcape: I don't know for certain but the function comment refers to "p.20 of the [Whitepaper](https://www.stellar.org/papers/stellar-consensus-protocol.pdf)," which describes how nodes are supposed to compute what I assume is intended as a deterministic random priority (using a SHA256 hash in the paper, Keccak in our case) for each of their neighbors.

iamalwaysuncomfortable: I'm sure it's correct, but I'm going to take a review of that.

iamalwaysuncomfortable: Neat, I'll take a review of that. I'm sure it's correct.

consensus/scp/src/slot.rs
```diff
@@ -494,10 +496,11 @@ impl<V: Value, ValidationError: Display> Slot<V, ValidationError> {
...
             // (weight256 = <max uint256> * <num> / <denom>)
             let (num, denom) = self.weight(node_id);
-            let mut tmp = bigint::U512::from(bigint::U256::max_value());
-            tmp = tmp.saturating_mul(bigint::U512::from(num));
-            tmp = tmp.overflowing_div(bigint::U512::from(denom)).0;
-            let weight256 = bigint::U256::from(tmp);
+            let mut tmp = U512::from(U256::max_value());
+            tmp = tmp.saturating_mul(U512::from(num));
+            tmp /= U512::from(denom);
+            let weight256 = U256::try_from(tmp)
```
remoun, line 502: How about we use `U256::full_mul`? It's description says `Multiplies two 256-bit integers to produce full 512-bit integer. Overflow is not possible.`

```suggestion
            let weight512: U512 = U256::max_value().full_mul(num.into()) / denom;
            let weight256 = U256::try_from(weight512)
```

remoun (approved):

consensus/scp/src/slot.rs
iamalwaysuncomfortable: From my look at the underlying crate full mull is an overflowing multiplication instead of a saturating multiplication which is not what we want here.

remoun: It's not overflowing because it puts the U256 operands in U512's before multiplying, and `2^256 * 2^256 == 2^512`

jcape: I'm not sure how I feel about this change. I appreciate it's less code, but the code that's there now is resilient to, e.g. changes in `self.weight()` that would make `num` or `denom` bigger than `U256::max_value()`.

remoun: I'm not sure what you mean. `weight()` returns `(u32, u32)`, which is a long way away from overflowing a `U256`. FWIW  `U256::full_mul()` only accepts `U256`, and `U256::from` is only implemented for narrower/smaller types.

With that said, I'm not attached to making this change.

PR #1921, opened 2022-05-04
Add cargo-sort to the pre-commit hook

jcape:
### Motivation

Our current CI uses cargo-sort to ensure our Cargo.toml dependencies are sorted. This adds a check for this to the pre-commit githook.

### In this PR
- Update precommit hook to install/use cargo sort.
- Redirect sort output to dev null.
- Do version checks first, then do tests, pipe some junk to /dev/null.
- Don't print "OK" for rustfmt check, version is sufficient.

### DIFF
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -85,9 +85,7 @@ fi
 
 printf "${PREFIX} Checking for rustfmt ... "
 cargo fmt --version
-if [ $? == 0 ]; then
-	printf "${SUCCESS}\n"
-else
+if [ $? != 0 ]; then
 	printf "${FAILURE}\n"
 	popd >/dev/null 2>&1
 	exit 1
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -100,7 +100,6 @@ else
 	FAILED=1
 fi
 
-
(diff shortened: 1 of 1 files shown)


hooks/pre-commit
```diff
@@ -68,16 +68,39 @@ else
...
 if [ $? == 0 ]; then
 	printf "${SUCCESS}\n"
 else
 	printf "${FAILURE}\n"
-	popd >/dev/null 2>&1
-	exit 1
+	FAILED=1
 fi
 
+
```
iamalwaysuncomfortable, line 103:
```suggestion
```

hooks/pre-commit
```diff
@@ -68,16 +68,39 @@ else
...
+printf "${PREFIX} Checking that Cargo.toml dependencies are sorted... "
+cargo sort --workspace --grouped --check >/dev/null
 if [ $? == 0 ]; then
 	printf "${SUCCESS}\n"
 else
 	printf "${FAILURE}\n"
-	popd >/dev/null 2>&1
-	exit 1
+	FAILED=1
```
iamalwaysuncomfortable, line 100: Nice catch

iamalwaysuncomfortable (approved):
This is a good cleanup

remoun (approved):

Issue #1892, opened 2022-05-01
Decide what to do with mobilecoinfoundation/fog#137


jcape:
This PR should probably be ported over to the combined repo.

PR #1924, opened 2022-05-05
Reduce log level of failure to publish report to DB

awygle:
Soundtrack of this PR: [The Impressions - It's All Right](https://www.youtube.com/watch?v=Y1LLwC7N1h8)

### Motivation

Bringing forward changes which were in progress before moving the fog repo to the unified repo.

### In this PR
* Reduce log level of failure to publish report to DB

Context from [mobilecoinfoundation/fog#147]:
```
Sometimes, we lose connection to the sql DB, and the report server write operation fails.

This should not be logged as an error because it's not actionable and errors get converted to sentry alerts.
```

Fixes [mobilecoinfoundation/fog#147] and [mobilecoinfoundation/mobilecoin#1892].

### DIFF
--- a/fog/ingest/server/src/controller.rs
+++ b/fog/ingest/server/src/controller.rs
@@ -488,10 +488,11 @@ where
                                 block.index,
                                 err
                             );
-                            std::thread::sleep(std::time::Duration::from_secs(retry_seconds));
-                            retry_seconds = std::cmp::min(retry_seconds + 1, 30);
                         }
                     }
+
+                    std::thread::sleep(std::time::Duration::from_secs(retry_seconds));
+                    retry_seconds = std::cmp::min(retry_seconds + 1, 30);
                 }
             }
 
--- a/fog/ingest/server/src/controller.rs
+++ b/fog/ingest/server/src/controller.rs
@@ -469,7 +469,19 @@ where
                             }
                             break;
                         }
+                        Err(Error::PublishReport) => {
+                            // This is already logged adequately within
+                            // self.publish_report
+                            // 
+                            // Failing to publish a report is not fatal, because we
+                            // attempt to publish
+                            // on every block, and even if it only succeeds half
+                            // the time, it wont make much difference in the pubkey
+                            // expiry window.
+                        }
                         Err(err) => {
+                            // If the error is something other than PublishReport then we should log it
+                            // at a high severity level
                             log::error!(
                                 self.logger,
                                 "Could not publish ingest report at block index {}, will retry: {}",
@@ -1236,6 +1248,7 @@ where
         };
         let report_id = self.config.fog_report_id.as_ref();
 
+        log::info!(self.logger, "publishing report to DB");
         self.recovery_db
             .set_report(ingress_public_key, report_id, &report_data)
             .map(|x| {
@@ -1243,9 +1256,12 @@ where
                 x
             })
             .map_err(|err| {
-                log::error!(
+                // This is a warning because transient report DB write errors are not a cause
+                // for action, as long as this write succeeds eventually it's
+                // fine.
+                log::warn!(
                     self.logger,
-                    "Could not publish report and check on ingress key status: {}",
+                    "Could not publish report to DB and check on ingress key status: {}",
                     err
                 );
                 // Note: At this revision, we don't have generic constraints for converting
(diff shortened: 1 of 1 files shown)


fog/ingest/server/src/controller.rs
```diff
@@ -469,7 +469,19 @@ where
...
+                        Err(Error::PublishReport) => {
+                            // This is already logged adequately within
+                            // self.publish_report
+                            //
+                            // Failing to publish a report is not fatal, because
+                            // we attempt to publish
+                            // on every block, and even if it only succeeds half
+                            // the time, it wont make much difference in the
+                            // pubkey expiry window.
+                        }
```
iamalwaysuncomfortable, line 481: I feel like this would be the simplest  way to handle this
```suggestion
                        // catch publish errors to prevent log duplication
                        Err(Error::PublishReport) => _ ,
```

iamalwaysuncomfortable (approved):
I think the comment in the post error catch for a shorter one would be the simplest way to improve the PR. 

A more elegant way might be to split the publisher error in the PR into two more contextual errors and do the log::warn! about the publisher error within the Error::PublishReport branch of the error.

Something to the tone of:
```rust
    /// Incorrect ingress key: {0}
    BadIngressKey(CompressedRistrettoPublic)
    /// Report publication/ingress key checkup failed: {0}
    PublishReport(String),
```

This more detailed way is entirely optional though.

fog/ingest/server/src/controller.rs
garbageslam: I think this has a problem which is that, we will no longer sleep when `PublishReport` errors occur. that seems not great, it's usually better to have some kind of backoff if there is a connection error (which is what this is usually)

is this actually what we did in the other repo?

garbageslam (dismissed):
we should keep sleeping when `PublishReport` error occurs, as we did before

fog/ingest/server/src/controller.rs
iamalwaysuncomfortable: I didn't notice the sleep right there - good point. Given the OK case breaks out of the loop above, the sleep could just be moved outside of the match statement and added to the end.

iamalwaysuncomfortable: So something like: 

```rust
                loop {
                    match self.publish_report(&ingress_pubkey, &mut state) {
                        Ok(ingress_key_status) => {
                            // If our key is retired, and the index we want to scan is past expiry,
                            // early return. Note, we don't even NEED to scan
                            // when block.index == pubkey_expiry, because the
                            // semantic of pubkey expiry is that it bounds the tombstone block, and
                            // a transaction cannot land in its
                            // tombstone block. But scanning the
                            // tombstone block as well may help
                            // deal with off-by-one errors somewhere else, and doesn't really hurt.
                            if ingress_key_status.retired
                                && block.index > ingress_key_status.pubkey_expiry
                            {
                                log::warn!(self.logger, "When preparing to process block index {}, we discovered that our ingress key is expired: {:?}. Switching to idle and nuking keys.", block.index, ingress_key_status);
                                state.set_idle();
                                self.new_egress_key(&mut state)
                                    .expect("Failure to rotate egress key can't be recovered from");
                                return;
                            }
                            break;
                        }
                        Err(Error::PublishReport) => {}
                        Err(err) => {
                            log::error!(
                                self.logger,
                                "Could not publish ingest report at block index {}, will retry: {}",
                                block.index,
                                err
                            );
                        }
                    }
                    std::thread::sleep(std::time::Duration::from_secs(retry_seconds));
                    retry_seconds = std::cmp::min(retry_seconds + 1, 30);
```

iamalwaysuncomfortable:
Bump to @awygle on this - what are your thoughts on how to proceed forward? This is a positive FOG SLA item 🥇 that I think is almost in reach.

fog/ingest/server/src/controller.rs
remoun: Or we just wrap the `error!` call in an `if err != Error::PublishReport`

awygle: I don't think we should reduce the context provided by the comment - without it it's very unclear why we should treat PublishReport separately.

I have restored the sleep in the PublishReport case.

PR #1933, opened 2022-05-06
Gift Code Memo Builder

iamalwaysuncomfortable:
### Motivation

Build transaction builders for gift codes. Tracks #1735.

### In this PR
* A transaction builder for each gift code memo type

### Future Work
* Create a protobuf message schema for TxOutGiftCode and relevant conversions in mc-api
* Add a gift code tx path in transaction builder
* Make code paths for libsignal ffis & mobilecoind to use to build gift code transactions

### DIFF
--- a/transaction/core/src/memo.rs
+++ b/transaction/core/src/memo.rs
@@ -223,7 +223,7 @@ derive_serde_from_repr_bytes!(MemoPayload);
 derive_prost_message_from_repr_bytes!(MemoPayload);
 
 /// An error which can occur when handling memos
-#[derive(Debug, Display, PartialEq, Eq)]
+#[derive(Debug, Display, Eq, PartialEq)]
 pub enum MemoError {
     /// Wrong length for memo payload: {0}
     BadLength(usize),
--- a/transaction/core/src/memo.rs
+++ b/transaction/core/src/memo.rs
@@ -223,7 +223,7 @@ derive_serde_from_repr_bytes!(MemoPayload);
 derive_prost_message_from_repr_bytes!(MemoPayload);
 
 /// An error which can occur when handling memos
-#[derive(Display, Debug)]
+#[derive(Debug, Display, PartialEq, Eq)]
 pub enum MemoError {
     /// Wrong length for memo payload: {0}
     BadLength(usize),
--- a/transaction/core/src/tx_error.rs
+++ b/transaction/core/src/tx_error.rs
@@ -2,8 +2,8 @@
 
 //! Errors that can occur when creating a new TxOut
 
-use crate::AmountError;
-use alloc::string::String;
+use crate::{AmountError, MemoError};
+use alloc::{format, string::String};
 use displaydoc::Display;
 use mc_crypto_keys::KeyError;
 
@@ -71,8 +71,31 @@ pub enum NewMemoError {
     MultipleOutputs,
     /// Missing output
     MissingOutput,
+    /// Missing required input to build the memo: {0}
+    MissingInput(String),
     /// Mixed Token Ids are not supported in these memos
     MixedTokenIds,
+    /// Destination memo is not supported
+    DestinationMemoNotAllowed,
+    /// Improperly configured input: {0}
+    BadInputs(String),
+    /// Creation
+    Creation(MemoError),
+    /// Utf-8 did not properly decode
+    Utf8Decoding,
     /// Other: {0}
     Other(String),
+}
+
+impl From<MemoError> for NewMemoError {
+    fn from (src: MemoError) -> Self {
+        match src {
+            MemoError::Utf8Decoding => Self::Utf8Decoding,
+            MemoError::BadLength(byte_len) => {
+                Self::BadInputs(
+                    format!("Input of length: {} exceeded max byte length", byte_len)
+                )
+            }
+        }
+    }
 }
--- a/transaction/core/src/tx_error.rs
+++ b/transaction/core/src/tx_error.rs
@@ -56,7 +56,7 @@ impl From<AmountError> for ViewKeyMatchError {
 /// We have included error codes for some known useful error conditions.
 /// For a custom MemoBuilder, you can try to reuse those, or use the Other
 /// error code.
-#[derive(Debug, Display, PartialEq, Eq)]
+#[derive(Debug, Display, Eq, PartialEq)]
 pub enum NewMemoError {
     /// Limits for '{0}' value exceeded
     LimitsExceeded(&'static str),
--- a/transaction/core/src/tx_error.rs
+++ b/transaction/core/src/tx_error.rs
@@ -4,6 +4,7 @@
 
 use crate::{AmountError, MemoError};
 use alloc::{format, string::String};
+use core::str::Utf8Error;
 use displaydoc::Display;
 use mc_crypto_keys::KeyError;
 
@@ -96,5 +97,11 @@ impl From<MemoError> for NewMemoError {
                 byte_len
             )),
         }
+    }
+}
+
+impl From<Utf8Error> for NewMemoError {
+    fn from(_: Utf8Error) -> Self {
+        Self::Utf8Decoding
     }
 }
--- a/transaction/core/src/tx_error.rs
+++ b/transaction/core/src/tx_error.rs
@@ -88,14 +88,13 @@ pub enum NewMemoError {
 }
 
 impl From<MemoError> for NewMemoError {
-    fn from (src: MemoError) -> Self {
+    fn from(src: MemoError) -> Self {
         match src {
             MemoError::Utf8Decoding => Self::Utf8Decoding,
-            MemoError::BadLength(byte_len) => {
-                Self::BadInputs(
-                    format!("Input of length: {} exceeded max byte length", byte_len)
-                )
-            }
+            MemoError::BadLength(byte_len) => Self::BadInputs(format!(
+                "Input of length: {} exceeded max byte length",
+                byte_len
+            )),
         }
     }
 }
--- a/transaction/std/src/memo_builder/gift_code_cancellation_memo_builder.rs
+++ b/transaction/std/src/memo_builder/gift_code_cancellation_memo_builder.rs
@@ -3,7 +3,10 @@
 //! Defines the Memo Builder for the gift code cancellation memo (0x0202)
 //! specified in MCIP #32
 
-use super::{memo::{UnusedMemo, GiftCodeCancellationMemo}, ReservedDestination, MemoBuilder};
+use super::{
+    memo::{GiftCodeCancellationMemo, UnusedMemo},
+    MemoBuilder, ReservedDestination,
+};
 use mc_account_keys::PublicAddress;
 use mc_transaction_core::{Amount, MemoContext, MemoPayload, NewMemoError};
 
@@ -95,5 +98,14 @@ impl MemoBuilder for GiftCodeCancellationMemoBuilder {
                 "Missing global index of TxOut to be cancelled".into(),
             ));
         }
+    }
+}
+
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_gift_code() {
+        // Tests forthcoming
     }
 }
--- a/transaction/std/src/memo_builder/gift_code_cancellation_memo_builder.rs
+++ b/transaction/std/src/memo_builder/gift_code_cancellation_memo_builder.rs
@@ -117,7 +117,7 @@ impl MemoBuilder for GiftCodeCancellationMemoBuilder {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::test_utils::build_zero_value_change_memo;
+    use crate::test_utils::{build_change_memo_with_amount, build_zero_value_change_memo};
     use std::convert::TryInto;
 
     #[test]
@@ -166,6 +166,21 @@ mod tests {
             memo_payload,
             Err(NewMemoError::MultipleChangeOutputs)
         ));
+    }
+
+    #[test]
+    fn test_gift_code_cancellation_memo_fails_for_nonzero_amount() {
+        // Create memo builder
+        let mut builder = GiftCodeCancellationMemoBuilder::default();
+
+        // Set the cancellation index
+        let index = 666;
+        builder.set_gift_code_tx_out_index(index);
+
+        // Build the memo payload
+        let amount = Amount::new(666, 0.into());
+        let memo_payload = build_change_memo_with_amount(&mut builder, amount);
+        assert!(matches!(memo_payload, Err(NewMemoError::BadInputs(_))));
     }
 
     #[test]
--- a/transaction/std/src/memo_builder/gift_code_cancellation_memo_builder.rs
+++ b/transaction/std/src/memo_builder/gift_code_cancellation_memo_builder.rs
@@ -19,7 +19,7 @@ use mc_transaction_core::{Amount, MemoContext, MemoPayload, NewMemoError};
 /// This memo builder builds a gift code cancellation memo (0x0202). Gift code
 /// cancellation is defined as the sender sending the gift code TxOut at the
 /// gift code subaddress back to their default address prior to it being spent
-/// by the receiver. When that happens a zero valued TxOut is sent to the gift
+/// by the receiver. When that happens a change TxOut is sent to the gift
 /// code sender's change subaddress with a gift code cancellation memo that
 /// stores the index of the TxOut originally sent to the gift code subaddress
 /// when the gift code was funded.
@@ -69,21 +69,20 @@ impl MemoBuilder for GiftCodeCancellationMemoBuilder {
         Ok(())
     }
 
-    /// Gift code destination memos are not allowed - all gift code
-    /// memos accompany TxOuts sent to the change address
+    /// Gift code cancellation memos write blank memos to destination TxOut(s)
     fn make_memo_for_output(
         &mut self,
         _amount: Amount,
         _recipient: &PublicAddress,
         _memo_context: MemoContext,
     ) -> Result<MemoPayload, NewMemoError> {
-        Err(NewMemoError::DestinationMemoNotAllowed)
+        Ok(UnusedMemo {}.into())
     }
 
     /// Build a memo for a gift code change output
     fn make_memo_for_change_output(
         &mut self,
-        amount: Amount,
+        _amount: Amount,
         _change_destination: &ReservedDestination,
         _memo_context: MemoContext,
     ) -> Result<MemoPayload, NewMemoError> {
@@ -96,11 +95,6 @@ impl MemoBuilder for GiftCodeCancellationMemoBuilder {
         if self.gift_code_tx_out_global_index.is_none() {
             return Err(NewMemoError::MissingInput(
                 "Must specify gift code TxOut global index".into(),
-            ));
-        }
-        if amount.value > 0 {
-            return Err(NewMemoError::BadInputs(
-                "Cancellation memo TxOut should be zero valued".into(),
             ));
         }
         self.wrote_change_memo = true;
@@ -117,7 +111,7 @@ impl MemoBuilder for GiftCodeCancellationMemoBuilder {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::test_utils::{build_change_memo_with_amount, build_zero_value_change_memo};
+    use crate::test_utils::build_zero_value_change_memo;
     use std::convert::TryInto;
 
     #[test]
@@ -166,21 +160,6 @@ mod tests {
             memo_payload,
             Err(NewMemoError::MultipleChangeOutputs)
         ));
-    }
-
-    #[test]
-    fn test_gift_code_cancellation_memo_fails_for_nonzero_amount() {
-        // Create memo builder
-        let mut builder = GiftCodeCancellationMemoBuilder::default();
-
-        // Set the cancellation index
-        let index = 666;
-        builder.set_gift_code_tx_out_index(index);
-
-        // Build the memo payload
-        let amount = Amount::new(666, 0.into());
-        let memo_payload = build_change_memo_with_amount(&mut builder, amount);
-        assert!(matches!(memo_payload, Err(NewMemoError::BadInputs(_))));
     }
 
     #[test]
--- a/transaction/std/src/memo_builder/gift_code_cancellation_memo_builder.rs
+++ b/transaction/std/src/memo_builder/gift_code_cancellation_memo_builder.rs
@@ -15,8 +15,8 @@ use mc_transaction_core::{Amount, MemoContext, MemoPayload, NewMemoError};
 /// | -->0x0202<--    | Gift Code Cancellation Memo |
 /// This memo builder builds a gift code cancellation memo (0x0202). Gift code
 /// cancellation is defined as the sender sending the gift code TxOut at the
-/// gift code subaddress back to their change address prior to it being spent
-/// by the receiver. When that happens a gift code cancellation memo is
+/// gift code subaddress to their change address prior to it being spent by
+/// the receiver. When that happens a gift code cancellation memo is
 /// written to the change TxOut that stores the index of the TxOut originally
 /// sent to the gift code subaddress when the gift code was funded.
 #[derive(Clone, Debug)]
@@ -54,8 +54,7 @@ impl MemoBuilder for GiftCodeCancellationMemoBuilder {
         Err(NewMemoError::DestinationMemoNotAllowed)
     }
 
-    /// Write the cancellation memo to the change output the gift code is
-    /// refunded to
+    /// Write the cancellation memo to the change output
     fn make_memo_for_change_output(
         &mut self,
         _amount: Amount,
--- a/transaction/std/src/memo_builder/gift_code_cancellation_memo_builder.rs
+++ b/transaction/std/src/memo_builder/gift_code_cancellation_memo_builder.rs
@@ -1,0 +1,99 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
+//! Defines the Memo Builder for the gift code cancellation memo (0x0202)
+//! specified in MCIP #32
+
+use super::{memo::{UnusedMemo, GiftCodeCancellationMemo}, ReservedDestination, MemoBuilder};
+use mc_account_keys::PublicAddress;
+use mc_transaction_core::{Amount, MemoContext, MemoPayload, NewMemoError};
+
+/// There are three possible gift code memo types specified in MCIP #32
+/// | Memo type bytes | Name                                              |
+/// | -----------     | -----------                                       |
+/// |    0x0002       | Gift Code Sender Memo                             |
+/// |    0x0201       | Gift Code Funding Memo                            |
+/// | -->0x0202<--    | Gift Code Cancellation Memo                       |
+/// This memo builder builds a gift code cancellation memo (0x0202). Gift code
+/// cancellation is defined as the sender sending the gift code TxOut at the
+/// gift code subaddress back to their default address prior to it being spent
+/// by the receiver. When that happens a zero valued TxOut is sent to the gift
+/// code sender's change subaddress with a gift code cancellation memo that
+/// stores the index of the TxOut originally sent to the gift code subaddress
+/// when the gift code was funded.
+#[derive(Clone, Debug)]
+pub struct GiftCodeCancellationMemoBuilder {
+    // Index of the gift code TxOut that was originally funded
+    gift_code_tx_out_global_index: Option<u64>,
+    // Whether or not to enable change memos
+    gift_code_change_memo_enabled: bool,
+    // Whether we've already written the change memo
+    wrote_change_memo: bool,
+}
+
+impl GiftCodeCancellationMemoBuilder {
+    /// Create a new cancellation gift code memo builder
+    pub fn new() -> Self {
+        Self {
+            gift_code_tx_out_global_index: None,
+            gift_code_change_memo_enabled: true,
+            wrote_change_memo: false,
+        }
+    }
+    /// Set the index of the gift code TxOut that was cancelled
+    pub fn set_gift_code_tx_out_index(&mut self, tx_out_global_index: u64) {
+        self.gift_code_tx_out_global_index = Some(tx_out_global_index)
+    }
+    /// Clear the index
+    pub fn clear_gift_code_tx_out_index(&mut self) {
+        self.gift_code_tx_out_global_index = None;
+    }
+    /// Enable change memos
+    pub fn enable_gift_code_change_memo(&mut self) {
+        self.gift_code_change_memo_enabled = true;
+    }
+    /// Disable change memos
+    pub fn disable_gift_code_change_memo(&mut self) {
+        self.gift_code_change_memo_enabled = false;
+    }
+}
+
+impl MemoBuilder for GiftCodeCancellationMemoBuilder {
+    /// Set the fee
+    fn set_fee(&mut self, _fee: Amount) -> Result<(), NewMemoError> {
+        Ok(())
+    }
+
+    /// Gift code destination memos are not allowed - all gift code
+    /// memos accompany TxOuts sent to the change address
+    fn make_memo_for_output(
+        &mut self,
+        _amount: Amount,
+        _recipient: &PublicAddress,
+        _memo_context: MemoContext,
+    ) -> Result<MemoPayload, NewMemoError> {
+        Err(NewMemoError::DestinationMemoNotAllowed)
+    }
+
+    /// Build a memo for a gift code change output
+    fn make_memo_for_change_output(
+        &mut self,
+        _amount: Amount,
+        _change_destination: &ReservedDestination,
+        _memo_context: MemoContext,
+    ) -> Result<MemoPayload, NewMemoError> {
+        if !self.gift_code_change_memo_enabled {
+            return Ok(UnusedMemo {}.into());
+        }
+        if self.wrote_change_memo {
+            return Err(NewMemoError::MultipleChangeOutputs);
+        }
+        self.wrote_change_memo = true;
+        if let Some(tx_out_global_index) = self.gift_code_tx_out_global_index.take() {
+            return Ok(GiftCodeCancellationMemo::from(tx_out_global_index).into());
+        } else {
+            return Err(NewMemoError::MissingInput(
+                "Missing global index of TxOut to be cancelled".into(),
+            ));
+        }
+    }
+}
--- a/transaction/std/src/memo_builder/gift_code_cancellation_memo_builder.rs
+++ b/transaction/std/src/memo_builder/gift_code_cancellation_memo_builder.rs
@@ -3,10 +3,7 @@
 //! Defines the Memo Builder for the gift code cancellation memo (0x0202)
 //! specified in MCIP #32
 
-use super::{
-    memo::{GiftCodeCancellationMemo, UnusedMemo},
-    MemoBuilder, ReservedDestination,
-};
+use super::{memo::GiftCodeCancellationMemo, MemoBuilder, ReservedDestination};
 use mc_account_keys::PublicAddress;
 use mc_transaction_core::{Amount, MemoContext, MemoPayload, NewMemoError};
 
@@ -18,48 +15,26 @@ use mc_transaction_core::{Amount, MemoContext, MemoPayload, NewMemoError};
 /// | -->0x0202<--    | Gift Code Cancellation Memo |
 /// This memo builder builds a gift code cancellation memo (0x0202). Gift code
 /// cancellation is defined as the sender sending the gift code TxOut at the
-/// gift code subaddress back to their default address prior to it being spent
-/// by the receiver. When that happens a change TxOut is sent to the gift
-/// code sender's change subaddress with a gift code cancellation memo that
-/// stores the index of the TxOut originally sent to the gift code subaddress
-/// when the gift code was funded.
+/// gift code subaddress back to their change address prior to it being spent
+/// by the receiver. When that happens a gift code cancellation memo is
+/// written to the change TxOut that stores the index of the TxOut originally
+/// sent to the gift code subaddress when the gift code was funded.
 #[derive(Clone, Debug)]
 pub struct GiftCodeCancellationMemoBuilder {
     // Index of the gift code TxOut that was originally funded
-    gift_code_tx_out_global_index: Option<u64>,
-    // Whether or not to enable change memos
-    gift_code_change_memo_enabled: bool,
+    gift_code_tx_out_global_index: u64,
     // Whether we've already written the change memo
     wrote_change_memo: bool,
 }
 
-// Create an empty GiftCodeCancellationMemoBuilder
-impl Default for GiftCodeCancellationMemoBuilder {
-    fn default() -> Self {
+impl GiftCodeCancellationMemoBuilder {
+    /// Initialize memo builder with the index of the originally
+    /// funded gift code TxOut
+    pub fn new(gift_code_tx_out_global_index: u64) -> Self {
         Self {
-            gift_code_tx_out_global_index: None,
-            gift_code_change_memo_enabled: true,
+            gift_code_tx_out_global_index,
             wrote_change_memo: false,
         }
-    }
-}
-
-impl GiftCodeCancellationMemoBuilder {
-    /// Set the index of the gift code TxOut that was cancelled
-    pub fn set_gift_code_tx_out_index(&mut self, tx_out_global_index: u64) {
-        self.gift_code_tx_out_global_index = Some(tx_out_global_index)
-    }
-    /// Clear the index
-    pub fn clear_gift_code_tx_out_index(&mut self) {
-        self.gift_code_tx_out_global_index = None;
-    }
-    /// Enable change memos
-    pub fn enable_gift_code_change_memo(&mut self) {
-        self.gift_code_change_memo_enabled = true;
-    }
-    /// Disable change memos
-    pub fn disable_gift_code_change_memo(&mut self) {
-        self.gift_code_change_memo_enabled = false;
     }
 }
 
@@ -69,42 +44,29 @@ impl MemoBuilder for GiftCodeCancellationMemoBuilder {
         Ok(())
     }
 
-    /// Gift code cancellation memos write blank memos to destination TxOut(s)
+    /// Destination memos for gift code cancellation memos are not allowed
     fn make_memo_for_output(
         &mut self,
         _amount: Amount,
         _recipient: &PublicAddress,
         _memo_context: MemoContext,
     ) -> Result<MemoPayload, NewMemoError> {
-        Ok(UnusedMemo {}.into())
+        Err(NewMemoError::DestinationMemoNotAllowed)
     }
 
-    /// Build a memo for a gift code change output
+    /// Write the cancellation memo to the change output the gift code is
+    /// refunded to
     fn make_memo_for_change_output(
         &mut self,
         _amount: Amount,
         _change_destination: &ReservedDestination,
         _memo_context: MemoContext,
     ) -> Result<MemoPayload, NewMemoError> {
-        if !self.gift_code_change_memo_enabled {
-            return Ok(UnusedMemo {}.into());
-        }
         if self.wrote_change_memo {
             return Err(NewMemoError::MultipleChangeOutputs);
         }
-        if self.gift_code_tx_out_global_index.is_none() {
-            return Err(NewMemoError::MissingInput(
-                "Must specify gift code TxOut global index".into(),
-            ));
-        }
         self.wrote_change_memo = true;
-        if let Some(tx_out_global_index) = self.gift_code_tx_out_global_index.take() {
-            Ok(GiftCodeCancellationMemo::from(tx_out_global_index).into())
-        } else {
-            Err(NewMemoError::MissingInput(
-                "Missing global index of TxOut to be cancelled".into(),
-            ))
-        }
+        Ok(GiftCodeCancellationMemo::from(self.gift_code_tx_out_global_index).into())
     }
 }
 
@@ -116,12 +78,9 @@ mod tests {
 
     #[test]
     fn test_gift_code_cancellation_memo_built_successfully_with_index() {
-        // Create memo builder
-        let mut builder = GiftCodeCancellationMemoBuilder::default();
-
-        // Set the cancellation index
+        // Set the cancellation index and create memo builder
         let index = 666;
-        builder.set_gift_code_tx_out_index(index);
+        let mut builder = GiftCodeCancellationMemoBuilder::new(index);
 
         // Build the memo payload and get the data
         let memo_payload = build_zero_value_change_memo(&mut builder).unwrap();
@@ -133,83 +92,19 @@ mod tests {
     }
 
     #[test]
-    fn test_gift_code_cancellation_memo_fails_without_index() {
-        // Instantiate memo builder
-        let mut builder = GiftCodeCancellationMemoBuilder::default();
-
-        // Build the memo payload
-        let memo_payload = build_zero_value_change_memo(&mut builder);
-
-        // Assert we've created the correct error
-        assert!(matches!(memo_payload, Err(NewMemoError::MissingInput(_))));
-    }
-
-    #[test]
     fn test_gift_code_cancellation_memo_fails_for_more_than_one_change_memo() {
-        // Create memo builder
-        let mut builder = GiftCodeCancellationMemoBuilder::default();
-
-        // Set the cancellation index
+        // Set the cancellation index and create memo builder
         let index = 666;
-        builder.set_gift_code_tx_out_index(index);
+        let mut builder = GiftCodeCancellationMemoBuilder::new(index);
 
-        // Build the memo payload
+        // Attempt to build two change outputs
         build_zero_value_change_memo(&mut builder).unwrap();
         let memo_payload = build_zero_value_change_memo(&mut builder);
+
+        // Assert failure for the second output
         assert!(matches!(
             memo_payload,
             Err(NewMemoError::MultipleChangeOutputs)
         ));
-    }
-
-    #[test]
-    fn test_gift_code_cancellation_memo_fields_are_set_and_cleared_properly() {
-        // Create memo builder
-        let mut builder = GiftCodeCancellationMemoBuilder::default();
-
-        // Set the cancellation index, and then replace it
-        let index = 666;
-        builder.set_gift_code_tx_out_index(index);
-        let replacement_index = 420;
-        builder.set_gift_code_tx_out_index(replacement_index);
-
-        // Build the memo payload and get the data
-        let memo_payload = build_zero_value_change_memo(&mut builder).unwrap();
-        let memo_data = memo_payload.get_memo_data();
-
-        // Verify memo data
-        let derived_index = u64::from_le_bytes(memo_data[0..8].try_into().unwrap());
-        assert_eq!(replacement_index, derived_index);
-
-        // Create another memo builder
-        let mut builder_2 = GiftCodeCancellationMemoBuilder::default();
-
-        // Set the cancellation index and then clear it
-        let index_2 = 666;
-        builder_2.set_gift_code_tx_out_index(index_2);
-        builder_2.clear_gift_code_tx_out_index();
-
-        // Build the memo payload and get the data
-        let memo_payload_2 = build_zero_value_change_memo(&mut builder_2);
-        assert!(matches!(memo_payload_2, Err(NewMemoError::MissingInput(_))));
-    }
-
-    #[test]
-    fn test_gift_code_cancellation_memo_writes_unused_if_change_disabled() {
-        // Create memo builder
-        let mut builder = GiftCodeCancellationMemoBuilder::default();
-
-        // Set the cancellation index
-        let index = 666;
-        builder.set_gift_code_tx_out_index(index);
-
-        // Disable change outputs
-        builder.disable_gift_code_change_memo();
-
-        // Build the memo payload and get the data
-        let memo_payload = build_zero_value_change_memo(&mut builder).unwrap();
-
-        // Verify memo data is blank
-        assert_eq!(memo_payload.get_memo_data(), &[0u8; 64]);
     }
 }
--- a/transaction/std/src/memo_builder/gift_code_cancellation_memo_builder.rs
+++ b/transaction/std/src/memo_builder/gift_code_cancellation_memo_builder.rs
@@ -11,11 +11,11 @@ use mc_account_keys::PublicAddress;
 use mc_transaction_core::{Amount, MemoContext, MemoPayload, NewMemoError};
 
 /// There are three possible gift code memo types specified in MCIP #32
-/// | Memo type bytes | Name                                              |
-/// | -----------     | -----------                                       |
-/// |    0x0002       | Gift Code Sender Memo                             |
-/// |    0x0201       | Gift Code Funding Memo                            |
-/// | -->0x0202<--    | Gift Code Cancellation Memo                       |
+/// | Memo type bytes | Name                        |
+/// | -----------     | -----------                 |
+/// |    0x0002       | Gift Code Sender Memo       |
+/// |    0x0201       | Gift Code Funding Memo      |
+/// | -->0x0202<--    | Gift Code Cancellation Memo |
 /// This memo builder builds a gift code cancellation memo (0x0202). Gift code
 /// cancellation is defined as the sender sending the gift code TxOut at the
 /// gift code subaddress back to their default address prior to it being spent
--- a/transaction/std/src/memo_builder/gift_code_cancellation_memo_builder.rs
+++ b/transaction/std/src/memo_builder/gift_code_cancellation_memo_builder.rs
@@ -83,7 +83,7 @@ impl MemoBuilder for GiftCodeCancellationMemoBuilder {
     /// Build a memo for a gift code change output
     fn make_memo_for_change_output(
         &mut self,
-        _amount: Amount,
+        amount: Amount,
         _change_destination: &ReservedDestination,
         _memo_context: MemoContext,
     ) -> Result<MemoPayload, NewMemoError> {
@@ -92,6 +92,16 @@ impl MemoBuilder for GiftCodeCancellationMemoBuilder {
         }
         if self.wrote_change_memo {
             return Err(NewMemoError::MultipleChangeOutputs);
+        }
+        if self.gift_code_tx_out_global_index.is_none() {
+            return Err(NewMemoError::MissingInput(
+                "Must specify gift code TxOut global index".into(),
+            ));
+        }
+        if amount.value > 0 {
+            return Err(NewMemoError::BadInputs(
+                "Cancellation memo TxOut should be zero valued".into(),
+            ));
         }
         self.wrote_change_memo = true;
         if let Some(tx_out_global_index) = self.gift_code_tx_out_global_index.take() {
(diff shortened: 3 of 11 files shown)


transaction/std/src/memo_builder/gift_code_memo_builder.rs
```diff
@@ -0,0 +1,181 @@
...
+    /// A note in the form of null terminated utf-8 bytes can be applied to
+    /// gift code funding or sender memos. These notes end up in the change
+    /// outputs and indicate what the memo was for. Set this memo here. This
+    /// method will enforce the 60 and 64 byte length for Funding Memos and
+    /// Sender Memos respectively. If the memo type is a cancellation memo
+    /// and a note is set, any value set here will be ignored as the
+    /// cancellation memo does not include a note.
+    pub fn set_note(&mut self, note: &'static str) -> Result<(), MemoError> {
+        let max_note_length = if let GiftCodeMemoType::Funding = self.gift_code_memo_type {
+            60
```
nick-mobilecoin, line 78: This seems like duplication of magic numbers.  Suggest using  `GiftCodeFundingMemo::NOTE_DATA_LEN`, and creating a companion `GiftCodeSenderMemo::NOTE_DATA_LEN`.  
Not really fond of creating a new const on the sender memo just for the builder, but this repeating of these values here seems like a bad code smell.
i.e. what happens if the memo size gets bumped to 128 and this gets changed but the memos themselves get missed.

transaction/std/src/memo_builder/gift_code_memo_builder.rs
```diff
@@ -0,0 +1,181 @@
...
+        Self {
+            gift_code_memo_type,
+            note: "",
+            gift_code_tx_out_public_key: None,
+            gift_code_tx_out_global_index: None,
+            gift_code_change_memo_enabled: true,
+            wrote_change_memo: false,
+        }
+    }
+    /// A note in the form of null terminated utf-8 bytes can be applied to
```
nick-mobilecoin, line 69: It may just be my unfamiliarity with rust but the "null terminated utf-8 bytes" made me do a double take as I looked at the input type of `str`.  Is this not captured in the definition of `str`?  It leaves one wondering if `60` bytes is meant to include the null or not.  Looking at the funding and sender memos It looks like it doesn't include the null terminator

transaction/std/src/memo_builder/gift_code_memo_builder.rs
```diff
@@ -0,0 +1,181 @@
...
+        &mut self,
+        _amount: Amount,
+        _recipient: &PublicAddress,
+        _memo_context: MemoContext,
+    ) -> Result<MemoPayload, NewMemoError> {
+        Err(NewMemoError::GiftCodeDesintationMemoNotAllowed)
+    }
+
+    /// Build a memo for a gift code change output (to ourselves)
+    fn make_memo_for_change_output(
```
nick-mobilecoin, line 140: I appreciate the value of breaking up changes, but I'm not seeing anything exercising/testing this behavior.  Is there  a plan to help ensure correct behavior going forward?

transaction/std/src/memo_builder/gift_code_memo_builder.rs
```diff
@@ -0,0 +1,181 @@
...
+    /// Gift code destination memos are not allowed - RTH transaction builders
+    /// should be used alongside this builder to create recoverable
+    /// transaction history if desired
+    fn make_memo_for_output(
+        &mut self,
+        _amount: Amount,
+        _recipient: &PublicAddress,
+        _memo_context: MemoContext,
+    ) -> Result<MemoPayload, NewMemoError> {
+        Err(NewMemoError::GiftCodeDesintationMemoNotAllowed)
```
nick-mobilecoin, line 136:
```suggestion
        Err(NewMemoError::GiftCodeDestinationMemoNotAllowed)
```

nick-mobilecoin (changes requested):
The two main blockers are the error code spelling and what looks to be a deviation from the MCIP

transaction/core/src/tx_error.rs
```diff
@@ -69,10 +69,14 @@ pub enum NewMemoError {
     InvalidRecipient,
     /// Multiple outputs are not supported
     MultipleOutputs,
+    /// Missing required input to build the memo
```
eranrund, line 72:
```suggestion
    /// Missing required input to build the memo: {0}
```

transaction/std/src/memo_builder/gift_code_memo_builder.rs
```diff
@@ -0,0 +1,181 @@
...
+/// does not write any destination memos. For normal recoverable transaction
+/// history (RTH), the RTH memo builder should be used in conjuction with this
+/// memo builder.
+#[derive(Clone, Debug)]
+pub struct GiftCodeMemoBuilder {
+    // The gift code memo type we'll be writing
+    gift_code_memo_type: GiftCodeMemoType,
+    // The utf-8 formatted note that appears on the change address when a gift code is
+    // or cancelled
+    note: &'static str,
```
eranrund, line 46: What if the note comes at runtime, e.g. from the user inputting something interactively?

transaction/std/src/memo_builder/gift_code_memo_builder.rs
```diff
@@ -0,0 +1,181 @@
...
+        }
+    }
+    /// A note in the form of null terminated utf-8 bytes can be applied to
+    /// gift code funding or sender memos. These notes end up in the change
+    /// outputs and indicate what the memo was for. Set this memo here. This
+    /// method will enforce the 60 and 64 byte length for Funding Memos and
+    /// Sender Memos respectively. If the memo type is a cancellation memo
+    /// and a note is set, any value set here will be ignored as the
+    /// cancellation memo does not include a note.
+    pub fn set_note(&mut self, note: &'static str) -> Result<(), MemoError> {
```
eranrund, line 76: This shouldn't be `&'static str`, that limits it to strings embedded in the code.

transaction/std/src/memo_builder/gift_code_memo_builder.rs
```diff
@@ -0,0 +1,181 @@
...
+}
+
+impl MemoBuilder for GiftCodeMemoBuilder {
+    /// Set the fee
+    fn set_fee(&mut self, _fee: Amount) -> Result<(), NewMemoError> {
+        Ok(())
+    }
+
+    /// Gift code destination memos are not allowed - RTH transaction builders
+    /// should be used alongside this builder to create recoverable
```
eranrund, line 128: How would that work given that `TransactionBuilder` can only accept a single `MemoBuilder`?

transaction/std/src/memo_builder/gift_code_memo_builder.rs
```diff
@@ -0,0 +1,181 @@
...
+        }
+        if self.wrote_change_memo {
+            return Err(NewMemoError::MultipleChangeOutputs);
+        }
+        match self.gift_code_memo_type {
+            GiftCodeMemoType::Funding => {
+                if let Some(tx_out_public_key) = self.gift_code_tx_out_public_key.take() {
+                    self.wrote_change_memo = true;
+                    Ok(GiftCodeFundingMemo::new(tx_out_public_key, self.note)
+                        .unwrap()
```
eranrund, line 157: Shouldn't have unwrap calls here.

transaction/std/src/memo_builder/gift_code_memo_builder.rs
```diff
@@ -0,0 +1,181 @@
...
+                    Ok(GiftCodeCancellationMemo::from(tx_out_global_index).into())
+                } else {
+                    Err(NewMemoError::MissingInput(
+                        "Missing global index of TxOut to be cancelled".into(),
+                    ))
+                }
+            }
+            GiftCodeMemoType::Sender => {
+                self.wrote_change_memo = true;
+                Ok(GiftCodeSenderMemo::try_from(self.note).unwrap().into())
```
eranrund, line 177: Shouldn't have unwraps here.

eranrund (changes requested):
It would be very helpful to see a test case that uses this memo builder for the three possible memo types, and ensure it plays correctly with the `TransactionBuilder`.

transaction/std/src/memo_builder/gift_code_memo_builder.rs
iamalwaysuncomfortable: This is is the design consideration I'm hoping to discuss as it affects the way future memo builders would work. The MCIP as written specifies we should be able to write destination memos alongside recoverable transaction history.

However, as you point out, it can only take ONE memo builder at the moment - so IMHO - the considerations are:
1. Do we want RTH to work as normal when generating special transaction/memo types?
- If we RTH to work as normal alongside other memo types (like this one), that means we build the capability to use two memo builders alongside each other OR the memo builder for special transaction type (burn, gift code, etc.) duplicates the work of RTH, which is kind of wasteful.

2. Are okay with RTH not working exactly the same for special memo types?
- If we're fine with this, then we can use a single memo builder no issue

Personally I think RTH working as normal for all/most transaction/memo types would keep the most consistent accounting if we somehow want to audit. However, it may not be necessary. What are thoughts?

transaction/core/src/tx_error.rs
```diff
@@ -69,10 +69,14 @@ pub enum NewMemoError {
...
+    MissingInput(String),
     /// Missing output
     MissingOutput,
     /// Mixed Token Ids are not supported in these memos
     MixedTokenIds,
+    /// Gift Code Destination Memos are not supported
+    GiftCodeDesintationMemoNotAllowed,
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: This we'll suss out soon, the design is up in there about whether a destination memo for RTH should be written by the builder or whether it should be handed off to the RTH builder. Once this is sussed out I'll more properly name it :)

transaction/std/src/memo_builder/gift_code_memo_builder.rs
```diff
@@ -0,0 +1,181 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: Hrm, I wonder what our policy is on this? Can any Mobilecoin heavies with greater knowledge of the copyright policy weigh in on this?

transaction/std/src/memo_builder/gift_code_memo_builder.rs
```diff
@@ -0,0 +1,181 @@
...
+    Funding,
+    /// Gift Code Cancellation Memo 0x0202
+    Cancellation,
+}
+
+/// This memo builder writes gift code memos to the change address which can be
+/// used by apps that offer gift codes as feature to track the status of gift
+/// codes as well as certain contextual information about them. This builder
+/// writes 3 possible gift code memo types. Only one memo which can be specified
+/// at the creation of the memo builder. It is worth noting that all memos
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: Ah yeah that's actually a great catch - I think as written, there' was a little bit of ambiguity around this, but I think that you're right that 4 distinct memos need to be written and one is destined for the change address. Thanks for catching that error.

Because there's also:
>Any TxOut's on the default subaddress are in-bound transactions, which may have 0x01.. authenticated sender memos which can be validated.
>Any TxOut's on the change subaddress contain recoverable transaction history memos that we can trust
> - 0x0200 Destination memo indicates an outbound transfer, and includes the recipient and the amount
> - 0x0201 Gift code funding memo indicates that we funded a gift code, and indicates which TxOut is the gift code
> - 0x0202 Gift code cancellation memo indicates that we cancelled a gift code that we previously funded
> - 0x0002 Gift code sender memo indicates that a gift code was claimed by the recipient and was sent by the sender
> - 0x0000 Unused indicates that we claimed a gift code but we aren't sure who it came from

Which leaves it somewhat amb

transaction/std/src/memo_builder/gift_code_memo_builder.rs
iamalwaysuncomfortable: This is my "learning Rust" curve showing. Thanks for pointing out that embedded doesn't cover all.

transaction/std/src/memo_builder/gift_code_memo_builder.rs
iamalwaysuncomfortable: Yeah there will be - but I want to get feedback on the design first to ensure its sanity checked.

transaction/std/src/memo_builder/gift_code_memo_builder.rs
iamalwaysuncomfortable: There's discussion on the Gift Codes PR about making it "null terminated unless max length"

transaction/std/src/memo_builder/gift_code_memo_builder.rs
iamalwaysuncomfortable: This is a great point!

transaction/std/src/memo_builder/gift_code_memo_builder.rs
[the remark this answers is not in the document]
iamalwaysuncomfortable: Yeah but directly below that it states - stating that it must go to the change output.

> When claiming a gift code, the gift code output is spent in a self-payment to the change subaddress. This transaction has a single output, the change output, which has a memo of type 0x0002 Gift code sender memo.

transaction/std/src/memo_builder/gift_code_memo_builder.rs
eranrund: A `String` is not `'static`:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b44493d157b3220be01831330cbe83ea

transaction/std/src/memo_builder/gift_code_memo_builder.rs
eranrund: I don't think this is aligned with my understanding of what RTH is. According to https://github.com/mobilecoinfoundation/mcips/blob/main/text/0032-fog-compatible-gift-codes.md#recoverable-transaction-history-and-memos, when creating gift codes we should be writing a memo for both the change output (`0x0201 Gift code funding memo`) and the actual gift code output (`0x0002 Gift code sender memo`).
It appears that the former is handled in this PR in `make_memo_for_change_output`, but the latter is not in `make_memo_for_output`.

I think when someone wants to send out a gift code, they will be creating a TransactionBuilder with the `GiftCodeMemoBuilder` (or perhaps it should be much more specific - a `GiftCodeFundingMemoBuilder`). They then add the single output that is the actual gift code output. There's no need for the "standard" (https://github.com/mobilecoinfoundation/mobilecoin/blob/b6af448d1ff5f980f05ac641aa087a99744d0e96/transaction/std/src/memo_builder/rth_memo_builder.rs) memos in this case, if I understood everything correctly.

The one thing I don't like here is that the user of the `TransactionBuilder` needs to be aware of the specific steps they need to take in order to generate a valid gift code, namely constructing the `TransactionBuilder` with the right memo builder, and calling `add_output` only once (which should be enforced by the memo builder) with the correct destination (the gift code subaddress).

It might be useful to provide a `GiftCodeFundingBuilder` object that embeds all this logic in it.

Either way, I'd like to see a unit test that uses the gift code memo builder(s) (might be worth splitting into twodifferent builders since the behavior is actually very different for the scenarios of funding and cancellation) with the `TransactionBuilder` to produce actual `Tx`s that contain the expected set of outputs and memos.

iamalwaysuncomfortable: It seems to make sense that the 0x0002 goes to the gift code output but it actually is specified to go to the change output. A little strange, but that's what's in there. We could potentially revisit that in the MCIP.

> When claiming a gift code, the gift code output is spent in a self-payment to the change subaddress. This transaction has a single output, the change output, which has a memo of type 0x0002 Gift code sender memo. 

Also noted about the tx builder user having having to be aware of what state they're at potentially causing problems - I will think about this.

Also thanks for the design input. I do intend to build tests, but I wanted to first sanity check the initial architecture with the hive mind before I started building too heavily in a specific direction. I will go ahead and address all the feedback here and add thorough tests.

eranrund: I see, interesting, I misunderstood that about `0x0002` but it does make sense.
So the output that goes to the gift code subaddress does not actually have a memo?

In that case it seems like funding a giftcode should allow for two outputs (change + regular output), with the regular one not having a memo and canceling/claiming a gift code should only allow for a change output. Is that correct?

iamalwaysuncomfortable: > So the output that goes to the gift code subaddress does not actually have a memo?

Yeah as written that's correct. It could potentially have a different design, but we'd probably want to get larger stakeholder buy in about changing the MCIP.

> In that case it seems like funding a giftcode should allow for two outputs (change + regular output), with the regular one not having a memo and canceling/claiming a gift code should only allow for a change output. Is that correct?

That is correct!

transaction/std/src/memo_builder/gift_code_memo_builder.rs
iamalwaysuncomfortable: Noted

transaction/std/src/memo_builder/gift_code_memo_builder.rs
iamalwaysuncomfortable: Also noted

transaction/std/src/memo_builder/gift_code_memo_builder.rs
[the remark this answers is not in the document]
nick-mobilecoin:
```mermaid
graph LR
    subgraph Gift
    B[Gift code output]
    C[Change output]
    end
    A[Funds] -->|TxOut| B
    A --> |Change| C
    C --> A
    B --> |TxOut| D(Claim)
```
(so cool that mermaid works in PR comments)
It may be my lack of understanding here.  It seems that there is an intermediate `TxOut` here, the "Gift code output" in the diagram.  My understanding is 
* "Change output" gets the funding memo.
* "Gift code output" gets the sender memo.
* "Claim" would get the cancellation memo, of the originator claims it.

I'm think `make_memo_for_change_output()` will only be called for the "Change output" and the self claim.  TBH not 100% on the self claim, it will be to the change subaddress, but does that force it through the entire change output logic...

transaction/std/src/memo_builder/gift_code_memo_builder.rs
iamalwaysuncomfortable: I think it'll be fine to have the MEMO_LEN const for the sender memo as as it works right now, all bytes are reserved for utf-8 so 128 bytes would be reserved. But yeah excellent catch on the constant usage being best practice.

transaction/std/src/memo_builder/gift_code_memo_builder.rs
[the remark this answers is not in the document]
iamalwaysuncomfortable: Mermaid JS in github is pretty awesome right? We should definitely be doing this more on MCIPs.

Here's roughly what the MCIP specifies
```mermaid
graph RL
    subgraph Sender_Account
    subgraph Sender_Gift_Code_Subaddress
    C[Gift_Code_TxOut]
    end
    subgraph Sender_Change_Address
    D[Zero_Value_Gift_Code_Funding_Memo_TxOut]
    E[Zero_Value_Gift_Cancellation_TxOut_with_Memo]
    end
    subgraph Sender_Default_Subaddress
    A[Available_TxIns]
    B[Cancellation_TxOut]
    end
    end
    subgraph Receiver_Account
    subgraph Receiver_Change_Address
    F[Redeemed_Gift Code_TxOut] 
    end
    subgraph Receiver_Gift_Code_Address
    G[Redeemed_Gift Code_TxOut] 
    end
    end

A -->|Funding - Funds| C
A -->|Funding - Memo| D
C -->|Cancellation - Funds| B
C -->|Cancellation - Memo| E
C -->|Redemption - Funds| G
C -->|Redemption - Memo| F
```

transaction/core/src/tx_error.rs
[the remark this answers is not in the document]
iamalwaysuncomfortable: Probably just DestinationMemoNotAllowed - I feel like since there are now a variety of memo types (most frequently used Change/Destination) including it in. "NotAllowed" could allude to a lot of things, and I think there's no harm in being extra clear.

transaction/std/src/memo_builder/gift_code_memo_builder.rs
iamalwaysuncomfortable: Tests now incoming!

transaction/std/src/memo_builder/gift_code_memo_builder.rs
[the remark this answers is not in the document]
nick-mobilecoin: I don't think the `Redemption - Funds` should be going to the gift code sub-address.  My understanding is that should only be used for _outbound_ in flight gifts.

I think the MCIP has some contradictions in it, as you mention it says:
> When cancelling a gift code, the gift code output is paid back to self. This transaction has a single output, the change output, which goes to the *change subaddress*

and
> When claiming a gift code, the gift code output is spent in a self-payment to the change subaddress. This transaction has a single output, the change output, which has a memo of type 0x0002 Gift code sender memo

However up in the guide level explanation it says:
> Until it is spent, the app can potentially cancel the gift code by paying the TxOut back to themself on subaddress 0.

and in the reference level explanation
> The app may cancel a gift code by paying it back to the primary subaddress.

transaction/std/src/memo_builder/gift_code_memo_builder.rs
nick-mobilecoin: The issue I saw with the MEMO_LEN is that the `GiftCodeFundingMemo` also has a `MEMO_LEN` which is the full 64 bytes, it may be tempting for someone to change that if they see `NOTE_DATA_LEN` from one and `MEMO_LEN` from another

transaction/std/src/memo_builder/gift_code_memo_builder.rs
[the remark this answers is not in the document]
eranrund: I wonder if it actually matters where the cancellation TxOut goes to (subaddress 0/primary or the change subaddress), since the way you tell it is a gift code cancellation is by looking at the memo. The only reason why I think having it go to the change subaddress is preferable, is because that subaddress is private so no-one could mess with your history.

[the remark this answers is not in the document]
iamalwaysuncomfortable: > I don't think the Redemption - Funds should be going to the gift code sub-address. My understanding is that should only be used for outbound in flight gifts.

Ya you're actually correct about this, it goes to the change address.

As far as the cancellation memo ambiguity, I think paying it back to the primary subaddress makes more sense given transaction builders can then use it in future txout selection.

transaction/std/src/memo_builder/gift_code_memo_builder.rs
iamalwaysuncomfortable: I've refactored them into separate transaction builders to clear up the ambiguity

transaction/std/src/memo_builder/gift_code_memo_builder.rs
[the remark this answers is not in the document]
eranrund: The transaction builder can use txouts from the change subaddress, it doesn't care which subaddress the outputs went to (if it couldn't then change would not be spendable...).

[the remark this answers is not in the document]
iamalwaysuncomfortable: I pull a modification to the MCIP to reflect this.

transaction/core/src/memo.rs
```diff
@@ -223,7 +223,7 @@ derive_serde_from_repr_bytes!(MemoPayload);
 derive_prost_message_from_repr_bytes!(MemoPayload);
 
 /// An error which can occur when handling memos
-#[derive(Display, Debug)]
+#[derive(Debug, Display, PartialEq, Eq)]
```
remoun, line 226: Prefer sorted
```suggestion
#[derive(Debug, Display, Eq, PartialEq)]
```

transaction/std/src/memo_builder/gift_code_cancellation_memo_builder.rs
```diff
@@ -0,0 +1,113 @@
...
+        self.gift_code_tx_out_global_index = None;
+    }
+    /// Enable change memos
+    pub fn enable_gift_code_change_memo(&mut self) {
+        self.gift_code_change_memo_enabled = true;
+    }
+    /// Disable change memos
+    pub fn disable_gift_code_change_memo(&mut self) {
+        self.gift_code_change_memo_enabled = false;
+    }
```
remoun, line 63: This API seems overly verbose. Consider a more 'fluent' style.

The separate enable/disable methods would make updating the builder based on a toggle UI require a conditional that calls one of two methods, vs. passing the value to one method.

```suggestion
    // Allows callers to pass N, Some(N), or None.
    pub fn gift_code_tx_out_index(mut self, tx_out_global_index: impl Into<Option<u64>>) -> Self {
        self.gift_code_tx_out_global_index = tx_out_global_index.into()
        self
    }
    /// Toggle change memos
    pub fn gift_code_change_memo(mut self, enabled: bool) -> Self {
        self.gift_code_change_memo_enabled = enabled;
        self
    }
```

transaction/std/src/memo_builder/gift_code_cancellation_memo_builder.rs
```diff
@@ -0,0 +1,113 @@
...
+};
+use mc_account_keys::PublicAddress;
+use mc_transaction_core::{Amount, MemoContext, MemoPayload, NewMemoError};
+
+/// There are three possible gift code memo types specified in MCIP #32
+/// | Memo type bytes | Name                                              |
+/// | -----------     | -----------                                       |
+/// |    0x0002       | Gift Code Sender Memo                             |
+/// |    0x0201       | Gift Code Funding Memo                            |
+/// | -->0x0202<--    | Gift Code Cancellation Memo                       |
```
remoun, line 18: Nit: The table does not need to be so wide
```suggestion
/// | Memo type bytes | Name                          |
/// | -----------     | -----------                   |
/// |    0x0002       | Gift Code Sender Memo         |
/// |    0x0201       | Gift Code Funding Memo        |
/// | -->0x0202<--    | Gift Code Cancellation Memo   |
```

transaction/std/src/memo_builder/gift_code_cancellation_memo_builder.rs
iamalwaysuncomfortable: It's done this way to be kept consistent with the other memo builders. impl Into is a pattern I'm not super fund of because it doesn't promote the caller thinking about that they're dealing with an option param.

transaction/std/src/memo_builder/gift_code_sender_memo_builder.rs
```diff
@@ -0,0 +1,276 @@
...
+            return Err(NewMemoError::MultipleChangeOutputs);
+        };
+        // Prevent callers from writing memo if last note set attempt was a failure
+        if self.attempted_invalid_note {
+            return Err(NewMemoError::BadInputs(
+                "Tried to set a note longer than 64 bytes".into(),
+            ));
+        }
+        self.wrote_change_memo = true;
+        Ok(GiftCodeSenderMemo::new(self.note.as_str())?.into())
```
nick-mobilecoin, line 119: nit: 
```suggestion
        Ok(GiftCodeSenderMemo::new(&self.note)?.into())
```

transaction/std/src/memo_builder/gift_code_sender_memo_builder.rs
```diff
@@ -0,0 +1,276 @@
...
+        let mut builder = GiftCodeSenderMemoBuilder::default();
+
+        // Attempt to set an invalid note
+        let note_bytes = [b'6'; GiftCodeSenderMemo::MEMO_DATA_LEN + 1];
+        let note = std::str::from_utf8(&note_bytes).unwrap();
+        let result = builder.set_gift_code_sender_note(note);
+        assert!(matches!(result, Err(NewMemoError::BadInputs(_))));
+
+        // Reset with valid note
+        let valid_note = "Im within the byte length, but validity is a subjective concept";
```
nick-mobilecoin, line 267: Should check boundary case of note at maximum length:
```suggestion
        let valid_note_bytes = [b'6'; GiftCodeSenderMemo::MEMO_DATA_LEN];
        let valid_note = std::str::from_utf8(&valid_note_bytes).unwrap();
```

transaction/std/src/memo_builder/gift_code_funding_memo_builder.rs
```diff
@@ -0,0 +1,444 @@
...
+        let mut rng: StdRng = SeedableRng::from_seed([0u8; 32]);
+        let change_tx_public_key = RistrettoPublic::from_random(&mut rng);
+        let note = "It's MEMO TIME!!";
+        builder
+            .set_gift_code_tx_out_public_key(&change_tx_public_key)
+            .unwrap();
+        builder.set_funding_note(note).unwrap();
+
+        // Build a memo
+        let mut rng: StdRng = SeedableRng::from_seed([0u8; 32]);
```
nick-mobilecoin, line 246: Is there a reason not to use `build_change_memo_with_amount()`?

transaction/std/src/test_utils.rs
```diff
@@ -201,3 +206,72 @@ pub fn get_transaction<RNG: RngCore + CryptoRng, FPR: FogPubkeyResolver + Clone>
...
+        } else {
+            GiftCodeFundingMemo::MEMO_DATA_LEN
+        };
+
+        std::str::from_utf8(&memo_data[GiftCodeFundingMemo::HASH_DATA_LEN..index]).unwrap()
+    }
+
+    /// Get the first four bytes of the Blake2b hash of a gift code TxOut public
+    /// key
+    pub fn tx_out_public_key_short_hash(
```
nick-mobilecoin, line 267: This probably belongs on the funding memo code, but is there any concern grabbing the first 4 bytes of a hash?
I don't know enough about the hash implementations, but could this result in the hashing logic ignoring the last n bytes of the input?  i.e. why do a 512bit hash on a value if only 32 bits are used, why not just do a 32 bit hash?
Again outside the scope of this review just a general question

transaction/std/src/memo_builder/gift_code_cancellation_memo_builder.rs
```diff
@@ -0,0 +1,236 @@
...
+    use crate::test_utils::{build_change_memo_with_amount, build_zero_value_change_memo};
+    use std::convert::TryInto;
+
+    #[test]
+    fn test_gift_code_cancellation_memo_built_successfully_with_index() {
+        // Create memo builder
+        let mut builder = GiftCodeCancellationMemoBuilder::default();
+
+        // Set the cancellation index
+        let index = 666;
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: :trollface:

transaction/std/src/memo_builder/gift_code_cancellation_memo_builder.rs
```diff
@@ -0,0 +1,236 @@
...
+        builder.set_gift_code_tx_out_index(index);
+
+        // Build the memo payload and get the data
+        let memo_payload = build_zero_value_change_memo(&mut builder).unwrap();
+        let memo_data = memo_payload.get_memo_data();
+
+        // Verify memo data
+        let derived_index = u64::from_le_bytes(memo_data[0..8].try_into().unwrap());
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: Thank you! I tend to prefer overstating :D

transaction/std/src/memo_builder/gift_code_funding_memo_builder.rs
```diff
@@ -0,0 +1,444 @@
...
+            return Err(NewMemoError::BadInputs(
+                "Tried to set a note longer than 64 bytes".into(),
+            ));
+        }
+        if self.gift_code_tx_out_public_key.as_ref() == Some(memo_context.tx_public_key) {
+            return Err(NewMemoError::BadInputs("The public_key in the memo should be the TxOut at the gift code subaddress and not that of the memo TxOut".into()));
+        }
+        if let Some(tx_out_public_key) = self.gift_code_tx_out_public_key.take() {
+            self.wrote_change_memo = true;
+            Ok(GiftCodeFundingMemo::new(&tx_out_public_key, self.note.as_str())?.into())
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: I think here I prefer the verbosity just to show people what's exactly happening, but point noted

transaction/std/src/memo_builder/gift_code_funding_memo_builder.rs
```diff
@@ -0,0 +1,444 @@
...
+        builder
+            .set_gift_code_tx_out_public_key(&change_tx_public_key)
+            .unwrap();
+        builder.set_funding_note(note).unwrap();
+
+        // Build a memo
+        let mut rng: StdRng = SeedableRng::from_seed([0u8; 32]);
+        let alice = AccountKey::random_with_fog(&mut rng);
+        let alice_address_book = ReservedDestination::from(&alice);
+        let change_amount = Amount::new(666, 666.into());
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: I re-read the MCIP and it doesn't actually state that change outputs have to be zero (and if you're funding it with a TxOut of an amount larger than the gift code, you'll need the change to be non-empty). So I removed the restriction.

transaction/std/src/memo_builder/gift_code_funding_memo_builder.rs
```diff
@@ -0,0 +1,444 @@
...
+        let derived_note = MemoDecoder::decode_funding_note(memo_data);
+        let expected_hash = MemoDecoder::tx_out_public_key_short_hash(&gift_code_public_key);
+        assert_eq!(blank_note, derived_note);
+        assert_eq!(
+            expected_hash,
+            memo_data[0..GiftCodeFundingMemo::HASH_DATA_LEN]
+        );
+
+        // Create another memo builder
+        let mut builder_2 = GiftCodeFundingMemoBuilder::default();
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: I had thought about that but, given these functionalities are within the same class, list was already long and felt like grouping them would help reader fatigue.

transaction/std/src/memo_builder/gift_code_funding_memo_builder.rs
```diff
@@ -0,0 +1,444 @@
...
+        builder_2.set_funding_note(note).unwrap();
+        builder_2.clear_gift_code_tx_out_public_key();
+
+        // Build the memo payload and ensure we can't build it without the tx public key
+        let memo_payload_2 = build_zero_value_change_memo(&mut builder_2);
+        assert!(matches!(memo_payload_2, Err(NewMemoError::MissingInput(_))));
+    }
+
+    #[test]
+    fn test_gift_code_funding_memo_writes_unused_if_change_disabled() {
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: I'd prefer to keep the above naming because then you can run all gift code tests with `cargo test test_gift_code`

transaction/std/src/memo_builder/gift_code_funding_memo_builder.rs
```diff
@@ -0,0 +1,444 @@
...
+        builder
+            .set_gift_code_tx_out_public_key(&gift_code_public_key)
+            .unwrap();
+        let result = builder.set_funding_note(bad_note);
+
+        // Ensure set method errors correctly
+        assert!(matches!(result, Err(NewMemoError::BadInputs(_))));
+
+        // Set a correct memo and ensure we can build a valid memo
+        let correct_note = "I'm also bad mwa-ha (but in a healthy, fun way)";
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: Yeah, that's probably better!

transaction/std/src/memo_builder/gift_code_sender_memo_builder.rs
iamalwaysuncomfortable: Also probably better!

transaction/std/src/test_utils.rs
iamalwaysuncomfortable: >  i.e. why do a 512bit hash on a value if only 32 bits are used, why not just do a 32 bit hash?

Because 32 bits would have a much larger possible set of collisions for different inputs. With a 512 bit output space, that set of possible colliding inputs is much larger. Using the first 4 bytes also increases the collision space, but not more than only using 32 bits. As far as why we used only 4 bytes of the 64 bytes of the hash in the MCIP - it's basically because we wanted to have as much space as possible for the text bytes.

It's a tradeoff we made :), but one that's unlikely to be attacked. If you do find some magic input that m

transaction/std/src/memo_builder/gift_code_funding_memo_builder.rs
iamalwaysuncomfortable: Yes because in that function, there's no way to specify which public key goes into it. This tests ensures that the public key set onto the memo isn't that of the change output and I needed a way to do that here. Thus why it's the only test that doesn't use build_change_memo_with_amount()

transaction/std/src/memo_builder/gift_code_sender_memo_builder.rs
iamalwaysuncomfortable: I think I'm going to keep it this way to express what's happening. The & is concise but hides more of what's happening.

transaction/std/src/memo_builder/gift_code_funding_memo_builder.rs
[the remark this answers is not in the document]
nick-mobilecoin: that's good to keep in mind, I need to learn more about how cargo test works.  I was hoping you could call out the file itself

transaction/std/src/test_utils.rs
nick-mobilecoin: > Using the first 4 bytes also increases the collision space, but not more than only using 32 bits.
I think this is better wording of what I was trying to ask, "is using the first 4 bytes the same collision space as using a dedicated 32 bit hash, or is it a larger collision space?"
I think for this use case it probably doesn't matter too much, just my lack of knowledge on how hash functions work under the hood.

transaction/std/src/memo_builder/gift_code_cancellation_memo_builder.rs
```diff
@@ -0,0 +1,215 @@
...
+        }
+        self.wrote_change_memo = true;
+        if let Some(tx_out_global_index) = self.gift_code_tx_out_global_index.take() {
+            Ok(GiftCodeCancellationMemo::from(tx_out_global_index).into())
+        } else {
+            Err(NewMemoError::MissingInput(
+                "Missing global index of TxOut to be cancelled".into(),
+            ))
+        }
+    }
```
remoun, line 108: Optional/DRY: Consolidate the MissingInput check
```suggestion
        if let Some(tx_out_global_index) = self.gift_code_tx_out_global_index.take() {
            self.wrote_change_memo = true;
            Ok(GiftCodeCancellationMemo::from(tx_out_global_index).into())
        } else {
            Err(NewMemoError::MissingInput(
                "Missing global index of TxOut to be cancelled".into(),
            ))
        }
    }
```

remoun (approved):

transaction/std/src/memo_builder/gift_code_funding_memo_builder.rs
```diff
@@ -0,0 +1,426 @@
...
+    /// Disable change memos
+    pub fn disable_gift_code_change_memo(&mut self) {
+        self.gift_code_change_memo_enabled = false;
+    }
+}
+
+impl MemoBuilder for GiftCodeFundingMemoBuilder {
+    /// Set the fee
+    fn set_fee(&mut self, _fee: Amount) -> Result<(), NewMemoError> {
+        Ok(())
```
eranrund, line 116: After checking with @garbageslam we think that we should actually track the fee being paid in the funding memo.

transaction/std/src/memo_builder/gift_code_sender_memo_builder.rs
```diff
@@ -0,0 +1,276 @@
...
+pub struct GiftCodeSenderMemoBuilder {
+    // Utf-8 formatted note within the memo that can be up to 64 bytes long
+    note: String,
+    // Whether or not to enable change memos
+    gift_code_change_memo_enabled: bool,
+    // Whether we've already written the change memo
+    wrote_change_memo: bool,
+    // Whether our last note setting attempt was invalid and the note hasn't
+    // been reset with a valid note or cleared
+    attempted_invalid_note: bool,
```
eranrund, line 36: I don't see what this gives us. Its the callers responsibility to handle a failure of `set_gift_code_sender_note` and not continue building the transaction.

transaction/std/src/memo_builder/gift_code_cancellation_memo_builder.rs
```diff
@@ -0,0 +1,215 @@
...
+/// |    0x0002       | Gift Code Sender Memo       |
+/// |    0x0201       | Gift Code Funding Memo      |
+/// | -->0x0202<--    | Gift Code Cancellation Memo |
+/// This memo builder builds a gift code cancellation memo (0x0202). Gift code
+/// cancellation is defined as the sender sending the gift code TxOut at the
+/// gift code subaddress back to their default address prior to it being spent
+/// by the receiver. When that happens a change TxOut is sent to the gift
+/// code sender's change subaddress with a gift code cancellation memo that
+/// stores the index of the TxOut originally sent to the gift code subaddress
+/// when the gift code was funded.
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: The MCIP left it unclear. Originally I was thinking it would be more consistent to have it go back to the funding address, but especially point number 2. I agree we should send it to the change address.

transaction/std/src/memo_builder/gift_code_funding_memo_builder.rs
```diff
@@ -0,0 +1,426 @@
...
+    gift_code_tx_out_public_key: Option<RistrettoPublic>,
+    // Whether or not to enable change memo
+    gift_code_change_memo_enabled: bool,
+    // Whether we've already written the change memo
+    wrote_change_memo: bool,
+    // Whether we've written an unused memo to the destination
+    wrote_destination_memo: bool,
+    // Whether our last note setting attempt was invalid and the note hasn't
+    // been reset with a valid note or cleared
+    attempted_invalid_note: bool,
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: I don't see the harm of having the redundancy here.

transaction/std/src/memo_builder/gift_code_funding_memo_builder.rs
iamalwaysuncomfortable: Sure. At this point though I'm inclined to make that a change to the MCIP and allow people a day to make comments on it because that puts gift codes into RTH territory & takes a way a decent amount of space from the memo - and given Moby has driven a lot of this, that change is something we should have other teams who will be using it (mainly Moby/mobile teams) weigh in on.

I'd like to address that in a separate PR scoped to adding information to the memos.

transaction/std/src/memo_builder/gift_code_sender_memo_builder.rs
iamalwaysuncomfortable: I'm not sure how the extra redundancy hurts either. Given the records on chain are immutable, another check doesn't hurt.

transaction/std/src/memo_builder/gift_code_cancellation_memo_builder.rs
```diff
@@ -0,0 +1,215 @@
...
+#[derive(Clone, Debug)]
+pub struct GiftCodeCancellationMemoBuilder {
+    // Index of the gift code TxOut that was originally funded
+    gift_code_tx_out_global_index: Option<u64>,
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: I mainly did it to make it consistent a consistent flow with the other builders providing opt-outs of certain flows. But I agree in general with memo builders being "all-or-nothing" and doing the thing they are designed to do and using the unused memo builder if you're not planning to use it.

transaction/std/src/memo_builder/gift_code_funding_memo_builder.rs
```diff
@@ -0,0 +1,426 @@
...
+    /// Set the TxOut public_key of the gift code TxOut sent to the
+    /// reserved gift code subaddress.
+    ///
+    /// IMPORTANT NOTE: Do NOT pass the public_key of the change TxOut
+    /// that the gift code memo is attached to as an argument. Doing
+    /// so will result in an error when attempting to build the memo.
+    pub fn set_gift_code_tx_out_public_key(
+        &mut self,
+        tx_out_public_key: &RistrettoPublic,
+    ) -> Result<(), NewMemoError> {
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: Yeah actually -- agreed!

transaction/std/src/memo_builder/gift_code_cancellation_memo_builder.rs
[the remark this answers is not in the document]
iamalwaysuncomfortable: In other words, I'm going remove the optional change output from all the gift code memo builders.

transaction/std/src/memo_builder/gift_code_sender_memo_builder.rs
```diff
@@ -0,0 +1,276 @@
...
+            return Err(NewMemoError::BadInputs(
+                "Sender note cannot be longer than 64 bytes".into(),
+            ));
+        }
+        self.note = note.into();
+        self.attempted_invalid_note = false;
+        Ok(())
+    }
+    /// Clear the gift code sender note
+    pub fn clear_sender_note(&mut self) {
```
garbageslam, line 68: IMO, the API is simpler if it's like this:

* `Default` is not implemented
* A `::new` which takes a `String` note is available
* The note provided must be valid, or `::new` returns an error
* We don't need to have `set` and `clear` functions, or track the `attempted_invalid_note` state, because the constructor enforces the invariants.

transaction/std/src/test_utils.rs
```diff
@@ -201,3 +206,72 @@ pub fn get_transaction<RNG: RngCore + CryptoRng, FPR: FogPubkeyResolver + Clone>
...
+    //Build memo
+    builder.make_memo_for_change_output(change_amount, &alice_address_book, memo_context)
+}
+
+/// Utilities for decoding and verifying memo data
+pub struct MemoDecoder;
+
+impl MemoDecoder {
+    /// Decode a sender note from bytes
+    pub fn decode_sender_note(memo_data: &[u8; GiftCodeSenderMemo::MEMO_DATA_LEN]) -> &str {
```
garbageslam, line 240: hmm so, why do we need this, can we just do `GiftCodeSenderMemo::try_from(MemoPayload)` ?

transaction/std/src/memo_builder/gift_code_cancellation_memo_builder.rs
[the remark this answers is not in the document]
garbageslam: FWIW I agree with Eran, if you don't have this number then the reader won't be able to make sense of what gift code was cancelled. So it should be required -- IMO we should not implement `Default`, we should require the user of this class to provide the number in a `new` function

This is similar to what we did in the `BurnRedemptionMemoBuilder`.

In the `RTHMemoBuilder` we didn't require the sender memo credential, but that was only because we thought there might be a use-case where a user chooses not to use the sender memo thing, but still wants change outputs. (Partly this was from discussions with Koe.) And so when that data was omitted it writes unused memo. If we didn't think that behavior was desired then probably `RTHMemoBuilder` would require a sender memo credential to be constructed.

In this case though, I think we can't actually use the memo correctly at all if we don't have this number.

garbageslam (changes requested):
IMO the funding and cancellation builders should not be allowed to be constructed without the data needed to create a valid memo, to make it harder to misuse them and get bad memos out of them. I think their APIs can then be simplified a bit

transaction/std/src/memo_builder/gift_code_sender_memo_builder.rs
iamalwaysuncomfortable: I think that's definitely better, I wrote the transaction builder tests today and I think I wasn't quite seeing how the flow progressed. I definitely agree with this now.

iamalwaysuncomfortable: In fact I think none of the memo builders actually NEED any get or set methods.

transaction/std/src/memo_builder/gift_code_funding_memo_builder.rs
eranrund: We discussed this over Slack and it will be addressed in a followup PR.

transaction/std/src/memo_builder/gift_code_sender_memo_builder.rs
```diff
@@ -0,0 +1,178 @@
...
+/// | -->0x0002<--    | Gift Code Sender Memo       |
+/// |    0x0201       | Gift Code Funding Memo      |
+/// |    0x0202       | Gift Code Cancellation Memo |
+/// This memo builder builds the gift code sender memo (0x0002). A gift code
+/// considered redeemed when the Receiver uses the TxOut spend private key
+/// of the gift code TxOut they received from the Sender to send the TxOut
+/// from the sender's gift code subaddress to their own change subaddress.
+/// The sender memo is written into the TxOut the receiver sends to the
+/// change address and includes an optional Utf-8 note up to 64 bytes long
+/// the Receiver can  use to record any information they desire about the
```
eranrund, line 22:
```suggestion
/// the Receiver can use to record any information they desire about the
```

transaction/std/src/memo_builder/gift_code_sender_memo_builder.rs
```diff
@@ -0,0 +1,178 @@
...
+}
+
+impl GiftCodeSenderMemoBuilder {
+    /// Initialize memo builder with a utf-8 note (up to 64 bytes), This
+    /// method will enforce the 64 byte limit with a NewMemoErr if the
+    /// note passed is longer than 64 bytes.
+    pub fn new(note: &str) -> Result<Self, NewMemoError> {
+        if note.len() > GiftCodeSenderMemo::MEMO_DATA_LEN {
+            return Err(NewMemoError::BadInputs(
+                "Sender note cannot be longer than 64 bytes".into(),
```
eranrund, line 39: Should use `MEMO_DATA_LEN` instead of 64

transaction/std/src/transaction_builder.rs
```diff
@@ -574,7 +574,7 @@ pub(crate) fn create_output_with_fog_hint<RNG: CryptoRng + RngCore>(
     if !block_version.masked_token_id_feature_is_supported() {
         tx_out.masked_amount.masked_token_id.clear();
     }
-
+    //
```
eranrund, line 577:
```suggestion

```

transaction/std/src/transaction_builder.rs
```diff
@@ -3104,4 +3105,505 @@ pub mod transaction_builder_tests {
...
+            assert!(
+                subaddress_matches_tx_out(&sender, GIFT_CODE_SUBADDRESS_INDEX, funding_output)
+                    .unwrap()
+            );
+            assert!(subaddress_matches_tx_out(
+                &sender,
+                CHANGE_SUBADDRESS_INDEX,
+                funding_change_output
+            )
+            .unwrap());
```
eranrund, line 3201: These are unnecessary since the `.find()` calls above would not find anything and `.expect` will end the test

transaction/std/src/transaction_builder.rs
```diff
@@ -3104,4 +3105,505 @@ pub mod transaction_builder_tests {
...
+                {
+                    MemoType::GiftCodeFunding(memo) => {
+                        assert!(memo.public_key_matches(funding_output_public_key),);
+                        assert_eq!(memo.funding_note().unwrap(), note,);
+                    }
+                    _ => {
+                        panic!("unexpected memo type")
+                    }
+                }
+                match MemoType::try_from(&funding_output_memo)
```
eranrund, line 3240: This can be shortened to `assert!(matches!(` (or `assert_matches!` is thats already a dependency of this crate)

transaction/std/src/transaction_builder.rs
```diff
@@ -3104,4 +3105,505 @@ pub mod transaction_builder_tests {
...
+            };
+
+            // Values we pretend we get from locating the TxOut using the global index
+            let masked_amount = funding_output.masked_amount.clone();
+
+            // Construct the sender Tx from the combo of "located" and "sent" information
+            let (sending_input_amount, blinding) = masked_amount
+                .get_value(&tx_out_gift_code.shared_secret)
+                .unwrap();
+            println!("{:?}", sending_input_amount);
```
eranrund, line 3277: Debug code leftover

transaction/std/src/transaction_builder.rs
```diff
@@ -3104,4 +3105,505 @@ pub mod transaction_builder_tests {
...
+    fn test_gift_code_transactions() {
+        let mut rng: StdRng = SeedableRng::from_seed([1u8; 32]);
+        let block_version = BlockVersion::MAX;
+        let token_id = TokenId::from(5);
+        let fog_resolver = MockFogResolver::default();
+        let sender = AccountKey::random(&mut rng);
+        let receiver = AccountKey::random(&mut rng);
+        let (funding_input_amt, funding_output_amt, fee) = (1000, 10, 1);
+        let sender_reserved_destinations = ReservedDestination::from(&sender);
+        let receiver_reserved_destinations = ReservedDestination::from(&receiver);
```
eranrund, line 3120: Any reason this is plural and not singular `destination`?

transaction/std/src/transaction_builder.rs
```diff
@@ -3104,4 +3105,505 @@ pub mod transaction_builder_tests {
...
+                .outputs
+                .iter()
+                .find(|tx_out| {
+                    subaddress_matches_tx_out(&receiver, CHANGE_SUBADDRESS_INDEX, tx_out).unwrap()
+                })
+                .expect("Didn't find sender's output");
+
+            validate_tx_out(block_version, change).unwrap();
+
+            assert!(subaddress_matches_tx_out(&receiver, CHANGE_SUBADDRESS_INDEX, change).unwrap());
```
eranrund, line 3345: Not needed, it will never trigger

transaction/std/src/transaction_builder.rs
```diff
@@ -3104,4 +3105,505 @@ pub mod transaction_builder_tests {
...
+                .iter()
+                .find(|tx_out| {
+                    subaddress_matches_tx_out(&sender, CHANGE_SUBADDRESS_INDEX, tx_out).unwrap()
+                })
+                .expect("Didn't find sender's output");
+
+            validate_tx_out(block_version, change).unwrap();
+
+            // Ensure it was sent to the change address
+            assert!(subaddress_matches_tx_out(&sender, CHANGE_SUBADDRESS_INDEX, change).unwrap());
```
eranrund, line 3421: Not needed

transaction/std/src/test_utils.rs
```diff
@@ -201,3 +262,28 @@ pub fn get_transaction<RNG: RngCore + CryptoRng, FPR: FogPubkeyResolver + Clone>
 
     transaction_builder.build(rng)
 }
+
+/// Build simulated change memo with zero amount
+pub fn build_zero_value_change_memo(
```
eranrund, line 267: I don't see this being used anywhere

transaction/std/src/test_utils.rs
```diff
@@ -201,3 +262,28 @@ pub fn get_transaction<RNG: RngCore + CryptoRng, FPR: FogPubkeyResolver + Clone>
...
+    builder: &mut impl MemoBuilder,
+) -> Result<MemoPayload, NewMemoError> {
+    build_change_memo_with_amount(builder, Amount::new(0, 0.into()))
+}
+
+/// Build simulated change memo with amount
+pub fn build_change_memo_with_amount(
```
eranrund, line 274: Same, is this used?

transaction/std/src/transaction_builder.rs
```diff
@@ -3104,4 +3105,505 @@ pub mod transaction_builder_tests {
...
+            let masked_amount = funding_output.masked_amount.clone();
+
+            // Construct the sender Tx from the combo of "located" and "sent" information
+            let (sending_input_amount, blinding) = masked_amount
+                .get_value(&tx_out_gift_code.shared_secret)
+                .unwrap();
+            println!("{:?}", sending_input_amount);
+            let sending_output_amount = Amount::new(sending_input_amount.value - fee, token_id);
+
+            let (ring, real_index) = get_ring_for_txout(
```
eranrund, line 3280: Is it possible to instead use `get_ring` (https://github.com/mobilecoinfoundation/mobilecoin/blob/master/transaction/std/src/test_utils.rs#L65) and replace the TxOut at the `real_index` with `funding_output`? Not sure why the new `get_ring_for_txout` is needed and if we can reduce the amount of test utilities that generate fake data, it will benefit us long term.

eranrund (changes requested):
Thanks for all your work on this. I still have some questions, but nothing too big it seems.
Nice work on the transaction builder tests!

transaction/std/src/memo_builder/gift_code_funding_memo_builder.rs
```diff
@@ -0,0 +1,316 @@
...
+#[derive(Clone, Debug)]
+pub struct GiftCodeFundingMemoBuilder {
+    // Utf-8 note within the memo that can be up to 60 bytes long
+    note: String,
+    // TxOut Public Key of the gift code TxOut sent to the gift code subaddress
+    gift_code_tx_out_public_key: Option<RistrettoPublic>,
+    // Whether we've already written the change memo
+    wrote_change_memo: bool,
+    // Whether we've already written the gift code TxOut
+    wrote_destination_memo: bool,
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: Yeah I had thought about this. While this is true it's implied by being none, if someone needs to maintain this who isn't an original author of it comes along, this makes it more explicit/readable. For the maintainability aspect, I'd vote for retaining it?

transaction/std/src/memo_builder/gift_code_funding_memo_builder.rs
```diff
@@ -0,0 +1,316 @@
...
+            return Err(NewMemoError::MultipleChangeOutputs);
+        }
+        if self.gift_code_tx_out_public_key.as_ref() == Some(memo_context.tx_public_key) {
+            return Err(NewMemoError::BadInputs("The public_key in the memo should be the TxOut at the gift code subaddress and not that of the memo TxOut".into()));
+        }
+        if let Some(tx_out_public_key) = self.gift_code_tx_out_public_key.take() {
+            self.wrote_change_memo = true;
+            Ok(GiftCodeFundingMemo::new(&tx_out_public_key, self.note.as_str())?.into())
+        } else {
+            Err(NewMemoError::MissingInput(
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: Yeah I suppose it would be.

transaction/std/src/transaction_builder.rs
iamalwaysuncomfortable: Indeeed

transaction/std/src/memo_builder/gift_code_funding_memo_builder.rs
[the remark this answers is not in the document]
eranrund: I don't think I agree with that, but don't feel strongly. I think having less `if` statements (one of which will never actually be true due to this redundancy) is more maintainable.

transaction/std/src/transaction_builder.rs
iamalwaysuncomfortable: Because the struct holds methods for multiple destinations?

iamalwaysuncomfortable: The struct itself should probably be renamed ReservedAddresses

transaction/std/src/test_utils.rs
iamalwaysuncomfortable: They're used in the memo builder tests to avoid a ton of boilerplate. I can see them being useful in future memo builder tests

transaction/std/src/test_utils.rs
iamalwaysuncomfortable: Same as above

transaction/std/src/transaction_builder.rs
iamalwaysuncomfortable: I built it to ensure the gift code flow where you're spending a TxOut that belongs to another account worked. I could remove the general method and just put the boilerplate into the actual test file.

transaction/std/src/memo_builder/gift_code_sender_memo_builder.rs
iamalwaysuncomfortable: Noted!

transaction/std/src/memo_builder/gift_code_funding_memo_builder.rs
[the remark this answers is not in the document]
garbageslam: @iamalwaysuncomfortable generally I would argue that less mutable state is more maintainable

if you want to make it more explicit / readable, because you think, `self.wrote_destination_memo` is more readable than `self.tx_out_public_key.is_some()`, another thing you could do is make this a private function instead, so its `self.wrote_destination_memo()` ?

then you don't have an extra variable that someone could forget to update while refactoring this.

transaction/std/src/memo_builder/gift_code_funding_memo_builder.rs
```diff
@@ -0,0 +1,316 @@
...
+/// valued) change TxOut is sent to the sender's reserved change subaddress
+/// with the gift code funding memo attached. The funding memo will include
+/// the first 4 bytes of the hash of the gift code TxOut sent to the
+/// sender's reserved gift code subaddress and 60 bytes for an optional
+/// utf-8 memo.
+#[derive(Clone, Debug)]
+pub struct GiftCodeFundingMemoBuilder {
+    // Utf-8 note within the memo that can be up to 60 bytes long
+    note: String,
+    // TxOut Public Key of the gift code TxOut sent to the gift code subaddress
```
garbageslam, line 32:
```suggestion
    // TxOut Public Key of the gift code TxOut sent to the gift code subaddress
    // This is populated when the output is created
```

garbageslam (approved):
LGTM

Issue #1869, opened 2022-04-27
Simplify Tx protobufs


garbageslam:
One issue encountered in review of #1827 is that, it seems like the protobuf schema we have for `SignatureRctBulletproofs` is not very good:

* There are several different lists of "repeated" elements in the protobuf that "must be the same length", and actually, "must be the same length as something in the TxPrefix", a completely different structure
* We had to add more such lists to accommodate the change

Fundamentally this is happening because, there is a missing struct type here -- there is one concept, which I think is "PseudoOutput", or alternatively, "Input", that has a single associated ring signature, and a single associated pseudo output commitment, and a single associated token id (and a single associated ring).

The immediate impact of this is, messy code with a lot of case analysis: the signature verification starts by checking that lots of obscure lists have the same size, and then later proceeds with lots of parallel indexing into these lists.

This impact is compounded because, in many places in the code we use `.iter().zip(...)` to iterate over pairs of these lists.

The zip function in the standard library [has a curious behavior](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.zip): If the two iterators don't have the same length, then the zipped iterator stops short, without panicking or returning an error.

This means for instance that, if we zip over a list of RingMLSAG's and a list pseudo output commitments, but there are fewer commitments than RingMLSAG's or vice versa, then some of the RingMLSAG's or pseudo output commitments might not be checked for validity at all, which could lead to a money-printing bug.

This means that each of the places where we check that two lists have the same size, are right now, security critical.

There are a few things we could do to try to reduce risk:
1. Restructure the `Tx` protobufs. We could try to move to a `SignatureRctBulletproofs2` object which has one list of "PseudoOutput", containing each item that is needed to validate a pseudo-output and RingMLSAG, one list of "output token id", one list of range proofs.  (I started doing an experimental commit like this still in the SignatureRctBulletproofs object, but it seemed like not a great direction: https://github.com/garbageslam/mobilecoin/pull/6)

This still involves some list length checking, because there are things in the TxPrefix that have to have the same length as these things in the signature part. The TxPrefix contains a list of inputs and ouptuts.

[...]

PR #1934, opened 2022-05-06
Proof of concept for zip-exact iterator

garbageslam:
This is one way of trying to address #1869 

Briefly, our protobuf schemas for transactions are quite messy right now. There are many "repeated" lists of items that need to be the same size (TxIn, RingMLSAG, pseudo output token id). Serialization does not enforce this so it falls to the transaction validation logic.

In the `signature_rct_bulletproofs` file, we have numerous checks that lists that are supposed to be the same size are the same, at the beginning of the file. Later, we zip these lists together.

What would happen if one of these length checks were accidentally skipped, or lost in a refactor?
Then, someone could submit a malformed Tx that has e.g. fewer TxIn than MLSAG objects.

Usually, we process these lists by using rust iterators and `zip`. However `zip` has some potentially dangerous behavior here -- if two lists are not the same size, zip just ignores the extra elements of the longer one, and does not panic or return an error.

It's not hard to imagine that e.g. this could lead to a money printing bug if e.g. the checking of an MLSAG object is just skipped by the transaction validation logic, due to this behavior of zip.

It would be very nice if we could just fix all our Tx protos and group data logically together instead of in parallel lists. However that's a fairly large and disruptive change right now.

An alternative is to make another version of zip that panics or returns an error in this scenario and use that in this file.

We chose not to make it panic. The reasoning is, in the scenario where this is supposed to be helping (a length check was missed), a panic still takes down the enclave, and creates a DOS vector. While not as bad as a money printing bug, it's still bad. So we decided to try to return an error instead.

### DIFF
--- a/transaction/core/src/ring_signature/rct_bulletproofs.rs
+++ b/transaction/core/src/ring_signature/rct_bulletproofs.rs
@@ -19,6 +19,7 @@ use mc_common::HashSet;
 use mc_crypto_digestible::{DigestTranscript, Digestible, MerlinTranscript};
 use mc_crypto_keys::{CompressedRistrettoPublic, RistrettoPrivate};
 use mc_util_serial::prost::Message;
+use mc_util_zip_exact::{zip_exact, ZipExactError};
 use rand_core::{CryptoRng, RngCore};
 use serde::{Deserialize, Serialize};
 
@@ -269,19 +270,26 @@ impl SignatureRctBulletproofs {
             for (token_id, range_proof) in token_ids.iter().zip(self.range_proofs.iter()) {
                 let generator = generator_cache.get(*token_id);
 
-                let commitments: Vec<CompressedRistretto> = self
-                    .pseudo_output_commitments
-                    .iter()
-                    .zip(self.pseudo_output_token_ids.iter())
-                    .chain(output_commitments.iter().zip(self.output_token_ids.iter()))
-                    .filter_map(|(compressed_commitment, this_token_id)| {
-                        if token_id == this_token_id {
-                            Some(compressed_commitment.point)
-                        } else {
-                            None
-                        }
-                    })
-                    .collect();
+                let commitments: Vec<CompressedRistretto> = zip_exact(
+                    self.pseudo_output_commitments.iter(),
+                    self.pseudo_output_token_ids.iter(),
+                )
+                .chain(zip_exact(
+                    output_commitments.iter(),
+                    self.output_token_ids.iter(),
+                ))
+                .filter_map(|result| {
+                    result
+                        .map(|(compressed_commitment, this_token_id)| {
+                            if token_id == this_token_id {
+                                Some(compressed_commitment.point)
+                            } else {
+                                None
+                            }
+                        })
+                        .transpose()
+                })
+                .collect::<Result<_, ZipExactError>>()?;
 
                 if commitments.is_empty() {
                     return Err(Error::NoCommitmentsForTokenId(*token_id));
--- a/util/zip-exact/src/lib.rs
+++ b/util/zip-exact/src/lib.rs
@@ -1,0 +1,114 @@
+#![no_std]
+
+use core::{
+    cmp::{self, Ordering::Equal},
+    fmt::{self, Debug, Display},
+    iter::{Fuse, FusedIterator},
+};
+
+// A variation on ZipLongest originally written by SimonSapin,
+// and dedicated to itertools https://github.com/rust-lang/rust/pull/19283
+//
+// This version iterates two other iterators simultaneously, but returns an
+// error if they do not have the same length, rather than failing silently.
+
+/// An iterator which iterates two other iterators simultaneously
+///
+/// This iterator is *fused*.
+///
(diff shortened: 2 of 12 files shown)


util/zip-exact/src/lib.rs
```diff
@@ -0,0 +1,114 @@
...
+            (Some(a), Some(b)) => Some(Ok((a, b))),
+            _ => Some(Err(ZipExactError {})),
+        }
+    }
+
+    #[inline]
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let (a_lower, a_upper) = self.a.size_hint();
+        let (b_lower, b_upper) = self.b.size_hint();
+        let lower = cmp::min(a_lower, b_lower);
```
remoun, line 65: Nit: `use core::cmp::min`

util/zip-exact/src/lib.rs
```diff
@@ -0,0 +1,114 @@
...
+}
+
+impl<T, U> FusedIterator for ZipExact<T, U>
+where
+    T: Iterator,
+    U: Iterator,
+{
+}
+
+#[derive(Copy, Clone, Default, Debug)]
```
remoun, line 107: Nit: Sort
```suggestion
#[derive(Clone, Copy, Debug, Default)]
```

util/zip-exact/src/lib.rs
```diff
@@ -0,0 +1,114 @@
...
+pub struct ZipExactError {}
+
+impl Display for ZipExactError {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        write!(fmt, "ZipExactError")
```
remoun, line 112: How about making this a bit more descriptive?
```suggestion
        write!(fmt, "zip_exact was called with iterators of mismatched length")
```

transaction/core/src/ring_signature/rct_bulletproofs.rs
```diff
@@ -269,19 +270,26 @@ impl SignatureRctBulletproofs {
...
+                        .map(|(compressed_commitment, this_token_id)| {
+                            if token_id == this_token_id {
+                                Some(compressed_commitment.point)
+                            } else {
+                                None
+                            }
+                        })
+                        .transpose()
+                })
+                .collect::<Result<_, ZipExactError>>()?;
```
garbageslam, line 292: How do people feel about this code? Is this readable?

I feel like I'm not very happy with it, the need to deal with errors returned by the iterator seems to cause a syntax explosion.

Maybe it's better to just use a simple for loop than to try to use iter and collect here?

```
for result in zip_exact(...).chain(zip_exact(...)) {
    let (compressed_commtiment, this_token_id) = result?;

    if token_id == this_token_id {
       compressed_commitments.push(...)
    }
}
```

LMK what you think

garbageslam:
There may also be a simpler approach that would work -- maybe we don't need a fully-generic `zip_exact` because I think we are almost always just iterating over `Vec`. So maybe `zip_exact` can make more assumptions about the arguments e.g. they know their exact length (is this a double-ended iterator?), and then just do like

```
fn zip_exact(a, b) -> Result<Zip<A, B>> {
   if a.len() != b.len() { return Err(ZipExactError); }
   a.zip(b)
}
```

iamalwaysuncomfortable:
This is important to understand:
`Then, someone could submit a malformed Tx that has e.g. fewer TxIn than MLSAG objects.`

Could you explain what the vulnerability would be if this were to happen? And WHY it would happen? 

`It's not hard to imagine that e.g. this could lead to a money printing bug if e.g. the checking of an MLSAG object is just skipped by the transaction validation logic, due to this behavior of zip.`

How does money printing arise if an mlsag object is skipped?

transaction/core/src/ring_signature/rct_bulletproofs.rs
iamalwaysuncomfortable: It feels like in most langs including rust, there's a drive to say "shorter, more language idiomatic code is better" but sometimes it's not in these type of cases where other factors come into play, and I feel like here a clear for loop IS better.

garbageslam:
So to make it concrete, there are a lot of length checks here: https://github.com/mobilecoinfoundation/mobilecoin/blob/a59be4c402b2b71aeb90ac4c4bef29d833501537/transaction/core/src/ring_signature/rct_bulletproofs.rs#L157

garbageslam:
> How does money printing arise if an mlsag object is skipped?

There are all these things that have to be checked for every item, like the validity of the MLSAG, the existence of the item in the ledger. If an MLSAG is not checked for validity, maybe it isn't valid. If it isn't valid, maybe the key image value is random junk. If the key image value is junk but the transaction is still accepted, then I can probably spend my txout an infinite number of times.

iamalwaysuncomfortable (approved):
Agreed with the for loop suggestion that clarifies code + added some sample documentation to make it exceedingly clear what security practices need to be abided by

garbageslam:
closing in favor of https://github.com/mobilecoinfoundation/mobilecoin/pull/1969

Issue #1890, opened 2022-04-29
Add constant for the subaddress which will be used for gift codes (equal to u64::MAX - 2)


iamalwaysuncomfortable:
Add blockchain constant that specified u64::max - 2 is reserved for gift code outputs

PR #1936, opened 2022-05-06
Accomodate Arbitrary Platform Reserved Addresses

iamalwaysuncomfortable:
### Motivation/In This PR

The motivation for [MCIP #36: Reserved Subaddresses](https://github.com/mobilecoinfoundation/mcips/blob/main/text/0036-reserved-subaddresses.md) states:
_In developing new features, we have discovered that sometimes, we want to reserve particular subaddress indices for specific purposes. This is particularly helpful for mobile wallets which are designed to be used with "linked devices" that share your private keys. By using the subaddress index to represent information about a TxOut, they can ensure that other linked devices will all see this information, without needing to create another synchronization mechnaism._

These addresses are index along the following scheme: 
| Subaddress index | Purpose         |
| ---------------- | --------------- |
| `u64::MAX`       | Invalid / None      |
| `u64::MAX - 1`   | Change subaddress [MCIP #4](https://github.com/mobilecoinfoundation/mcips/pull/4)       |
| `u64::MAX - 2`   | Gift code subaddress [MCIP #32](https://github.com/mobilecoinfoundation/mcips/pull/32) |

The supporting `ChangeAddress` object provides data fields for storing the actual derived addresses corresponding to the change subaddress and primary addresses. Given the list of official reserved subaddresses has grown, this PR:

* Expands the `ChangeAddress` object to hold all of the reserved sub-addresses and is renamed to `ReservedAddress`.  
* Adds the gift code reserved subaddresses specified in MCIPs 32 & 36

This PR addresses issue: #1890


android-bindings/src/bindings.rs
```diff
@@ -1747,8 +1747,8 @@ pub unsafe extern "C" fn Java_com_mobilecoin_lib_TransactionBuilder_add_1change_
                 env.get_rust_field(source_account_key, RUST_OBJ_FIELD)?;
 
             let value = jni_big_int_to_u64(env, value)?;
-            let change_destination: ChangeDestination =
-                ChangeDestination::from(&*source_account_key);
+            let change_destination: ReservedDestination =
+                ReservedDestination::from(&*source_account_key);
```
remoun, line 1751: Nit: DRY
```suggestion
            let change_destination = ReservedDestination::from(&*source_account_key);
```

transaction/std/src/reserved_destination.rs
```diff
@@ -0,0 +1,73 @@
...
+        change_subaddress_index: Option<u64>,
+        gift_code_subaddress_index: Option<u64>,
+    ) -> Self {
+        let change_subaddress = if let Some(change_subaddress) = change_subaddress_index {
+            acct.subaddress(change_subaddress)
+        } else {
+            acct.change_subaddress()
+        };
+
+        let gift_code_subaddress = if let Some(gift_code_subaddress) = gift_code_subaddress_index {
```
remoun, line 61: Optional/Equivalent: Use `Into<Option<u64>>` to allow callers to omit `Some()`

```suggestion
        // allows callers to pass `n`, `Some(n)` or `None`
        change_subaddress_index: impl Into<Option<u64>>,
        gift_code_subaddress_index: impl Into<Option<u64>>,
    ) -> Self {
        let change_subaddress = if let Some(change_subaddress) = change_subaddress_index.into() {
            acct.subaddress(change_subaddress)
        } else {
            acct.change_subaddress()
        };

        let gift_code_subaddress = if let Some(gift_code_subaddress) = gift_code_subaddress_index.into() {
```

remoun (approved):
LGTM

transaction/std/src/reserved_destination.rs
iamalwaysuncomfortable: Good suggestion. Let me think on this one, given the sensitivity/implications of the systems this touches, I think I may want to keep the Some() to make it explicit to the caller that they're making an active change to system defaults.

transaction/std/src/reserved_destination.rs
```diff
@@ -0,0 +1,73 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
+//! Object representing a destination for change from a transaction
```
nick-mobilecoin, line 3: don't have a good suggestion, but looks like there is a stray use of "change" from the rename

transaction/std/src/reserved_destination.rs
```diff
@@ -0,0 +1,73 @@
...
+    /// account. It is the account to which all change outputs are actually
+    /// sent. The account owner is able to confirm that an output is change
+    /// by checking that it matches to the change subaddress.
+    /// This should always be the change_subaddress for some AccountKey.
+    pub change_subaddress: PublicAddress,
+
+    /// A secret reserved subaddress to which gift code ouputs are sent
+    /// when they are created. Similarly to the change_subaddress, gift code
+    /// creators can check to see gift codes they've created by checking outputs
+    /// at this address
```
nick-mobilecoin, line 33: nit
```suggestion
    /// at this subaddress
```

nick-mobilecoin (approved):

transaction/std/src/reserved_destination.rs
iamalwaysuncomfortable:
```suggestion
//! Object containing the public destination addresses of MobileCoin standard reserved subaddresses
```

transaction/std/src/reserved_destination.rs
```diff
@@ -0,0 +1,73 @@
...
+            primary_address: src.default_subaddress(),
+            change_subaddress: src.change_subaddress(),
+            gift_code_subaddress: src.gift_code_subaddress(),
+        }
+    }
+}
+
+impl ReservedDestination {
+    /// Set alternate subaddresseses for reserved addresses. This is useful in some things
+    /// like mobilecoind
```
iamalwaysuncomfortable, line 49:
```suggestion
    /// Set alternate subaddresseses for reserved addresses. This is useful in
    /// some things like mobilecoind
```

android-bindings/src/bindings.rs
iamalwaysuncomfortable: Changed!

transaction/std/src/reserved_destination.rs
iamalwaysuncomfortable: Indeed, addressed!

transaction/std/src/reserved_destination.rs
iamalwaysuncomfortable: Added!

PR #1941, opened 2022-05-07
Amount -> MaskedAmount

dolanbernard:
Refactor Android Bindings Amount to MaskedAmount
Updating Android SDK protobufs

### Motivation

This is needed in order for the name of the Java object to match the name of the Rust struct and avoid confusion caused by the AmountData -> Amount Rust refactor.


garbageslam (approved):
Lgtm

PR #1952, opened 2022-05-09
unmask_value -> unmask_amount

dolanbernard:
Renamed unmask_value to unmask_amount
unmask_amount now returns both unmasked value and unmasked token id

### Motivation

This is needed for the SDK to recover the token ID of a masked amount

### Future Work
Parameterize tx builder for token ID


iamalwaysuncomfortable (approved):
Looks good

remoun (approved):
LGTM

Issue #1946, opened 2022-05-09
Add mobilecoind codepaths for implementing gift codes


iamalwaysuncomfortable:
Mobilecoind and the associated mobilecoind-json should both have codepaths for sending gift codes

iamalwaysuncomfortable:
Deprioritized

Issue #1945, opened 2022-05-09
Add libmobilecoin & android ffi functions for building gift codes


iamalwaysuncomfortable:
Mobile apps which build upon the mobilecoin sdks are projected to be the largest consumers of the gift code functionality as they're perfect candidates for sharing a mobilecoin gift with someone who doesn't yet have an account. To accommodate this, code paths for the libmobilecoin (iOS) and android ffi interfaces should be written to enable the mobile SDKs to take advantage of this functionality.

The following should be done to achieve this for iOS/Android
- Add "mc_memo_builder" ffi functions which create gift code memos
- Add "gift_code_transaction_builder" ffi functions which allow specific gift code transactions to be constructed
- Potentially add "gift_code_memo_verification" that can verify gift code messages received are still valid

Issue #1852, opened 2022-04-25
Update `InputCredentials` object to make shared secret easier to derive


iamalwaysuncomfortable:
Update `InputCredentials` object (or perhaps make an `InputCredentialsV2` object) which takes the shared secret and not the view private key. Potentially to ease migration, `InputCredentials` can be made convertible a potential `InputCredentialsV2` and transaction builder can take the new object.

PR #1953, opened 2022-05-10
Add proto schema for GiftCode and rust equivalent object 

iamalwaysuncomfortable:
Add a new protobuf for the gift code message payload which conforms to the new design for gift codes specified in [#MCIP 32](https://github.com/garbageslam/mcips/blob/fog-compatible-gift-codes/text/0032-fog-compatible-gift-codes.md) and rust equivalent object.

### Motivation

A message format needs to be used to communicate that MobileCoin account holder has sent a gift code TxOut with enough information for the receiver to find the TxOut, unmask its amount, and spend the TxOut.

### Future Work
[Add libmobilecoin and android ffis for building gift codes](https://github.com/mobilecoinfoundation/mobilecoin/issues/1945)
[Add mobilecoind codepaths for creating & sending gift codes](https://github.com/mobilecoinfoundation/mobilecoin/issues/1946)
[Update InputCredentials](https://github.com/mobilecoinfoundation/mobilecoin/issues/1852)

### DIFF
--- a/api/proto/printable.proto
+++ b/api/proto/printable.proto
@@ -49,7 +49,7 @@ message TransferPayload {
 /// un-blind the amount, and spend the TxOut. This can be used to give
 /// MobileCoin to a both FOG users & non-FOG users who may not yet have
 // a MobileCoin account enabled
-message GiftCode {
+message TxOutGiftCode {
 
   // The global index of the TxOut that has been gifted. This allows
   // the receiver to find & uniquely identify the TxOut
@@ -70,5 +70,5 @@ message PrintableWrapper { oneof wrapper {
     external.PublicAddress public_address = 1;
     PaymentRequest payment_request = 2;
     TransferPayload transfer_payload = 3;
-    GiftCode gift_code = 4;
+    TxOutGiftCode tx_out_gift_code = 4;
 }}
--- a/api/proto/printable.proto
+++ b/api/proto/printable.proto
@@ -45,20 +45,17 @@ message TransferPayload {
     bytes bip39_entropy = 4;
 }
 
-/// Message encoding information about a TxOut required to locate a TxOut,
+/// Message encoding information required to locate a TxOut,
 /// un-blind the amount, and spend the TxOut. This can be used to give
-/// MobileCoin to a both FOG users & non-FOG users who may not yet have
+/// MobileCoin to both FOG & non-FOG users who may not yet have
 // a MobileCoin account enabled
 message TxOutGiftCode {
-
   // The global index of the TxOut that has been gifted. This allows
   // the receiver to find & uniquely identify the TxOut
   uint64 global_index = 1;
-
   // The one-time private key which can be used to spend the TxOut
   external.RistrettoPrivate onetime_private_key = 2;
-
-  // The shared secret used to unblind the amount of the TxOut
+  // The shared secret used to un-blind the amount of the TxOut
   external.CompressedRistretto shared_secret = 3;
 
 }
--- a/api/proto/printable.proto
+++ b/api/proto/printable.proto
@@ -45,6 +45,24 @@ message TransferPayload {
     bytes bip39_entropy = 4;
 }
 
+/// Message encoding information about a TxOut required to locate a TxOut,
+/// un-blind the amount, and spend the TxOut. This can be used to give
+/// MobileCoin to a both FOG users & non-FOG users who may not yet have
+// a MobileCoin account enabled
+message GiftCode {
+
+  // The global index of the TxOut that has been gifted. This allows
+  // the receiver to find & uniquely identify the TxOut
+  uint64 global_index = 1;
+
+  // The one-time private key which can be used to spend the TxOut
+  external.RistrettoPrivate onetime_private_key = 2;
+
+  // The shared secret used to unblind the amount of the TxOut
+  external.CompressedRistretto shared_secret = 3;
+
+}
+
 /// This wraps all of the above messages using "oneof", allowing us to
 /// have a single encoding scheme and extend as necessary simply by adding
 /// new messages without breaking backwards compatibility
@@ -52,4 +70,5 @@ message PrintableWrapper { oneof wrapper {
     external.PublicAddress public_address = 1;
     PaymentRequest payment_request = 2;
     TransferPayload transfer_payload = 3;
+    GiftCode gift_code = 4;
 }}
--- a/api/src/convert/mod.rs
+++ b/api/src/convert/mod.rs
@@ -44,6 +44,9 @@ mod verification_report;
 mod verification_signature;
 mod watcher;
 
+// printable
+mod tx_out_gift_code;
+
 pub use self::error::ConversionError;
 
 use std::path::PathBuf;
--- a/api/src/convert/tx_out_gift_code.rs
+++ b/api/src/convert/tx_out_gift_code.rs
@@ -26,7 +26,7 @@ impl TryFrom<&printable::TxOutGiftCode> for mc_transaction_core::TxOutGiftCode {
     fn try_from(src: &printable::TxOutGiftCode) -> Result<Self, Self::Error> {
         let global_index = src.get_global_index();
         let onetime_private_key = RistrettoPrivate::try_from(src.get_onetime_private_key())?;
-        let compressed_shared_secret = CompressedRistrettoPublic::try_from(src.get_shared_secret())?
+        let compressed_shared_secret = CompressedRistrettoPublic::try_from(src.get_shared_secret())?;
         let shared_secret = RistrettoPublic::try_from(&compressed_shared_secret)?;
 
         Ok(Self {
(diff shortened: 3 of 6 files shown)


eranrund (changes requested):

api/src/convert/mod.rs
```diff
@@ -44,6 +44,9 @@ mod verification_report;
 mod verification_signature;
 mod watcher;
 
+// printable
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: Because the mods are separated above by the proto modules they addressed

api/proto/printable.proto
```diff
@@ -45,11 +45,30 @@ message TransferPayload {
     bytes bip39_entropy = 4;
 }
 
+/// Message encoding information about a TxOut required to locate a TxOut,
+/// un-blind the amount, and spend the TxOut. This can be used to give
+/// MobileCoin to a both FOG users & non-FOG users who may not yet have
+// a MobileCoin account enabled
+message GiftCode {
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: Yeah I changed it to gift code as I was thinking this would be our way of doing gift codes in the future, but it could be feasible that other types might arise. I'm happy to revert.

transaction/core/src/gift_code.rs
```diff
@@ -0,0 +1,58 @@
...
+    ) -> Result<Amount, AmountError> {
+        let (_, amount) =
+            MaskedAmount::reconstruct(masked_value, masked_token_id, &self.shared_secret)?;
+        Ok(amount)
+    }
+}
+
+impl PartialEq for GiftCode {
+    fn eq(&self, other: &Self) -> bool {
+        self.global_index == other.global_index && self.shared_secret == other.shared_secret
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: Because that doesn't have an equal function impl'd. I believe for the reason that we don't want people doing equalities or ordering operations on private keys (unless they're constant time) as that could give away information to people on what the value of the private keys are.

api/proto/printable.proto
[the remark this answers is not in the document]
eranrund: I don't have any strong feelings about this.

transaction/core/src/gift_code.rs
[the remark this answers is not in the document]
garbageslam: You could call ct_eq I think

api/proto/printable.proto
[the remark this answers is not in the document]
garbageslam: Ideally this would match the filename

transaction/core/src/gift_code.rs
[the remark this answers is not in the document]
eranrund: I don't think that this applies in this particular case, and it's odd that you can have two objects with different keys that end up appearing equal. You can compare the byte representation of the private key.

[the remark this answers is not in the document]
iamalwaysuncomfortable: Yeah that is the odd case, but potentially calling ct_eq - unless the byte comparison is O(1) or something? This is the spend key so it makes sense to guard it, I'll think on this.

api/proto/printable.proto
```diff
@@ -45,11 +45,30 @@ message TransferPayload {
     bytes bip39_entropy = 4;
 }
 
+/// Message encoding information about a TxOut required to locate a TxOut,
```
remoun, line 48:
```suggestion
/// Message encoding information required to locate a TxOut,
```

api/proto/printable.proto
```diff
@@ -45,11 +45,30 @@ message TransferPayload {
...
+/// un-blind the amount, and spend the TxOut. This can be used to give
+/// MobileCoin to a both FOG users & non-FOG users who may not yet have
```
remoun, line 50:
```suggestion
/// MobileCoin to both FOG users & non-FOG users who may not yet have
```

api/proto/printable.proto
```diff
@@ -45,11 +45,30 @@ message TransferPayload {
...
+// a MobileCoin account enabled
+message TxOutGiftCode {
+
```
remoun, line 53:
```suggestion
```

api/proto/printable.proto
```diff
@@ -45,11 +45,30 @@ message TransferPayload {
...
+  // The global index of the TxOut that has been gifted. This allows
+  // the receiver to find & uniquely identify the TxOut
+  uint64 global_index = 1;
+
+  // The one-time private key which can be used to spend the TxOut
+  external.RistrettoPrivate onetime_private_key = 2;
+
+  // The shared secret used to unblind the amount of the TxOut
+  external.CompressedRistretto shared_secret = 3;
+
```
remoun, line 63:
```suggestion
```

transaction/core/src/lib.rs
```diff
@@ -21,6 +21,7 @@ use mc_crypto_keys::{KeyError, RistrettoPrivate, RistrettoPublic};
 mod amount;
 mod blockchain;
 mod domain_separators;
+mod tx_out_gift_code;
 mod memo;
 mod token;
 mod tx_error;
```
remoun, line 27: Nit: Alphabetize
```suggestion
mod memo;
mod token;
mod tx_error;
mod tx_out_gift_code;
```

transaction/core/src/tx_out_gift_code.rs
```diff
@@ -0,0 +1,76 @@
...
+
+    /// Un-blind amount given a masked value and token id
+    pub fn unblind_amount_with_masked_token_id(
+        &self,
+        masked_value: u64,
+        masked_token_id: &[u8],
+    ) -> Result<Amount, AmountError> {
+        let (_, amount) =
+            MaskedAmount::reconstruct(masked_value, masked_token_id, &self.shared_secret)?;
+        Ok(amount)
```
remoun, line 56: Optional/Equivalent: use `Result::map`
```suggestion
        MaskedAmount::reconstruct(masked_value, masked_token_id, &self.shared_secret)
            .map(|(_, amount)| amount)
```

remoun (approved):

api/proto/printable.proto
iamalwaysuncomfortable: Noted, changed along with some other grammar

api/proto/printable.proto
iamalwaysuncomfortable: Noted, changed along with some other grammar

transaction/core/src/lib.rs
iamalwaysuncomfortable: Done!

transaction/core/src/tx_out_gift_code.rs
iamalwaysuncomfortable: Done along with the method above!

api/src/convert/tx_out_gift_code.rs
```diff
@@ -0,0 +1,91 @@
...
+
+/// Convert from printable::TxOutGiftCode --> mc_transaction_core::TxOutGiftCode
+impl TryFrom<&printable::TxOutGiftCode> for mc_transaction_core::TxOutGiftCode {
+    type Error = ConversionError;
+
+    fn try_from(src: &printable::TxOutGiftCode) -> Result<Self, Self::Error> {
+        let global_index = src.get_global_index();
+        let onetime_private_key = RistrettoPrivate::try_from(src.get_onetime_private_key())?;
+        let compressed_shared_secret = CompressedRistrettoPublic::try_from(src.get_shared_secret())
+            .map_err(|_| ConversionError::KeyCastError)?;
```
eranrund, line 30: Why is map_err needed here at all? It appears that the [conversions](https://github.com/mobilecoinfoundation/mobilecoin/blob/master/api/src/convert/compressed_ristretto.rs#L60) are already returning a `ConversionError`?

iamalwaysuncomfortable, line 30:
```suggestion
        let compressed_shared_secret = CompressedRistrettoPublic::try_from(src.get_shared_secret())?
```

iamalwaysuncomfortable: So they are. Removed

eranrund (approved):

api/src/convert/tx_out_gift_code.rs
```diff
@@ -0,0 +1,90 @@
...
+}
+
+/// Convert from printable::TxOutGiftCode --> mc_transaction_core::TxOutGiftCode
+impl TryFrom<&printable::TxOutGiftCode> for mc_transaction_core::TxOutGiftCode {
+    type Error = ConversionError;
+
+    fn try_from(src: &printable::TxOutGiftCode) -> Result<Self, Self::Error> {
+        let global_index = src.get_global_index();
+        let onetime_private_key = RistrettoPrivate::try_from(src.get_onetime_private_key())?;
+        let compressed_shared_secret = CompressedRistrettoPublic::try_from(src.get_shared_secret())?
```
iamalwaysuncomfortable, line 29:
```suggestion
        let compressed_shared_secret = CompressedRistrettoPublic::try_from(src.get_shared_secret())?;
```

PR #1965, opened 2022-05-12
Make fee map enforce divisibility by minimum fee. Also check this in enclave impl

garbageslam:
We determined that, if the minimum fee of a token is not divisible by the global minimum, it can cause an information leakage, because paying the minimum fee for that token does not lead to a priority of 128 like everything else.

There are several possible ways to try to mitigate this, but right now the least disruptive (in terms of changing business logic, and MCIPs) is to enforce this rule in the fee map validation. (And for good measure we check that in the enclave init function.)

This does not cause any problems for any existing currencies like MOB, because the mob minimum fee is 4 * 10^9, and we are only requiring that it is divisible by 2^7, which it is. It is likely that most currencies will have a minimum fee value with > 6 zeroes or so, and so they will easily satisfy this requirement.

### DIFF
--- a/consensus/enclave/api/src/fee_map.rs
+++ b/consensus/enclave/api/src/fee_map.rs
@@ -150,8 +150,7 @@ pub enum Error {
     /// Token `{0}` has invalid fee (too small) `{1}`
     InvalidFeeTooSmall(TokenId, u64),
 
-    /// Token `{0}` has invalid fee (not divisible by smallest minimum fee)
-    /// `{1}`
+    /// Token `{0}` has invalid fee (not divisible) `{1}`
     InvalidFeeNotDivisible(TokenId, u64),
 
     /// Token `{0}` is missing from the fee map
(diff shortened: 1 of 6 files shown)


remoun (approved):

consensus/enclave/api/src/fee_map.rs
```diff
@@ -112,7 +112,14 @@ impl FeeMap {
...
         {
-            return Err(Error::InvalidFee(*token_id, *fee));
+            return Err(Error::InvalidFeeTooSmall(*token_id, *fee));
+        }
+
+        if let Some((token_id, fee)) = minimum_fees
+            .iter()
+            .find(|(_token_id, fee)| (**fee % (1 << SMALLEST_MINIMUM_FEE_LOG2)) != 0)
+        {
+            return Err(Error::InvalidFeeNotDivisible(*token_id, *fee));
```
iamalwaysuncomfortable, line 122: Mainly the same concerns I expressed on the original PR, that requiring integer multiples of 128 given specific tokenomics could be awkward or maybe expensive if token prices vs. fiat go way up. Given that's **purely** hypothetical at the given moment though, this seems a-ok.

iamalwaysuncomfortable (approved):
Mainly the same concerns I expressed on the original PR, that requiring integer multiples of 128 given specific tokenomics could be awkward or maybe expensive if token prices vs. fiat go way up. Given that's purely hypothetical at the given moment though, this seems a-ok.

eranrund (approved):

remoun (approved):

PR #1975, opened 2022-05-13
Rename ReservedDestination To AddressBook

iamalwaysuncomfortable:
* Renamed ReservedDestination to AddressBook + Updated Dependencies

### Motivation

The number of Reserved Addresses has increased and likely will continue to do so. A previous PR created a "reserved destination" object (i.e. renamed from "ChangeAddress") for the purpose of having a way to access any given account's special reserved sub-addresses. That name however is mildly unclear as to its purpose. This suggested rename adds more clarity to the purpose of the object.

### DIFF
--- a/android-bindings/src/bindings.rs
+++ b/android-bindings/src/bindings.rs
@@ -52,8 +52,8 @@ use mc_transaction_core::{
 };
 
 use mc_transaction_std::{
-    AuthenticatedSenderMemo, AuthenticatedSenderWithPaymentRequestIdMemo, DestinationMemo,
-    InputCredentials, MemoBuilder, MemoPayload, RTHMemoBuilder, ReservedDestination,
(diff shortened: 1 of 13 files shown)


garbageslam (changes requested):
I think the name "Address Book" is confusing because it sounds like a contact list, which is not what this is.

Maybe `ReservedAddresses`?

iamalwaysuncomfortable:
> I think the name "Address Book" is confusing because it sounds like a contact list, which is not what this is.
> 
> Maybe `ReservedAddresses`?

Yeah actually good point. Probably will just close the PR.

iamalwaysuncomfortable:
> I think the name "Address Book" is confusing because it sounds like a contact list, which is not what this is.
> 
> Maybe `ReservedAddresses`?

Yeah this a good point - I think that actually negates the need for the PR.

PR #1979, opened 2022-05-16
Make the test client exercise the signed contingent input functionality

garbageslam:
The goal here is to exercise the MCIP #31 functionality in a dev network

This makes the test-client unit-testing mode take multiple token ids instead of just one. Then it runs transfer tests for all of these first, then it does atomic swap transfer tests using the first two tokens (if there are at least two).

Still to do:
- [x] get a passing deployment in a dev network


iamalwaysuncomfortable (approved):

fog/test-client/src/config.rs
```diff
@@ -129,6 +129,10 @@ pub struct TestClientConfig {
     /// What token id to use for the test
     #[clap(long, env = "MC_TOKEN_ID", default_value = "0")]
     pub token_id: TokenId,
+
+    /// Additional token ids to use for the test
+    #[clap(long, env = "MC_EXTRA_TOKEN_IDS", use_value_delimiter = true)]
+    pub extra_token_ids: Vec<TokenId>,
```
garbageslam, line 135: lmk if you guys think it's better to just have `token_ids`, the thing that's annoying is I can't set a default value for it then afaict.

but if we coordinate with ops then we can do a breaking change here i think, which would make the API simpler

iamalwaysuncomfortable: It would be better, but not a huge concern. If it has an ops dependency, something we should address as a separate concern.

fog/test-client/src/test_client.rs
```diff
@@ -676,37 +870,85 @@ impl TestClient {
...
+                let source_client = clients[source_index].clone();
+                let target_client = clients[target_index].clone();
+
+                let transaction = self.test_transfer(
+                    token_id.clone(),
+                    source_client.clone(),
+                    source_index,
+                    target_client,
+                    target_index,
+                )?;
```
iamalwaysuncomfortable, line 896: Potentially add another test with slightly more complex swap list?

iamalwaysuncomfortable (approved):

fog/sample-paykit/src/client.rs
```diff
@@ -362,6 +360,232 @@ impl Client {
...
+    ) -> Result<Tx> {
+        mc_common::trace_time!(self.logger, "MobileCoinClient.build_swap_transaction");
+
+        // Validate the sci
+        sci.validate()?;
+
+        // Check if its key image alreay landed
+        let res = self.fog_key_image.check_key_images(&[sci.key_image()])?;
+        if res.results[0].key_image_result_code == KeyImageResultCode::Spent as u32 {
+            return Err(Error::SciExpired);
```
eranrund, line 471: I'm not sure how useful this is, due to the racy nature of SCIs. It adds some latency to the building, but probably nice for debugging. Nothing actionable, just thinking out loud.

fog/sample-paykit/src/client.rs
```diff
@@ -362,6 +360,232 @@ impl Client {
...
+        tx_builder.add_presigned_input(sci)?;
+
+        // Pay the leftover to ourselves
+        tx_builder.add_change_output(leftover, &change_destination, rng)?;
+
+        // Contribute our own inputs that may be required
+        for (token_id, value) in outlay {
+            // Arbitrarily choose 3 as the maximum number of inputs
+            // TODO: Should be based on fee scaling and fee choice
+            const TARGET_NUM_INPUTS: usize = 3;
```
eranrund, line 563: Any reason for this to not be `MAX_INPUTS - 1`? (1 being used by the SCI)

fog/sample-paykit/src/client.rs
```diff
@@ -720,11 +869,104 @@ fn build_transaction_helper<T: RngCore + CryptoRng, FPR: FogPubkeyResolver>(
...
+            ring,
+            membership_proofs,
+            source_account_key,
+        )?);
+    }
+
+    Ok(())
+}
+
+fn input_credentials_helper(
```
eranrund, line 912: Worth documenting what this is helping with.

fog/test-client/src/config.rs
eranrund: I think you can set a default by making it not required, and when you use it check if it's empty and if so default to whatever you want. Since there's already a `token_ids()` method it can be changed to have this behavior. I would just make the `token_ids` field in this struct not `pub` so this behavior is always enforced.

fog/test-client/src/test_client.rs
```diff
@@ -665,6 +661,204 @@ impl TestClient {
         Ok(transfer_data.transaction)
     }
 
+    /// Conduct an atomic swap transfer between two clients
+    /// Returns the transaction and the block count of the node it was submitted
+    /// to.
+    ///
+    /// The target client builds a signed contingent input, then the source
+    /// client incorporates this into a Tx and submits it.
+    /// This only builds and submits the transaction, it does not confirm it
```
eranrund, line 670:
```suggestion
    /// This only builds and submits the transaction, it does not confirm it.
```

fog/test-client/src/test_client.rs
```diff
@@ -665,6 +661,204 @@ impl TestClient {
...
+            "Attempting to swap {} of {} and {} ({})",
+            self.policy.transfer_amount,
+            token_id1,
+            token_id2,
+            source_client.consensus_service_address()
+        );
+
+        // First do a balance check to flush out any spent txos
+        let tracer = tracer!();
+        tracer.in_span("pre_swap_balance_check", |_cx| {
```
eranrund, line 695:
```suggestion
        tracer.in_span("pre_swap_src_balance_check", |_cx| {
```

fog/test-client/src/test_client.rs
```diff
@@ -665,6 +661,204 @@ impl TestClient {
...
+            source_client
+                .check_balance()
+                .map_err(TestClientError::CheckBalance)
+        })?;
+        tracer.in_span("pre_swap_balance_check", |_cx| {
```
eranrund, line 700:
```suggestion
        tracer.in_span("pre_swap_dst_balance_check", |_cx| {
```

eranrund (approved):
LGTM once CI is green. Thank you! It's nice to see how SCIs gets used with the sample paykit, the API is user-friendly.

fog/sample-paykit/src/client.rs
```diff
@@ -362,6 +360,232 @@ impl Client {
...
+            .map(FogUri::from_str)
+            .collect::<core::result::Result<Vec<_>, _>>()?;
+        let fog_responses = self
+            .fog_report_conn
+            .fetch_fog_reports(fog_uris.into_iter())?;
+        let fog_resolver = FogResolver::new(fog_responses, &self.fog_verifier)?;
+
+        // Check amount found, calculate change
+        let input_amount = input.amount;
+        if offered.token_id != input_amount.token_id || offered.value > input_amount.value {
```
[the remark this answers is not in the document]
garbageslam: thank you

fog/sample-paykit/src/client.rs
garbageslam: yeah i guess this would make more sense in a non-fog client that has the ledger locally

fog/sample-paykit/src/client.rs
```diff
@@ -362,6 +360,232 @@ impl Client {
...
+    ) -> Result<SignedContingentInput> {
+        mc_common::trace_time!(self.logger, "MobileCoinClient.build_swap_proposal");
+
+        // Only one input can be used, otherwise defragmentation is required
+        let inputs = self.tx_data.get_transaction_inputs(offered, 1)?;
+        let inputs: Vec<(OwnedTxOut, TxOutMembershipProof)> = self.get_proofs(&inputs)?;
+        let rings: Vec<Vec<(TxOut, TxOutMembershipProof)>> = self.get_rings(&inputs, rng)?;
+
+        assert_eq!(inputs.len(), 1);
+        assert_eq!(rings.len(), 1);
```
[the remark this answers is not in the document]
garbageslam: I think it's okay here because these are post conditions of the `get_transaction_data` function, so the only thing that makes sense is either to assert, or don't check it and just assume it's true. I'm okay with removing the asserts if you want that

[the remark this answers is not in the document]
remoun: Ah, OK. `assert` SGTM then

fog/sample-paykit/src/client.rs
garbageslam: noted in code comment

fog/sample-paykit/src/client.rs
```diff
@@ -362,6 +360,232 @@ impl Client {
...
+            .map(FogUri::from_str)
+            .collect::<core::result::Result<Vec<_>, _>>()?;
+        let fog_responses = self
+            .fog_report_conn
+            .fetch_fog_reports(fog_uris.into_iter())?;
+        let fog_resolver = FogResolver::new(fog_responses, &self.fog_verifier)?;
+
+        let block_version = BlockVersion::try_from(self.tx_data.get_latest_block_version())?;
+
+        let change_destination = ReservedDestination::from(&self.account_key);
```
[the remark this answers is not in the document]
garbageslam: I guess I went a little further -- since there's only one fog uri here, ours, we don't need iter, filtermap, and Vec, we can just get the fog report url from the account key.

fog/test-client/src/config.rs
garbageslam: you were right remoun, I had some misconceptions about how this works.

fog/test-client/src/test_client.rs
garbageslam: yeah we could do that -- i think it's worth just getting the basic functionality to work first

garbageslam:
This passed deployment here: https://deploybot.development-eu1.mobilecoin.com/job/mobilecoin-internal/job/deploy%252Fbuildtest/37/

fog/sample-paykit/src/client.rs
garbageslam: so, i got this number from the `build_transaction` function in this client already (i'm not sure why it was 3 there)

but in this case, it does make sense for it to be less than `MAX_INPUTS - 1`, because there may be several token ids at work here, and so if they each produce `MAX_INPUTS-1` then we will go over `MAX_INPUTS`.

I'm inclined to just leave it for now

fog/sample-paykit/src/client.rs
```diff
@@ -362,6 +360,280 @@ impl Client {
...
+        rng: &mut T,
+    ) -> Result<Tx> {
+        mc_common::trace_time!(self.logger, "MobileCoinClient.build_swap_transaction");
+
+        // Validate the sci
+        sci.validate()?;
+
+        // Check if its key image alreay landed
+        // Note: An actual fog wallet may not want to do this part, because it adds to
+        // latency, but if you have a local ledger copy you should definitely do this
```
eranrund, line 472: Nit: Might be worth noting that this is still racy - i.e. if this test passes it is no guarantee that the SCI won't expire between now and when it is submitted.

eranrund (approved):
Super cool

PR #1999, opened 2022-05-18
make permission denied log level be info (#1997)

garbageslam:
Logging at error causes a sentry alert, and so if a client like
fog distro is in a retry loop submitting transactions and getting
this error, it can blow out our sentry quota

---

This cherry-picks #1997 from master, there were no conflicts


iamalwaysuncomfortable (approved):
This is good, what it should maybe be accompanied by is metrics on the clients that would be generating this error so that ops can set an alert. Or perhaps if there ARE metrics that are close enough to this already, just setting a prometheus/grafana alert.

eranrund (approved):

jgreat (commented):
I like this change.  In the new release automation I'm not wiring up sentry alerts for development environments.

PR #2008, opened 2022-05-19
Rename ReservedDestination to ReservedSubaddresses

iamalwaysuncomfortable:
*ReservedDestination renamed to ReservedSubaddresses

### Motivation

ReservedDestination is a struct that resolves reserved subaddresses for a given set of account keys. Its naming has led to some confusion. This rename clarifies the purpose of the struct by.


remoun (approved):
LGTM

nick-mobilecoin (approved):
I can't think of a better naming alternative, the plural `ReservedSubaddresses` makes one do a double take. i.e. `ReservedSubaddresses::from()` is that a single item or some kind of collection?

Approving as I can't think of a better alternative and it may just be my newness that the name causes a double take.

garbageslam (approved):
looks good

PR #2013, opened 2022-05-19
Make mobilecoind python able to load both root entropy and mnemonic accounts

garbageslam:
```
    We decided that we probably don't have to make it load the old non-b58
    pub address file format, because we don't need it to be able to send to
    old accounts, only send from old accounts to new accounts. So we didn't
    change that.
```


eranrund (approved):

wjuan-mob (approved):

garbageslam:
Thanks william for collab with me on this

PR #2019, opened 2022-05-20
make test-client and fog-distro able to read keyfiles in both formats

garbageslam:
this was requested by ops to ease release testing

### DIFF
--- a/util/keyfile/src/lib.rs
+++ b/util/keyfile/src/lib.rs
@@ -16,7 +16,7 @@ use bip39::Mnemonic;
 use mc_account_keys::{AccountKey, PublicAddress, RootIdentity};
 use mc_api::printable::PrintableWrapper;
 use std::{
-    convert::TryInto,
+    convert::{TryFrom, TryInto},
     fs::File,
     io::{Read, Write},
     path::Path,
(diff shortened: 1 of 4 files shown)


iamalwaysuncomfortable (approved):

util/keyfile/src/lib.rs
```diff
@@ -53,14 +53,36 @@ pub fn read_root_entropy_keyfile_data<R: Read>(buffer: R) -> Result<RootIdentity
...
+/// from disk
 pub fn read_keyfile<P: AsRef<Path>>(path: P) -> Result<AccountKey, Error> {
     read_keyfile_data(File::open(path)?)
 }
 
-/// Read user root identity from any implementor of `Read`
+/// Read an account key file in either format
 pub fn read_keyfile_data<R: Read>(buffer: R) -> Result<AccountKey, Error> {
-    Ok(serde_json::from_reader::<R, UncheckedMnemonicAccount>(buffer)?.try_into()?)
+    let value = serde_json::from_reader::<R, serde_json::Value>(buffer)?;
```
wjuan-mob, line 74: Should this function be calling read mnemonic_keyfile_data and read_root_entropy_keyfile_data?

wjuan-mob (approved):

util/keyfile/src/lib.rs
garbageslam: Not in my opinion, it could simply be, read any keydata, and if more are created they can be added

garbageslam: sorry misread your comment,

I think it's simpler to parse to json first, inspect it, then decide whether to parse the json one way or the other. because you can't "reset" a `Read` buffer afaik

garbageslam:
this passed deploy

util/keyfile/src/lib.rs
garbageslam: i mean this is basically the same way we made the python code work

PR #2020, opened 2022-05-20
make test-client and fog-distro able to read keyfiles in both formats (#2019)

garbageslam:
* make test-client and fog-distro able to read keyfiles in both formats

* clippy

this is a cherrypick of #2019 , there were no conflicts


wjuan-mob (approved):

jcape (approved):
LGTM, we should put a stake in the ground about being able to use mnemonics at some point...

PR #2027, opened 2022-05-23
Add a new "cargo-check-json" command to mob

DianaNites:
### Motivation

This change allows external tools, such as rust-analyzer, to run cargo check inside the docker environment, improving the development experience for people who can't install the SGX SDK locally.

Changing the RA `rust-analyzer.checkOnSave.overrideCommand` setting to `["./mob", "cargo-check-json"]` allows RA to use this and provide IDE support

### DIFF
--- a/mob
+++ b/mob
@@ -87,11 +87,11 @@ import subprocess
 import sys
 
 parser = argparse.ArgumentParser(prog="mob", description="Perform an action or get a prompt in docker build environment")
-parser.add_argument("action", choices=["prompt", "image", "default-tag", "check"], help="""
+parser.add_argument("action", choices=["prompt", "image", "default-tag", "cargo-check-json"], help="""
                     (prompt) Run bash in the build environment
                     (image) Only create the builder image, don't even run bash
                     (default-tag) Print on stdout the default image tag, to support automation
-                    (check) Run cargo check
+                    (cargo-check-json) Run cargo check
                     """)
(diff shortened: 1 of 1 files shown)


mob
```diff
@@ -438,14 +439,21 @@ if args.container_name is not None:
 # Add image name and command
 docker_run.extend([tag])
 
-docker_run.extend(["/bin/bash"])
+if args.action == "check":
+    docker_run.extend(["/root/.cargo/bin/cargo", "check", "--workspace", "--message-format=json", "--all-targets"])
```
remoun, line 443: How about we make args like `--message-format=json` propagated from `sys.argv`?

garbageslam (approved):
LGTM, thank you

mob
remoun: Fair enough. How about we rename this command to `cargo-check-json`?

mob
```diff
@@ -424,7 +425,7 @@ elif args.ssh_agent:
     docker_run.extend(["--volume", os.environ["SSH_AUTH_SOCK"] + ":" + "/tmp/ssh_auth_sock"])
 
 # debug options allow you to attach gdb when debugging failing tests
-if args.action == "prompt":
+if args.action == "prompt" or args.action == "cargo-check-json":
```
remoun, line 428: Optional/Equivalent: Extract consts for these action names.

remoun (approved):

iamalwaysuncomfortable (approved):

PR #2041, opened 2022-05-24
Add fees to gift code memos

iamalwaysuncomfortable:
### Summary of Changes
* Added fee bytes to gift code memos
* Added methods for setting/extracting fees
* Added extra tests & error checking

### Motivation

It was determined that fees should be added to gift code memos. This PR adds fee bytes to gift code memos and updates tests, checks, and documentation to reflect those changes.

### Points to discuss
* The fees are tracked in big endian bytes, but the index on the gift code cancellation memo was originally in little endian bytes. Shall we just make the index tracked in big endian bytes as well?

### Future Work
* Add method in transaction builder to allow gift codes to be funded
* Work Remaining in: #1735

### DIFF
--- a/transaction/core/src/tx_error.rs
+++ b/transaction/core/src/tx_error.rs
@@ -84,7 +84,7 @@ pub enum NewMemoError {
     Creation(MemoError),
     /// Utf-8 did not properly decode
     Utf8Decoding,
-    /// Max fee of {0} exceeded. Attempted to set fee amount: {1}
+    /// Attempted value: {1} > Max Value: {0}
     MaxFeeExceeded(u64, u64),
     /// Other: {0}
     Other(String),
--- a/transaction/core/src/tx_error.rs
+++ b/transaction/core/src/tx_error.rs
@@ -84,6 +84,8 @@ pub enum NewMemoError {
     Creation(MemoError),
     /// Utf-8 did not properly decode
     Utf8Decoding,
+    /// Max fee of {0} exceeded. Attempted to set fee amount: {1}
+    MaxFeeExceeded(u64, u64),
     /// Other: {0}
     Other(String),
 }
@@ -96,6 +98,9 @@ impl From<MemoError> for NewMemoError {
                 "Input of length: {} exceeded max byte length",
                 byte_len
             )),
+            MemoError::MaxFeeExceeded(max_fee, attempted_fee) => {
+                Self::MaxFeeExceeded(max_fee, attempted_fee)
+            }
         }
     }
 }
--- a/transaction/std/src/memo/gift_code_cancellation.rs
+++ b/transaction/std/src/memo/gift_code_cancellation.rs
@@ -5,13 +5,16 @@
 //! This was proposed for standardization in mobilecoinfoundation/mcips/pull/32
 
 use crate::{impl_memo_type_conversions, RegisteredMemoType};
+use mc_transaction_core::MemoError;
 use std::convert::TryInto;
 
 /// Memo representing the cancellation of a gift code. If a gift code is
 /// never redeemed, the sender may cancel it by sending the TxOut back
 /// to their primary address. This memo will be written to the
-/// reserved change address with 8 bytes reserved for a u64 that
-/// represents the index of the cancelled gift code TxOut
+/// reserved change address with 8 little endian bytes reserved for a u64
+/// that represents the index of the cancelled gift code TxOut and 7 big
+/// endian bytes reserved for recording the fee paid to cancel the gift
+/// code as a 56 bit number.
 #[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd)]
 pub struct GiftCodeCancellationMemo {
     /// The data representing the gift code memo
@@ -23,17 +26,52 @@ impl RegisteredMemoType for GiftCodeCancellationMemo {
 }
 
 impl GiftCodeCancellationMemo {
-    /// The length of the custom memo data
+    /// Number of bytes in the memo
     pub const MEMO_DATA_LEN: usize = 64;
 
-    /// Create a new gift code memo
-    pub fn new(global_index: u64) -> Self {
-        GiftCodeCancellationMemo::from(global_index)
+    /// Number of bytes used to represent the global index of the TxOut used to
+    /// fund the cancelled gift code
+    pub const INDEX_DATA_LEN: usize = 8;
+
+    /// Number of bytes used to represent the fee paid for gift code
+    /// cancellation
+    pub const FEE_DATA_LEN: usize = 7;
+
+    /// Maximum value of the fee we can record
+    pub const MAX_FEE: u64 = u64::MAX >> 8;
+
+    /// Create a new gift code cancellation memo
+    pub fn new(global_index: u64, fee: u64) -> Result<Self, MemoError> {
+        // Check if the fee we're setting is greater than the max fee
+        if fee > Self::MAX_FEE {
+            return Err(MemoError::MaxFeeExceeded(Self::MAX_FEE, fee));
+        }
+
+        let mut memo_data = [0u8; Self::MEMO_DATA_LEN];
+
+        // Put global index of the previously funded gift code into memo
+        memo_data[..Self::INDEX_DATA_LEN]
+            .copy_from_slice(&global_index.to_le_bytes()[..Self::INDEX_DATA_LEN]);
+
+        // Put fee into memo
+        memo_data[Self::INDEX_DATA_LEN..(Self::INDEX_DATA_LEN + Self::FEE_DATA_LEN)]
+            .copy_from_slice(&fee.to_be_bytes()[1..(Self::FEE_DATA_LEN + 1)]);
+
+        Ok(Self { memo_data })
     }
 
-    /// Get global TxOut index of the cancelled gift code
+    /// Get global index of the TxOut used to fund the gift code
     pub fn cancelled_gift_code_index(&self) -> u64 {
-        u64::from_le_bytes(self.memo_data[0..8].try_into().unwrap())
+        u64::from_le_bytes(self.memo_data[..8].try_into().unwrap())
+    }
+
+    /// Get fee amount paid to cancel the gift code
+    pub fn get_fee(&self) -> u64 {
+        let mut fee_bytes = [0u8; 8];
+        fee_bytes[1..8].copy_from_slice(
+            &self.memo_data[Self::INDEX_DATA_LEN..(Self::INDEX_DATA_LEN + Self::FEE_DATA_LEN)],
+        );
+        u64::from_be_bytes(fee_bytes)
     }
 }
 
@@ -63,50 +101,115 @@ impl_memo_type_conversions! { GiftCodeCancellationMemo }
 
 #[cfg(test)]
 mod tests {
-
     use super::*;
 
     #[test]
     fn test_gift_code_cancellation_memo_to_and_from_u64() {
-        let index: u64 = 666;
-        let memo = GiftCodeCancellationMemo::from(index);
+        // Set cancellation index and fee
+        let index = 666;
+        let fee = 20;
+        let memo = GiftCodeCancellationMemo::new(index, fee).unwrap();
 
         // Check recovered index is correct
         assert_eq!(memo.cancelled_gift_code_index(), index);
+
+        // Check recovered fee is correct
+        assert_eq!(memo.get_fee(), fee);
     }
 
     #[test]
     fn test_gift_code_cancellation_memo_at_min_max_bounds_succeed() {
-        let index_min: u64 = 0;
+        // Set fee and minimum and maximum indices
+        let index_min = 0;
         let index_max = u64::MAX;
-        let memo_min = GiftCodeCancellationMemo::new(index_min);
-        let memo_max = GiftCodeCancellationMemo::new(index_max);
+        let fee = 20;
+        let memo_min = GiftCodeCancellationMemo::new(index_min, fee).unwrap();
+        let memo_max = GiftCodeCancellationMemo::new(index_max, fee).unwrap();
 
         // Check recovered indices are correct
         assert_eq!(memo_min.cancelled_gift_code_index(), index_min);
         assert_eq!(memo_max.cancelled_gift_code_index(), index_max);
+
+        // Check recovered fees are correct
+        assert_eq!(memo_min.get_fee(), fee);
+        assert_eq!(memo_max.get_fee(), fee);
     }
 
     #[test]
     fn test_gift_code_cancellation_memo_from_corrupted_bytes_fails() {
+        // Insert bytes representing the index and the fee into an empty array
         let index: u64 = 666;
+        let fee: u64 = 20;
+        let fee_bytes = fee.to_be_bytes();
         let mut memo_bytes = [0u8; GiftCodeCancellationMemo::MEMO_DATA_LEN];
-        memo_bytes[0..8].copy_from_slice(&index.to_le_bytes());
+        memo_bytes[0..GiftCodeCancellationMemo::INDEX_DATA_LEN]
+            .copy_from_slice(&index.to_le_bytes());
+        memo_bytes[GiftCodeCancellationMemo::INDEX_DATA_LEN
+            ..(GiftCodeCancellationMemo::INDEX_DATA_LEN + GiftCodeCancellationMemo::FEE_DATA_LEN)]
+            .copy_from_slice(&fee_bytes[1..(GiftCodeCancellationMemo::FEE_DATA_LEN + 1)]);
+
+        // Corrupt the bytes
         memo_bytes[5] = 124;
+        memo_bytes[10] = 124;
+
+        // Recover the memo
         let memo = GiftCodeCancellationMemo::from(&memo_bytes);
 
-        // Check recovered index is erroneous
+        // Check recovered index is incorrect
         assert_ne!(memo.cancelled_gift_code_index(), index);
+
+        // Check recovered fee is incorrect
+        assert_ne!(memo.get_fee(), fee);
     }
 
     #[test]
     fn test_gift_code_cancellation_memo_from_valid_bytes_is_ok() {
+        // Insert bytes representing the index and the fee into an empty array
         let index: u64 = 666;
+        let fee: u64 = 20;
+        let fee_bytes = fee.to_be_bytes();
         let mut memo_bytes = [0u8; GiftCodeCancellationMemo::MEMO_DATA_LEN];
-        memo_bytes[0..8].copy_from_slice(&index.to_le_bytes());
+        memo_bytes[0..GiftCodeCancellationMemo::INDEX_DATA_LEN]
+            .copy_from_slice(&index.to_le_bytes());
+        memo_bytes[GiftCodeCancellationMemo::INDEX_DATA_LEN
+            ..(GiftCodeCancellationMemo::INDEX_DATA_LEN + GiftCodeCancellationMemo::FEE_DATA_LEN)]
+            .copy_from_slice(&fee_bytes[1..(GiftCodeCancellationMemo::FEE_DATA_LEN + 1)]);
+
+        // Recover the memo
         let memo = GiftCodeCancellationMemo::from(&memo_bytes);
 
-        // Check recovered index is okay
+        // Check recovered index is correct
         assert_eq!(memo.cancelled_gift_code_index(), index);
+
+        // Check recovered fee is correct
+        assert_eq!(memo.get_fee(), fee);
+    }
+
+    #[test]
+    fn test_gift_code_cancellation_memo_fee_boundaries() {
+        // Create index and create fees close to the maximum
+        let index = 666;
+        let fee_max_minus_one = GiftCodeCancellationMemo::MAX_FEE - 1;
+        let fee_max = GiftCodeCancellationMemo::MAX_FEE;
+        let fee_max_plus_one: u64 = GiftCodeCancellationMemo::MAX_FEE + 1;
+
+        // Attempt to instantiate cancellation memos
+        let memo_max_minus_one = GiftCodeCancellationMemo::new(index, fee_max_minus_one).unwrap();
+        let memo_max = GiftCodeCancellationMemo::new(index, fee_max).unwrap();
+        let memo_max_plus_one = GiftCodeCancellationMemo::new(index, fee_max_plus_one);
+
+        // Check recovered index is correct from memos initialized with valid fees
+        assert_eq!(memo_max_minus_one.cancelled_gift_code_index(), index);
+        assert_eq!(memo_max.cancelled_gift_code_index(), index);
+
+        // Check recovered fee is correct from memos initialized with valid fees
+        assert_eq!(memo_max_minus_one.get_fee(), fee_max_minus_one);
+        assert_eq!(memo_max.get_fee(), fee_max);
+
+        // Check memo initialized with bad fee fails
+        assert!(matches!(
+            memo_max_plus_one,
+            Err(MemoError::MaxFeeExceeded(_, _))
+        ));
     }
 }
--- a/transaction/std/src/memo/gift_code_cancellation.rs
+++ b/transaction/std/src/memo/gift_code_cancellation.rs
@@ -68,7 +68,7 @@ impl GiftCodeCancellationMemo {
     /// Get fee amount paid to cancel the gift code
     pub fn get_fee(&self) -> u64 {
         let mut fee_bytes = [0u8; 8];
-        // Copy the 7 fee bytes into a u64 array, leaving the most significant bit 0
+        // Copy the 7 fee bytes into a u64 array, leaving the most significant byte 0
         fee_bytes[1..].copy_from_slice(
             &self.memo_data[Self::INDEX_DATA_LEN..(Self::INDEX_DATA_LEN + Self::FEE_DATA_LEN)],
         );
--- a/transaction/std/src/memo/gift_code_cancellation.rs
+++ b/transaction/std/src/memo/gift_code_cancellation.rs
@@ -10,7 +10,7 @@ use mc_transaction_core::MemoError;
 /// Memo representing the cancellation of a gift code. If a gift code is
 /// never redeemed, the sender may cancel it by sending the TxOut back
 /// to their primary address. This memo will be written to the
-/// reserved change address with 8 little endian bytes reserved for a u64
+/// reserved change address with 8 big endian bytes reserved for a u64
 /// that represents the index of the cancelled gift code TxOut and 7 big
 /// endian bytes reserved for recording the fee paid to cancel the gift
 /// code as a 56 bit number.
@@ -60,7 +60,7 @@ impl GiftCodeCancellationMemo {
 
     /// Get global index of the TxOut used to fund the gift code
     pub fn cancelled_gift_code_index(&self) -> u64 {
-        let mut index_bytes = [0u8; 8];
+        let mut index_bytes = [0u8; Self::INDEX_DATA_LEN];
         index_bytes.copy_from_slice(&self.memo_data[..Self::INDEX_DATA_LEN]);
         u64::from_be_bytes(index_bytes)
     }
@@ -68,6 +68,7 @@ impl GiftCodeCancellationMemo {
     /// Get fee amount paid to cancel the gift code
     pub fn get_fee(&self) -> u64 {
         let mut fee_bytes = [0u8; 8];
+        // Copy the 7 fee bytes into a u64 array, leaving the most significant bit 0
         fee_bytes[1..].copy_from_slice(
             &self.memo_data[Self::INDEX_DATA_LEN..(Self::INDEX_DATA_LEN + Self::FEE_DATA_LEN)],
         );
--- a/transaction/std/src/memo/gift_code_cancellation.rs
+++ b/transaction/std/src/memo/gift_code_cancellation.rs
@@ -6,7 +6,6 @@
 
 use crate::{impl_memo_type_conversions, RegisteredMemoType};
 use mc_transaction_core::MemoError;
-use std::convert::TryInto;
 
 /// Memo representing the cancellation of a gift code. If a gift code is
 /// never redeemed, the sender may cancel it by sending the TxOut back
@@ -50,25 +49,26 @@ impl GiftCodeCancellationMemo {
         let mut memo_data = [0u8; Self::MEMO_DATA_LEN];
 
         // Put global index of the previously funded gift code into memo
-        memo_data[..Self::INDEX_DATA_LEN]
-            .copy_from_slice(&global_index.to_le_bytes()[..Self::INDEX_DATA_LEN]);
+        memo_data[..Self::INDEX_DATA_LEN].copy_from_slice(&global_index.to_be_bytes());
 
         // Put fee into memo
         memo_data[Self::INDEX_DATA_LEN..(Self::INDEX_DATA_LEN + Self::FEE_DATA_LEN)]
-            .copy_from_slice(&fee.to_be_bytes()[1..(Self::FEE_DATA_LEN + 1)]);
+            .copy_from_slice(&fee.to_be_bytes()[1..]);
 
         Ok(Self { memo_data })
     }
 
     /// Get global index of the TxOut used to fund the gift code
     pub fn cancelled_gift_code_index(&self) -> u64 {
-        u64::from_le_bytes(self.memo_data[..8].try_into().unwrap())
+        let mut index_bytes = [0u8; 8];
+        index_bytes.copy_from_slice(&self.memo_data[..Self::INDEX_DATA_LEN]);
+        u64::from_be_bytes(index_bytes)
     }
 
     /// Get fee amount paid to cancel the gift code
     pub fn get_fee(&self) -> u64 {
         let mut fee_bytes = [0u8; 8];
-        fee_bytes[1..8].copy_from_slice(
+        fee_bytes[1..].copy_from_slice(
             &self.memo_data[Self::INDEX_DATA_LEN..(Self::INDEX_DATA_LEN + Self::FEE_DATA_LEN)],
         );
         u64::from_be_bytes(fee_bytes)
@@ -86,14 +86,6 @@ impl From<&[u8; Self::MEMO_DATA_LEN]> for GiftCodeCancellationMemo {
 impl From<GiftCodeCancellationMemo> for [u8; GiftCodeCancellationMemo::MEMO_DATA_LEN] {
     fn from(src: GiftCodeCancellationMemo) -> [u8; GiftCodeCancellationMemo::MEMO_DATA_LEN] {
         src.memo_data
-    }
-}
-
-impl From<u64> for GiftCodeCancellationMemo {
-    fn from(src: u64) -> Self {
-        let mut memo_data = [0u8; Self::MEMO_DATA_LEN];
-        memo_data[0..8].copy_from_slice(&src.to_le_bytes());
-        Self { memo_data }
     }
 }
 
@@ -136,44 +128,17 @@ mod tests {
     }
 
     #[test]
-    fn test_gift_code_cancellation_memo_from_corrupted_bytes_fails() {
-        // Insert bytes representing the index and the fee into an empty array
-        let index: u64 = 666;
-        let fee: u64 = 20;
-        let fee_bytes = fee.to_be_bytes();
-        let mut memo_bytes = [0u8; GiftCodeCancellationMemo::MEMO_DATA_LEN];
-        memo_bytes[0..GiftCodeCancellationMemo::INDEX_DATA_LEN]
-            .copy_from_slice(&index.to_le_bytes());
-        memo_bytes[GiftCodeCancellationMemo::INDEX_DATA_LEN
-            ..(GiftCodeCancellationMemo::INDEX_DATA_LEN + GiftCodeCancellationMemo::FEE_DATA_LEN)]
-            .copy_from_slice(&fee_bytes[1..(GiftCodeCancellationMemo::FEE_DATA_LEN + 1)]);
-
-        // Corrupt the bytes
-        memo_bytes[5] = 124;
-        memo_bytes[10] = 124;
-
-        // Recover the memo
-        let memo = GiftCodeCancellationMemo::from(&memo_bytes);
-
-        // Check recovered index is incorrect
-        assert_ne!(memo.cancelled_gift_code_index(), index);
-
-        // Check recovered fee is incorrect
-        assert_ne!(memo.get_fee(), fee);
-    }
-
-    #[test]
     fn test_gift_code_cancellation_memo_from_valid_bytes_is_ok() {
         // Insert bytes representing the index and the fee into an empty array
         let index: u64 = 666;
         let fee: u64 = 20;
         let fee_bytes = fee.to_be_bytes();
         let mut memo_bytes = [0u8; GiftCodeCancellationMemo::MEMO_DATA_LEN];
-        memo_bytes[0..GiftCodeCancellationMemo::INDEX_DATA_LEN]
-            .copy_from_slice(&index.to_le_bytes());
+        memo_bytes[..GiftCodeCancellationMemo::INDEX_DATA_LEN]
+            .copy_from_slice(&index.to_be_bytes());
         memo_bytes[GiftCodeCancellationMemo::INDEX_DATA_LEN
             ..(GiftCodeCancellationMemo::INDEX_DATA_LEN + GiftCodeCancellationMemo::FEE_DATA_LEN)]
-            .copy_from_slice(&fee_bytes[1..(GiftCodeCancellationMemo::FEE_DATA_LEN + 1)]);
+            .copy_from_slice(&fee_bytes[1..]);
 
         // Recover the memo
         let memo = GiftCodeCancellationMemo::from(&memo_bytes);
@@ -207,9 +172,9 @@ mod tests {
         assert_eq!(memo_max.get_fee(), fee_max);
 
         // Check memo initialized with bad fee fails
-        assert!(matches!(
+        assert_eq!(
             memo_max_plus_one,
-            Err(MemoError::MaxFeeExceeded(_, _))
-        ));
+            Err(MemoError::MaxFeeExceeded(fee_max, fee_max_plus_one))
+        );
     }
 }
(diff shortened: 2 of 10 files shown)


transaction/std/src/memo_builder/gift_code_funding_memo_builder.rs
```diff
@@ -48,16 +51,21 @@ impl GiftCodeFundingMemoBuilder {
...
 }
 
 impl MemoBuilder for GiftCodeFundingMemoBuilder {
     /// Set the fee
-    fn set_fee(&mut self, _fee: Amount) -> Result<(), NewMemoError> {
+    fn set_fee(&mut self, fee: Amount) -> Result<(), NewMemoError> {
+        if self.wrote_change_memo {
+            return Err(NewMemoError::FeeAfterChange);
+        }
+        self.fee = fee;
```
nick-mobilecoin, line 68: If any corrective action is done here it may affect my comment in the cancellation memo of let `make_memo_for_change_output` propagate the error.
In the `new()` method above the note was verified for size and an error occurred immediately, however here the `fee` isn't checked and instead an error is propagated later, this lacks symmetry for the users.  Should always check at the builder interface or should always delegate to the memo type

transaction/std/src/memo_builder/gift_code_sender_memo_builder.rs
```diff
@@ -171,9 +197,26 @@ mod tests {
...
         // Create Memo Builder with an input longer than allowed
-        let note_bytes = [b'6'; GiftCodeSenderMemo::MEMO_DATA_LEN + 1];
+        let note_bytes = [b'6'; GiftCodeSenderMemo::NOTE_DATA_LEN + 1];
         let note = std::str::from_utf8(&note_bytes).unwrap();
         let builder = GiftCodeSenderMemoBuilder::new(note);
         assert!(matches!(builder, Err(NewMemoError::BadInputs(_))));
     }
+
+    #[test]
+    fn test_gift_code_sender_memo_builder_fee_token_cannot_be_different_from_change_token() {
```
nick-mobilecoin, line 207: similar to other builders consider a test case for default fee

nick-mobilecoin (changes requested):
Most things are fairly minor the main thing we need to flush out is byte order it should be consistent, unless we can justify deviating in the same byte buffer.

transaction/std/src/memo/gift_code_cancellation.rs
```diff
@@ -23,17 +26,52 @@ impl RegisteredMemoType for GiftCodeCancellationMemo {
...
-    /// Create a new gift code memo
-    pub fn new(global_index: u64) -> Self {
-        GiftCodeCancellationMemo::from(global_index)
+    /// Number of bytes used to represent the global index of the TxOut used to
+    /// fund the cancelled gift code
+    pub const INDEX_DATA_LEN: usize = 8;
+
+    /// Number of bytes used to represent the fee paid for gift code
+    /// cancellation
+    pub const FEE_DATA_LEN: usize = 7;
```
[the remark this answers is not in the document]
garbageslam: i think we only needed to save 1 byte, so it was suggested to record 7 bytes for the fee instead of 8.

(https://github.com/mobilecoinfoundation/mcips/pull/4)

transaction/std/src/memo/gift_code_cancellation.rs
```diff
@@ -23,17 +26,52 @@ impl RegisteredMemoType for GiftCodeCancellationMemo {
...
+            .copy_from_slice(&fee.to_be_bytes()[1..(Self::FEE_DATA_LEN + 1)]);
+
+        Ok(Self { memo_data })
     }
 
-    /// Get global TxOut index of the cancelled gift code
+    /// Get global index of the TxOut used to fund the gift code
     pub fn cancelled_gift_code_index(&self) -> u64 {
-        u64::from_le_bytes(self.memo_data[0..8].try_into().unwrap())
+        u64::from_le_bytes(self.memo_data[..8].try_into().unwrap())
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: I try to find all of these last night, but alas I failed - thanks for catching it.

transaction/std/src/memo_builder/gift_code_sender_memo_builder.rs
```diff
@@ -27,29 +27,36 @@ pub struct GiftCodeSenderMemoBuilder {
...
     // Whether we've already written the change memo
     wrote_change_memo: bool,
+    // Fee paid to redeem gift code
+    fee: Amount,
 }
 
 impl GiftCodeSenderMemoBuilder {
     /// Initialize memo builder with a utf-8 note (up to 64 bytes), This
     /// method will enforce the 64 byte limit with a NewMemoErr if the
     /// note passed is longer than 64 bytes.
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: Yeah missed here, but I'm reticent to reference constants in the comments as reads kind of awkward

transaction/std/src/memo_builder/gift_code_funding_memo_builder.rs
```diff
@@ -168,46 +189,58 @@ mod tests {
...
         // Test blank note is okay
         {
             let mut builder = GiftCodeFundingMemoBuilder::new(blank_note).unwrap();
 
-            // Build the memo payload and get the data
-            let memo_payload = build_gift_code_memos(&mut builder, &gift_code_public_key).unwrap();
+            // Set fee and build the memo payload
+            builder.set_fee(fee).unwrap();
+            let memo_payload =
+                build_gift_code_memos(&mut builder, &gift_code_public_key, fee).unwrap();
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: Yeah noted on the first point :).

On the second, the only place where gift code memo builders are used is in transaction builders and set fee is always called when transaction builders are created, so I'm going to keep the set fee in build_gift_code_memos as it mimics the behavior of the transaction builder better.

transaction/std/src/memo_builder/gift_code_sender_memo_builder.rs
```diff
@@ -27,29 +27,36 @@ pub struct GiftCodeSenderMemoBuilder {
...
 }
 
 impl MemoBuilder for GiftCodeSenderMemoBuilder {
     /// Set the fee
-    fn set_fee(&mut self, _fee: Amount) -> Result<(), NewMemoError> {
+    fn set_fee(&mut self, fee: Amount) -> Result<(), NewMemoError> {
+        if self.wrote_change_memo {
+            return Err(NewMemoError::FeeAfterChange);
+        }
+        self.fee = fee;
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: Same comment as above

transaction/std/src/memo_builder/gift_code_sender_memo_builder.rs
iamalwaysuncomfortable: Same as above, default fee is never used memo builders

transaction/std/src/memo_builder/gift_code_cancellation_memo_builder.rs
```diff
@@ -57,53 +67,92 @@ impl MemoBuilder for GiftCodeCancellationMemoBuilder {
...
+        )
     }
 }
 
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::test_utils::build_zero_value_change_memo;
-    use std::convert::TryInto;
+    use crate::test_utils::build_change_memo_with_amount;
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: Same as the other cases, memo builder's method of set_fee is always called in the transaction builder, so testing the initialized default doesn't test any intended behavior.

transaction/std/src/memo/gift_code_cancellation.rs
```diff
@@ -23,17 +26,52 @@ impl RegisteredMemoType for GiftCodeCancellationMemo {
...
+        // Check if the fee we're setting is greater than the max fee
+        if fee > Self::MAX_FEE {
+            return Err(MemoError::MaxFeeExceeded(Self::MAX_FEE, fee));
+        }
+
+        let mut memo_data = [0u8; Self::MEMO_DATA_LEN];
+
+        // Put global index of the previously funded gift code into memo
+        memo_data[..Self::INDEX_DATA_LEN]
+            .copy_from_slice(&global_index.to_le_bytes()[..Self::INDEX_DATA_LEN]);
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: Okay cool, I agree. We came up with LE order in initial discussions, but I %100 agree for consistency is better.

transaction/std/src/memo/gift_code_cancellation.rs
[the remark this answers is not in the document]
iamalwaysuncomfortable: I originally thought smaller byte amounts might be fine as well, but fees are paid in picomob there are 1^12 picomob in supply, so 5 bytes would only be able to record up to 1 mob. Other token combos of large supplies + extremely small minimum units could quickly fill up 5 or 6 bytes.

transaction/std/src/memo/gift_code_cancellation.rs
```diff
@@ -23,17 +26,52 @@ impl RegisteredMemoType for GiftCodeCancellationMemo {
...
+
+    /// Number of bytes used to represent the fee paid for gift code
+    /// cancellation
+    pub const FEE_DATA_LEN: usize = 7;
+
+    /// Maximum value of the fee we can record
+    pub const MAX_FEE: u64 = u64::MAX >> 8;
+
+    /// Create a new gift code cancellation memo
+    pub fn new(global_index: u64, fee: u64) -> Result<Self, MemoError> {
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: I had thought about this but ultimately decided against it given the value itself in the struct is stored in u64.

transaction/std/src/memo/gift_code_cancellation.rs
```diff
@@ -63,50 +101,115 @@ impl_memo_type_conversions! { GiftCodeCancellationMemo }
...
+        // Recover the memo
         let memo = GiftCodeCancellationMemo::from(&memo_bytes);
 
-        // Check recovered index is okay
+        // Check recovered index is correct
         assert_eq!(memo.cancelled_gift_code_index(), index);
+
+        // Check recovered fee is correct
+        assert_eq!(memo.get_fee(), fee);
+    }
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: Yeah you're right about this

transaction/std/src/memo/gift_code_cancellation.rs
```diff
@@ -63,50 +101,115 @@ impl_memo_type_conversions! { GiftCodeCancellationMemo }
...
+
+        // Check recovered fee is correct from memos initialized with valid fees
+        assert_eq!(memo_max_minus_one.get_fee(), fee_max_minus_one);
+        assert_eq!(memo_max.get_fee(), fee_max);
+
+        // Check memo initialized with bad fee fails
+        assert!(matches!(
+            memo_max_plus_one,
+            Err(MemoError::MaxFeeExceeded(_, _))
+        ));
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: You're also right about this!

transaction/std/src/memo_builder/gift_code_cancellation_memo_builder.rs
```diff
@@ -34,13 +39,18 @@ impl GiftCodeCancellationMemoBuilder {
...
             wrote_change_memo: false,
+            fee: Amount::new(Mob::MINIMUM_FEE, Mob::ID),
         }
     }
 }
 
 impl MemoBuilder for GiftCodeCancellationMemoBuilder {
     /// Set the fee
-    fn set_fee(&mut self, _fee: Amount) -> Result<(), NewMemoError> {
+    fn set_fee(&mut self, fee: Amount) -> Result<(), NewMemoError> {
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: This question has been raised before and the answer was "no" because these builders work in a way here they're called multiple places along the transaction building journey instead of configured all at once upfront.

transaction/std/src/memo_builder/gift_code_cancellation_memo_builder.rs
```diff
@@ -34,13 +39,18 @@ impl GiftCodeCancellationMemoBuilder {
...
+        if self.wrote_change_memo {
+            return Err(NewMemoError::FeeAfterChange);
+        }
+        self.fee = fee;
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: It could yes, that's actually a good suggestion that I'll make. But it should definitely be in this method instead of ::new()

transaction/std/src/memo/gift_code_cancellation.rs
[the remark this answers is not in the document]
iamalwaysuncomfortable: Changed!

transaction/std/src/memo/gift_code_cancellation.rs
```diff
@@ -23,17 +26,52 @@ impl RegisteredMemoType for GiftCodeCancellationMemo {
...
+    /// Get global index of the TxOut used to fund the gift code
     pub fn cancelled_gift_code_index(&self) -> u64 {
-        u64::from_le_bytes(self.memo_data[0..8].try_into().unwrap())
+        u64::from_le_bytes(self.memo_data[..8].try_into().unwrap())
+    }
+
+    /// Get fee amount paid to cancel the gift code
+    pub fn get_fee(&self) -> u64 {
+        let mut fee_bytes = [0u8; 8];
+        fee_bytes[1..8].copy_from_slice(
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: Addressed!

transaction/std/src/memo/gift_code_cancellation.rs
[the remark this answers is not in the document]
iamalwaysuncomfortable: Changed to be :)

transaction/std/src/memo/gift_code_sender.rs
```diff
@@ -27,23 +29,64 @@ impl RegisteredMemoType for GiftCodeSenderMemo {
...
         };
 
-        str::from_utf8(&self.memo_data[0..index]).map_err(Into::into)
+        str::from_utf8(&self.memo_data[Self::FEE_DATA_LEN..index]).map_err(Into::into)
+    }
+
+    /// Get fee amount paid
+    pub fn get_fee(&self) -> u64 {
+        let mut fee_bytes = [0u8; 8];
+        fee_bytes[1..(GiftCodeSenderMemo::FEE_DATA_LEN + 1)]
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: Done :)

transaction/std/src/memo_builder/gift_code_cancellation_memo_builder.rs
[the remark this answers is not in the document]
garbageslam: nope, our builders in this crate are not exactly following the rust pattern, and that's fine

transaction/std/src/memo/gift_code_cancellation.rs
```diff
@@ -23,17 +25,53 @@ impl RegisteredMemoType for GiftCodeCancellationMemo {
...
+            .copy_from_slice(&fee.to_be_bytes()[1..]);
+
+        Ok(Self { memo_data })
     }
 
-    /// Get global TxOut index of the cancelled gift code
+    /// Get global index of the TxOut used to fund the gift code
     pub fn cancelled_gift_code_index(&self) -> u64 {
-        u64::from_le_bytes(self.memo_data[0..8].try_into().unwrap())
+        let mut index_bytes = [0u8; 8];
```
remoun, line 63: Optional/Equivalent: Use `INDEX_DATA_LEN`

```suggestion
        let mut index_bytes = [0u8; Self::INDEX_DATA_LEN];
```

transaction/std/src/memo/gift_code_cancellation.rs
```diff
@@ -5,13 +5,15 @@
...
 use crate::{impl_memo_type_conversions, RegisteredMemoType};
-use std::convert::TryInto;
+use mc_transaction_core::MemoError;
 
 /// Memo representing the cancellation of a gift code. If a gift code is
 /// never redeemed, the sender may cancel it by sending the TxOut back
 /// to their primary address. This memo will be written to the
-/// reserved change address with 8 bytes reserved for a u64 that
-/// represents the index of the cancelled gift code TxOut
+/// reserved change address with 8 little endian bytes reserved for a u64
```
remoun, line 13:
```suggestion
/// reserved change address with 8 big endian bytes reserved for a u64
```

transaction/std/src/memo/gift_code_cancellation.rs
```diff
@@ -23,17 +25,53 @@ impl RegisteredMemoType for GiftCodeCancellationMemo {
...
-        u64::from_le_bytes(self.memo_data[0..8].try_into().unwrap())
+        let mut index_bytes = [0u8; 8];
+        index_bytes.copy_from_slice(&self.memo_data[..Self::INDEX_DATA_LEN]);
+        u64::from_be_bytes(index_bytes)
+    }
+
+    /// Get fee amount paid to cancel the gift code
+    pub fn get_fee(&self) -> u64 {
+        let mut fee_bytes = [0u8; 8];
+        fee_bytes[1..].copy_from_slice(
```
remoun, line 71: Please document why we're skipping the first byte. I'm guessing it's because the memo uses 7 bytes.

transaction/core/src/tx_error.rs
```diff
@@ -84,6 +84,8 @@ pub enum NewMemoError {
     Creation(MemoError),
     /// Utf-8 did not properly decode
     Utf8Decoding,
+    /// Max fee of {0} exceeded. Attempted to set fee amount: {1}
+    MaxFeeExceeded(u64, u64),
```
garbageslam, line 88: In the other memo builder we were using the `LimitsExceeded` error with string `"fee"`, but we could change those to use this error instead?

transaction/core/src/tx_error.rs
```diff
@@ -84,6 +84,8 @@ pub enum NewMemoError {
     Creation(MemoError),
     /// Utf-8 did not properly decode
     Utf8Decoding,
+    /// Max fee of {0} exceeded. Attempted to set fee amount: {1}
```
garbageslam, line 87:
```suggestion
    /// Attempted to set fee to {1} > {0} max fee
```
Maybe a little terser?

transaction/std/src/memo/gift_code_funding.rs
```diff
@@ -30,36 +32,68 @@ impl RegisteredMemoType for GiftCodeFundingMemo {
...
-        // Compute TxOut hash and store it into the memo data
-        memo_data[0..Self::HASH_DATA_LEN]
+
+        // Compute TxOut hash and put it into the memo data
+        memo_data[..Self::HASH_DATA_LEN]
             .copy_from_slice(&tx_out_public_key_short_hash(tx_out_public_key));
 
+        // Put fee into memo
+        memo_data[Self::HASH_DATA_LEN..Self::NOTE_OFFSET]
+            .copy_from_slice(&fee.to_be_bytes()[1..(Self::FEE_DATA_LEN + 1)]);
```
remoun, line 77: Optional/Equivalent: Use an inclusive range.
```suggestion
            .copy_from_slice(&fee.to_be_bytes()[1..=Self::FEE_DATA_LEN]);
```

transaction/std/src/memo/gift_code_funding.rs
```diff
@@ -68,14 +102,14 @@ impl GiftCodeFundingMemo {
             .memo_data
             .iter()
             .enumerate()
-            .position(|(i, b)| i >= Self::HASH_DATA_LEN && b == &0u8)
+            .position(|(i, b)| i >= Self::NOTE_OFFSET && b == &0u8)
```
remoun, line 105: Optional/Equivalent: Use `skip()` to skip past the offset, and use `*b` instead of `&0`
```suggestion
            .skip(Self::NOTE_OFFSET)
            .position(|b| *b == 0u8)
```

transaction/std/src/memo/gift_code_cancellation.rs
```diff
@@ -23,17 +25,53 @@ impl RegisteredMemoType for GiftCodeCancellationMemo {
...
+            &self.memo_data[Self::INDEX_DATA_LEN..(Self::INDEX_DATA_LEN + Self::FEE_DATA_LEN)],
```
remoun, line 72: Since this `(Self::INDEX_DATA_LEN + Self::FEE_DATA_LEN)` repeats, consider making another const, e.g. `const END_OFFSET: usize = Self::INDEX_DATA_LEN + Self::FEE_DATA_LEN;`

transaction/std/src/memo/gift_code_funding.rs
```diff
@@ -136,70 +174,89 @@ mod tests {
...
 
         // Check that the public key can be verified
         assert!(memo.public_key_matches(&key));
+
+        // Check that the fee is correct
+        assert_eq!(memo.get_fee(), fee);
+
+        // Check that a blank note is extracted correctly
+        let note = "";
+        assert_eq!(memo.funding_note().unwrap(), note);
```
remoun, line 215: Nit Inline
```suggestion
        assert_eq!(memo.funding_note().unwrap(), "");
```

transaction/std/src/memo/gift_code_funding.rs
```diff
@@ -136,70 +174,89 @@ mod tests {
...
     }
 
     #[test]
-    fn test_gift_code_funding_memo_with_only_null_memo_bytes_is_okay() {
+    fn test_gift_code_funding_memo_note_with_only_null_memo_bytes_is_okay() {
         // Initialize hash bytes
         let mut rng: StdRng = SeedableRng::from_seed([1u8; 32]);
         let key = RistrettoPublic::from_random(&mut rng);
+        let fee: u64 = 666;
```
remoun, line 195: Optional/Nit: Use a suffix instead of declaring a type (throughout this PR)
```suggestion
        let fee = 666u64;
```

transaction/std/src/memo/gift_code_funding.rs
```diff
@@ -30,36 +32,68 @@ impl RegisteredMemoType for GiftCodeFundingMemo {
...
+    /// Check if a given public key matches the hash of the gift code TxOut
     pub fn public_key_matches(&self, tx_out_public_key: &RistrettoPublic) -> bool {
-        tx_out_public_key_short_hash(tx_out_public_key) == self.memo_data[0..Self::HASH_DATA_LEN]
+        tx_out_public_key_short_hash(tx_out_public_key) == self.memo_data[..Self::HASH_DATA_LEN]
+    }
+
+    /// Get fee amount paid to fund the gift code
+    pub fn get_fee(&self) -> u64 {
+        let mut fee_bytes = [0u8; 8];
+        fee_bytes[1..(GiftCodeFundingMemo::FEE_DATA_LEN + 1)]
```
remoun, line 94:
```suggestion
        fee_bytes[1..=GiftCodeFundingMemo::FEE_DATA_LEN]
```

transaction/std/src/memo/gift_code_funding.rs
```diff
@@ -233,80 +291,107 @@ mod tests {
...
 
     #[test]
-    fn test_gift_code_funding_memo_created_with_corrupted_bytes_fail() {
-        // Initialize note and hash bytes
-        let note = str::from_utf8(&[b'6'; GiftCodeFundingMemo::NOTE_DATA_LEN - 1]).unwrap();
+    fn test_gift_code_funding_memo_fees_close_to_max_fee_process_as_expected() {
+        // Create notes near max length
+        const MAX_FEE: u64 = GiftCodeFundingMemo::MAX_FEE;
+        const MAX_FEE_MINUS_ONE: u64 = GiftCodeFundingMemo::MAX_FEE - 1;
+        const MAX_FEE_PLUS_ONE: u64 = GiftCodeFundingMemo::MAX_FEE + 1;
```
remoun, line 340: DRY: Reuse the previous const, or just use `+ 1`/`- 1` since that's shorter.

```suggestion
        const MAX_FEE_MINUS_ONE: u64 = MAX_FEE - 1;
        const MAX_FEE_PLUS_ONE: u64 = MAX_FEE + 1;
```

transaction/std/src/memo/gift_code_sender.rs
```diff
@@ -27,23 +29,62 @@ impl RegisteredMemoType for GiftCodeSenderMemo {
...
+            return Err(MemoError::BadLength(note.len()));
+        }
+
+        let mut memo_data = [0u8; Self::MEMO_DATA_LEN];
+
+        // Put fee into memo
+        memo_data[0..Self::FEE_DATA_LEN].copy_from_slice(&fee.to_be_bytes()[1..]);
+
+        // Put note into memo
+        memo_data[Self::FEE_DATA_LEN..(Self::FEE_DATA_LEN + note.len())]
```
remoun, line 61: Is the upper bound necessary?
```suggestion
        memo_data[Self::FEE_DATA_LEN..]
```

transaction/std/src/memo/gift_code_sender.rs
```diff
@@ -27,23 +29,62 @@ impl RegisteredMemoType for GiftCodeSenderMemo {
...
     }
 
     /// Get the sender note
     pub fn sender_note(&self) -> Result<&str, MemoError> {
-        let index = if let Some(terminator) = &self.memo_data.iter().position(|b| b == &0u8) {
+        let index = if let Some(terminator) = &self
+            .memo_data
+            .iter()
+            .enumerate()
+            .position(|(i, b)| i >= Self::FEE_DATA_LEN && b == &0u8)
```
remoun, line 73: Optional/Equivalent: Use `skip()` to skip past the offset, and use `*b` instead of `&0`
```suggestion
            .skip(Self::FEE_DATA_LEN)
            .position(|b| *b == 0u8)
```

transaction/std/src/memo/gift_code_sender.rs
```diff
@@ -154,36 +195,58 @@ mod tests {
...
+        let fee: u64 = 666;
+        let fee_bytes = fee.to_be_bytes();
         let note = str::from_utf8(&note_bytes).unwrap();
 
         // Create memo from note bytes
         let mut memo_bytes = [0u8; GiftCodeSenderMemo::MEMO_DATA_LEN];
-        memo_bytes[0..(GiftCodeSenderMemo::MEMO_DATA_LEN - 1)].copy_from_slice(&note_bytes);
+        memo_bytes[..GiftCodeSenderMemo::FEE_DATA_LEN].copy_from_slice(&fee_bytes[1..]);
+        memo_bytes[GiftCodeSenderMemo::FEE_DATA_LEN
+            ..(GiftCodeSenderMemo::FEE_DATA_LEN + GiftCodeSenderMemo::NOTE_DATA_LEN - 1)]
```
remoun, line 216: IIUC we can omit the upper bound
```suggestion
        memo_bytes[GiftCodeSenderMemo::FEE_DATA_LEN..]
```

transaction/core/src/tx_error.rs
iamalwaysuncomfortable: I i'd propose maybe something like 

```rust
LimitsExceeded('static &str, u64, u64)
```

i.e. domain/max val/attempted val - thoughts?

transaction/std/src/memo_builder/gift_code_funding_memo_builder.rs
```diff
@@ -297,13 +338,48 @@ mod tests {
...
+        let fee = Amount::new(1, 9001.into());
+
+        // Set a fee with a different token id
+        builder.set_fee(fee).unwrap();
+
+        // Attempt to build the memo
+        let memo_payload = build_gift_code_memos(&mut builder, &gift_code_public_key, fee);
+
+        // Ensure memo creation fails
+        assert!(matches!(memo_payload, Err(NewMemoError::MixedTokenIds)))
```
remoun, line 366: Nit: This can use `assert_eq`
```suggestion
        assert_eq!(memo_payload, Err(NewMemoError::MixedTokenIds))
```

transaction/std/src/memo_builder/gift_code_cancellation_memo_builder.rs
```diff
@@ -57,53 +73,109 @@ impl MemoBuilder for GiftCodeCancellationMemoBuilder {
...
+        let memo_payload_2 = build_change_memo_with_amount(&mut builder, amount);
+
+        // Ensure we can recover the index and fee for the first change output
+        let memo = GiftCodeCancellationMemo::from(memo_payload.get_memo_data());
+        assert_eq!(index, memo.cancelled_gift_code_index());
+        assert_eq!(fee.value, memo.get_fee());
 
-        // Assert failure for the second output
+        // Assert failure for the second attempted change output
         assert!(matches!(
```
remoun, line 142: Let's use `assert_matches` (`use assert_matches::assert_matches;`), since this crate already depends on that helper.

transaction/std/src/memo_builder/gift_code_cancellation_memo_builder.rs
```diff
@@ -57,53 +73,109 @@ impl MemoBuilder for GiftCodeCancellationMemoBuilder {
...
+        let fee = Amount::new(1, 9001.into());
+
+        // Set a fee with a different token id
+        builder.set_fee(fee).unwrap();
+
+        // Attempt to build the memo
+        let memo_payload = build_change_memo_with_amount(&mut builder, amount);
+
+        // Ensure memo creation fails
+        assert!(matches!(memo_payload, Err(NewMemoError::MixedTokenIds)))
```
remoun, line 162: Nit: `assert_eq`
```suggestion
        assert_eq!(memo_payload, Err(NewMemoError::MixedTokenIds))
```

transaction/std/src/memo_builder/gift_code_sender_memo_builder.rs
```diff
@@ -171,9 +203,43 @@ mod tests {
...
+        let fee = Amount::new(1, 9001.into());
+
+        // Set a fee with a different token id
+        builder.set_fee(fee).unwrap();
+
+        // Attempt to build the memo
+        let memo_payload = build_change_memo_with_amount(&mut builder, amount);
+
+        // Ensure memo creation fails
+        assert!(matches!(memo_payload, Err(NewMemoError::MixedTokenIds)))
```
remoun, line 226: Nit: `assert_eq`
```suggestion
        assert_eq!(memo_payload, Err(NewMemoError::MixedTokenIds))
```

transaction/std/src/memo/gift_code_cancellation.rs
remoun: Actually this one can probably omit the upper bound
```suggestion
            &self.memo_data[Self::INDEX_DATA_LEN..],
```

transaction/std/src/memo/gift_code_sender.rs
iamalwaysuncomfortable: That would be a mismatch in the number of bytes being copied

transaction/std/src/memo_builder/gift_code_funding_memo_builder.rs
iamalwaysuncomfortable: Done

transaction/std/src/memo_builder/gift_code_cancellation_memo_builder.rs
iamalwaysuncomfortable: Will use assert_eq!

transaction/std/src/memo_builder/gift_code_sender_memo_builder.rs
iamalwaysuncomfortable: Done

transaction/std/src/memo/gift_code_sender.rs
iamalwaysuncomfortable: This actually fails. I believe because skip returns a new iterator with a different indexing

transaction/std/src/memo/gift_code_sender.rs
iamalwaysuncomfortable: Lower bound, and yeah because otherwise we're trying to copy a note of varied length into exactly 57 bytes

transaction/std/src/memo/gift_code_funding.rs
iamalwaysuncomfortable: In going between different memo builders, it's easy to get confused which one you're dealing with, so I chose the longer path

transaction/std/src/memo/gift_code_funding.rs
iamalwaysuncomfortable: I prefered to use the variable to keep it consistent with the rest of the tests

transaction/std/src/memo/gift_code_funding.rs
iamalwaysuncomfortable: I think the previous is a bit more readable

transaction/std/src/memo/gift_code_cancellation.rs
iamalwaysuncomfortable: The bound is definitely necessary here otherwise it's trying to copy 57 bytes into 7

transaction/std/src/memo/gift_code_funding.rs
iamalwaysuncomfortable: I tested this, it fails.

transaction/std/src/memo/gift_code_funding.rs
iamalwaysuncomfortable: Done. I actually didn't know about this, thanks for the tip!

transaction/std/src/memo/gift_code_funding.rs
iamalwaysuncomfortable: Thanks for this tip!

transaction/std/src/memo/gift_code_cancellation.rs
iamalwaysuncomfortable: Yeah the memo used 7, so it leaves the most significant bit zero. Added a comment orating that

transaction/std/src/memo/gift_code_cancellation.rs
iamalwaysuncomfortable: Good nit!

transaction/std/src/memo/gift_code_cancellation.rs
iamalwaysuncomfortable: Done!

transaction/core/src/tx_error.rs
iamalwaysuncomfortable: Went with
```rust
    /// Attempted value: {1} > Max Value: {0}
```

Can modify LimitExceeded to include the numeric data points if you like that idea.

transaction/std/src/memo/gift_code_cancellation.rs
iamalwaysuncomfortable: As far as END_OFFSET, I thikn because there's only 2 constants at play here, keeping the exact math is more telling

transaction/std/src/memo/gift_code_cancellation.rs
```diff
@@ -23,17 +25,54 @@ impl RegisteredMemoType for GiftCodeCancellationMemo {
...
-        u64::from_le_bytes(self.memo_data[0..8].try_into().unwrap())
+        let mut index_bytes = [0u8; Self::INDEX_DATA_LEN];
+        index_bytes.copy_from_slice(&self.memo_data[..Self::INDEX_DATA_LEN]);
+        u64::from_be_bytes(index_bytes)
+    }
+
+    /// Get fee amount paid to cancel the gift code
+    pub fn get_fee(&self) -> u64 {
+        let mut fee_bytes = [0u8; 8];
+        // Copy the 7 fee bytes into a u64 array, leaving the most significant bit 0
```
nick-mobilecoin, line 71: nit:
```suggestion
        // Copy the 7 fee bytes into a u64 array, leaving the most significant byte 0
```

transaction/std/src/memo/gift_code_funding.rs
```diff
@@ -30,36 +32,69 @@ impl RegisteredMemoType for GiftCodeFundingMemo {
...
+    /// Check if a given public key matches the hash of the gift code TxOut
     pub fn public_key_matches(&self, tx_out_public_key: &RistrettoPublic) -> bool {
-        tx_out_public_key_short_hash(tx_out_public_key) == self.memo_data[0..Self::HASH_DATA_LEN]
+        tx_out_public_key_short_hash(tx_out_public_key) == self.memo_data[..Self::HASH_DATA_LEN]
+    }
+
+    /// Get fee amount paid to fund the gift code
+    pub fn get_fee(&self) -> u64 {
+        let mut fee_bytes = [0u8; 8];
+        // Copy the 7 fee bytes into a u64 array, leaving the most significant bit 0
```
nick-mobilecoin, line 94: nit:
```suggestion
        // Copy the 7 fee bytes into a u64 array, leaving the most significant byte 0
```

transaction/std/src/memo/gift_code_sender.rs
```diff
@@ -27,23 +29,63 @@ impl RegisteredMemoType for GiftCodeSenderMemo {
...
         };
 
-        str::from_utf8(&self.memo_data[0..index]).map_err(Into::into)
+        str::from_utf8(&self.memo_data[Self::FEE_DATA_LEN..index]).map_err(Into::into)
+    }
+
+    /// Get fee amount paid
+    pub fn get_fee(&self) -> u64 {
+        let mut fee_bytes = [0u8; 8];
+        // Copy the 7 fee bytes into a u64 array, leaving the most significant bit 0
```
nick-mobilecoin, line 86: nit:
```suggestion
        // Copy the 7 fee bytes into a u64 array, leaving the most significant byte 0
```

transaction/std/src/memo_builder/gift_code_cancellation_memo_builder.rs
```diff
@@ -57,53 +73,106 @@ impl MemoBuilder for GiftCodeCancellationMemoBuilder {
...
+
+        // Ensure memo creation fails
+        assert_eq!(memo_payload, Err(NewMemoError::MixedTokenIds));
+    }
+
+    #[test]
+    fn test_gift_code_cancellation_memo_builder_set_fee_fails_when_exceeding_max_fee() {
+        let index = 666;
+        let mut builder = GiftCodeCancellationMemoBuilder::new(index);
+        let fee = Amount::new(u64::MAX, 0.into());
```
nick-mobilecoin, line 166: This should be testing boundary conditions:
* `GiftCodeCancellationMemo::MAX_FEE`
* `GiftCodeCancellationMemo::MAX_FEE + 1`

transaction/std/src/memo_builder/gift_code_funding_memo_builder.rs
```diff
@@ -290,20 +332,52 @@ mod tests {
...
+
+        // Ensure memo creation fails
+        assert_eq!(memo_payload, Err(NewMemoError::MixedTokenIds));
+    }
+
+    #[test]
+    fn test_gift_code_funding_memo_builder_set_fee_fails_when_exceeding_max_fee() {
+        let note = "It's MEMO TIME!!";
+        let mut builder = GiftCodeFundingMemoBuilder::new(note).unwrap();
+        let fee = Amount::new(u64::MAX, 0.into());
```
nick-mobilecoin, line 371: This should be testing boundary conditions:
* `GiftCodeFundingMemo::MAX_FEE`
* `GiftCodeFundingMemo::MAX_FEE + 1`

transaction/std/src/memo_builder/gift_code_sender_memo_builder.rs
```diff
@@ -154,26 +183,61 @@ mod tests {
...
+
+        // Ensure memo creation fails
+        assert_eq!(memo_payload, Err(NewMemoError::MixedTokenIds))
+    }
+
+    #[test]
+    fn test_gift_code_sender_memo_builder_set_fee_fails_when_exceeding_max_fee() {
+        let note = "It's MEMO TIME!!";
+        let mut builder = GiftCodeSenderMemoBuilder::new(note).unwrap();
+        let fee = Amount::new(u64::MAX, 0.into());
```
nick-mobilecoin, line 231: This should be testing boundary conditions:
* `GiftCodeSenderMemo::MAX_FEE`
* `GiftCodeSenderMemo::MAX_FEE + 1`

nick-mobilecoin (approved):
mostly some minor nits

transaction/std/src/memo_builder/gift_code_sender_memo_builder.rs
iamalwaysuncomfortable: Addressed!

transaction/std/src/memo_builder/gift_code_funding_memo_builder.rs
iamalwaysuncomfortable: Addressed!

transaction/std/src/memo/gift_code_sender.rs
iamalwaysuncomfortable: Addressed!

transaction/std/src/memo/gift_code_funding.rs
iamalwaysuncomfortable: Addressed!

transaction/std/src/memo/gift_code_cancellation.rs
iamalwaysuncomfortable: Addressed!

transaction/core/src/tx_error.rs
garbageslam: sounds good to me!

iamalwaysuncomfortable: Actually would you mind if I did this in a separate PR? It involves changing some of the plumbing of the other memo builders and I'd like to catch any breakage in that scope.

transaction/std/src/memo_builder/gift_code_cancellation_memo_builder.rs
```diff
@@ -57,53 +73,112 @@ impl MemoBuilder for GiftCodeCancellationMemoBuilder {
...
+        amount: Amount,
         _change_destination: &ReservedSubaddresses,
         _memo_context: MemoContext,
     ) -> Result<MemoPayload, NewMemoError> {
         if self.wrote_change_memo {
             return Err(NewMemoError::MultipleChangeOutputs);
         }
+        // fee and change amount token id must match
+        if self.fee.token_id != amount.token_id {
+            return Err(NewMemoError::MixedTokenIds);
```
garbageslam, line 85: nice

garbageslam (approved):
LGTM, thank you!

Issue #1403, opened 2022-01-31
Publishing AVRs with Ledger contents


jcape:
It is a strict improvement of our existing methodology to publish our AVRs alongside our blocks. This eliminates much of the initial necessity for an independent watcher, as it's effectively publish the "certificate" alongside the data it is meant to guarantee.

jcape:
Publishing AVRs alongside blocks is considered a minimum-to-ride for a block streaming API.

PR #2052, opened 2022-05-25
Add `BlockMetadata` to consensus, ledger and watcher

remoun:
Builds on #2058.

### Motivation

* #1403
* https://github.com/mobilecoinfoundation/mcips/pull/43

### Future Work
* #2104


transaction/core/src/blockchain/quorum_set.rs
```diff
@@ -0,0 +1,187 @@
...
+};
+use mc_common::{NodeID, ResponderId};
+use mc_crypto_digestible::Digestible;
+use mc_crypto_keys::Ed25519Public;
+use prost::{Message, Oneof};
+use serde::{Deserialize, Serialize};
+
+/// The quorum set defining the trusted set of peers.
+#[derive(Clone, Deserialize, Digestible, Message, Ord, PartialOrd, Serialize)]
+pub struct QuorumSet {
```
remoun, line 20: I opted to define a minimal/simple `QuorumSet` after grappling with `scp::QuorumSet`, trying to make it implement `Message`, which `LedgerDB` requires to serialize values. Note that this PR also includes `impl [Try]From` to convert back and forth between them.

ledger/db/src/ledger_db.rs
```diff
@@ -538,6 +558,15 @@ impl LedgerDB {
             )?;
         }
 
+        if let Some(metadata) = metadata {
+            db_transaction.put(
+                self.block_metadata,
+                &u64_to_key_bytes(block.index),
+                &encode(metadata),
+                WriteFlags::empty(),
+            )?;
```
remoun, line 567: This is a key part of this PR

consensus/service/src/byzantine_ledger/worker.rs
```diff
@@ -873,7 +873,17 @@ impl<
...
+            block.id.clone(),
+            Some((&self.scp_node.quorum_set()).into()),
+            self.enclave.get_ias_report().ok(),
+        );
+        // TODO: Is this the right key for signing the metadata?
+        let metadata =
+            SignedBlockMetadata::from_contents_and_keypair(metadata, &self.msg_signer_key)
+                .expect("Failed to sign block metadata");
+
+        BlockData::new(block, block_contents, signature, metadata)
```
remoun, line 886: This is a key part of this PR

api/src/convert/archive_block.rs
```diff
@@ -1,115 +1,102 @@
...
+            .map(SignedBlockMetadata::try_from)
+            .transpose()?;
 
-        Ok(mc_transaction_core::BlockData::new(
-            block,
-            block_contents,
-            signature,
-        ))
+        if block.contents_hash == block_contents.hash() {
+            Ok(BlockData::new(block, block_contents, signature, metadata))
```
iamalwaysuncomfortable, line 59: If it's not already in one of the try statements above, the block id should be computed & matched. 

```suggestion
        if block.contents_hash == block_contents.hash() && block.is_block_id_valid() {
            Ok(BlockData::new(block, block_contents, signature, metadata))
```

iamalwaysuncomfortable (commented):
Left an initial comment, will look with fresh eyes in the morning

consensus/service/src/byzantine_ledger/worker.rs
```diff
@@ -873,47 +873,45 @@ impl<
...
-        BlockData::new(block, block_contents, Some(signature))
+        let metadata = BlockMetadata::new(
+            block.id.clone(),
+            Some((&self.scp_node.quorum_set()).into()),
+            self.enclave.get_ias_report().ok(),
+        );
+        // TODO: Is this the right key for signing the metadata?
+        let metadata =
+            SignedBlockMetadata::from_contents_and_keypair(metadata, &self.msg_signer_key)
+                .expect("Failed to sign block metadata");
```
iamalwaysuncomfortable, line 884: If we're signing the Block Metadata, this is definitely the unique message signing key that all consensus nodes are required to have (and send other peers to be considered part of the Quorum process). So if we're saying we're putting Block Metadata on chain signed by consensus nodes, I think this is definitely the key we want.

A subtle, yet significant TODO in this process is finding a mechanism to track a record of these signing keys. If we are forgetting the public key history of the consensus notes, it somewhat defeats the purpose of signing the block as we have no way to verify that these were the authoritative keys at a later point.

consensus/service/src/byzantine_ledger/worker.rs
```diff
@@ -873,47 +873,45 @@ impl<
...
 
-        BlockData::new(block, block_contents, Some(signature))
+        let metadata = BlockMetadata::new(
+            block.id.clone(),
+            Some((&self.scp_node.quorum_set()).into()),
+            self.enclave.get_ias_report().ok(),
+        );
+        // TODO: Is this the right key for signing the metadata?
+        let metadata =
+            SignedBlockMetadata::from_contents_and_keypair(metadata, &self.msg_signer_key)
```
iamalwaysuncomfortable, line 883: Another subtle note is I don't know if from_contents_and_keypair signs %100 correctly. Comments on that in the method definition below.

transaction/core/src/blockchain/block_metadata.rs
```diff
@@ -0,0 +1,123 @@
...
+        }
+    }
+
+    /// Instantiate a [SignedBlockMetadata] by signing the given [BlockMetadata]
+    /// with the given [Ed25519Pair].
+    pub fn from_contents_and_keypair(
+        contents: BlockMetadata,
+        key_pair: &Ed25519Pair,
+    ) -> Result<Self, SignatureError> {
+        let signature = key_pair.try_sign(&contents.digest())?;
```
iamalwaysuncomfortable, line 99: I'm unsure whether we should be using `try_sign_digest` or `try_sign` here. 

Notable
- `digest()` makes a hash of a `message` and I'm unsure whether the `digest` should be considered the `message` itself or whether the `digest` is a `pre-hash` of the message? Perhaps we should be signing a specific byte representation of `BlockMetaData` first? Any advice from @isis-mc or @garbageslam would be welcome here.
- I'm also unsure whether we should have any domain separators added to the digest first. I think we might want that?

iamalwaysuncomfortable (commented):
Notes on the cryptography considerations - will review the rest later.

transaction/core/src/blockchain/quorum_set.rs
```diff
@@ -0,0 +1,187 @@
...
+// This wrapper struct is required because of a peculiarity of prost: you cannot
+// have repeated oneof fields (like a Vec<QuorumSetMember>), so we wrap
+// QuorumSetMember in a struct which implements prost::Message. Unfortunately
+// protobuf also doesn't allow for required oneof fields, so the inner value has
+// to be optional. In practice we expect it to always be Some.
+#[derive(
+    Clone, Deserialize, Digestible, Eq, Hash, Message, Ord, PartialEq, PartialOrd, Serialize,
+)]
+pub struct QuorumSetMemberWrapper {
+    #[prost(oneof = "QuorumSetMember", tags = "1,2")]
```
eranrund, line 39: Why two tags?

eranrund (commented):
Nice work, overall looks great. Did this go through a k8 deployment?

iamalwaysuncomfortable (commented):
This overall looks pretty great. My main questions are:
1. The crypto being used in signing the block metadata needs to be reviewed as I think we may want a domain separator and to double check the block_metadata digest is the correct message to hash
2a. Couple questions about the re-organization. I'd vote for leaving the transaction core crate methods where they are because there's work towards re-organizing that. 
2b. I'm okay with the verifier being moved into a types crate, but others might not be. I don't think our strategy for how to isolate types is formed, so it's a good proposal for maybe how to do that.
3. append_blocks could maybe avoid explicitly annotated lifetimes and we should sanity we're not running into any lifetime errors there vis-a-vis: https://github.com/pretzelhammer/rust-blog/blob/master/posts/common-rust-lifetime-misconceptions.md
4. We should create a tracking issue for all the todos in this PR

api/proto/blockchain.proto
```diff
@@ -70,17 +71,42 @@ message BlockSignature {
     uint64 signed_at = 3;
 }
 
+message BlockMetadata {
+    /// The Block ID.
+    BlockID block_id = 1;
```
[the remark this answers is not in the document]
jcape: The idea is that the signature should be proving that a given quorum set and enclave were used when verifying a given block, and using the ID means you can prove that your "block 42" is the same as mine.

The idea is that you receive all of this at once, and check it as it comes in.

transaction/core/src/blockchain/quorum_set.rs
remoun: Making `GenericNodeId` include `prost::Message` proved problematic -- `mc_common::NodeID` and `mc_common::ResponderId` would also need to implement `Message`, and we don't currently have a `prost` dependency in `common`.

The other main issue was the impedance mismatch between how Prost handles `oneof` (hence `mc_transaction_core::QuorumSetMemberWrapper`) and how `scp::QuorumSet` is structured. I don't remember exactly what I got stuck on there, but I do remember it feeling gnarly, and giving up on that approach.

transaction/core/src/blockchain/block_data.rs
```diff
@@ -1,20 +1,18 @@
 // Copyright (c) 2018-2022 The MobileCoin Foundation
 
-use crate::{Block, BlockContents, BlockSignature};
-use prost::Message;
+use crate::{Block, BlockContents, BlockSignature, SignedBlockMetadata};
 use serde::{Deserialize, Serialize};
 
 /// An object that holds all data included in and associated with a block.
-#[derive(Clone, PartialEq, Eq, Serialize, Deserialize, Message)]
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
```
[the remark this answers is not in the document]
remoun: Oops, re-adding. It was a holdover from #1823, which removed `Message` so I could use `scp::QuorumSet`. Persisting to `LedgerDB` proved that `Message` is necessary.

transaction/core/src/blockchain/quorum_set.rs
remoun: `oneof` keys fields by tag number, and `QuorumSetMember` has two variants.

ledger/db/src/ledger_db.rs
```diff
@@ -141,11 +145,12 @@ impl Ledger for LedgerDB {
...
-    fn append_block(
+    fn append_block<'b>(
         &mut self,
-        block: &Block,
-        block_contents: &BlockContents,
-        signature: Option<BlockSignature>,
+        block: &'b Block,
+        block_contents: &'b BlockContents,
+        signature: Option<&'b BlockSignature>,
+        metadata: Option<&'b SignedBlockMetadata>,
```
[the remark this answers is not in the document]
remoun: I had that initially, but passing `&None` felt clunky. I can switch back if `&None` is acceptable.

consensus/service/src/byzantine_ledger/worker.rs
```diff
@@ -873,47 +873,45 @@ impl<
         // The enclave cannot provide a timestamp, so this happens in untrusted.
         signature.set_signed_at(chrono::Utc::now().timestamp() as u64);
 
-        BlockData::new(block, block_contents, Some(signature))
+        let metadata = BlockMetadata::new(
+            block.id.clone(),
+            Some((&self.scp_node.quorum_set()).into()),
+            self.enclave.get_ias_report().ok(),
```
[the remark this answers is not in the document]
remoun: Added `log::crit!`

consensus/service/src/mint_tx_manager/mod.rs
```diff
@@ -436,7 +439,10 @@ mod mint_config_tx_tests {
             &block_contents,
         );
 
-        ledger.append_block(&block, &block_contents, None).unwrap();
+        // FIXME: Add metadata, too.
```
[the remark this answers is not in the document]
remoun: I've pruned many of them, and am trying to remove all of them with this PR. If any remain, I'll file a tracking issue.

attest/verifier/types/src/verification.rs
```diff
@@ -0,0 +1,126 @@
...
+    /// DER-formatted bytes, from the X-IASReport-Signing-Certificate HTTP
+    /// header.
+    #[prost(bytes, repeated, tag = 2)]
+    pub chain: Vec<Vec<u8>>,
+
+    /// The raw report body JSON, as a byte sequence
+    #[prost(string, required, tag = 3)]
+    #[digestible(never_omit)]
+    pub http_body: String,
+}
```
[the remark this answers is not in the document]
remoun: I opted to split this crate per James' suggestion to minimize the additional footprint for these types. Suggestions welcome for where this crate might be better situated.

transaction/core/src/tx.rs
```diff
@@ -2,36 +2,72 @@
...
+    output: &TxOut,
+) -> Result<bool, KeyError> {
+    let sub_addr_spend = recover_public_subaddress_spend_key(
+        acct.view_private_key(),
+        &RistrettoPublic::try_from(&output.target_key)?,
+        &RistrettoPublic::try_from(&output.public_key)?,
+    );
+    Ok(sub_addr_spend == RistrettoPublic::from(&acct.subaddress_spend_private(subaddress_index)))
+}
+
```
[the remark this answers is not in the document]
remoun: OK, reverted these changes.

transaction/core/src/blockchain/block_metadata.rs
remoun: Let's continue this discussion on https://github.com/mobilecoinfoundation/mobilecoin/pull/2058#discussion_r882283633

iamalwaysuncomfortable (approved):

consensus/service/src/byzantine_ledger/worker.rs
remoun: #2104 tracks this

consensus/service/src/byzantine_ledger/worker.rs
```diff
@@ -874,68 +874,87 @@ impl<
         // The enclave cannot provide a timestamp, so this happens in untrusted.
         signature.set_signed_at(chrono::Utc::now().timestamp() as u64);
 
-        // FIXME: Add metadata.
-        BlockData::new(block, block_contents, signature, None)
+        let metadata = self.get_block_metadata(block.id.clone());
+
+        BlockData::new(block, block_contents, signature, metadata)
```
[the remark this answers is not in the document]
remoun: Agreed! Done.

remoun:
This PR passed a test deploy.

ledger/sync/src/ledger_sync/ledger_sync_service.rs
```diff
@@ -188,27 +188,20 @@ impl<L: Ledger, BC: BlockchainConnection + 'static, TF: TransactionsFetcher + 's
...
-            "Appended {} blocks to ledger",
-            blocks_and_contents.len()
-        );
+        mc_common::trace_time!(self.logger, "Appended {} blocks to ledger", blocks.len());
 
-        for (block, contents) in blocks_and_contents {
+        for block_data in blocks {
             let append_block_start = SystemTime::now();
-            self.ledger.append_block(block, contents, None)?;
+            self.ledger.append_block_data(block_data)?;
```
remoun, line 204: Perhaps we shouldn't append the signature here, and possibly not even the metadata; see discussion at https://github.com/mobilecoinfoundation/mcips/pull/43/files#r898367411

blockchain/types/tests/digest.rs
```diff
@@ -212,55 +190,65 @@ fn origin_block_digestible_ast() {
...
+#[test]
+fn block_contents_digestible_v0() {
     assert_eq!(
-        results[1].1.hash().0,
+        test_blockchain(BlockVersion::ZERO),
         [
-            62, 59, 155, 219, 6, 106, 54, 48, 148, 22, 117, 39, 251, 161, 22, 50, 158, 7, 5, 104,
-            182, 182, 7, 2, 156, 24, 0, 200, 68, 170, 76, 242
+            [
+                203, 245, 7, 219, 1, 210, 70, 210, 93, 109, 135, 251, 188, 10, 120, 155, 108, 240,
```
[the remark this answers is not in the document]
remoun: Because I changed how the blocks are generated, and I couldn't find a way to make it compatible with the old blocks generations. So the options here are to change the bytes, or copy the old code that generated blocks here, which seems unwieldy.

consensus/service/src/byzantine_ledger/worker.rs
```diff
@@ -874,67 +874,84 @@ impl<
...
+        let contents = BlockMetadataContents::new(
+            block_id.clone(),
+            Some(self.scp_node.quorum_set()),
+            verification_report.ok(),
+        );
+
+        let metadata_result =
+            BlockMetadata::from_contents_and_keypair(contents, &self.msg_signer_key);
+        if let Err(err) = &metadata_result {
+            log::crit!(
```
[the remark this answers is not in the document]
garbageslam: The panic handler itself does a log crit, so it would make more sense to panic than to explicitly log crit and then panic

remoun:
I pruned the test-utils refactoring and cleanup from this PR, for easier review. I'll send those out in another PR.

watcher/src/block_data_store.rs
```diff
@@ -19,44 +21,54 @@ use url::Url;
...
     pub block_hash: Vec<u8>,
 
-    /// 32 bytes hash of BlockContent.
-    #[prost(bytes, required, tag = "2")]
+    /// 32 bytes hash of [BlockContents].
+    #[prost(bytes, required, tag = 2)]
     pub block_contents_hash: Vec<u8>,
 
+    /// Hash of [BlockMetadata].
+    #[prost(bytes, required, tag = 4)]
```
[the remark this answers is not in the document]
remoun: Good catch! I'll make it an optional 32 bytes.

ledger/db/src/ledger_db.rs
```diff
@@ -38,6 +39,7 @@ pub const MAX_LMDB_DATABASES: u32 = 27; // maximum number of databases in the lm
 pub const COUNTS_DB_NAME: &str = "ledger_db:counts";
 pub const BLOCKS_DB_NAME: &str = "ledger_db:blocks";
 pub const BLOCK_SIGNATURES_DB_NAME: &str = "ledger_db:block_signatures";
+pub const BLOCK_METADATA_DB_NAME: &str = "ledger_db:block_metadata";
```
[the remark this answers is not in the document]
remoun: Hmm, yeah. I lowered it to 19 and tests still pass -- I count 18 `Database`s under `ledger/db/`, and 1 for `MetadataStore`.

jcape (commented):
The `hash()` method crept back in, otherwise looks good/will approve.

Thanks for trimming it down to size! :-)

eranrund:
Is there ever a case where we would `append_block` with metadata but without a signature? My feeling is no, so perhaps we should not allow that (this will require changing a whole bunch of tests so it doesn't have to be in this PR)

eranrund (commented):
LGTM pending @jcape's comments. I know we are currently having deployment issues, but it would be really nice if this is deployed and tested on an actual network before it gets merged.
It would also be good to test that LedgerDB/WatcherDB can successfully open and work on databases created prior to this merge. There's currently no automated test for this and it will be suboptimal if we find out that for whatever reason something here is incompatible with existing data.

ledger/db/src/ledger_trait.rs
```diff
@@ -14,14 +16,26 @@ use mockall::*;
...
+    /// Appends a block with contents, signature and metadata.
+    fn append_block<'b>(
         &mut self,
-        block: &Block,
-        block_contents: &BlockContents,
-        signature: Option<BlockSignature>,
+        block: &'b Block,
+        block_contents: &'b BlockContents,
+        signature: Option<&'b BlockSignature>,
+        metadata: Option<&'b BlockMetadata>,
```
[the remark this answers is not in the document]
remoun: I'm not sure; rustc is unhappy with the reference in an Option.

Without it I get an `error[E0637]: & without an explicit lifetime name cannot be used here` so I made all the references use the same lifetime.

Compiler output:
```
error[E0637]: `&` without an explicit lifetime name cannot be used here
  --> ledger/db/src/ledger_trait.rs:24:27
   |
24 |         signature: Option<&BlockSignature>,
   |                           ^ explicit lifetime name needed here

error[E0637]: `&` without an explicit lifetime name cannot be used here
  --> ledger/db/src/ledger_trait.rs:25:26
   |
25 |         metadata: Option<&BlockMetadata>,
   |                          ^ explicit lifetime name needed here

error[E0106]: missing lifetime specifier
  --> ledger/db/src/ledger_trait.rs:24:27
   |
24 |         signature: Option<&BlockSignature>,
   |                           ^ expected named lifetime parameter
   |
help: consider introducing a named lifetime parameter
   |
17 ~ #[automock]<'a>
18 | pub trait Ledger: Send {
19 |     /// Appends a block with contents, signature and metadata.
20 |     fn append_block<'b>(
21 |         &mut self,
22 |         block: &Block,
```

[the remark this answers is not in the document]
jcape: Blech, OK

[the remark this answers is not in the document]
garbageslam: oh so it's because of automock's code gen? that makes sense

eranrund (approved):
LGTM but would really prefer if this goes through k8 and a manual test of verifying existing Water/Ledger DBs work fine after this change, both when blocks with and without metadata appear.
Thank you!

remoun:
> LGTM but would really prefer if this goes through k8 and a manual test of verifying existing Water/Ledger DBs work fine after this change, both when blocks with and without metadata appear. Thank you!

Agreed, I want to run both those tests before merging this. Currently blocked on broken deploys.

remoun:
I'm pulling out the bit that makes block metadata required, since that's the tricky part of this PR that we want to test more.

remoun:
> I'm pulling out the bit that makes block metadata required, since that's the tricky part of this PR that we want to test more.

Re-added since it's scoped to BlockVersion 3, and we're still at 0. We still need to address the catch-up mode, which I'll probably do in a follow-up PR.

eranrund:
> > I'm pulling out the bit that makes block metadata required, since that's the tricky part of this PR that we want to test more.
> 
> Re-added since it's scoped to BlockVersion 3, and we're still at v0. We still need to address the catch-up mode, which I'll probably do in a follow-up PR.

What happens to mobilecoind instances that are deployed on test networks with block version 3 turned on?

remoun:
> What happens to mobilecoind instances that are deployed on test networks with block version 3 turned on?

... They would fail. OK, reverting the required bit.

remoun:
PTAL. I switched the `open_db` calls to `create_db` to work with existing DBs.

I'd started adding a LedgerDB migration but it ended up being just a `create_db` call, and found `create_db` works with existing `Database`s; I could add the migration if that is strongly preferred.

ledger/sync/src/ledger_sync/ledger_sync_service.rs
remoun: #2216 handles this

watcher/src/block_data_store.rs
```diff
@@ -19,44 +22,54 @@ use url::Url;
...
     /// Block signature (optional).
     // The signature is unique (we do not expect to encounter duplicate signatures)
     // so we store it inside here.
-    #[prost(message, tag = "3")]
+    #[prost(message, tag = 3)]
     pub signature: Option<BlockSignature>,
+
+    /// Hash of [BlockMetadata].
+    #[prost(bytes, optional, tag = 4)]
+    pub block_metadata_hash: Option<Vec<u8>>,
```
eranrund, line 54: I missed this in previous reviews, but I don't think this should be storing the metadata hash, similarly to how it doesn't store the signature hash.
The idea of storing hashes of the `Block` and `BlockContents` is that they are expected to be identical amongst all instances of a given block, but the signatures will vary for each source - or to phrase differently, for each node we "watch" we will be getting a list of `StoredBlockData` for the blocks produced by that node. For a given block index, the Block and BlockContents will very very likely be the same, so we avoid storing them multiple times in the database and instead store the hash here, and have a separate table to map hash -> actual contents.
We don't do that for the signature since that is expected to be unique, and in most cases the metadata will also be unique - so it should be stored here as `Option<BlockMetadata>`. This will likely simplify the changes here since there won't be a need for the new `BLOCK_METADATA_BY_HASH_DB_NAME` database.
Apologies for not catching it earlier.

eranrund (changes requested):

watcher/src/block_data_store.rs
remoun: Good catch!

> Apologies for not catching it earlier.

No worries; that's why I pared down this PR.

ledger/db/src/ledger_db.rs
```diff
@@ -32,12 +33,15 @@ use std::{
 };
 
 pub const MAX_LMDB_FILE_SIZE: usize = 1 << 40; // 1 TB
-pub const MAX_LMDB_DATABASES: u32 = 27; // maximum number of databases in the lmdb file
+
+/// maximum number of [Database]s in the lmdb file
+pub const MAX_LMDB_DATABASES: u32 = 19;
```
garbageslam, line 38: did we really go down from 27 to 19 here? or this number was just wrong before?

remoun: The number was wrong before. I tried iteratively lowering it, 18 failed but 19 passed.

jcape (approved):

PR #2061, opened 2022-05-26
dev-faucet

garbageslam:
This adds a thin HTTP server called `mobilecoind-dev-faucet`, similar to `mobilecoind-json` but much smaller, which functions as a faucet and can be deployed in dev networks.

This an alternative way that clients can get funds to run tests, which may simplify a lot of the things we do in deployment around funding a large number of accounts of several varieties as part of the deployment. 

---

The server has a background thread which periodically checks if it should split off smaller TxOut's from the main faucet balance, and queues pre-split TxOut's in a tokio queue.

It should be able to handle a relatively high volume of concurrent faucet requests, so if e.g. SDK unit tests hit the faucet many times it won't fail.

I tested this version locally using the local_network.py script, it seems to be working as expected.


mobilecoind-dev-faucet/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+Requesting payment:
+
+```
+$ curl -s localhost:9090/ -d '{"b58_address": "c7f04fcd40d093ca6578b13d790df0790c96e94a77815e5052993af1b9d12923"}' -X POST -H 'Content-type: application/json'
+{"success":true}
+```
+
+```
+$ curl -s localhost:9090/ -d '{"b58_address": "c7f04fcd40d093ca6578b13d790df0790c96e94a77815e5052993af1b9d12923", token_id = "1"}' -X POST -H 'Content-type: application/json'
```
eranrund, line 61:
```suggestion
$ curl -s localhost:9090/ -d '{"b58_address": "c7f04fcd40d093ca6578b13d790df0790c96e94a77815e5052993af1b9d12923", "token_id": "1"}' -X POST -H 'Content-type: application/json'
```

mobilecoind-dev-faucet/src/bin/main.rs
```diff
@@ -0,0 +1,289 @@
...
+    fn lock_and_check_inflight_tx_state(
+        &self,
+    ) -> Result<MutexGuard<Option<SubmitTxResponse>>, String> {
+        let guard = self.inflight_tx_state.lock().expect("mutex poisoned");
+        if let Some(prev_tx) = guard.as_mut() {
+            let mut tries = 10;
+            loop {
+                let resp = self
+                    .mobilecoind_api_client
+                    .get_network_status(&prev_tx)
```
eranrund, line 152: I don't think this is the right call...

eranrund (commented):
Looks good so far, thank you for doing this!

mobilecoind-dev-faucet/src/bin/main.rs
garbageslam: i'm trying to base it on the other PR, so in the `last-block-info` PR as it currently exists, I think this would be the right call. but if we redo that then we might have to change this

mobilecoind-dev-faucet/README.md
garbageslam: thank you!

mobilecoind-dev-faucet/src/bin/main.rs
garbageslam: oh i'm sorry, this is the wrong call, you're right, I'm supposed to call the get tx status thing. whoops

mobilecoind-dev-faucet/src/worker.rs
```diff
@@ -0,0 +1,585 @@
...
+                .collect();
+
+            // First make sure we have enough funds for what we want to do, so we don't spam
+            // errors when we are depleted, and so that faucet users can know
+            // that retries won't help.
+            if non_target_value_utxos
+                .iter()
+                .map(|utxo| utxo.value)
+                .sum::<u64>()
+                < self.target_value * (MAX_OUTPUTS - 1)
```
eranrund, line 497: Might need to account for the fee as well here for this to be accurate.

mobilecoind-dev-faucet/src/worker.rs
```diff
@@ -0,0 +1,585 @@
...
+            // (-1 is for a change output)
+            let mut outlay = mc_mobilecoind_api::Outlay::new();
+            outlay.set_receiver(public_address.clone());
+            outlay.set_value(self.target_value);
+
+            // Generate a Tx
+            let mut req = mc_mobilecoind_api::GenerateTxRequest::new();
+            req.set_sender_monitor_id(monitor_id.to_vec());
+            req.set_token_id(*self.token_id);
+            req.set_input_list(RepeatedField::from_vec(non_target_value_utxos));
```
eranrund, line 520: `.into()` should work instead of `RepeatedField::from_vec`

eranrund (commented):
Some small nits but other than that looking good.

mobilecoind-dev-faucet/src/data_types.rs
```diff
@@ -0,0 +1,157 @@
...
+    fn from(src: JsonU64) -> u64 {
+        src.0
+    }
+}
+
+impl AsRef<u64> for JsonU64 {
+    fn as_ref(&self) -> &u64 {
+        &self.0
+    }
+}
```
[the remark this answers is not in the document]
garbageslam: good idea

mobilecoind-dev-faucet/src/bin/main.rs
```diff
@@ -0,0 +1,317 @@
...
+    /// The account key holding our funds
+    pub account_key: AccountKey,
+    /// The bytes of our monitor id, which holds the faucet's funds
+    pub monitor_id: Vec<u8>,
+    /// The public address of the faucet, which someone can use to replenish the
+    /// faucet
+    pub monitor_b58_address: String,
+    /// The amounts the faucet attempts to pay for each token id
+    /// This is initialized to network fee * amount factor at startup
+    pub faucet_amounts: HashMap<TokenId, u64>,
```
iamalwaysuncomfortable, line 79: Maybe more clarification these are the payout amounts?

```suggestion
    pub faucet_payout_amounts: HashMap<TokenId, u64>,
```

OR 

```suggestion
    pub faucet_payouts: HashMap<TokenId, u64>,
```

mobilecoind-dev-faucet/src/worker.rs
```diff
@@ -0,0 +1,602 @@
...
+//!
+//! Other threads may dequeue UTXOs from such a queue and then use them.
+//! The queue also carries a one-shot channel which can give the worker a
+//! SubmitTxResponse, or signal that an error occurred. The worker then follows
+//! up on what happens with the UTXO, and eventually requeues it if the
+//! transaction doesn't resolve successfully.
+//!
+//! The worker uses its own thread, but uses async-friendly tokio primitives.
+//! The worker does not require to be launched from the context of a tokio
+//! runtime.
```
iamalwaysuncomfortable, line 19: It kind of seems like we're doing a custom implementation of futures and an executor here? This is pretty impressive what you made here - but a lot of it also seems like it could be done using Rust's normal async primitives + an executor like Tokio/Block-on/Async-std.

We're avoiding using the Tokio executor - but why exactly? It seems like there's a boilerplate could be avoided by functionality already existing in Tokio and other executors like spawned tasks, threads, and a ton of helper structs that provide things like queues that can work in an async context:
https://docs.rs/futures/0.3.21/futures/future/index.html
https://docs.rs/futures/0.3.21/futures/stream/index.html

I.e. like these for qeueing up tasks:
https://docs.rs/futures/0.3.21/futures/stream/struct.FuturesUnordered.html
or 
https://docs.rs/futures/0.3.21/futures/stream/struct.FuturesOrdered.html

I think we lose some level of maintaibility here, ability to work with native rust concurrency, and also some pre-done optimization that other executors have. 

If this already works and the faucet is ready to go, we can/should go with this, but it could in the future potentially we could go towards a more "native async" approach.

garbageslam: I don't think I made a custom implementation of futures or an executor here -- I think all I did was create a background worker thread. 

I considered doing it another way:
* Replace `std::thread::spawn` with `tokio::task::spawn`
* Replace `std::thread::sleep` with `tokio::task::sleep(..).await`.

The main reason that I didn't do that is because I wasn't sure if I'm allowed to await futures that come from grpcio in a task spawned by the tokio stuff (that was partly why I asked questions earlier).

Anyways, it seems like there are fewer questions that arise if I just make it an `std::thread`, and the worker doesn't really need parallelism or more than one thread. So I wasn't sure what was gained by making it a tokio task instead.

Note also that, most of our rust web services right now, including consensus, fog ingest, fog view, etc. all spawn background threads this way and communicate either using queues, or sometimes just std lib mutexes. So I guess I just felt very comfortable making the background worker an std thread.

The other thing was, by making it a tokio task, it's possible that the worker thread can be starved if there are a lot requests that come in, because all the OS threads could be assigned to the rocket task I guess. Maybe it doesn't really matter. But it seemed a little easier to think about in what circumstances the worker could get blocked if I don't make it use the same thread pool as the HTTP server.

---

I'm sure that there are things that can be improved about the worker thread if it is async, the main one being, instead of sleeping a fixed amount of time, you could do some kind of select over the one-shot receivers and the timer, so that it wakes up immediately if there might be something to do. You could also do that using an std condition variable I think but select is better for various reasons. But I'm not sure if having the worker thread wake up immediately in those scenarios matters. And just using `std::thread` this way seemed simpler to me.

---

Anyways if you see a major benefit to changing it to `tokio::Task` then I'm happy to change this, but I'm curious if in your mind this is more than substituting `std::thread::spawn` with `tokio::task` and `std::thread::sleep` with `tokio::time::sleep.await`.

I don't see that the Streams stuff is relevant here -- if it became necessary to use those APIs, we could turn the mpsc queue endpoint into a stream and then use it that way: https://docs.rs/tokio-stream/0.1.8/tokio_stream/wrappers/struct.ReceiverStream.html

iamalwaysuncomfortable: Also thank you for doing this, this is an *excellent* feature

mobilecoind-dev-faucet/src/data_types.rs
```diff
@@ -0,0 +1,111 @@
...
+
+/// Related to (but not the same as) mobilecoind_api::SubmitTxResponse
+///
+/// This json includes a "success" field and an "err_str" field, so it is
+/// effectively like an enum over `SubmitTxResponse` and `String`.
+///
+/// The `From` conversions set `success` to true or false appropriately.
+/// In the success case, we only include the receiver tx receipt list, because
+/// the faucet user cannot make use of the sender tx receipt.
+#[derive(Deserialize, Serialize, Default, Debug)]
```
remoun, line 90: Nit: Sort
```suggestion
#[derive(Debug, Default, Deserialize, Serialize)]
```

mobilecoind-dev-faucet/src/bin/main.rs
```diff
@@ -0,0 +1,353 @@
...
+    #[clap(long, default_value = "20", env = "MC_TARGET_QUEUE_DEPTH")]
+    pub target_queue_depth: usize,
+
+    /// Worker poll period in milliseconds.
+    #[clap(long, default_value = "100", env = "MC_WORKER_POLL_PERIOD_MS")]
+    pub worker_poll_period_ms: u64,
+}
+
+/// Connection to the mobilecoind client
+struct State {
```
remoun, line 67: Consider moving this `State` into a `mod` so we could test it

mobilecoind-dev-faucet/src/bin/main.rs
```diff
@@ -0,0 +1,353 @@
...
+        .merge(("port", config.listen_port))
+        .merge(("address", config.listen_host.clone()));
+
+    let state = State::new(&config, &logger).expect("Could not initialize");
+
+    let _rocket = rocket::custom(figment)
+        .mount("/", routes![post, status])
+        .manage(state)
+        .launch()
+        .await?;
```
remoun, line 351: Consider extracting a helper for this Rocket/server setup, a la https://github.com/mobilecoinfoundation/mobilecoin/blob/master/fog/overseer/server/src/server.rs

mobilecoind-dev-faucet/src/worker.rs
```diff
@@ -0,0 +1,602 @@
...
+            worker_token_states.push(state);
+            receivers.insert(*token_id, receiver);
+        }
+
+        let stop_requested = Arc::new(AtomicBool::default());
+        let thread_stop_requested = stop_requested.clone();
+
+        let logger = logger.new(o!("thread" => "worker"));
+
+        let join_handle = Some(std::thread::spawn(move || {
```
remoun, line 218: Please move this thread worker logic into a separate method

mobilecoind-dev-faucet/src/worker.rs
```diff
@@ -0,0 +1,602 @@
...
+        // If it is new and has the target value, then queue it
+        let mut output_list_key_images = HashSet::<KeyImage>::default();
+
+        for utxo in resp.output_list.iter() {
+            // Sanity check the token id
+            if utxo.token_id != self.token_id {
+                continue;
+            }
+
+            if utxo.value != self.target_value {
```
remoun, line 453: Please document what this check is for.

remoun (commented):
LGTM -- one question, otherwise nits.

mobilecoind-json/Cargo.toml
```diff
@@ -21,13 +22,13 @@ protobuf = "2.27.1"
...
 serde_derive = "1.0"
-serde_with = "1.13"
 
 [dev-dependencies]
 mc-crypto-keys = { path = "../crypto/keys" }
 mc-ledger-db = { path = "../ledger/db" }
 mc-transaction-core = { path = "../transaction/core" }
 mc-transaction-core-test-utils = { path = "../transaction/core/test-utils" }
 mc-util-from-random = { path = "../util/from-random" }
+mc-util-serial = { path = "../util/serial", features = ["std"] }
```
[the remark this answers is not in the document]
remoun: Right, so `mc-util-serial` becomes a `[dependency]`, which you already added, which makes this `[dev-dependency]` redundant :)

mobilecoind-dev-faucet/src/bin/main.rs
garbageslam: After looking at this, I guess I'm not sure what the point of this is.

If I want to test the assembly of the server and rocket, then I think one way is, in `mobilecoind-dev-faucet/tests`, I could add integration tests that start the server by running the compiled binary, and then poke it over http etc. But this will also be quite complex because you need a consensus network for this test to work. So at some point it seems easier to make this like one of the tests that runs `local_network.py`.

I guess I don't see that it's interesting to build another binary that starts another http server using rocket, with the same routes as this one, but isn't actually the same binary. So if I'm going to make a re-usable helper out of this, who is the other user that is consuming it?

I think for the fog overseer there is a difference that, the integration tests for that server do actually start a bunch of fog ingest server on the stack in the test code: 

https://github.com/mobilecoinfoundation/mobilecoin/blob/9d7f88efe181ee498c07f62767ff56aed18a032c/fog/overseer/server/tests/inactive_outstanding_key_idle_does_not_have_key.rs#L30
https://github.com/mobilecoinfoundation/mobilecoin/blob/9d7f88efe181ee498c07f62767ff56aed18a032c/fog/overseer/server/tests/utils/mod.rs#L27

But I don't think this is currently how we test the consensus network + mobilecoind. Consensus network has a ton of threads and starting 5 consensus nodes on the stack isn't really going to work. We used to have a lot of tests like that and it was pretty flaky and eventually got removed. (Maybe, we should test fog ingest clusters differently too, but our approach there seems to be working, so if it ain't broke maybe don't fix it?)

remoun: Fair enough. I thought it might be interesting to exercise the faucet with a mock server, but I see your point that this is more akin to `local_network.py`

garbageslam: I see -- I think we could make a mock of mobilecoind in a tests/ file, and test the faucet using this:

https://doc.rust-lang.org/cargo/commands/cargo-test.html#target-selection

> Binary targets are automatically built if there is an integration test or benchmark. This allows an integration test to execute the binary to exercise and test its behavior. The CARGO_BIN_EXE_<name> [environment variable](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates) is set when the integration test is built so that it can use the [env macro](https://doc.rust-lang.org/std/macro.env.html) to locate the executable.

I think it has some benefit, like it's actually testing the final binary, and it doesn't have to build the rocket part again for the test

mobilecoind-dev-faucet/src/worker.rs
iamalwaysuncomfortable: Yeah I'm good with the way this is. It works (and is impressive engineering..) and it solves a pretty significant problem in our release cycles, so it should go in.

As to the engineering philosophy bit, I think there's a bit of an art to mixing threads and executors. Std::thread + an async executor is pretty common in practice I think and a valid approach. Sync code can pass as effectively async through the mspc channels between threads because the channels act as a stream that that queue up done tasks as is done here.  But there are other ways and I think my main point is let's huddle more as a team on async design in the design phase of things like this as I think that's a path towards having a team mastery of the async paradigm.

Thanks for building this!

mobilecoind-dev-faucet/src/bin/main.rs
```diff
@@ -0,0 +1,73 @@
...
+    req: Json<JsonFaucetRequest>,
+) -> Json<JsonSubmitTxResponse> {
+    Json(match state.handle_post(&req).await {
+        Ok(resp) => resp.into(),
+        Err(err_str) => JsonSubmitTxResponse {
+            success: false,
+            err_str: Some(err_str),
+            ..Default::default()
+        },
+    })
```
remoun, line 28: Optional/Equivalent: Consider moving this error handling into an `impl From<Result<mc_mobilecoind_api::SubmitTxResponse, String> for JsonSubmitTxResponse`, then this method can be
```suggestion
    Json(state.handle_post(&req).await.into())
```

mobilecoind-dev-faucet/src/lib.rs
```diff
@@ -0,0 +1,296 @@
...
+        Ok(JsonFaucetStatus {
+            success: true,
+            err_str: None,
+            b58_address: self.monitor_b58_address.clone(),
+            faucet_payout_amounts: self
+                .faucet_payout_amounts
+                .iter()
+                .map(convert_balance_pair)
+                .collect(),
+            balances: balances.iter().map(convert_balance_pair).collect(),
```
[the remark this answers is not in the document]
garbageslam: i tried this in an earlier revision, but `convert_balance_pair` only either works with references, or not with references. i figured having less code is more valuable than avoiding some copies, since we need it to work somewhere else with references.

mobilecoind-dev-faucet/src/lib.rs
```diff
@@ -0,0 +1,296 @@
...
+        let resp = self
+            .mobilecoind_api_client
+            .submit_tx_async(&req)
+            .map_err(|err| format!("Failed to submit Tx: {}", err))?
+            .await
+            .map_err(|err| format!("Submit Tx ended in error: {}", err))?;
+
+        // Tell the worker that this utxo was submitted, so that it can track and
+        // recycle the utxo if this payment fails
+        let _ = utxo_record.sender.send(resp.clone());
```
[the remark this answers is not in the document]
garbageslam: hmm, so this error occurs if the one-shot receiver was dropped by the worker thread before we sent this.

i feel like, this represents some kind of internal error or race, but we still submitted a transaction successfully. if we propagate this error, then we are telling the faucet user that the tx failed, but that may not be true -- at this point I think we submitted something successfully but had an internal error passing a message.

i feel like the right thing to do is log but not propagate, because i think the right thing for the faucet to do is return `success: true` and let the client track what happens the payment if they want to

mobilecoind-dev-faucet/src/worker.rs
```diff
@@ -0,0 +1,632 @@
...
+        // Get the "initial" ledger block count
+        let block_count = loop {
+            match client.get_ledger_info(&Default::default()) {
+                Ok(resp) => break resp.block_count,
+                Err(err) => {
+                    log::error!(logger, "Could not get ledger info: {:?}", err);
+                }
+            }
+            std::thread::sleep(worker_poll_period);
+        };
```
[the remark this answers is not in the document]
garbageslam: I guess that it will log errors at a rate of `worker_poll_period`, unless / until mobilecoind recovers

That seems like desirable behavior, what would you rather see?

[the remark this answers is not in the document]
garbageslam: We could backoff on the timeout so that we don't log at faster than 1 Hz or something?

mobilecoind-dev-faucet/src/worker.rs
```diff
@@ -0,0 +1,632 @@
...
+                log::trace!(
+                    logger,
+                    "Queueing a utxo: key_image = {:?}, value = {}",
+                    key_image,
+                    utxo.value
+                );
+                let (tracker, record) = UtxoTracker::new(utxo.clone());
+                // Add to queue depth before push, because we subtract after pop
+                self.queue_depth.fetch_add(1, Ordering::SeqCst);
+                let _ = self.sender.send(record);
```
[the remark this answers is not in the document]
garbageslam: I guess I'm not sure if it's interesting to try to handle this error

This error occurs if we try to add a utxo to the queue, but we cannot. The queue is an unbounded queue, so according to docs it only happens if `close` is called or if the receiver end is dropped.

Close is not called, so the only way this can happen is if the Worker handle struct is dropped. But drop for that function calls join for the handle to this thread. So this really should never happen.

If the receiver end of the queue is dropped, the worker thread cannot do anything useful anymore. If we make this error be map err and returned to caller, then the error will just be logged infinitely in a loop. It might be reasonable to try to say, if this error is encountered, we should try to exit this thread gracefully. But this will also be a lot of plumbing, and I'm not sure it really improves the overall health of the program, because the faucet will stop working if the worker has joined, and the faucet doesn't currently have any way to restart the worker.

Actually I guess the most sensible thing to do is panic, since the worker thread expects this not to ever fail, and the worker thread can't really do anything useful anymore if the queue is closed. I would be open to changing this to a panic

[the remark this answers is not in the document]
remoun: Panicking for this exceptional case SGTM

mobilecoind-dev-faucet/src/data_types.rs
```diff
@@ -0,0 +1,188 @@
...
+use serde_derive::{Deserialize, Serialize};
+use std::collections::HashMap;
+
+/// A request to the faucet to fund an address
+#[derive(Debug, Default, Deserialize, Serialize)]
+pub struct JsonFaucetRequest {
+    /// The address to fund
+    pub b58_address: String,
+    /// The token id to fund. Assumed 0 if omitted.
+    pub token_id: Option<JsonU64>,
```
eranrund, line 17: I think you can skip the `Option<>` and use `#[serde(default)]`

mobilecoind-dev-faucet/src/lib.rs
```diff
@@ -0,0 +1,324 @@
...
+        req.set_account_key((&self.account_key).into());
+        req.set_input_list(vec![utxo_record.utxo].into());
+        req.set_receiver(public_address.clone());
+        req.set_token_id(*token_id);
+
+        let resp = self
+            .mobilecoind_api_client
+            .generate_tx_from_tx_out_list_async(&req)
+            .map_err(|err| format!("Failed to build Tx: {}", err))?
+            .await
```
eranrund, line 258: I might be wrong, but if I understand correctly whats happening here then it might be worth noting that the request handler is going to be executed from a tokio runner context, while this await is going to block on whatever runner grpcio is using. As such maybe its not worth using the `async` variant of the API client?

eranrund (approved):
LGTM, although there might be an issue with mixing the grpcio calls and tokio executor. The recommended way according to some issues I found on the grpcio GH repository is to use some form of channel to bridge between the two.
With that said, we don't particularly care about slowing down the http request handling here so I think it is okay if they block for the duration of the mobilecoind GRPC calls.
Not blocking this on approval since this isn't critical code, but I do suggest documenting this and maybe not using the `_async` call variants at all inside tokio async functions.

mobilecoind-dev-faucet/src/lib.rs
remoun: No, each suspended Future tracks a waker to resume when it should be polled again, e.g. when data arrives (impls have to call the waker when `Future::poll` returns `Poll::Pending`). They can be running on different executors, including two Tokio runtimes. So the gRPC futures will likely be driven by the gRPC executor (in C) and then these handlers would use that data while running in a Tokio executor

garbageslam: I had similar thoughts and I asked in the fog-core team standup (since there are a lot of async experts on that team now) if this is a bug, to my understanding, this isn't a bug and this is the way we're supposed to do it.

Here's busyjay on the issue: https://github.com/tikv/grpc-rs/issues/520

* tokio and grpcio have different thread pools and different executors in this program, the grpcio one is based in some C code.
* calls to `tokio::task::spawn` (or any tokio function that calls that) must be made in the context of the tokio executor, tokio will panic like `grpc-poll-0' panicked at 'not currently running on the Tokio runtime.` (https://github.com/tikv/grpc-rs/issues/338)
* awaiting a grpcio future in a tokio context is not bad though, if a tokio task goes to sleep on a grpcio future, i think the grpcio executor will ultimately call the "waker" from tokio via a generic and compatible API, and this is expected to work
* using an async API here is good because in theory this thread can be used by tokio to handle more requests
* even if we switched to calling a sync API, it is likely that the sync API works by calling the async api and then using `.await` (https://docs.rs/grpcio/latest/src/grpcio/client.rs.html#41)

that last point may be a good argument for switching to calling the sync version again, since it is calling await under the hood for us anyways, but we would have less code / less error handling boilerplate and still reap the benefits of async

remoun: > * using an async API here is good because in theory this thread can be used by tokio to handle more requests

That's precisely the benefit of async APIs vs blocking.

> * even if we switched to calling a sync API, it is likely that the sync API works by calling the async api and then using `.await` (https://docs.rs/grpcio/latest/src/grpcio/client.rs.html#41)
> 
> that last point may be a good argument for switching to calling the sync version again, since it is calling await under the hood for us anyways, but we would have less code / less error handling boilerplate and still reap the benefits of async

Note that they're calling `block_on` to block the current thread, which is very different from `.await`

remoun (approved):

mobilecoind-dev-faucet/src/lib.rs
```diff
@@ -0,0 +1,324 @@
...
+            monitor_public_address,
+            minimum_fees,
+            faucet_payout_amounts.clone(),
+            config.target_queue_depth,
+            Duration::from_millis(config.worker_poll_period_ms),
+            logger,
+        );
+
+        let logger = logger.new(o!("thread" => "req-handler"));
+
```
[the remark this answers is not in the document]
garbageslam: okay sounds good

Issue #1866, opened 2022-04-27
Re-enable the test for sorted outputs


garbageslam:
We had to disable the test for sorted outputs, because:
* It was never previously exercised on a block version where the sorted output requirement is active
* It turns out the test tx only have one TxOut in it, so it is sorted for trivial reasons even when the reverse sorter is installed. This causes the test for sorting check effectiveness to pass even though the test expects it to fail.

I think we can fix this test if we go into the `get_transaction` function in `mc_transaction_core_test_utils` and make it add a second output, e.g. a change output, because then I think the test will work as intended

(Maybe worth capturing in a testing fast-follow-on ticket?

_Originally posted by @sugargoat in https://github.com/mobilecoinfoundation/mobilecoin/pull/1827#discussion_r859976028_)

garbageslam:
FYI the relevant test is being moved in this PR: https://github.com/mobilecoinfoundation/mobilecoin/pull/1875

PR #2062, opened 2022-05-27
added proper testing for sorted outputs for BlockVersion version 3 an…

integral-llc:
…d higher

### Motivation

#1866 

needed unit tests for sorted outputs fpr BV > 3

### Future Work

[Soundtrack of this PR]()


transaction/core/tests/validation.rs
```diff
@@ -7,826 +7,842 @@ extern crate alloc;
...
-        Err(TransactionValidationError::MissingMemo)
-    );
-
-    let (tx, _) = create_test_tx(BlockVersion::ONE);
-    let tx_out = tx.prefix.outputs.first().unwrap();
-
-    assert!(tx_out.e_memo.is_some());
-    assert_eq!(validate_memo_exists(tx_out), Ok(()));
-}
+mod tests {
```
garbageslam, line 10: This is normally done for unit tests in a source file, but now, this is a separate tests module at crate level. There's no need to write mod tests here, this file is already entirely tests. So we should just not do this

garbageslam (changes requested):
Looks fine, but we don't need the `mod tests` that was added

garbageslam (approved):
Lgtm

remoun (approved):

PR #2064, opened 2022-05-27
Parameterize TransactionBuilder for Token ID

dolanbernard:
Parameterize TransactionBuilder Android bindings for token ID

### Motivation

This is needed in order for the Android SDK to handle multiple different token IDs

### DIFF
--- a/android-bindings/src/bindings.rs
+++ b/android-bindings/src/bindings.rs
@@ -46,9 +46,8 @@ use mc_transaction_core::{
         create_shared_secret, recover_onetime_private_key, recover_public_subaddress_spend_key,
     },
     ring_signature::KeyImage,
-    TokenId,
     tx::{Tx, TxOut, TxOutConfirmationNumber, TxOutMembershipProof},
-    Amount, BlockVersion, CompressedCommitment, MaskedAmount,
+    Amount, BlockVersion, CompressedCommitment, MaskedAmount, TokenId,
 };
 
 use mc_transaction_std::{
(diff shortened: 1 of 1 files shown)


android-bindings/src/bindings.rs
```diff
@@ -1666,19 +1668,17 @@ pub unsafe extern "C" fn Java_com_mobilecoin_lib_TransactionBuilder_init_1jni(
...
         let block_version = BlockVersion::try_from(block_version as u32).unwrap();
-        // Note: RTHMemoBuilder can be selected here, but we will only actually
-        // write memos if block_version is large enough that memos are supported.
-        // If block version is < 2, then transaction builder will filter out memos.
         let memo_builder_box: Box<dyn MemoBuilder + Send + Sync> =
             env.take_rust_field(memo_builder_box, RUST_OBJ_FIELD)?;
-        // FIXME #1595: The token id should be a parameter and not hard coded to Mob
-        // here
-        let fee_amount = Amount::new(Mob::MINIMUM_FEE, Mob::ID);
+        let token_id = TokenId::try_from(token_id as u64).unwrap();
```
garbageslam, line 1680: i think this can just be `TokenId::from`, all u64's are potentially valid token ids

garbageslam (approved):
LGTM

android-bindings/src/bindings.rs
jcape: Yeah, the problem is `jlong ~= i64`

PR #2067, opened 2022-05-28
Split out `mc_transaction_core::blockchain` into a separate `mc_blockchain_types` crate

remoun:
Restructure blockchain types, as part of an ongoing restructuring of `mc_transaction_core`


transaction/core/src/lib.rs
```diff
@@ -50,10 +48,9 @@ pub use tx::MemoContext;
 pub use tx_error::{NewMemoError, NewTxError, ViewKeyMatchError};
 pub use tx_out_gift_code::TxOutGiftCode;
 
-// Re-export Amount and TokenId from transaction-types, and certain types from
-// RingSignature crate
+// Re-export from transaction-types, and some from RingSignature crate.
 pub use mc_crypto_ring_signature::{Commitment, CompressedCommitment};
-pub use mc_transaction_types::{Amount, TokenId};
+pub use mc_transaction_types::*;
```
garbageslam, line 53: it may be a good idea to also re-export `mc-blockchain-types` here?

garbageslam: actually we do we even depend on that? I guess we don't use it?

transaction/core/Cargo.toml
```diff
@@ -56,6 +56,7 @@ proptest = { version = "1.0", default-features = false, features = ["default-cod
 rand = "0.8"
 tempdir = "0.3"
 
+mc-blockchain-types = { path = "../../blockchain/types" }
```
garbageslam, line 59: it looks to me that this dependency is unused, maybe we can remove it?

garbageslam: oh i see, it is a dev-dependency.

this is still not a good thing, this caused us to require an "adapt-hack" trick in the past which made test code hard to use.

maybe we can move the `tests/` integration tests to `mc-blockchain-types` ?

garbageslam (approved):
it looks to me that we have a circular dependency:

* mc-blockchain-types depends on mc-transaction-core
* mc-transaction-core has a dev-dependency on mc-blockchain-types

i wonder if there is any way to simplify this?

i get that, the root of the problem is `mc-transaction-core-test-utils`, because that depends on `core` and core has a dev-dependency on that. and those functions will depend on blockchain-types so now the transaction core tests need that.

if you can think of any way to improve this it would be a good thing. approving since i don't have a clear idea how to improve it myself, other than re-writing `mc-transaction-core-test-utils` so that it doesn't use anything outside of transaction-core

remoun:
> it looks to me that we have a circular dependency:
> 
> * mc-blockchain-types depends on mc-transaction-core
> * mc-transaction-core has a dev-dependency on mc-blockchain-types
> 
> i wonder if there is any way to simplify this?

Yes, I managed to remove the dev-dep.

> i get that, the root of the problem is `mc-transaction-core-test-utils`, because that depends on `core` and core has a dev-dependency on that. and those functions will depend on blockchain-types so now the transaction core tests need that.
> 
> if you can think of any way to improve this it would be a good thing. approving since i don't have a clear idea how to improve it myself, other than re-writing `mc-transaction-core-test-utils`

I'm also moving the block-related helpers from `mc_transaction_core_test_utils` into `mc_blockchain_test_utils` in #2052

nick-mobilecoin:
Sorry forgot to write why I didn't approve, it looks like there are 2 instances of `TxHash` now, that's the only real blocker I see

transaction/types/src/tx_hash.rs
```diff
@@ -0,0 +1,76 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
+use alloc::vec::Vec;
```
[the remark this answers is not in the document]
remoun: Good catch! This was left over from when I tried moving all of `TxOut` into `transaction/types`

api/src/convert/error.rs
```diff
@@ -43,18 +46,24 @@ impl From<RingSigError> for ConversionError {
...
 
-impl From<mc_crypto_keys::KeyError> for ConversionError {
-    fn from(src: mc_crypto_keys::KeyError) -> Self {
+impl From<KeyError> for ConversionError {
+    fn from(src: KeyError) -> Self {
         Self::Key(src)
     }
 }
 
+impl From<SignatureError> for ConversionError {
```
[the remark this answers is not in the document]
remoun: To DRY calls like `verify(...).map(|_| ConversionError::InvalidSignature)`. It turns out there's only one of those now, but I'm adding another with `BlockMetadata` and will probably clean up other signature-type conversions to use this, in a follow-up PR.

blockchain/types/src/lib.rs
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
+//! Blockchain data structures.
```
iamalwaysuncomfortable, line 3: Is there anything else going in the blockchain crate? Or just the types? I think if it's just the units going here and nothing else immediately planned, it doesn't warrant a new Blockchain folder I think. If it's only the types going in this folder in the near future - it should probably be in `transaction/blockchain/types` or `transaction/types/blockchain`

Types keep popping up I think we should chat a bit as a team about the Taxonomy of that. Like perhaps we have a types workspace/folder with the chain types?

Like perhaps we want something like a types workspace where all the types are organized? It seems like a single Blockchain folder with just a types folder creates too much fragmentation.

blockchain/types/src/lib.rs
```diff
@@ -0,0 +1,26 @@
...
+pub use crate::{
+    block::{compute_block_id, Block, BlockIndex, MAX_BLOCK_VERSION},
+    block_contents::{BlockContents, BlockContentsHash},
+    block_data::BlockData,
+    block_id::BlockID,
+    block_signature::BlockSignature,
+    error::ConvertError,
+};
+
+pub use mc_transaction_types::BlockVersion;
```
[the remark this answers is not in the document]
iamalwaysuncomfortable: It's originally defined in transaction_core because it's a key flag which defines what kind of operations (like minting, multiple tokens, etc.) can be performed and thus what things to include in the proofs/how the proofs work out (i.e. doing the bulletproofs/range proofs in multiple tokens vs. just MOB). But also Blockversion is used to do logic by multiple actors in the system, so it seems like it could go here just fine and this would be a dependency of core?

iamalwaysuncomfortable (changes requested):
The moving of the types is a good idea - but I don't think it's right to create a whole other top level Blockchain folder for it unless we have a plan for what else is going to move there.

Are we moving other things into the Blockchain folder? If not, suggestions on taxonomy for the Blockchain types would be `transaction/types/blockchain` or a top level `types` folder that we migrate all relevant types into.

blockchain/types/src/lib.rs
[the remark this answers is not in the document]
remoun: No, because `mc_blockchain_types` depends on `mc_transaction_core` (mainly for `TxOut`) and `mc_transaction_core` needs `BlockVersion` for a bunch of things. I'm not sure why `BlockVersion` is in `mc_transaction_types`, though.

api/src/convert/archive_block.rs
```diff
@@ -20,73 +22,64 @@ impl From<&mc_transaction_core::BlockData> for blockchain::ArchiveBlock {
...
             .transpose()?;
 
         if let Some(signature) = signature.as_ref() {
-            signature
-                .verify(&block)
-                .map_err(|_| ConversionError::InvalidSignature)?;
+            signature.verify(&block)?;
         }
 
         if block.contents_hash != block_contents.hash() {
```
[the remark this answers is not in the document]
remoun: Added.

blockchain/types/src/lib.rs
remoun: FWIW I'm also adding `mc_blockchain_test_utils` in #2058.

iamalwaysuncomfortable (approved):

blockchain/types/src/lib.rs
iamalwaysuncomfortable: Oke, I voiced similar grumblings to James about this being not quite the right taxonomy, but am accepting of being outvoted in this instance 😺 - approved

PR #2088, opened 2022-06-03
fix block-version 0 behavior in consensus enclave

garbageslam:
this is required because the mint_output function rejects any new
token id if new token ids are not allowed yet, but the fee aggregation
code would sometimes, under load, try to mint a fee output of alternate
token ids, because of privacy considerations in MCIP #25

If not fixed, then if 1.2.0 is started with bv = 0, and has alternate tokens in the `tokens.toml` file, then the consensus nodes will crash if more than one Tx appears in the block.

```
OH NO, WE CRASHED :( thread ByzantineLedgerNodeID(peer1.mc-v1-2-0-without-upgrade.development.mobilecoin.com:443:1af604b8ef998e89b939bffca5731e377122a78bfdd844a1b5e50c7534979321) on /usr/bin/consensus-service
Details: panicked at 'form_block failed: FormBlock("Cannot mint outputs for non-MOB tokens until they are supported")', /__w/mobilecoin/mobilecoin/consensus/service/src/byzantine_ledger/worker.rs:871:14
```

### DIFF
--- a/consensus/enclave/impl/src/lib.rs
+++ b/consensus/enclave/impl/src/lib.rs
@@ -826,9 +826,18 @@ impl ConsensusEnclave for SgxConsensusEnclave {
         //
         // Note: The "take" here only makes sense because sort_zeros_to_end was called.
         // This ensures that everything we are ignoring via take has zero value.
+        let num_fee_outputs = if config.block_version.masked_token_id_feature_is_supported() {
+            min(total_fees.len(), transactions.len())
+        } else {
+            // Before token ids are supported, 0 is the only valid token id.
+            // We must not attempt to create any other fee outputs or mint_output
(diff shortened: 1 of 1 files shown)


iamalwaysuncomfortable (approved):

jgreat (approved):
Looks good!

I branched off your branch to feature/fix-bv-0-behavior and let the CD do the full upgrade cycle.  https://github.com/mobilecoinfoundation/mobilecoin/actions/runs/2440117851

I took the binaries from that build and deployed them to a namespace to test at v1.2.0 v0 and it passed all the block version 0 tests including fog-distribution which caused the consensus node to panic before this change. https://github.com/mobilecoinfoundation/mobilecoin/runs/6739561560

Issue #2081, opened 2022-06-02
Tracking issue for MCIP #42 POC


cbeck88:
This is a tracking issue for work connected to [MCIP 42](https://github.com/mobilecoinfoundation/mcips/pull/42) proof of concept

The POC work likely has three steps:

- [x] Demonstrate viability for changing the TxOut -> Blinding factor derivations to include Amount Shared Secret as described, and a compatibility path for this https://github.com/mobilecoinfoundation/mobilecoin/pull/2096
- [x] Add partial fill rules and verification as described, in the TxIn's and transaction validation, and create a code path in the signed contingent input builder to exercise this functionality. Put this under unit tests https://github.com/mobilecoinfoundation/mobilecoin/pull/2484
- [x] Integrate this into the fog-test-client similarly as before and exercise the functionality in a deployed network. https://github.com/mobilecoinfoundation/mobilecoin/pull/2629
- [x] Address action items from MCIP review: https://github.com/mobilecoinfoundation/mcips/pull/42#issuecomment-1267303792 https://github.com/mobilecoinfoundation/mobilecoin/pull/2654
- [ ] Fixup transaction-builder `add_presigned_partial_fill_input` API. It's more intuitive to pass the fill amount rather than the fractional change amount. EDIT: I think this should be a won't fix, we can make a friendlier API at another abstraction layer. As I started trying to make this change I realized I'm making it a lot more complicated.

PR #2094, opened 2022-06-06
Make `TxOut::new` take a block version

### LINKED ISSUE #1749: `TxOut::new` should take `BlockVersion` as a parameter
There are a lot of places where we call `TxOut::new`, and then adjust the resulting structure to conform to block version rules, like removing a memo, or possibly, removing a masked amount.

As we add more features, these adjustments will proliferate. And we had some bugs during compat testing that were being caused by, a `TxOut::new` call somewhere that was not respecting current block version rules.

Most likely we should just make `TxOut::new` take `BlockVersion` as a parameter so that you have to specify the rules up-front and all the adjustments are in the constructor. This could also be a separate constructor, but making this the only constructor may be better overall for maintainability.

garbageslam:
This makes it easier to make `TxOut` that conform to certain
structural rules, which is important for compatibility testing.

Fixes #1749

This will make it easier to execute #2081

### DIFF
--- a/consensus/enclave/impl/src/lib.rs
+++ b/consensus/enclave/impl/src/lib.rs
@@ -991,7 +991,7 @@ fn mint_output<T: Digestible>(
 ) -> Result<TxOut> {
     // Check if the token id makes sense right now
     if !block_version.masked_token_id_feature_is_supported() && amount.token_id != 0 {
-        return Err(Error::FormBlock(
+        return Err(Error::BlockVersion(
             "Cannot mint outputs for non-MOB tokens at this block version".to_string(),
         ));
     }
--- a/consensus/enclave/impl/src/lib.rs
+++ b/consensus/enclave/impl/src/lib.rs
@@ -1010,16 +1010,17 @@ fn mint_output<T: Digestible>(
     };
 
     // Create a single TxOut
-    let mut output = TxOut::new(amount, recipient, &tx_private_key, Default::default())
-        .map_err(|e| Error::FormBlock(format!("AmountError: {:?}", e)))?;
+    let output = TxOut::new(
+        block_version,
+        amount,
+        recipient,
+        &tx_private_key,
+        Default::default(),
+    )
+    .map_err(|e| Error::FormBlock(format!("NewTxError: {}", e)))?;
 
-    // The output must conform to block version rules
-    if !block_version.e_memo_feature_is_supported() {
-        output.e_memo = None;
-    }
-
+    // Check if the token id makes sense right now
     if !block_version.masked_token_id_feature_is_supported() {
-        output.masked_amount.masked_token_id.clear();
         if amount.token_id != 0 {
             return Err(Error::FormBlock(
                 "Cannot mint outputs for non-MOB tokens until they are supported".to_string(),
@@ -1543,6 +1544,7 @@ mod tests {
             for token_id in [token_id1, token_id2] {
                 for _ in 0..5 {
                     let spendable_output = TxOut::new(
+                        block_version,
                         Amount {
                             value: 10_000_000_000,
                             token_id,
--- a/consensus/enclave/impl/src/lib.rs
+++ b/consensus/enclave/impl/src/lib.rs
@@ -989,6 +989,13 @@ fn mint_output<T: Digestible>(
     amount: Amount,
     counter: usize,
 ) -> Result<TxOut> {
+    // Check if the token id makes sense right now
+    if !block_version.masked_token_id_feature_is_supported() && amount.token_id != 0 {
+        return Err(Error::FormBlock(
+            "Cannot mint outputs for non-MOB tokens until they are supported".to_string(),
+        ));
+    }
+
     // Create a determinstic private key based on the block contents.
     let tx_private_key = {
         let mut hash_value = [0u8; 32];
@@ -1010,25 +1017,14 @@ fn mint_output<T: Digestible>(
     };
 
     // Create a single TxOut
-    let output = TxOut::new(
+    TxOut::new(
         block_version,
         amount,
         recipient,
         &tx_private_key,
         Default::default(),
     )
-    .map_err(|e| Error::FormBlock(format!("NewTxError: {}", e)))?;
-
-    // Check if the token id makes sense right now
-    if !block_version.masked_token_id_feature_is_supported() {
-        if amount.token_id != 0 {
-            return Err(Error::FormBlock(
-                "Cannot mint outputs for non-MOB tokens until they are supported".to_string(),
-            ));
-        }
-    }
-
-    Ok(output)
+    .map_err(|e| Error::FormBlock(format!("NewTxError: {}", e)))
 }
 
 #[cfg(test)]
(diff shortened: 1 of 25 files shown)


consensus/enclave/impl/src/lib.rs
```diff
@@ -989,6 +989,13 @@ fn mint_output<T: Digestible>(
     amount: Amount,
     counter: usize,
 ) -> Result<TxOut> {
+    // Check if the token id makes sense right now
+    if !block_version.masked_token_id_feature_is_supported() && amount.token_id != 0 {
+        return Err(Error::FormBlock(
+            "Cannot mint outputs for non-MOB tokens until they are supported".to_string(),
```
remoun, line 995: Optional/Equivalent: Note that they are not supported at this block version
```suggestion
            "Cannot mint outputs for non-MOB tokens at this block version".to_string(),
```

consensus/enclave/impl/src/lib.rs
```diff
@@ -989,6 +989,13 @@ fn mint_output<T: Digestible>(
     amount: Amount,
     counter: usize,
 ) -> Result<TxOut> {
+    // Check if the token id makes sense right now
+    if !block_version.masked_token_id_feature_is_supported() && amount.token_id != 0 {
+        return Err(Error::FormBlock(
```
remoun, line 994: This is (arguably) a `BlockVersion` error
```suggestion
        return Err(Error::BlockVersion(
```

transaction/core/src/tx.rs
```diff
@@ -333,56 +334,74 @@ impl TxOut {
...
     ) -> Result<Self, NewTxError> {
         let target_key = create_tx_out_target_key(tx_private_key, recipient).into();
         let public_key = create_tx_out_public_key(tx_private_key, recipient.spend_public_key());
 
         let shared_secret = create_shared_secret(recipient.view_public_key(), tx_private_key);
 
-        let masked_amount = MaskedAmount::new(amount, &shared_secret)?;
+        let mut masked_amount = MaskedAmount::new(amount, &shared_secret)?;
 
         let memo_ctxt = MemoContext {
```
remoun, line 386: Consider moving this declaration into the `if block_version.e_memo_feature_is_supported()` block below.

consensus/service/src/mint_tx_manager/mod.rs
```diff
@@ -366,7 +366,7 @@ mod mint_config_tx_tests {
         let parent_block = ledger.get_block(n_blocks - 1).unwrap();
 
         let block_contents = BlockContents {
-            outputs: vec![create_test_tx_out(&mut rng)],
+            outputs: vec![create_test_tx_out(BlockVersion::MAX, &mut rng)],
```
remoun, line 369: Consider adding a `const BLOCK_VERSION = BlockVersion::MAX` to make it obvious we're using the same block version for all these calls.

remoun (approved):
LGTM, with some suggestions. Thank you!

fog/ingest/enclave/impl/tests/tx_processing.rs
```diff
@@ -59,6 +59,7 @@ fn test_ingest_enclave(logger: Logger) {
                 let tx_private_key = RistrettoPrivate::from_random(&mut rng);
                 let e_fog_hint = FogHint::from(&bob_public_address).encrypt(&fog_pubkey, &mut rng);
                 TxOut::new(
+                    BlockVersion::MAX,
```
iamalwaysuncomfortable, line 62: Is it worth for some of these tests to also support previous block versions as well?

test-vectors/tx-out-records/build.rs
```diff
@@ -100,6 +102,7 @@ fn generate_tx_out_record_data() -> TxOutRecordData {
             let e_fog_hint =
                 FogHint::from(&recipient_public_address).encrypt(&fog_pubkey, &mut rng);
             TxOut::new(
+                BlockVersion::TWO,
```
iamalwaysuncomfortable, line 105: Why TWO here and not MAX?

transaction/core/src/tx_error.rs
```diff
@@ -15,6 +15,8 @@ pub enum NewTxError {
     Amount(AmountError),
     /// Memo: {0}
     Memo(NewMemoError),
+    /// Token Id not allowed at this block version
+    TokenIdNotAllowedAtBlockVersion,
```
iamalwaysuncomfortable, line 19: Suggestion: give a tell about which BlockVersion it was. It may be obvious to us which versions don't support TokenIds, but this'd be helpful to people new to developing against the chain.

```suggestion
    /// Token Id not allowed at block version: {0}
    TokenIdNotAllowedAtBlockVersion(BlockVersion),
```

transaction/std/src/transaction_builder.rs
```diff
@@ -368,22 +353,16 @@ impl<FPR: FogPubkeyResolver> TransactionBuilder<FPR> {
...
             |memo_ctxt| {
-                if block_version.e_memo_feature_is_supported() {
-                    Some(mb.make_memo_for_output(
-                        amount,
-                        &reserved_subaddresses.gift_code_subaddress,
-                        memo_ctxt,
-                    ))
-                    .transpose()
-                } else {
-                    Ok(None)
```
iamalwaysuncomfortable, line 385: I %100 support removing the option around the memo builder. Good change.

iamalwaysuncomfortable (approved):
Approval with 
1. A question on whether tests should be covering previous block versions
2. Error enum nit

test-vectors/tx-out-records/build.rs
garbageslam: Because this is a test vector, so the idea is, we build some tx data, and commit it to the repo as a test vector. Then as a unit test we check that it's hash matches some specific value, so that we can avoid making breaking changes by accidentally changing the hash of old data.

transaction/core/src/tx.rs
```diff
@@ -333,56 +334,74 @@ impl TxOut {
...
     /// * `memo_fn` - A callback taking MemoContext, which produces a
     ///   MemoPayload, or a NewMemo error
     pub fn new_with_memo(
+        block_version: BlockVersion,
         amount: Amount,
         recipient: &PublicAddress,
         tx_private_key: &RistrettoPrivate,
         hint: EncryptedFogHint,
-        memo_fn: impl FnOnce(MemoContext) -> Result<Option<MemoPayload>, NewMemoError>,
+        memo_fn: impl FnOnce(MemoContext) -> Result<MemoPayload, NewMemoError>,
```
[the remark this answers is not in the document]
garbageslam: What's happening is, we're moving the logic around "check if block version allows for memos" from this function, and into the function that calls this function (`TxOut::new_with_memo`), so this guy will only be called if `TxOut::new_with_memo` thinks it needs to be called. So there's no need to return `Option` from here now.

fog/ingest/enclave/impl/tests/tx_processing.rs
garbageslam: yeah I guess so

consensus/enclave/impl/src/lib.rs
garbageslam: Okay, I agree -- I think it's good to make `form_block` return `FormBlock` errors though because it helps trace the error when it ultimately bubbles up from untrusted side. So maybe when form_block calls `mint_output` we can map this to `FormBlock`, but `mint_output` itself returns `BlockVersion` error in this case.

garbageslam:
rebased on master for conflicts

PR #2096, opened 2022-06-07
Masked amount v2

garbageslam:
This is an experiment to see if making `MaskedAmount` into an enum called `MaskedAmount` and then deriving `::prost::Oneof` is a tenable direction for introducing masked amount v2

edit: This experiment seems to have been successful, and looks like the way to go

The purpose of masked amount v2 is to update the derivation pathway that determines blinding factors that appear in masked amount, in connection to:
* the Pedersen commitment
* the masked_value
* the masked_token_id

The old derivation pathway derived all of these blinding factors from Blake2b hashes of the TxOut shared secret using different domain separators.

The new derivation pathway has an intermediate key called Amount shared secret, derived by a sha512 hash from TxOut shared secret. The blinding factors are then derived using HKDF sha512 from that.

This allows someone who owns a TxOut to selectively reveal the amount and token id in a way that a counterparty can verify, without also revealing the TxOut shared secret, the encrypted memo, etc.

This can then be used to implement partial fill atomic swaps, because the enclave can require that the amount shared secret is disclosed when necessary to verify partial fill rules.

### DIFF
--- a/api/src/convert/amount.rs
+++ b/api/src/convert/amount.rs
@@ -1,11 +1,17 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
 //! Convert to/from external::Amount
 
 use crate::{external, ConversionError};
-use mc_transaction_core::{CompressedCommitment, MaskedAmount};
+use mc_transaction_core::{CompressedCommitment, MaskedAmount, MaskedAmountV1, MaskedAmountV2};
 use mc_util_repr_bytes::ReprBytes;
 
-impl From<&MaskedAmount> for external::MaskedAmount {
-    fn from(source: &MaskedAmount) -> Self {
+// Note:
+// external::MaskedAmount is a proto message
+// external::TxOut_oneof_masked_amount is a proto oneof
+
+impl From<&MaskedAmountV1> for external::MaskedAmount {
+    fn from(source: &MaskedAmountV1) -> Self {
         let commitment_bytes = source.commitment.to_bytes().to_vec();
         let mut amount = external::MaskedAmount::new();
         amount.mut_commitment().set_data(commitment_bytes);
@@ -15,18 +21,73 @@ impl From<&MaskedAmount> for external::MaskedAmount {
     }
 }
 
-impl TryFrom<&external::MaskedAmount> for MaskedAmount {
+impl From<&MaskedAmountV2> for external::MaskedAmount {
+    fn from(source: &MaskedAmountV2) -> Self {
+        let commitment_bytes = source.commitment.to_bytes().to_vec();
+        let mut amount = external::MaskedAmount::new();
+        amount.mut_commitment().set_data(commitment_bytes);
+        amount.set_masked_value(source.masked_value);
+        amount.set_masked_token_id(source.masked_token_id.clone());
+        amount
+    }
+}
+
+impl From<&MaskedAmount> for external::TxOut_oneof_masked_amount {
+    fn from(source: &MaskedAmount) -> Self {
+        match source {
+            MaskedAmount::V1(masked_amount) => {
+                external::TxOut_oneof_masked_amount::v1(masked_amount.into())
+            }
+            MaskedAmount::V2(masked_amount) => {
+                external::TxOut_oneof_masked_amount::v2(masked_amount.into())
+            }
+        }
+    }
+}
+
+impl TryFrom<&external::MaskedAmount> for MaskedAmountV1 {
     type Error = ConversionError;
 
     fn try_from(source: &external::MaskedAmount) -> Result<Self, Self::Error> {
         let commitment = CompressedCommitment::try_from(source.get_commitment())?;
         let masked_value = source.get_masked_value();
         let masked_token_id = source.get_masked_token_id();
-        let amount = MaskedAmount {
+        let amount = MaskedAmountV1 {
             commitment,
             masked_value,
             masked_token_id: masked_token_id.to_vec(),
         };
         Ok(amount)
+    }
+}
+
+impl TryFrom<&external::MaskedAmount> for MaskedAmountV2 {
+    type Error = ConversionError;
+
+    fn try_from(source: &external::MaskedAmount) -> Result<Self, Self::Error> {
+        let commitment = CompressedCommitment::try_from(source.get_commitment())?;
+        let masked_value = source.get_masked_value();
+        let masked_token_id = source.get_masked_token_id().to_vec();
+        let amount = MaskedAmountV2 {
+            commitment,
+            masked_value,
+            masked_token_id,
+        };
+        Ok(amount)
+    }
+}
+
+impl TryFrom<&external::TxOut_oneof_masked_amount> for MaskedAmount {
+    type Error = ConversionError;
+
+    fn try_from(source: &external::TxOut_oneof_masked_amount) -> Result<Self, Self::Error> {
+        match source {
+            external::TxOut_oneof_masked_amount::v1(masked_amount) => {
+                Ok(MaskedAmount::V1(masked_amount.try_into()?))
+            }
+            external::TxOut_oneof_masked_amount::v2(masked_amount) => {
+                Ok(MaskedAmount::V2(masked_amount.try_into()?))
+            }
+        }
     }
 }
--- a/api/src/convert/amount.rs
+++ b/api/src/convert/amount.rs
@@ -36,10 +36,10 @@ impl From<&MaskedAmount> for external::TxOut_oneof_masked_amount {
     fn from(source: &MaskedAmount) -> Self {
         match source {
             MaskedAmount::V1(masked_amount) => {
-                external::TxOut_oneof_masked_amount::v1(masked_amount.into())
+                external::TxOut_oneof_masked_amount::masked_amount_v1(masked_amount.into())
             }
             MaskedAmount::V2(masked_amount) => {
-                external::TxOut_oneof_masked_amount::v2(masked_amount.into())
+                external::TxOut_oneof_masked_amount::masked_amount_v2(masked_amount.into())
             }
         }
     }
@@ -82,10 +82,10 @@ impl TryFrom<&external::TxOut_oneof_masked_amount> for MaskedAmount {
 
     fn try_from(source: &external::TxOut_oneof_masked_amount) -> Result<Self, Self::Error> {
         match source {
-            external::TxOut_oneof_masked_amount::v1(masked_amount) => {
+            external::TxOut_oneof_masked_amount::masked_amount_v1(masked_amount) => {
                 Ok(MaskedAmount::V1(masked_amount.try_into()?))
             }
-            external::TxOut_oneof_masked_amount::v2(masked_amount) => {
+            external::TxOut_oneof_masked_amount::masked_amount_v2(masked_amount) => {
                 Ok(MaskedAmount::V2(masked_amount.try_into()?))
             }
         }
--- a/api/src/convert/tx_out.rs
+++ b/api/src/convert/tx_out.rs
@@ -1,3 +1,5 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
 //! Convert to/from external::TxOut
 
 use crate::{external, ConversionError};
@@ -9,8 +11,7 @@ impl From<&tx::TxOut> for external::TxOut {
     fn from(source: &tx::TxOut) -> Self {
         let mut tx_out = external::TxOut::new();
 
-        let masked_amount = external::MaskedAmount::from(&source.masked_amount);
-        tx_out.set_masked_amount(masked_amount);
+        tx_out.masked_amount = source.masked_amount.as_ref().map(Into::into);
 
         let target_key_bytes = source.target_key.as_bytes().to_vec();
         tx_out.mut_target_key().set_data(target_key_bytes);
@@ -36,7 +37,11 @@ impl TryFrom<&external::TxOut> for tx::TxOut {
     type Error = ConversionError;
 
     fn try_from(source: &external::TxOut) -> Result<Self, Self::Error> {
-        let masked_amount = MaskedAmount::try_from(source.get_masked_amount())?;
+        let oneof_masked_amount = source
+            .masked_amount
+            .as_ref()
+            .ok_or(ConversionError::ObjectMissing)?;
+        let masked_amount = Some(MaskedAmount::try_from(oneof_masked_amount)?);
 
         let target_key_bytes: &[u8] = source.get_target_key().get_data();
         let target_key: CompressedRistrettoPublic = RistrettoPublic::try_from(target_key_bytes)
@@ -75,11 +80,8 @@ impl TryFrom<&external::TxOut> for tx::TxOut {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use generic_array::GenericArray;
-    use mc_crypto_keys::RistrettoPublic;
-    use mc_transaction_core::{
-        encrypted_fog_hint::ENCRYPTED_FOG_HINT_LEN, tokens::Mob, Amount, MaskedAmount, Token,
-    };
+    use mc_crypto_keys::RistrettoPrivate;
+    use mc_transaction_core::{tokens::Mob, Amount, BlockVersion, PublicAddress, Token};
     use mc_util_from_random::FromRandom;
     use rand::{rngs::StdRng, SeedableRng};
 
@@ -92,14 +94,14 @@ mod tests {
             value: 1u64 << 13,
             token_id: Mob::ID,
         };
-        let source = tx::TxOut {
-            masked_amount: MaskedAmount::new(amount, &RistrettoPublic::from_random(&mut rng))
-                .unwrap(),
-            target_key: RistrettoPublic::from_random(&mut rng).into(),
-            public_key: RistrettoPublic::from_random(&mut rng).into(),
-            e_fog_hint: (&[0u8; ENCRYPTED_FOG_HINT_LEN]).into(),
-            e_memo: None,
-        };
+        let source = tx::TxOut::new(
+            BlockVersion::ZERO,
+            amount,
+            &PublicAddress::from_random(&mut rng),
+            &RistrettoPrivate::from_random(&mut rng),
+            Default::default(),
+        )
+        .unwrap();
 
         let converted = external::TxOut::from(&source);
 
@@ -116,14 +118,14 @@ mod tests {
             value: 1u64 << 13,
             token_id: Mob::ID,
         };
-        let source = tx::TxOut {
-            masked_amount: MaskedAmount::new(amount, &RistrettoPublic::from_random(&mut rng))
-                .unwrap(),
-            target_key: RistrettoPublic::from_random(&mut rng).into(),
-            public_key: RistrettoPublic::from_random(&mut rng).into(),
-            e_fog_hint: (&[0u8; ENCRYPTED_FOG_HINT_LEN]).into(),
-            e_memo: Some((*GenericArray::from_slice(&[9u8; 66])).into()),
-        };
+        let source = tx::TxOut::new(
+            BlockVersion::MAX,
+            amount,
+            &PublicAddress::from_random(&mut rng),
+            &RistrettoPrivate::from_random(&mut rng),
+            Default::default(),
+        )
+        .unwrap();
 
         let converted = external::TxOut::from(&source);
 
--- a/fog/api/proto/view.proto
+++ b/fog/api/proto/view.proto
@@ -267,5 +267,12 @@ message TxOutRecord {
     /// This is omitted for TxOut's from before the upgrade that introduced memos.
     bytes tx_out_e_memo_data = 9;
     /// The masked token id associated to the amount field in the TxOut that was recovered
-    bytes tx_out_amount_masked_token_id = 10;
+    oneof tx_out_masked_amount_token_id {
+        /// The masked token id associated to the v1 amount field in the TxOut that was recovered
+        /// The presence of this field indicates that a MaskedAmountV1 object was serialized.
+        bytes v1 = 10;
+        /// The masked token id associated to the v2 amount field in the TxOut that was recovered
+        /// The presence of this field indicates that a MaskedAmountV2 object was serialized.
+        bytes v2 = 11;
+    }
 }
--- a/fog/api/proto/view.proto
+++ b/fog/api/proto/view.proto
@@ -267,12 +267,12 @@ message TxOutRecord {
     /// This is omitted for TxOut's from before the upgrade that introduced memos.
     bytes tx_out_e_memo_data = 9;
     /// The masked token id associated to the amount field in the TxOut that was recovered
-    oneof tx_out_masked_amount_token_id {
+    oneof tx_out_amount_masked_token_id {
         /// The masked token id associated to the v1 amount field in the TxOut that was recovered
         /// The presence of this field indicates that a MaskedAmountV1 object was serialized.
-        bytes tx_out_masked_amount_v1_token_id  = 10;
+        bytes tx_out_amount_masked_v1_token_id  = 10;
         /// The masked token id associated to the v2 amount field in the TxOut that was recovered
         /// The presence of this field indicates that a MaskedAmountV2 object was serialized.
-        bytes tx_out_masked_amount_v2_token_id = 11;
+        bytes tx_out_amount_masked_v2_token_id = 11;
     }
 }
--- a/fog/api/proto/view.proto
+++ b/fog/api/proto/view.proto
@@ -270,9 +270,9 @@ message TxOutRecord {
     oneof tx_out_masked_amount_token_id {
         /// The masked token id associated to the v1 amount field in the TxOut that was recovered
         /// The presence of this field indicates that a MaskedAmountV1 object was serialized.
-        bytes v1 = 10;
+        bytes tx_out_masked_amount_v1_token_id  = 10;
         /// The masked token id associated to the v2 amount field in the TxOut that was recovered
         /// The presence of this field indicates that a MaskedAmountV2 object was serialized.
-        bytes v2 = 11;
+        bytes tx_out_masked_amount_v2_token_id = 11;
     }
 }
--- a/fog/distribution/src/main.rs
+++ b/fog/distribution/src/main.rs
@@ -409,7 +409,8 @@ fn select_spendable_tx_outs(
                     get_tx_out_shared_secret(account.view_private_key(), &public_key);
 
                 let (amount, _blinding_factor) = tx_out
-                    .masked_amount
+                    .get_masked_amount()
+                    .unwrap_or_else(|err| panic!("TX missing masked value: {}", err))
                     .get_value(&shared_secret)
                     .unwrap_or_else(|err| {
                         panic!(
--- a/fog/distribution/src/main.rs
+++ b/fog/distribution/src/main.rs
@@ -410,7 +410,7 @@ fn select_spendable_tx_outs(
 
                 let (amount, _blinding_factor) = tx_out
                     .get_masked_amount()
-                    .unwrap_or_else(|err| panic!("TX missing masked value: {}", err))
+                    .expect("TxOut missing masked value")
                     .get_value(&shared_secret)
                     .unwrap_or_else(|err| {
                         panic!(
--- a/fog/ingest/enclave/impl/src/lib.rs
+++ b/fog/ingest/enclave/impl/src/lib.rs
@@ -43,7 +43,7 @@ use mc_fog_types::{
 use mc_oblivious_traits::{subtle::ConstantTimeEq, ORAMStorageCreator};
 use mc_sgx_compat::sync::Mutex;
 use mc_sgx_report_cache_api::{ReportableEnclave, Result as ReportableEnclaveResult};
-use mc_transaction_core::fog_hint::FogHint;
+use mc_transaction_core::{fog_hint::FogHint, tx::TxOut};
 use mc_util_from_random::FromRandom;
 use zeroize::Zeroize;
 
@@ -86,7 +86,7 @@ impl<OSC: ORAMStorageCreator<StorageDataSize, StorageMetaSize>> SgxIngestEnclave
     /// overflowed and we have to change the egress key and try again.
     /// Returns `None` if overflow occurs.
     fn attempt_ingest_txs(
-        chunk: &TxsForIngest,
+        chunk: &PreparedBlockData,
         ingress_key: &RistrettoPrivate,
         egress_key: &RistrettoPrivate,
         rng_store: &mut RngStore<OSC>,
@@ -96,7 +96,7 @@ impl<OSC: ORAMStorageCreator<StorageDataSize, StorageMetaSize>> SgxIngestEnclave
         let mut new_tx_rows = Vec::new();
 
         // Use the constant time fog hint decryption
-        for (index, txo) in chunk.redacted_txs.iter().enumerate() {
+        for (index, (txo, fog_tx_out)) in chunk.tx_outs.iter().enumerate() {
             let mut user_id = FogHint::new(RistrettoPublic::from_random(&mut rng));
             // Note: This is ignored because the semantic we want is, user_id should be
             // random if decryption failed, and ct_decrypt has no side-effects
@@ -143,13 +143,12 @@ impl<OSC: ORAMStorageCreator<StorageDataSize, StorageMetaSize>> SgxIngestEnclave
 
                 // Create a TxOutRecord, flattening the Txo data and getting extra data like
                 // global index, block index, timestamp.
-                let fog_tx_out = FogTxOut::from(txo);
                 let meta = FogTxOutMetadata {
                     global_index: chunk.global_txo_index + index as u64,
                     block_index: chunk.block_index,
                     timestamp: chunk.timestamp,
                 };
-                let txo_record = TxOutRecord::new(fog_tx_out, meta);
+                let txo_record = TxOutRecord::new(fog_tx_out.clone(), meta);
 
                 // Get the view-kew-encrypted payload for this TX
                 let plaintext = mc_util_serial::encode(&txo_record);
@@ -298,18 +297,25 @@ impl<OSC: ORAMStorageCreator<StorageDataSize, StorageMetaSize>> IngestEnclave
         let mut rng_store_lk = self.rng_store.lock()?;
         let rng_store = rng_store_lk.as_mut().expect("enclave was not initialized");
 
+        // Converting a TxOut to a FogTxOut can fail if the masked amount is missing,
+        // so flush all those errors out now.
+        let prepared_block_data = PreparedBlockData::try_from(chunk)?;
+
         // Try to ingest the new tx's
         loop {
-            if let Some(e_tx_out_records) =
-                Self::attempt_ingest_txs(&chunk, &*ingress_key, &*egress_key, rng_store)
-            {
+            if let Some(e_tx_out_records) = Self::attempt_ingest_txs(
+                &prepared_block_data,
+                &*ingress_key,
+                &*egress_key,
+                rng_store,
+            ) {
                 return Ok((e_tx_out_records, new_kex_rng_pubkey));
             } else {
                 // If attempt_ingest_txs fails, that means the rng store overflowed.
                 // If this happened too many times give up
                 if chunk_retries == 0 {
                     return Err(Error::ChunkTooBig(
-                        chunk.redacted_txs.len(),
+                        prepared_block_data.tx_outs.len(),
                         MAX_CHUNK_RETRIES,
                         rng_store.capacity(),
                     ));
@@ -382,4 +388,51 @@ impl<OSC: ORAMStorageCreator<StorageDataSize, StorageMetaSize>> IngestEnclave
 // Helper for sealing a key, which maps the error to IngestEnclaveError
 fn seal_private_key(src: &RistrettoPrivate) -> Result<SealedIngestKey> {
     Ok(IntelSealed::seal_raw(src.as_ref(), &[])?.as_ref().to_vec())
+}
+
+// Helper to work around TxOut => FogTxOut being fallible.
+// During attempt_ingest_txs, we must be constant-time and it complicates things
+// if operations can fail, because an adversary may be able to use that as
+// leverage.
+//
+// Instead we want to convert all TxOut to FogTxOut up front before calling
+// attempt_ingest_txs, so that all the failures happen before anything sensitive
+// happens.
+//
+// This is done by converting TxsForIngest to PreparedBlockData up front, and
+// then proceeding to process the prepared block data.
+#[derive(Clone, Eq, PartialEq, Debug)]
+struct PreparedBlockData {
+    /// The index of the block in the blockchain where this chunk of Txs
+    /// appeared
+    pub block_index: u64,
+    /// The number of txo's appearing in the entire blockchain before this chunk
+    /// This is needed to compute TxOutRecord::global_tx_out_index
+    pub global_txo_index: u64,
+    /// The txout's of this chunk, also formatted as FogTxOut
+    pub tx_outs: Vec<(TxOut, FogTxOut)>,
+    /// The timestamp of this block
+    pub timestamp: u64,
+}
+
+impl TryFrom<TxsForIngest> for PreparedBlockData {
+    type Error = Error;
+
+    fn try_from(src: TxsForIngest) -> Result<PreparedBlockData> {
+        let tx_outs = src
+            .redacted_txs
+            .into_iter()
+            .map(|txo| match FogTxOut::try_from(&txo) {
+                Ok(ftxo) => Ok((txo, ftxo)),
+                Err(err) => Err(Error::InvalidBlockData(err)),
+            })
+            .collect::<Result<_>>()?;
+
+        Ok(PreparedBlockData {
+            block_index: src.block_index,
+            global_txo_index: src.global_txo_index,
+            timestamp: src.timestamp,
+            tx_outs,
+        })
+    }
 }
--- a/fog/ingest/enclave/impl/tests/tx_processing.rs
+++ b/fog/ingest/enclave/impl/tests/tx_processing.rs
@@ -137,6 +137,7 @@ fn test_ingest_enclave(logger: Logger) {
                 let tx_out_record = bob_fog_credential
                     .decrypt_tx_out_result(tx_rows[idx].payload.clone())
                     .unwrap();
+
                 assert_eq!(tx_out_record.block_index, txs_for_ingest.block_index);
                 assert_eq!(
                     tx_out_record.tx_out_global_index,
@@ -144,7 +145,7 @@ fn test_ingest_enclave(logger: Logger) {
                 );
                 assert_eq!(
                     tx_out_record.get_fog_tx_out().unwrap(),
-                    FogTxOut::from(&tx_outs_for_bob[idx])
+                    FogTxOut::try_from(&tx_outs_for_bob[idx]).unwrap()
                 );
             }
 
@@ -305,7 +306,7 @@ fn test_ingest_enclave_malformed_txos(logger: Logger) {
                 );
                 assert_eq!(
                     tx_out_record.get_fog_tx_out().unwrap(),
-                    FogTxOut::from(&tx_outs[4 * index3])
+                    FogTxOut::try_from(&tx_outs[4 * index3]).unwrap()
                 );
 
                 assert!(
@@ -527,7 +528,7 @@ fn test_ingest_enclave_overflow(logger: Logger) {
                         "{:?} {:?} alice:{} bob:{}",
                         tx_out_record, tx_row, alice_found, bob_found,
                     );
-                    let expected_fog_tx_out = FogTxOut::from(&all_tx_outs[idx]);
+                    let expected_fog_tx_out = FogTxOut::try_from(&all_tx_outs[idx]).unwrap();
                     assert_eq!(
                         tx_out_record.get_fog_tx_out().unwrap(),
                         expected_fog_tx_out,
@@ -567,7 +568,7 @@ fn test_ingest_enclave_overflow(logger: Logger) {
                         "{:?} {:?} alice:{} bob:{}",
                         tx_out_record, tx_row, alice_found, bob_found,
                     );
-                    let expected_fog_tx_out = FogTxOut::from(&all_tx_outs[idx]);
+                    let expected_fog_tx_out = FogTxOut::try_from(&all_tx_outs[idx]).unwrap();
                     assert_eq!(
                         tx_out_record.get_fog_tx_out().unwrap(),
                         expected_fog_tx_out,
--- a/fog/sample-paykit/src/cached_tx_data/mod.rs
+++ b/fog/sample-paykit/src/cached_tx_data/mod.rs
@@ -595,42 +595,52 @@ impl CachedTxData {
                 block_data.global_txo_count - number_of_tx_outs_in_block;
 
             for (i, external_tx_out) in block_data.outputs.iter().enumerate() {
-                match TxOut::try_from(external_tx_out) {
-                    Ok(tx_out) => {
-                        let fog_tx_out: FogTxOut = FogTxOut::from(&tx_out);
-
-                        let fog_tx_out_metadata: FogTxOutMetadata = FogTxOutMetadata {
-                            global_index: first_tx_out_global_index + i as u64,
-                            block_index: block_data.index,
-                            timestamp: block_data.timestamp,
-                        };
-
-                        let tx_out_record: TxOutRecord =
-                            TxOutRecord::new(fog_tx_out, fog_tx_out_metadata);
-                        // Try to create an OwnedTxOut. If this fails, and it
-                        // will fail for the majority of TxOutRecords from these
-                        // missed blocks, then view key scanning failed, which
-                        // means that the user doesn't own this TxOut. Do this
-                        // here before adding it to the returned TxOutRecord
-                        // vector to prevent unnecssary logs to be emitted when
-                        // the TxOutRecords are consumed.
-                        if OwnedTxOut::new(
-                            tx_out_record.clone(),
-                            &self.account_key,
-                            &self.spsk_to_index,
-                        )
-                        .is_ok()
-                        {
-                            tx_out_records.push(tx_out_record);
-                        }
-                    }
+                let tx_out = match TxOut::try_from(external_tx_out) {
+                    Ok(tx_out) => tx_out,
                     Err(error) => {
                         log::warn!(
                             self.logger,
                             "TxOut could not be created from external.TxOut: {}",
                             error
                         );
+                        continue;
                     }
+                };
+
+                let fog_tx_out = match FogTxOut::try_from(&tx_out) {
+                    Ok(fog_tx_out) => fog_tx_out,
+                    Err(error) => {
+                        log::warn!(
+                            self.logger,
+                            "FogTxOut could not be created from TxOut: {}",
+                            error
+                        );
+                        continue;
+                    }
+                };
+
+                let fog_tx_out_metadata: FogTxOutMetadata = FogTxOutMetadata {
+                    global_index: first_tx_out_global_index + i as u64,
+                    block_index: block_data.index,
+                    timestamp: block_data.timestamp,
+                };
+
+                let tx_out_record = TxOutRecord::new(fog_tx_out, fog_tx_out_metadata);
+                // Try to create an OwnedTxOut. If this fails, and it
+                // will fail for the majority of TxOutRecords from these
+                // missed blocks, then view key scanning failed, which
+                // means that the user doesn't own this TxOut. Do this
+                // here before adding it to the returned TxOutRecord
+                // vector to prevent unnecssary logs to be emitted when
+                // the TxOutRecords are consumed.
+                if OwnedTxOut::new(
+                    tx_out_record.clone(),
+                    &self.account_key,
+                    &self.spsk_to_index,
+                )
+                .is_ok()
+                {
+                    tx_out_records.push(tx_out_record);
                 }
             }
         }
@@ -905,7 +915,7 @@ impl OwnedTxOut {
         let decompressed_tx_pub = RistrettoPublic::try_from(&tx_out.public_key)?;
         let shared_secret =
             get_tx_out_shared_secret(account_key.view_private_key(), &decompressed_tx_pub);
-        let (amount, _blinding) = tx_out.masked_amount.get_value(&shared_secret)?;
+        let (amount, _blinding) = tx_out.get_masked_amount()?.get_value(&shared_secret)?;
 
         // Calculate the subaddress spend public key for tx_out.
         let tx_out_target_key = RistrettoPublic::try_from(&tx_out.target_key)?;
--- a/fog/sample-paykit/src/cached_tx_data/mod.rs
+++ b/fog/sample-paykit/src/cached_tx_data/mod.rs
@@ -619,7 +619,7 @@ impl CachedTxData {
                     }
                 };
 
-                let fog_tx_out_metadata: FogTxOutMetadata = FogTxOutMetadata {
+                let fog_tx_out_metadata = FogTxOutMetadata {
                     global_index: first_tx_out_global_index + i as u64,
                     block_index: block_data.index,
                     timestamp: block_data.timestamp,
--- a/fog/types/src/view.rs
+++ b/fog/types/src/view.rs
@@ -210,10 +210,14 @@ impl From<&MaskedAmount> for TxOutAmountMaskedTokenId {
     fn from(src: &MaskedAmount) -> Self {
         match src {
             MaskedAmount::V1(masked_amount) => {
-                TxOutAmountMaskedTokenId::TxOutAmountMaskedTokenIdV1(masked_amount.masked_token_id.clone())
+                TxOutAmountMaskedTokenId::TxOutAmountMaskedTokenIdV1(
+                    masked_amount.masked_token_id.clone(),
+                )
             }
             MaskedAmount::V2(masked_amount) => {
-                TxOutAmountMaskedTokenId::TxOutAmountMaskedTokenIdV2(masked_amount.masked_token_id.clone())
+                TxOutAmountMaskedTokenId::TxOutAmountMaskedTokenIdV2(
+                    masked_amount.masked_token_id.clone(),
+                )
             }
         }
     }
--- a/fog/types/src/view.rs
+++ b/fog/types/src/view.rs
@@ -202,7 +202,9 @@ pub enum TxOutAmountMaskedTokenId {
 
 impl Default for TxOutAmountMaskedTokenId {
     fn default() -> Self {
-        Self::TxOutAmountMaskedTokenIdV2(Default::default())
+        // Note: This must default to empty bytes with tag V1, to support the
+        // backwards compatibility with before masked token id even existed.
+        Self::TxOutAmountMaskedTokenIdV1(Default::default())
     }
 }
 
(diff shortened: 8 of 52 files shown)


fog/api/proto/view.proto
```diff
@@ -268,4 +268,8 @@ message TxOutRecord {
     bytes tx_out_e_memo_data = 9;
     /// The masked token id associated to the amount field in the TxOut that was recovered
     bytes tx_out_amount_masked_token_id = 10;
+    /// The masked token id associated to the v2 amount field in the TxOut that was recovered
+    /// This field is mutually exclusive with `tx_out_amount_masked_token_id`. It's presense indicates that
```
remoun, line 272: Nit: typo
```suggestion
    /// This field is mutually exclusive with `tx_out_amount_masked_token_id`. It's presence indicates that
```

fog/api/proto/view.proto
```diff
@@ -268,4 +268,8 @@ message TxOutRecord {
...
+    /// a MaskedAmountV2 object was serialized, rather than a V1 object.
+    bytes tx_out_amount_v2_masked_token_id = 11;
```
remoun, line 274: Why not make this a `oneof`?

fog/types/src/view.rs
```diff
@@ -344,16 +354,35 @@ pub struct FogTxOut {
...
-        Self {
-            target_key: src.target_key,
-            public_key: src.public_key,
-            amount_masked_value: src.masked_amount.masked_value,
-            amount_masked_token_id: src.masked_amount.masked_token_id.clone(),
-            amount_commitment_data_crc32: Crc::<u32>::new(&crc::CRC_32_ISO_HDLC)
-                .checksum(src.masked_amount.commitment.point.as_bytes()),
-            e_memo: src.e_memo,
+        match src.get_masked_amount() {
+            Err(_) => Self {
```
remoun, line 359: this seems like it could silently introduce bad data. Why not make this `TryFrom`?

transaction/core/src/amount/mod.rs
```diff
@@ -5,85 +5,62 @@
 //! Amounts are implemented as Pedersen commitments. The associated private keys
 //! are "masked" using a shared secret.
 
-#![cfg_attr(test, allow(clippy::unnecessary_operation))]
+#![allow(missing_docs)]
```
remoun, line 8:
```suggestion
// FIXME: Remove when prost-generated for `derive(Oneof)` has the necessary
// doc comments: https://github.com/tokio-rs/prost/issues/237
#![allow(missing_docs)]
```

transaction/core/src/lib.rs
```diff
@@ -63,6 +63,9 @@ pub mod ring_signature {
 // Re-export the one-time keys module which historically lived in this crate
 pub use mc_crypto_ring_signature::onetime_keys;
 
+// Re-export some dependent types from mc-account-keys
+pub use mc_account_keys::PublicAddress;
```
remoun, line 67: This is definitely useful! Maybe also `AccountKey`?

transaction/core/src/tx.rs
```diff
@@ -376,12 +382,23 @@ impl TxOut {
         hint: EncryptedFogHint,
         memo_fn: impl FnOnce(MemoContext) -> Result<MemoPayload, NewMemoError>,
     ) -> Result<Self, NewTxError> {
+        // Conform to block version rules
+        if !block_version.masked_token_id_feature_is_supported() {
+            if amount.token_id != 0 {
```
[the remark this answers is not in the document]
garbageslam: okay, makes sense

transaction/core/src/amount/v1.rs
```diff
@@ -0,0 +1,341 @@
...
+
+    /// Reveals `masked_token_id`, with backwards compat
+    fn unmask_token_id(
+        masked_token_id: &[u8],
+        tx_out_shared_secret: &RistrettoPublic,
+    ) -> Result<u64, AmountError> {
+        match masked_token_id.len() {
+            0 => Ok(0),
+            TokenId::NUM_BYTES => {
+                let masked_token_id_val = u64::from_le_bytes(masked_token_id.try_into().unwrap());
```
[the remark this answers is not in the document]
garbageslam: I think it's okay -- the thing is, we know statically that the length of the slice is 8, so converting to a slice of length 8 cannot fail.

I never found a rust syntax that lets me do this in safe code without the possibility of panicking, so I just rely on the optimizer to eliminate the failure path here.

Right now we have a lot of `try_into().unwrap()` like this in enclave code as well.

---

How would you feel about something like this

```
match u64::from_le_bytes(...) {
    Ok(masked_token_id_val) => ...
    Err(_) => match masked_token_id.len() {
      0=> Ok(0)
      other => InvalidMaskedTokenIdLength(other)
    }
```

fog/api/proto/view.proto
garbageslam: In this case specifically, I really want to minimize the number of bytes on the wire that go into fog view, because that's sort of the most difficult ORAM in the whole project, and we are very close to another breakpoint.

Tags are stored using the "zig zag" encoding in protobuf, and I think 10 and 11 are the same number of bytes on the wire in this encoding.

I'm not sure exactly what happens on the wire if we use a one-of, but if it would require us to store another discriminant it would be bad. That's how I assumed it worked anyways, although this PR has taught me that actually each one-of possibility has its own tag. So maybe it would be fine to use a one-of.

fog/types/src/view.rs
garbageslam: I started with making this TryFrom, but this created problems in the fog ingest enclave implementation. Basically once we start processing TxOut's, we need the rest of the operation to be no-fail.

We could do all these conversions up-front before processing anything I guess.

I'll revisit this, thanks for the comment

transaction/core/src/amount/mod.rs
garbageslam: thanks

transaction/core/src/lib.rs
garbageslam: yah okay will do

transaction/core/src/amount/error.rs
```diff
@@ -1,9 +1,11 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
 //! Errors that can occur when handling an amount commitment.
 
 use displaydoc::Display;
 
 /// An error which can occur when handling an amount commitment.
-#[derive(Debug, Display, Eq, PartialEq)]
+#[derive(Debug, Display, Eq, PartialEq, Clone)]
```
remoun, line 8: Nit: Prefer keeping Derive's sorted.
```suggestion
#[derive(Clone, Debug, Display, Eq, PartialEq)]
```

transaction/core/src/amount/v2.rs
```diff
@@ -0,0 +1,337 @@
...
+        Crc::<u32>::new(&crc::CRC_32_ISO_HDLC).checksum(commitment.point.as_bytes())
+    }
+}
+
+/// Computes the amount shared secret from the tx_out shared secret
+fn get_amount_shared_secret(tx_out_shared_secret: &RistrettoPublic) -> [u8; 32] {
+    let mut hasher = Blake2b512::new();
+    hasher.update(&AMOUNT_SHARED_SECRET_DOMAIN_TAG);
+    hasher.update(&tx_out_shared_secret.to_bytes());
+    hasher.finalize()[0..32].try_into().unwrap()
```
remoun, line 215:
```suggestion
    // Safety: Blake2b512 is a 512-bit (64-byte) hash.
    hasher.finalize()[0..32].try_into().unwrap()
```

transaction/core/src/amount/v1.rs
```diff
@@ -0,0 +1,337 @@
...
+
+    proptest! {
+
+            #[test]
+            /// MaskedAmount::new() should return Ok for valid values and blindings.
+            fn test_new_ok(
+                value in any::<u64>(),
+                token_id in any::<u64>(),
+                shared_secret in arbitrary_ristretto_public()) {
+                    let amount = Amount { value, token_id: token_id.into() };
```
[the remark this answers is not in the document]
garbageslam: hmm maybe rebase did something funky, iiuc it skips the commit hook during rebase steps

garbageslam:
rebased on master to resolve conflicts

android-bindings/src/bindings.rs
```diff
@@ -321,7 +321,9 @@ pub unsafe extern "C" fn Java_com_mobilecoin_lib_Amount_init_1jni(
 
         // FIXME #1595: We should get a masked token id also, here we default to
         // 0 bytes, which is backwards compatible
-        let masked_amount = MaskedAmount {
+        // FIXME: This should probably be VersionedMaskedAmount instead, and we
+        // should pass enough data to decide whether to use V1 or V2
```
iamalwaysuncomfortable, line 325: What data would be passed to provide context? 

Potentially some amount of key material like Txout pubkey into a ::new() method for masked amount which can then do a reconstruction of the shared secret to see if it was derived from the original TxOut shared secret or the v2 shared secret? That would be a heavier, but unambigous way to do it?

Perhaps there's a cheaper way of doing that can do an automatic tell without having to pass in something explicitly telling which version to use (which would be operator-error prone)?

fog/types/src/view.rs
```diff
@@ -236,9 +236,14 @@ pub struct TxOutRecord {
...
     /// The masked token id associated to the amount field in the TxOut that
-    /// was recovered
+    /// was recovered. This is used when the masked amount has version 1.
     #[prost(bytes, tag = "10")]
     pub tx_out_amount_masked_token_id: Vec<u8>,
+
+    /// The masked token id associated to the amount field in the TxOut that
+    /// was recovered. This is used when the masked amount has version 2.
+    #[prost(bytes, tag = "11")]
+    pub tx_out_amount_v2_masked_token_id: Vec<u8>,
```
iamalwaysuncomfortable, line 246: What's motivating the choice to make a separate field for a v2 token id?

crypto/digestible/derive/src/lib.rs
```diff
@@ -547,6 +549,103 @@ fn try_digestible_enum(
     Ok(expanded.into())
 }
 
+fn try_digestible_enum_transparent(
+    ident: &Ident,
+    generics: &Generics,
+    variant_data: &DataEnum,
+) -> Result<TokenStream, &'static str> {
```
iamalwaysuncomfortable, line 556: What does this do? It looks to me like this is trying to digest the VersionedAmount enum in a zero knowledge way - is that correct? I know this is a research PR but, if you find this direction a worthy path to pursue, def document this one!

api/src/convert/tx_out.rs
```diff
@@ -117,14 +121,14 @@ mod tests {
...
-                .unwrap(),
-            target_key: RistrettoPublic::from_random(&mut rng).into(),
-            public_key: RistrettoPublic::from_random(&mut rng).into(),
-            e_fog_hint: (&[0u8; ENCRYPTED_FOG_HINT_LEN]).into(),
-            e_memo: Some((*GenericArray::from_slice(&[9u8; 66])).into()),
-        };
+        let source = tx::TxOut::new(
+            BlockVersion::MAX,
+            amount,
+            &PublicAddress::from_random(&mut rng),
```
iamalwaysuncomfortable, line 127: Is PublicAddress the right abstraction here? It's a more concise container but I think it's a confusing context because target & txout pubkey don't really match the category PublicAddress was designed for (i.e. a tuple of EC points uniquely addressing a user). 

This could be re-addressed to have the documentation of PublicAddress state that it can be used generalize to related pairs? Or make a separate struct like TxOutPublicAddress?

api/src/convert/amount.rs
```diff
@@ -16,18 +35,62 @@ impl From<&MaskedAmount> for external::MaskedAmount {
...
+        let amount = MaskedAmountV2 {
+            commitment,
+            masked_value,
+            masked_token_id,
+        };
+        Ok(amount)
+    }
+}
+
+impl TryFrom<&external::TxOut_oneof_masked_amount> for VersionedMaskedAmount {
```
iamalwaysuncomfortable, line 83: If we can invent around the challenges with this, this seems like the superior approach.

crypto/digestible/derive/src/lib.rs
garbageslam: yeah if it looks like this is working out, i think we need to pull out the digestible-derive changes into their own PR to ease review

android-bindings/src/bindings.rs
garbageslam: i think there's no way that can work

they are going to get from the proto they receive data about whether this is masked amount v1 or masked amount v2.

a guess and check operation will require at least one extra elliptic curve point operation, so that's a pretty massive computational overhead, when it's easy enough for them to just pass the data they got from the proto (or, make a proto-oriented API)

crypto/digestible/derive/src/lib.rs
iamalwaysuncomfortable: It does indeed look like it's working out

fog/types/src/view.rs
iamalwaysuncomfortable: That totally makes sense. I think it read a little odd, but I don't have a better suggestion at the moment.

garbageslam:
rebased again, most of the conflicsts where for mc-blockchain-types being split out

garbageslam:
closed by accident

fog/api/proto/view.proto
garbageslam: I'm going to change this to oneof as suggested, thanks :)

fog/ingest/enclave/impl/tests/tx_processing.rs
```diff
@@ -94,7 +94,7 @@ fn test_ingest_enclave(logger: Logger) {
             // Check that the tx row ciphertexts have the right size (but only at block
             // version max)
             if block_version == BlockVersion::MAX {
-                const EXPECTED_PAYLOAD_SIZE: usize = 237; // The observed tx_row.payload size
+                const EXPECTED_PAYLOAD_SIZE: usize = 239; // The observed tx_row.payload size
```
garbageslam, line 97: @remoun for some reason after I changed the TxOutRecord to use `oneof`, the size of this seems to have increased. i'm going to tinker around with this some more and make sure, but if true that's a reason not to use oneof

garbageslam: i guess it's probably fine either way, we're still at 239 + 1 + 16 = 256 bytes exactly

garbageslam: and circuit oram is coming in this release so the 256 bytes thing probably won't be as critical

eranrund: If the size increased could this potentially indicate the change is not backwards compatible?

ledger/db/src/ledger_db.rs
```diff
@@ -728,6 +728,14 @@ impl LedgerDB {
             return Err(Error::InvalidBlockContents);
         }
 
+        // Check that none of the outputs are missing masked amount (or, have a masked
+        // amount we don't understand)
+        for output in &block_contents.outputs {
+            if output.get_masked_amount().is_err() {
+                return Err(Error::MissingMaskedAmount);
+            }
+        }
```
eranrund, line 737: Consider `if block_contents.outputs().iter().any(|output| output.get_masked_amount().is_err())`

transaction/core/src/amount/mod.rs
```diff
@@ -5,85 +5,67 @@
 //! Amounts are implemented as Pedersen commitments. The associated private keys
 //! are "masked" using a shared secret.
 
-#![cfg_attr(test, allow(clippy::unnecessary_operation))]
+// FIXME: Remove when prost-generated for `derive(Oneof)` has the necessary
+// doc comments: https://github.com/tokio-rs/prost/issues/237
+#![allow(missing_docs)]
```
eranrund, line 10: Prost 0.11 was released that I think addresses this.

transaction/core/src/amount/v2.rs
```diff
@@ -0,0 +1,338 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
+//! A commitment to an output's amount, denominated in picoMOB.
+//!
+//! Amounts are implemented as Pedersen commitments. The associated private keys
+//! are "masked" using a shared secret.
```
eranrund, line 6: Might be worth mentioning what this adds/changes from V1

transaction/core/src/amount/v2.rs
```diff
@@ -0,0 +1,338 @@
...
+    (
+        u64::from_le_bytes(value_mask),
+        u64::from_le_bytes(token_id_mask),
+        Scalar::from_bytes_mod_order_wide(&scalar_blinding_bytes),
+    )
+}
+
+#[cfg(test)]
+mod amount_tests {
+    #![allow(clippy::unnecessary_operation)]
```
eranrund, line 255: This is already at the top of the file so is it necessary here?

transaction/core/src/memo.rs
```diff
@@ -224,7 +224,7 @@ derive_prost_message_from_repr_bytes!(MemoPayload);
 derive_debug_and_display_hex_from_as_ref!(MemoPayload);
 
 /// An error which can occur when handling memos
-#[derive(Debug, Display, Eq, PartialEq)]
+#[derive(Debug, Display, Eq, PartialEq, Clone)]
```
eranrund, line 227:
```suggestion
#[derive(Clone, Debug, Display, Eq, PartialEq)]
```

transaction/core/src/tx_error.rs
```diff
@@ -31,13 +29,22 @@ impl From<NewMemoError> for NewTxError {
     }
 }
 
+/// An error that occurs when handling a TxOut
+#[derive(Debug, Display, Clone)]
```
eranrund, line 33:
```suggestion
#[derive(Clone, Debug, Display)]
```

transaction/core/src/validation/error.rs
```diff
@@ -160,3 +163,11 @@ impl From<InputRuleError> for TransactionValidationError {
         Self::InputRule(src)
     }
 }
+
+impl From<TxOutConversionError> for TransactionValidationError {
+    fn from(src: TxOutConversionError) -> Self {
+        match src {
+            TxOutConversionError::UnknownMaskedAmountVersion => Self::UnknownMaskedAmountVersion,
```
eranrund, line 170: Any reason to do this versus include a `TxOutConversion(TxOutConversionError)` variant in the `TransactionValidationError` enum?

eranrund (commented):
Nice, overall LGTM but have a few small nits and questions.

Also wondering, the pattern of `tx_out.get_masked_amount().map_err(...).get_value()` (or `unwrap` instead of `map_err`) is repeated in many places. Maybe it is better to include a `get_masked_amount_value` method in `TxOut`?

Thank you!

fog/ingest/enclave/impl/tests/tx_processing.rs
garbageslam: Not sure. We should be able to test it using CD

transaction/core/src/amount/mod.rs
garbageslam: Cool

fog/ingest/enclave/impl/tests/tx_processing.rs
eranrund: Seems important...

garbageslam: let me explain, there were two separate changes:

1. Make TxOut use a `oneof` field where previously we had a masked amount message
2. Make TxOutRecord (used in fog) use a `oneof` field where previously we had a `bytes` type

(1) did not cause a change in size here. But (2) did, and I'm not sure why.

I'm not worried about an incompatibility in the blockchain part, because that did not affect the size, and we went through CD with that. Would like to better understand what actually caused the size increase at step (2), maybe i screwed up the protobuf annotation or something.

fog/ingest/enclave/impl/src/lib.rs
```diff
@@ -383,3 +389,45 @@ impl<OSC: ORAMStorageCreator<StorageDataSize, StorageMetaSize>> IngestEnclave
...
+        Ok(PreparedBlockData {
+            block_index: src.block_index,
+            global_txo_index: src.global_txo_index,
+            timestamp: src.timestamp,
+            tx_outs: src
+                .redacted_txs
+                .into_iter()
+                .map(|txo| match FogTxOut::try_from(&txo) {
+                    Ok(ftxo) => Ok((txo, ftxo)),
+                    Err(_) => Err(Error::InvalidBlockData),
```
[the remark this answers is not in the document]
garbageslam: i think right now it's like, this would require `AmountError` to support `serde` because these errors are serde serialized. but there are separate efforts going on to make a bunch of transaction core stuff be no-alloc friendly to support hardware wallets. so i'm worried about adding more serde here and creating more work here.

maybe we should just turn the error to string to sidestep the serde issue?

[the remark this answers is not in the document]
garbageslam: i guess i shold just make those errors serde as well, there is a lot of serde stuff in transaction core and whatever solution is used for all of that should apply just fine for these errors too

fog/types/src/view.rs
```diff
@@ -342,18 +374,34 @@ pub struct FogTxOut {
...
+                    MaskedTokenId::V1(
+                        masked_amount.masked_token_id.clone(),
+                    ), masked_amount.commitment_crc32())
+                },
+                VersionedMaskedAmount::V2(masked_amount) => {
+                    (masked_amount.masked_value, MaskedTokenId::V2(
+                        masked_amount.masked_token_id.clone(),
+                    ), masked_amount.commitment_crc32())
+                },
+        };
```
garbageslam, line 394: this is what i ended up with after drying it up, but i think maybe it can still be better, like if we add a function it could be:

```
let masked_amount = src.get_masked_amount().map_err(...)?;
let amount_masked_value = masked_amount.get_masked_value();
let commitment = masked_amount.get_commitment();
let amount_masked_token_id = match ... {

};
```

transaction/core/src/amount/v2.rs
```diff
@@ -0,0 +1,341 @@
...
+    /// third party who can then use this API to audit the value of
+    /// a TxOut without having permissions to do other things that require the
+    /// TxOut shared secret.
+    ///
+    /// # Arguments
+    /// * `amount_shared_secret` - The shared secret, derived by hashing TxOut
+    ///   shared secret
+    pub fn get_value_from_amount_shared_secret(
+        &self,
+        amount_shared_secret: &[u8; 32],
```
nick-mobilecoin, line 120: :wrench: This might be better as a follow on PR, iff desired
Since this is a public method it might be better to have a dedicated type for `amount_shared_secret` so that it's easier to see the flow and prevents one from accidentally passing in a random 512bit hash trying to get things working

transaction/core/src/amount/v2.rs
```diff
@@ -0,0 +1,341 @@
...
+
+    /// Returns the amount underlying the masked amount, given the amount shared
+    /// secret.
+    ///
+    /// Generally, the recipient knows the TxOut shared secret, and it's more
+    /// convenient to use the other API. This API allows that the sender or
+    /// recipient could selectively disclose the amount_shared_secret to a
+    /// third party who can then use this API to audit the value of
+    /// a TxOut without having permissions to do other things that require the
+    /// TxOut shared secret.
```
nick-mobilecoin, line 113: :upside_down_face: 
notes on the suggestion:
* Prefer explicit over implicit, thus the function name
* I very well could be wrong, but I think of the entire module/struct as the API versus an individual `function`.
```suggestion
    /// Generally, the recipient knows the TxOut shared secret, and it's more
    /// convenient to use [get_value()]. This function allows that the sender or
    /// recipient could selectively disclose the `amount_shared_secret` to a
    /// third party who can then use this function to audit the value of
    /// a TxOut without having permissions to do other things that require the
    /// TxOut shared secret.
```

transaction/core/src/amount/v2.rs
```diff
@@ -0,0 +1,341 @@
...
+    /// is malformed.
+    fn compute_commitment(
+        masked_value: u64,
+        masked_token_id: &[u8],
+        amount_shared_secret: &[u8; 32],
+    ) -> Result<(CompressedCommitment, Amount, Scalar), AmountError> {
+        let (value_mask, token_id_mask, blinding) = get_blinding_factors(amount_shared_secret);
+
+        let token_id = TokenId::from(match masked_token_id.len() {
+            0 => 0,
```
nick-mobilecoin, line 191: ~:question: My understanding is that v2 of the `VersionedMasked` is block 3, and token id was added in block 2, do we anticipate token id being 0 with this change?~

:x: Thinking about this more in `new()` the `amount.token_id` is *always* masked with `token_id_mask`.  I think of `0` got in here somehow this would not provide the right `token_id`

transaction/std/src/transaction_builder.rs
```diff
@@ -427,7 +427,8 @@ impl<FPR: FogPubkeyResolver> TransactionBuilder<FPR> {
             create_output_with_fog_hint(self.block_version, amount, recipient, hint, memo_fn, rng)?;
 
         let (amount, blinding) = tx_out
-            .masked_amount
+            .get_masked_amount()
+            .expect("TransactionBuilder created an invalid VersionedMaskedAmount")
```
nick-mobilecoin, line 431: :wrench: This was here before with the expect 2 lines below, but it seems inconsistent that this error will cause the program to abort, while other errors in this function keep operating.
Perhaps since there is now ` TxOutConversion(TxOutConversionError),` these might return the error.

nick-mobilecoin (changes requested):
Most are stylistic/questions I think the only blockers are:
* ensuring the correct use of blake with HKDF
* ensuring proper handling/masking of token_id 0

transaction/core/src/amount/mod.rs
```diff
@@ -93,30 +71,58 @@ impl MaskedAmount {
...
-            // spendable).
-            return Err(AmountError::InconsistentCommitment);
+        match &self {
+            Self::V1(masked_amount) => masked_amount.get_value(tx_out_shared_secret),
+            Self::V2(masked_amount) => masked_amount.get_value(tx_out_shared_secret),
+        }
+    }
+
+    /// Get the masked value field
+    pub fn get_masked_value(&self) -> &u64 {
```
remoun, line 86: [Rust style guide says to omit `get_` prefix](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter) (throughout this PR)

transaction/types/src/block_version.rs
```diff
@@ -120,6 +120,12 @@ impl BlockVersion {
         self.0 >= 3
     }
 
+    /// Masked amount v2 derivation [MCIP #42](https://github.com/mobilecoinfoundation/mcips/pull/42)
+    /// are introduced in block version 3
```
remoun, line 124: Nit: Summary first, then link to MCIP
```suggestion
    /// Masked amount V2 derivation introduced with block version 3.
    /// [MCIP #42](https://github.com/mobilecoinfoundation/mcips/pull/42)
```

fog/ingest/enclave/impl/src/lib.rs
```diff
@@ -143,13 +143,12 @@ impl<OSC: ORAMStorageCreator<StorageDataSize, StorageMetaSize>> SgxIngestEnclave
...
                 // Create a TxOutRecord, flattening the Txo data and getting extra data like
                 // global index, block index, timestamp.
-                let fog_tx_out = FogTxOut::from(txo);
                 let meta = FogTxOutMetadata {
                     global_index: chunk.global_txo_index + index as u64,
                     block_index: chunk.block_index,
                     timestamp: chunk.timestamp,
                 };
-                let txo_record = TxOutRecord::new(fog_tx_out, meta);
+                let txo_record = TxOutRecord::new(fog_tx_out.clone(), meta);
```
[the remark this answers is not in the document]
garbageslam: There is some discussion in the README of mc-oblivious repo

At this particular line, we're not worried about constant-time, because we aren't trying to be constant-time with respect to which fog_tx_out we are processing, the operator already knows / can infer that. We're just trying to be constant time with respect to the contents of the decrypted fog hint

transaction/core/src/amount/v2.rs
garbageslam: let's do that in a later PR, this one is already pretty big

fog/types/src/view.rs
```diff
@@ -182,33 +182,61 @@ pub struct TxOutSearchResult {
...
+
+    /// A v2 masked amount's token id bytes
+    /// This is used when the masked amount in the original TxOut has version 2
+    #[prost(message, tag = "11")]
+    V2(Vec<u8>),
+}
+
+impl Default for MaskedTokenId {
+    fn default() -> Self {
+        Self::V1(Default::default())
```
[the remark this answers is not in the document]
garbageslam: sure

fog/types/src/view.rs
```diff
@@ -342,18 +374,32 @@ pub struct FogTxOut {
...
+        let amount_masked_value = *masked_amount.get_masked_value();
+        let amount_commitment_data_crc32 = masked_amount.commitment_crc32();
+        let amount_masked_token_id = match masked_amount {
+            VersionedMaskedAmount::V1(masked_amount) => {
+                MaskedTokenId::V1(masked_amount.masked_token_id.clone())
+            }
+            VersionedMaskedAmount::V2(masked_amount) => {
+                MaskedTokenId::V2(masked_amount.masked_token_id.clone())
+            }
+        };
```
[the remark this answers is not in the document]
garbageslam: sure

api/src/convert/amount.rs
```diff
@@ -15,18 +34,62 @@ impl From<&MaskedAmount> for external::MaskedAmount {
     }
 }
 
-impl TryFrom<&external::MaskedAmount> for MaskedAmount {
+impl From<&VersionedMaskedAmount> for external::TxOut_oneof_masked_amount {
```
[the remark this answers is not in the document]
garbageslam: I'm okay to remove `Versioned` from this struct, I guess it will be clear from context that it is versioned

fog/types/src/view.rs
```diff
@@ -380,19 +426,35 @@ impl FogTxOut {
...
+            MaskedTokenId::V2(bytes) => {
+                let (masked_amount, _) = VersionedMaskedAmount::reconstruct_v2(
+                    self.amount_masked_value,
+                    bytes,
+                    &tx_out_shared_secret,
+                )
+                .map_err(FogTxOutError::Amount)?;
+                masked_amount
+            }
+        };
```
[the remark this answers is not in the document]
garbageslam: yeah I'm okay to do that

libmobilecoin/src/transaction.rs
```diff
@@ -62,8 +62,11 @@ pub extern "C" fn mc_tx_out_reconstruct_commitment(
 
         // FIXME #1596: McTxOutAmount should include the masked_token_id bytes, which
         // are 0 or 4 bytes For now zero to avoid breaking changes to FFI
+        // FIXME: We should either make this use VersionedMaskedAmount instead and pass
```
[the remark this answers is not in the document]
garbageslam: libmobilecoin has been removed from this repo, I will follow up with SDK team on this

mobilecoind-json/src/data_types.rs
```diff
@@ -491,21 +491,63 @@ pub struct JsonMaskedAmount {
...
+            )),
+            Some(other) => Err(format!("Unknown masked amount version: {}", other)),
         }
     }
 }
 
 #[derive(Deserialize, Serialize, Default, Debug, Clone)]
 pub struct JsonTxOut {
-    pub masked_amount: JsonMaskedAmount,
+    pub masked_amount: Option<JsonMaskedAmount>,
```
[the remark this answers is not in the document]
garbageslam: good point, will circle back on this

transaction/core/src/amount/v2.rs
```diff
@@ -0,0 +1,341 @@
...
+/// Computes the value mask, token id mask, and blinding factor for the
+/// commitment, in a masked amount.
+///
+/// # Arguments
+/// * `amount_shared_secret` - The amount shared secret, derived as a hash of
+///   `rB`.
+fn get_blinding_factors(amount_shared_secret: &[u8; 32]) -> (u64, u64, Scalar) {
+    // Use HKDF-SHA512 to produce blinding factors for value, token id, and
+    // commitment
+    let kdf = Hkdf::<Sha512>::new(Some(b"mc-amount-blinding-factors"), amount_shared_secret);
```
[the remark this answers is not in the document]
garbageslam: sure, will do

transaction/core/src/amount/v2.rs
```diff
@@ -0,0 +1,341 @@
...
+    /// # Arguments
+    /// * `amount` - The amount information to be masked
+    /// * `shared_secret` - The shared secret, e.g. `rB` for transaction private
+    ///   key `r` and recipient public key `B`.
+    #[inline]
+    pub fn new(
+        amount: Amount,
+        tx_out_shared_secret: &RistrettoPublic,
+    ) -> Result<Self, AmountError> {
+        let amount_shared_secret = get_amount_shared_secret(tx_out_shared_secret);
```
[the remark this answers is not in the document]
garbageslam: we shipped the original blockchain using hkdf blake2, it will break all the existing TxOuts if we try to change that for V1. In V2 we can do it correctly

transaction/core/src/amount/v2.rs
```diff
@@ -0,0 +1,341 @@
...
+        let (value_mask, token_id_mask, blinding) = get_blinding_factors(&amount_shared_secret);
+
+        // Pedersen generators
+        let generator = generators(*amount.token_id);
+
+        // Pedersen commitment `v*H_i + b*G`.
+        let commitment = CompressedCommitment::new(amount.value, blinding, &generator);
+
+        // The value is XORed with the first 8 bytes of the mask.
```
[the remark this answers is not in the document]
garbageslam: thanks

transaction/std/src/transaction_builder.rs
garbageslam: my thinking here was that this error should never occur in prod, it's a logic error if this operation fails and i'm basically making an assertion that it doesn't fail.

I'm okay to propagate it like you say, I think it doesn't make that much difference

mobilecoind-json/src/data_types.rs
```diff
@@ -491,21 +491,63 @@ pub struct JsonMaskedAmount {
...
+                masked_token_id: hex::encode(src.get_masked_token_id()),
+                version: None,
+            },
+            mc_api::external::TxOut_oneof_masked_amount::v2(src) => Self {
+                commitment: hex::encode(src.get_commitment().get_data()),
+                masked_value: JsonU64(src.get_masked_value()),
+                masked_token_id: hex::encode(src.get_masked_token_id()),
+                version: Some(2),
+            },
+        }
```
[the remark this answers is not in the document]
garbageslam: I tried this, but I reverted it because it doesn't actually build

I think the way it was is the best way it can actually be, the src is shadowed by what is matched in the match expression. We could try adding more helpers to this externally generated proto structure to make the code you wrote work, but I don't think it's really worth it for now

mobilecoind-json/src/data_types.rs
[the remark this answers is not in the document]
garbageslam: it's optional because otherwise there is no way to implement `From<external::TxOut>` for this type without panicking

```
impl From<&mc_api::external::TxOut> for JsonTxOut {
    fn from(src: &mc_api::external::TxOut) -> Self {
        Self {
            masked_amount: src.masked_amount.as_ref().map(Into::into),
            target_key: hex::encode(src.get_target_key().get_data()),
            public_key: hex::encode(src.get_public_key().get_data()),
            e_fog_hint: hex::encode(src.get_e_fog_hint().get_data()),
            e_memo: hex::encode(src.get_e_memo().get_data()),
        }
    }
}
```

fog/ingest/enclave/impl/tests/tx_processing.rs
garbageslam: I managed to figure this out:

```
    /// A v1 masked amount's token id bytes.
    /// This is used when the masked amount in the original TxOut has version 1
    /// Note: This tag must match the historical tag used for masked token id
    /// bytes
    #[prost(bytes, tag = "10")]
    V1(Vec<u8>),

    /// A v2 masked amount's token id bytes
    /// This is used when the masked amount in the original TxOut has version 2
    #[prost(message, tag = "11")]
    V2(Vec<u8>),
}
```

The above is wrong, I marked a bytes as a message.

I found this using protobuf inspector:

https://github.com/mildsunrise/protobuf-inspector

```
(base) chris@chris-ThinkPad-X1-Extreme-Gen-3:~/mobilecoinofficial$ protobuf_inspector < foo_proto_old
root:
    2 <64bit> = 0x5413595A7527A94F / 6058284168790911311 / 1.0332323e+97
    3 <chunk> = bytes (32)
        0000   28 DF 86 34 4C CB E0 B9 9D E6 CF DD B4 FB A4 0D 69 76 A9 D6 31 C2 0E 7F  (..4L...........iv..1...
        0018   F5 14 9A 4E CF 42 8D 46                                                  ...N.B.F
    4 <chunk> = bytes (32)
        0000   A4 B4 2C CA BD 98 32 1F 97 6F 95 7D 5C F2 02 6D 8F 31 B5 2B A3 C2 CC 25  ..,...2..o.}\..m.1.+...%
        0018   0F 82 2D A5 FC 3B C5 25                                                  ..-..;.%
    5 <64bit> = 0x000000000000006D / 109 / 5.3853155e-322
    6 <64bit> = 0x0000000000000001 / 1 / 4.9406565e-324
    7 <64bit> = 0x000000000000000A / 10 / 4.9406565e-323
    8 <32bit> = 0xEFC5EAB6 / -272241994 / -1.22505e+29
    9 <chunk> = bytes (66)
        0000   6D BB 45 75 D2 37 D0 7D A5 8F B5 DB 83 E0 86 50 DE 03 C6 E6 35 DB 09 86  m.Eu.7.}.......P....5...
        0018   B3 7F 16 E3 C4 31 68 EB 2B 50 0D E3 1B 69 74 2A 37 CE 29 0C 79 51 D8 FD  .....1h.+P...it*7.).yQ..
        0030   DF 85 00 16 C0 C8 92 2B DD CD 51 5A 5D 64 04 B8 5A EA                    .......+..QZ]d..Z.
    10 <chunk> = bytes (8)
        0000   88 D4 B1 3A D4 0D 88 F3                                                  ...:....

(base) chris@chris-ThinkPad-X1-Extreme-Gen-3:~/mobilecoinofficial$ protobuf_inspector < foo_proto_new
root:
    2 <64bit> = 0x95CDAD403242D4A3 / -7652269700599851869 / -1.1831762e-203
    3 <chunk> = bytes (32)
        0000   28 DF 86 34 4C CB E0 B9 9D E6 CF DD B4 FB A4 0D 69 76 A9 D6 31 C2 0E 7F  (..4L...........iv..1...
        0018   F5 14 9A 4E CF 42 8D 46                                                  ...N.B.F
    4 <chunk> = bytes (32)
        0000   A4 B4 2C CA BD 98 32 1F 97 6F 95 7D 5C F2 02 6D 8F 31 B5 2B A3 C2 CC 25  ..,...2..o.}\..m.1.+...%
        0018   0F 82 2D A5 FC 3B C5 25                                                  ..-..;.%
    5 <64bit> = 0x000000000000006D / 109 / 5.3853155e-322
    6 <64bit> = 0x0000000000000001 / 1 / 4.9406565e-324
    7 <64bit> = 0x000000000000000A / 10 / 4.9406565e-323
    8 <32bit> = 0xD16902BE / -781647170 / -6.25483e+10
    9 <chunk> = bytes (66)
        0000   6D BB 45 75 D2 37 D0 7D A5 8F B5 DB 83 E0 86 50 DE 03 C6 E6 35 DB 09 86  m.Eu.7.}.......P....5...
        0018   B3 7F 16 E3 C4 31 68 EB 2B 50 0D E3 1B 69 74 2A 37 CE 29 0C 79 51 D8 FD  .....1h.+P...it*7.).yQ..
        0030   DF 85 00 16 C0 C8 92 2B DD CD 51 5A 5D 64 04 B8 5A EA                    .......+..QZ]d..Z.
    11 <chunk> = message:
        1 <chunk> = bytes (8)
            0000   2B 79 59 A3 62 8B B3 9F                                                  +yY.b...

```

transaction/core/src/amount/v2.rs
garbageslam: I guess you're right, the 0 byte payload is the backwards compat path for before token id existed. But that compat path is only needed when the masked amount is version 1. A version 2 masked amount will never have an empty token id field. So we could safely remove this, it's leftover from copy paste. Thanks

transaction/core/src/amount/v2.rs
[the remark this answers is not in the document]
garbageslam: i see -- so the question is, Amount Shared Secret is new, so shouldn't it use SHA512 instead of blake2b here?

I think you are absolutely right, and I meant to change this. Thanks for spotting this

fog/types/src/view.rs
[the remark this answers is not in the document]
garbageslam: I don't see a good way to do that right now -- we have `reconstruct_v1` and `reconstruct_v2` which have the same signature.

We could have another `reconstruct` function that takes a version number? maybe as  u16?

fog/types/src/view.rs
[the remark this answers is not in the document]
garbageslam: I guess we already have a function for this:

```
pub fn get_masked_token_id(&self) -> &[u8]
```

I'm just going to call that

fog/types/src/view.rs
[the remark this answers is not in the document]
garbageslam: gonna punt on this

fog/types/src/view.rs
```diff
@@ -182,33 +182,61 @@ pub struct TxOutSearchResult {
     pub ciphertext: Vec<u8>,
 }
 
+/// An enum capturing the Oneof in the proto file around masked token id bytes
+/// This is an enum because the options are used to capture which version of
+/// masked amount was serialized.
+#[derive(Clone, Eq, Hash, Oneof, PartialEq)]
+pub enum MaskedTokenId {
```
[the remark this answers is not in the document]
garbageslam: I'm going to do the rename you proposed, so I think we don't need to change this

garbageslam:
rebased on master

transaction/std/src/transaction_builder.rs
garbageslam: changing this in a nice way means changing a bunch of surrounding code in a way that's orthogonal to this PR, so I'd rather do this in a later PR. Let's make a ticket if you want to track this

garbageslam:
this is being deployed here: https://github.com/mobilecoinfoundation/mobilecoin/pull/2483

fog/sample-paykit/src/cached_tx_data/mod.rs
```diff
@@ -595,42 +595,52 @@ impl CachedTxData {
...
+                        log::warn!(
+                            self.logger,
+                            "FogTxOut could not be created from TxOut: {}",
+                            error
+                        );
+                        continue;
+                    }
+                };
+
+                let fog_tx_out_metadata: FogTxOutMetadata = FogTxOutMetadata {
```
remoun, line 622: Nit: Unnecessary type
```suggestion
                let fog_tx_out_metadata = FogTxOutMetadata {
```

remoun (approved):

transaction/std/src/transaction_builder.rs
nick-mobilecoin: I may have latched on to the _different_ error cases too much and not scrutinized the `expect` statement
```rust
.expect("TransactionBuilder created an invalid MaskedAmount")
```
That implies the `TransactionBuilder` failed, thus an internal code error.  Maybe there is better wording, but I think the fact that it's referencing internal types should communicate an internal implementation failure, versus an external communication or malformed data error.

My initial thought was to add a `// SAFETY: ` comment like was done in the 512bit hash -> `[u8; 32]`.  But I think when used diligently, the `expect("reasoning")` should fulfill that more concisely and provide better error handling should it happen.

fog/api/proto/view.proto
```diff
@@ -267,5 +267,12 @@ message TxOutRecord {
     /// This is omitted for TxOut's from before the upgrade that introduced memos.
     bytes tx_out_e_memo_data = 9;
     /// The masked token id associated to the amount field in the TxOut that was recovered
-    bytes tx_out_amount_masked_token_id = 10;
+    oneof tx_out_masked_amount_token_id {
```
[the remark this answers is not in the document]
nick-mobilecoin: :wrench: no action or comment was made on the original mention of this.  
I'm fine if there is no desire to rename `tx_out_amount_masked_token_id` in `fog/types/src/view.rs`.  I just want to be sure it was intentional and not something that got lost.

nick-mobilecoin (approved):
LGTM

fog/api/proto/view.proto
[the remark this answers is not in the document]
garbageslam: i'm happy to change it to `amount_masked_token_id`, I think I just lost track of the issue. thanks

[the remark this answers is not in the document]
nick-mobilecoin: What I was thinking was `tx_out_amount_masked_token_id` got renamed to `tx_out_masked_amount_token_id` in the proto
So perhaps the names should usually be `<prefix>_masked_amount_<suffix>`, while right now they're `<prefix>_amount_masked_<suffix>`

fog/distribution/src/main.rs
```diff
@@ -409,7 +409,8 @@ fn select_spendable_tx_outs(
                     get_tx_out_shared_secret(account.view_private_key(), &public_key);
 
                 let (amount, _blinding_factor) = tx_out
-                    .masked_amount
+                    .get_masked_amount()
+                    .unwrap_or_else(|err| panic!("TX missing masked value: {}", err))
```
samdealy, line 413: Nit: I think you can replace this line with `expect("Tx missing masked value)` because `expect` will  append `: {ERROR MESSAGE}`. (https://doc.rust-lang.org/std/result/enum.Result.html#method.expect)

fog/api/proto/view.proto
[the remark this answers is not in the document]
garbageslam: done

samdealy (approved):
Reviewed the Fog Ingest code. LGTM

remoun (approved):
The PR description could use an update.

garbageslam:
saw this during CD:

```
2022-09-07 21:40:57.657790387 UTC INFO About to start LedgerServerConn to fog-ledger://fog.mc-masked-amount-v2.development.mobilecoin.com:443/, mc.app: test_client, mc.module: mc_fog_sample_paykit::client_builder, mc.src: fog/sample-paykit/src/client_builder.rs:138
2022-09-07 21:40:57.664170075 UTC INFO About to start LedgerServerConn to fog-ledger://fog.mc-masked-amount-v2.development.mobilecoin.com:443/, mc.app: test_client, mc.module: mc_fog_sample_paykit::client_builder, mc.src: fog/sample-paykit/src/client_builder.rs:138
2022-09-07 21:40:57.670487362 UTC INFO About to start LedgerServerConn to fog-ledger://fog.mc-masked-amount-v2.development.mobilecoin.com:443/, mc.app: test_client, mc.module: mc_fog_sample_paykit::client_builder, mc.src: fog/sample-paykit/src/client_builder.rs:138
2022-09-07 21:40:57.677419458 UTC INFO About to start LedgerServerConn to fog-ledger://fog.mc-masked-amount-v2.development.mobilecoin.com:443/, mc.app: test_client, mc.module: mc_fog_sample_paykit::client_builder, mc.src: fog/sample-paykit/src/client_builder.rs:138
2022-09-07 21:40:57.683596243 UTC INFO About to start LedgerServerConn to fog-ledger://fog.mc-masked-amount-v2.development.mobilecoin.com:443/, mc.app: test_client, mc.module: mc_fog_sample_paykit::client_builder, mc.src: fog/sample-paykit/src/client_builder.rs:138
2022-09-07 21:40:57.689797029 UTC INFO About to start LedgerServerConn to fog-ledger://fog.mc-masked-amount-v2.development.mobilecoin.com:443/, mc.app: test_client, mc.module: mc_fog_sample_paykit::client_builder, mc.src: fog/sample-paykit/src/client_builder.rs:138
2022-09-07 21:40:57.956351308 UTC WARN View key scanning failed, fog gave us a TXO that wasn't ours: Error decompressing FogTxOut: An invalid amount: The masked amount is missing, mc.app: test_client, mc.module: mc_fog_sample_paykit::cached_tx_data, mc.src: fog/sample-paykit/src/cached_tx_data/mod.rs:499
2022-09-07 21:40:57.956394809 UTC WARN View key scanning failed, fog gave us a TXO that wasn't ours: Error decompressing FogTxOut: An invalid amount: The masked amount is missing, mc.app: test_client, mc.module: mc_fog_sample_paykit::cached_tx_data, mc.src: fog/sample-paykit/src/cached_tx_data/mod.rs:499
2022-09-07 21:40:57.956401909 UTC WARN View key scanning failed, fog gave us a TXO that wasn't ours: Error decompressing FogTxOut: An invalid amount: The masked amount is missing, mc.app: test_client, mc.module: mc_fog_sample_paykit::cached_tx_data, mc.src: fog/sample-paykit/src/cached_tx_data/mod.rs:499
2022-09-07 21:40:57.956429309 UTC WARN View key scanning failed, fog gave us a TXO that wasn't ours: Error decompressing FogTxOut: An invalid amount: The masked amount is missing, mc.app: test_client, mc.module: mc_fog_sample_paykit::cached_tx_data, mc.src: fog/sample-paykit/src/cached_tx_data/mod.rs:499
2022-09-07 21:40:57.956434509 UTC WARN View key scanning failed, fog gave us a TXO that wasn't ours: Error decompressing FogTxOut: An invalid amount: The masked amount is missing, mc.app: test_client, mc.module: mc_fog_sample_paykit::cached_tx_data, mc.src: fog/sample-paykit/src/cached_tx_data/mod.rs:499
2022-09-07 21:40:57.956452110 UTC WARN View key scanning failed, fog gave us a TXO that wasn't ours: Error decompressing FogTxOut: An invalid amount: The masked amount is missing, mc.app: test_client, mc.module: mc_fog_sample_paykit::cached_tx_data, mc.src: fog/sample-paykit/src/cached_tx_data/mod.rs:499
2022-09-07 21:40:57.956456310 UTC WARN View key scanning failed, fog gave us a TXO that wasn't ours: Error decompressing FogTxOut: An invalid amount: The masked amount is missing, mc.app: test_client, mc.module: mc_fog_sample_paykit::cached_tx_data, mc.src: fog/sample-paykit/src/cached_tx_data/mod.rs:499
2022-09-07 21:40:57.956460710 UTC WARN View key scanning failed, fog gave us a TXO that wasn't ours: Error decompressing FogTxOut: An invalid amount: The masked amount is missing, mc.app: test_client, mc.module: mc_fog_sample_paykit::cached_tx_data, mc.src: fog/sample-paykit/src/cached_tx_data/mod.rs:499
2022-09-07 21:40:57.956464910 UTC WARN View key scanning failed, fog gave us a TXO that wasn't ours: Error decompressing FogTxOut: An invalid amount: The masked amount is missing, mc.app: test_client, mc.module: mc_fog_sample_paykit::cached_tx_data, mc.src: fog/sample-paykit/src/cached_tx_data/mod.rs:499
2022-09-07 21:40:57.956469810 UTC WARN View key scanning failed, fog gave us a TXO that wasn't ours: Error decompressing FogTxOut: An invalid amount: The masked amount is missing, mc.app: test_client, mc.module: mc_fog_sample_paykit::cached_tx_data, mc.src: fog/sample-paykit/src/cached_tx_data/mod.rs:499
2022-09-07 21:40:57.956473810 UTC WARN View key scanning failed, fog gave us a TXO that wasn't ours: Error decompressing FogTxOut: An invalid amount: The masked amount is missing, mc.app: test_client, mc.module: mc_fog_sample_paykit::cached_tx_data, mc.src: fog/sample-paykit/src/cached_tx_data/mod.rs:499
2022-09-07 21:40:57.956478110 UTC WARN View key scanning failed, fog gave us a TXO that wasn't ours: Error decompressing FogTxOut: An invalid amount: The masked amount is missing, mc.app: test_client, mc.module: mc_fog_sample_paykit::cached_tx_data, mc.src: fog/sample-paykit/src/cached_tx_data/mod.rs:499
2022-09-07 21:40:57.956482210 UTC WARN View key scanning failed, fog gave us a TXO that wasn't ours: Error decompressing FogTxOut: An invalid amount: The masked amount is missing, mc.app: test_client, mc.module: mc_fog_sample_paykit::cached_tx_data, mc.src: fog/sample-paykit/src/cached_tx_data/mod.rs:499
2022-09-07 21:40:57.956491110 UTC WARN View key scanning failed, fog gave us a TXO that wasn't ours: Error decompressing FogTxOut: An invalid amount: The masked amount is missing, mc.app: test_client, mc.module: mc_fog_sample_paykit::cached_tx_data, mc.src: fog/sample-paykit/src/cached_tx_data/mod.rs:499
2022-09-07 21:40:57.956497410 UTC WARN View key scanning failed, fog gave us a TXO that wasn't ours: Error decompressing FogTxOut: An invalid amount: The masked amount is missing, mc.app: test_client, mc.module: mc_fog_sample_paykit::cached_tx_data, mc.src: fog/sample-paykit/src/cached_tx_data/mod.rs:499
2022-09-07 21:40:57.956502710 UTC WARN View key scanning failed, fog gave us a TXO that wasn't ours: Error decompressing FogTxOut: An invalid amount: The masked amount is missing, mc.app: test_client, mc.module: mc_fog_sample_paykit::cached_tx_data, mc.src: fog/sample-paykit/src/cached_tx_data/mod.rs:499
2022-09-07 21:40:57.956507610 UTC WARN View key scanning failed, fog gave us a TXO that wasn't ours: Error decompressing FogTxOut: An invalid amount: The masked amount is missing, mc.app: test_client, mc.module: mc_fog_sample_paykit::cached_tx_data, mc.src: fog/sample-paykit/src/cached_tx_data/mod.rs:499
2022-09-07 21:40:57.956514010 UTC WARN View key scanning failed, fog gave us a TXO that wasn't ours: Error decompressing FogTxOut: An invalid amount: The masked amount is missing, mc.app: test_client, mc.module: mc_fog_sample_paykit::cached_tx_data, mc.src: fog/sample-paykit/src/cached_tx_data/mod.rs:499
2022-09-07 21:40:57.956519910 UTC WARN View key scanning failed, fog gave us a TXO that wasn't ours: Error decompressing FogTxOut: An invalid amount: The masked amount is missing, mc.app: test_client, mc.module: mc_fog_sample_paykit::cached_tx_data, mc.src: fog/sample-paykit/src/cached_tx_data/mod.rs:499
2022-09-07 21:40:57.956525111 UTC WARN View key scanning failed, fog gave us a TXO that wasn't ours: Error decompressing FogTxOut: An invalid amount: The masked amount is missing, mc.app: test_client, mc.module: mc_fog_sample_paykit::cached_tx_data, mc.src: fog/sample-paykit/src/cached_tx_data/mod.rs:499
2022-09-07 21:40:57.956529911 UTC WARN View key scanning failed, fog gave us a TXO that wasn't ours: Error decompressing FogTxOut: An invalid amount: The masked amount is missing, mc.app: test_client, mc.module: mc_fog_sample_paykit::cached_tx_data, mc.src: fog/sample-paykit/src/cached_tx_data/mod.rs:499
2022-09-07 21:40:57.956534311 UTC WARN View key scanning failed, fog gave us a TXO that wasn't ours: Error decompressing FogTxOut: An invalid amount: The masked amount is missing, mc.app: test_client, mc.module: mc_fog_sample_paykit::cached_tx_data, mc.src: fog/sample-paykit/src/cached_tx_data/mod.rs:499
2022-09-07 21:40:57.956556211 UTC WARN View key scanning failed, fog gave us a TXO that wasn't ours: Error decompressing FogTxOut: An invalid amount: The masked amount is missing, mc.app: test_client, mc.module: mc_fog_sample_paykit::cached_tx_data, mc.src: fog/sample-paykit/src/cached_tx_data/mod.rs:499
2022-09-07 21:40:57.956562611 UTC WARN View key scanning failed, fog gave us a TXO that wasn't ours: Error decompressing FogTxOut: An invalid amount: The masked amount is missing, mc.app: test_client, mc.module: mc_fog_sample_paykit::cached_tx_data, mc.src: fog/sample-paykit/src/cached_tx_data/mod.rs:499
2022-09-07 21:40:57.956567511 UTC WARN View key scanning failed, fog gave us a TXO that wasn't ours: Error decompressing FogTxOut: An invalid amount: The masked amount is missing, mc.app: test_client, mc.module: mc_fog_sample_paykit::cached_tx_data, mc.src: fog/sample-paykit/src/cached_tx_data/mod.rs:499
2022-09-07 21:40:57.956571811 UTC WARN View key scanning failed, fog gave us a TXO that wasn't ours: Error decompressing FogTxOut: An invalid amount: The masked amount is missing, mc.app: test_client, mc.module: mc_fog_sample_paykit::cached_tx_data, mc.src: fog/sample-paykit/src/cached_tx_data/mod.rs:499
2022-09-07 21:40:57.956576711 UTC WARN View key scanning failed, fog gave us a TXO that wasn't ours: Error decompressing FogTxOut: An invalid amount: The masked amount is missing, mc.app: test_client, mc.module: mc_fog_sample_paykit::cached_tx_data, mc.src: fog/sample-paykit/src/cached_tx_data/mod.rs:499
2022-09-07 21:40:57.956582011 UTC WARN View key scanning failed, fog gave us a TXO that wasn't ours: Error decompressing FogTxOut: An invalid amount: The masked amount is missing, mc.app: test_client, mc.module: mc_fog_sample_paykit::cached_tx_data, mc.src: fog/sample-paykit/src/cached_tx_data/mod.rs:499
2022-09-07 21:40:57.956588211 UTC WARN View key scanning failed, fog gave us a TXO that wasn't ours: Error decompressing FogTxOut: An invalid amount: The masked amount is missing, mc.app: test_client, mc.module: mc_fog_sample_paykit::cached_tx_data, mc.src: fog/sample-paykit/src/cached_tx_data/mod.rs:499
2022-09-07 21:40:57.956601612 UTC WARN View key scanning failed, fog gave us a TXO that wasn't ours: Error decompressing FogTxOut: An invalid amount: The masked amount is missing, mc.app: test_client, mc.module: mc_fog_sample_paykit::cached_tx_data, mc.src: fog/sample-paykit/src/cached_tx_data/mod.rs:499
2022-09-07 21:40:57.956608612 UTC WARN View key scanning failed, fog gave us a TXO that wasn't ours: Error decompressing FogTxOut: An invalid amount: The masked amount is missing, mc.app: test_client, mc.module: mc_fog_sample_paykit::cached_tx_data, mc.src: fog/sample-paykit/src/cached_tx_data/mod.rs:499
2022-09-07 21:40:57.956615212 UTC WARN View key scanning failed, fog gave us a TXO that wasn't ours: Error decompressing FogTxOut: An invalid amount: The masked amount is missing, mc.app: test_client, mc.module: mc_fog_sample_paykit::cached_tx_data, mc.src: fog/sample-paykit/src/cached_tx_data/mod.rs:499
2022-09-07 21:40:57.956622212 UTC WARN View key scanning failed, fog gave us a TXO that wasn't ours: Error decompressing FogTxOut: An invalid amount: The masked amount is missing, mc.app: test_client, mc.module: mc_fog_sample_paykit::cached_tx_data, mc.src: fog/sample-paykit/src/cached_tx_data/mod.rs:499
2022-09-07 21:40:57.956629512 UTC WARN View key scanning failed, fog gave us a TXO that wasn't ours: Error decompressing FogTxOut: An invalid amount: The masked amount is missing, mc.app: test_client, mc.module: mc_fog_sample_paykit::cached_tx_data, mc.src: fog/sample-paykit/src/cached_tx_data/mod.rs:499
2022-09-07 21:40:57.956633812 UTC WARN View key scanning failed, fog gave us a TXO that wasn't ours: Error decompressing FogTxOut: An invalid amount: The masked amount is missing, mc.app: test_client, mc.module: mc_fog_sample_paykit::cached_tx_data, mc.src: fog/sample-paykit/src/cached_tx_data/mod.rs:499
```

investigating

garbageslam:
Deployment almost worked but failed in a wierd way at the last step:

```

-- execute command:
   /test/minting-tx-test.sh \
  --key-dir /tmp/3-testing/keys \
  --token-id 1


-- sending mint tx for account key /tmp/3-testing/keys/account_keys_496.b58pub
response: result {code: NoMatchingMintConfig} block_count: 419 block_version: 3
command terminated with exit code 12

```

Going to try re-running it

garbageslam:
looks like the mint config issue went away on re-run, will investigate later

PR #2100, opened 2022-06-07
Add a new "cmd" argument to mob prompt

DianaNites:
### Motivation

Like #2027, this change allows external tools, such as rust-analyzer, to run `cargo check` inside the docker environment, improving the development experience for people who can't install the SGX SDK locally.

This approach is more flexible than that in #2027, and no longer hard-codes any specific command.

I believe this approach is reliable under argparse's non-standard argument parsing, but if that proves to not be the case another option could be to take the command line as a single string, potentially requiring the user to escape it.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -26,6 +26,39 @@ If you have SGX-enabled hardware (activated in BIOS, and with SGX kernel module 
 you can use `./mob prompt --hw` to get SGX in the container. Then you can both build and
 run the tests in `SGX_MODE=HW`. (See below for an explanation.)
 
+##### IDE Support using the docker environment
+
+You can use `./mob prompt --cmd <command>` to run `<command>` from within
+the docker environment. For IDE support, this can be used to run `cargo check`.
+
+An example workspace configuration for Rust Analyzer is provided below
+
+```json
+"rust-analyzer.checkOnSave.overrideCommand": [
+    "./mob",
+    "prompt",
+    "--cmd",
+    "cargo",
+    "check",
+    "--workspace",
+    "--message-format=json",
+    "--all-targets"
+],
+"rust-analyzer.cargo.buildScripts.overrideCommand": [
+    "./mob",
+    "prompt",
+    "--cmd",
+    "cargo",
+    "check",
+    "--workspace",
+    "--message-format=json",
+    "--all-targets"
+],
+"rust-analyzer.cargo.buildScripts.useRustcWrapper": false,
+"rust-analyzer.cargo.buildScripts.enable": false,
+"rust-analyzer.procMacro.enable": false,
+```
+
 #### No-docker build
 
 A docker-less build also works fine for development:
--- a/BUILD.md
+++ b/BUILD.md
@@ -31,7 +31,7 @@ run the tests in `SGX_MODE=HW`. (See below for an explanation.)
 You can use `./mob prompt --cmd <command>` to run `<command>` from within
 the docker environment. For IDE support, this can be used to run `cargo check`.
 
-An example workspace configuration for Rust Analyzer is provided below
+An example workspace configuration for Rust Analyzer is provided below:
 
 ```json
 "rust-analyzer.checkOnSave.overrideCommand": [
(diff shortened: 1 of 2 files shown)


garbageslam (approved):
this looks great, thank you!

BUILD.md
```diff
@@ -26,6 +26,39 @@ If you have SGX-enabled hardware (activated in BIOS, and with SGX kernel module
 you can use `./mob prompt --hw` to get SGX in the container. Then you can both build and
 run the tests in `SGX_MODE=HW`. (See below for an explanation.)
 
+##### IDE Support using the docker environment
+
+You can use `./mob prompt --cmd <command>` to run `<command>` from within
```
remoun, line 31: How about we make a `run` action instead? e.g. `mob run cargo check ...`

DianaNites: I wanted to, but due to limitations with argparse, I couldn't get it to work reliably. Depending on how it was called, in intuitive ways, it would conflict with the `container_name` positional argument, causing it to think `cargo` is the container name, for example, and then fail. Thats why I went with an argument instead.

It might be possible to work around this by using another library, or possibly more argparse features like subparsers/subcommands, but that would be a bigger change.

DianaNites: I wanted to, but due to limitations with argparse, I couldn't get it to work reliably. Depending on how it was called, in intuitive ways, it would conflict with the `container_name` positional argument, causing it to think `cargo` is the container name, for example, and then fail. Thats why I went with an argument instead.

It might be possible to work around this by using another library, or possibly more argparse features like subparsers/subcommands, but that would be a bigger change.

DianaNites: For example trying to use `parser.add_argument("cmd", nargs=argparse.REMAINDER)` and calling it as so `./mob prompt cargo check` results in the following arguments. It also does not work with `nargs="*"`

```python
Namespace(action='prompt', container_name='cargo', hw=False, ias_prod=False, dry_run=False, tag=None, no_pull=False, verbose=False, ssh_dir=False, ssh_agent=False, expose=None, publish=None, cmd=['check'])
```

Argparse just doesnt seem to properly support the bash/linux/posix? convention of `--`([see here](https://unix.stackexchange.com/questions/11376/what-does-double-dash-mean)), so it can't be used to signify the end of options in that manner.

remoun: I see. Would you mind trying the `run` action and removing `nargs` param for the `add_argument("container_name", ...)` call (line 95)? `nargs='?', default=None` doesn't really make sense.

remoun: On further thought, I guess that won't keep the `container_name` arg from slurping the first word. It seems more usable to have the Python code treat `args.container_name` as the first part of the command in `run` mode, but I won't block approval on that.

mob
```diff
@@ -440,12 +441,19 @@ docker_run.extend([tag])
...
+
 try:
     maybe_run(docker_run)
 except subprocess.CalledProcessError as exception:
     if args.action == 'prompt' and exception.returncode == 130:
         # This is normal exit of prompt
         sys.exit(0)
+    if args.action == 'prompt' and args.cmd is not None:
+        # Make sure custom commands have their expected return codes
+        sys.exit(exception.returncode)
```
remoun, line 456: IIUC we should check for `cmd` first, so we don't swallow an exit code of 130 as normal.

```suggestion
    if args.action == 'prompt' and args.cmd is not None:
        # Make sure custom commands have their expected return codes
        sys.exit(exception.returncode)
    if args.action == 'prompt' and exception.returncode == 130:
        # This is a normal exit of prompt
        sys.exit(0)
```

remoun (approved):

BUILD.md
DianaNites: That would mean its impossible to provide both arguments though. I'm not sure if it even makes sense to want to `run` with another `container_name`, though.

iamalwaysuncomfortable (approved):

DianaNites:
Checks for this PR seem to be a broken, preventing merging?

remoun:
> Checks for this PR seem to be a broken, preventing merging? Can't tell, nothing seems to be running or in the checks tab?

Yeah, CircleCI does that sometimes, and we don't know why. I don't see the test run in the GH nor CircleCI UI, so I guess your best bet to restart the tests is to push a commit.

DianaNites:
just noticing the spurious changed file in the commit and no idea why, it seems to keep showing up sometimes when i `cargo build`?

DianaNites:
should be fixed now

Issue #1842, opened 2022-04-22
We should make a test utils crate that replaces localized tempfile references


wjuan-mob:
We should make a test utils crate that uses `TempDir::new_in()` to put all these into the `OUT_DIR`, and replace all our localized `tempfile` users.

_Originally posted by @jcape in https://github.com/mobilecoinfoundation/mobilecoin/pull/1689#discussion_r856378857_

Issue #1467, opened 2022-02-09
De-duplicate test-utils


garbageslam:
Numerous core crates come also with a "test-utils" crate (although sometimes the `test-utils` come in a submodule).

`mc-transaction-core-test-utils`
`mc-ledger-db::test_utils`
`mobilecoind-test-utils`

Some of these test-utils crates have a lot of very similar (possibly the same?) functionality:

In `mc-transaction-core-test-utils`:
* `initialize_ledger` takes a ledger db object and adds a bunch of test blocks to it
* `create_transaction` creates a test transaction
* `get_blocks` creates a list of test blocks

In `mc-transaction-std`, we have separate, slightly different versions of `create_transaction` in the tests module.

In `mc-ledger-db-test-utils`, `get_test_ledger_blocks` is almost the same `transaction-core-test-utils::get_blocks`.

In `ledger/sync/src/test_app/main.rs` we have functions like `_make_ledger_long` which takes a ledger and adds a bunch of blocks to it. This is almost exactly the same as `mc_transaction_core_test_utils::initialize_ledger`.

In `mobilecoind::test_utils`, `add_block_to_ledger_db` is almost the same as `mc-transaction-core-test-utils::initialize_ledger` and the `_make_ledger_long`.

We could likely have 2 or 3 functions instead of 10. This would make it much easier to read and maintain this test code, if we just had one flexible "add blocks to ledger" function instead of 3 or 4 variations on a theme that no one can remember the differences between. It would also make it much easier in the future when we have to make changes to `TxOut` or `TransactionBuilder`, which will happen whenever we want to ship new features into the blockchain.

remoun:
I did some of this in #2052, which I hope to merge next week.

PR #2108, opened 2022-06-08
Replace plain tempfile uses

DianaNites:
Create and use a test-utils crate

### Motivation

As part of #1842, make a new test-utils crate and replace global tmp directory usage with `OUT_DIR`

### Future Work

Move/unify/deduplicate more testing utility code as part of #1467.


eranrund:
Any reason not to put this stuff inside the `mc-util-test-helper` crate (or maybe merge that one into the new test-utils crate)?

DianaNites:
Not that I know of, one should be merged into the other at some point, probably along with the various other `test-` crates in `util`, and the various `test-utils` submodules/subcrates in our other crates, but that'd end up as a much bigger PR.

eranrund:
> Not that I know of, one should be merged into the other at some point, probably along with the various other `test-` crates in `util`, and the various `test-utils` submodules/subcrates in our other crates, but that'd end up as a much bigger PR.

A lot of the`test-utils` subcrates/submodules will likely need to stay where they are, since they were created to work around circular dependency issues.
For example, we can't merge `mc-transaction-core-test-utils` into the `test-helper` crate, since that would mean `test-helper` would need to depend on `mc-transaction-core`. However, the tests in `mc-transaction-core` want to use stuff from the `test-helper` crate.

DianaNites:
Those issues shouldn't apply anymore since cargo has improved since then, me and james talked about this a bit.

Ever since edition2021, which all our crates use today, resolver=2 is the default and should prevent the mixing of dependencies and features that used to cause problems 2+ years ago.

garbageslam:
Color me skeptical -- the resolver=2 just fixes things about cargo fixing regular dependencies and build dependencies.

I don't think resolver=2 allows circular dependencies between crates to just magically work.

I think a lot of our test-utils crates should actually be modules in the crates they are trying to test, and they can be conditional on test flags. But transaction core test-utils is harder because it actually depends on ledger db and a bunch of other stuff that transaction core cannot depend on itself

DianaNites:
That could be the case, that'd be unfortunate. Problem for the other issue though, guess i need to figure out why in the world checks failed here

remoun:
> I think a lot of our test-utils crates should actually be modules in the crates they are trying to test, and they can be conditional on test flags. But transaction core test-utils is harder because it actually depends on ledger db and a bunch of other stuff that transaction core cannot depend on itself

FYI I opted to split up `mc_transaction_core_test_utils` to remove the `mc_ledger_db` dep in #2052, which brings us closer to moving it to `mc_transaction_core::test_utils`

util/test-utils/src/lib.rs
```diff
@@ -0,0 +1,14 @@
...
+use tempfile::{NamedTempFile, TempDir};
+
+/// Create a temporary directory in the directory specified by the
+/// cargo-provided `OUT_DIR` environment variable.
+///
+/// # Panics
+///
+/// - If `OUT_DIR` doesn't exist
+/// - If [`TempDir::new_in`] fails to create the directory.
+pub fn tempdir() -> TempDir {
```
remoun, line 11: I think we'll want a variant that takes a prefix, e.g. `tempdir_with_prefix()`. It can use [Builder::prefix()](https://docs.rs/tempfile/latest/tempfile/struct.Builder.html#method.prefix)

util/grpc/src/server_cert_reloader.rs
```diff
@@ -146,6 +146,7 @@ mod tests {
     use grpcio::{ChannelBuilder, ChannelCredentialsBuilder, EnvBuilder, Server, ServerBuilder};
     use mc_common::logger::test_with_logger;
     use mc_crypto_x509_test_vectors::{ok_self_signed_1, ok_self_signed_2};
+    use mc_util_test_utils::tempdir;
```
remoun, line 149: +1 for moving into `mc_util_test_helper`

util/keyfile/Cargo.toml
```diff
@@ -48,5 +48,5 @@ x509-signature = "0.5"
 [dev-dependencies]
 mc-crypto-x509-test-vectors = { path = "../../crypto/x509/test-vectors" }
 mc-util-test-helper = { path = "../../util/test-helper" }
-
 tempfile = "3.2"
```
remoun, line 51: IIUC we can now remove this.
```suggestion
```

remoun (changes requested):
It would be nice if we can also replace `tempdir` with these helpers: https://github.com/mobilecoinfoundation/mobilecoin/search?l=TOML&q=tempdir&type=code

There's also a couple of other crates that use `tempfile`: https://github.com/mobilecoinfoundation/mobilecoin/search?l=TOML&q=tempfile&type=code

util/keyfile/Cargo.toml
DianaNites: The bin-determinism test still depends on `tempfile::NamedTempFile`, so that needs to be replaced first

DianaNites:
The tests keep failing because the `OUT_DIR` variable isnt set during them. Cargo only seems to set `OUT_DIR` when a `build.rs` exists, so I don't think we can actually do this, or at the very least we need a fallback.

util/test-utils/src/lib.rs
```diff
@@ -0,0 +1,14 @@
...
+    let out = std::env::var("OUT_DIR").expect("Missing environment variable OUT_DIR");
+    TempDir::new_in(&out).expect(&format!("Could not create temporary directory {}", &out))
```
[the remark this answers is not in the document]
DianaNites: Why should `expect` be replaced? This seems to replicate what expect already does https://doc.rust-lang.org/stable/std/result/enum.Result.html#method.expect

Is this to avoid the unconditional `format!` call?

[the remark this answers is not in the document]
remoun: Yes, exactly. There's a Clippy check that requires this. If it doesn't trigger for this change, I'm fine with the `expect` call as written.

PR #2111, opened 2022-06-09
Bump version to 1.2.2

jcape:
- Bump all versions to 1.2.2
- Update CHANGELOG.md

### Motivation

Last step before tagging 1.2.2.


nick-mobilecoin (approved):
I think this needs to get bumped to `1.2.2` as well
```
android-bindings/lib-wrapper/android-bindings/publish.gradle
3:version '1.2.1'
```

remoun (approved):
While reviewing this I realized a couple of crates have incorrect versions, which I'm updating in #2122. We might want a similar change here.

Issue #2028, opened 2022-05-23
Fog View Router


jcape:
- [x] #2097
- [x] #2254
- [x] #2253 
- [x] #2233
- [x] #2256
- [x] #2255
- [x] #2435
- [x] #2395
- [x] #2836
- [x] #2837
- [x] #2930
- [x] #2838
- [x] #3177
- [x] It was determined that #3178 is not needed for the initial release

nick-mobilecoin:
Done in #3297

PR #2115, opened 2022-06-09
Implement create_multi_view_query_data API for Fog View Router

samdealy:
### Motivation

This is the second PR for the Fog View Router project. As described in the design [doc](https://docs.google.com/document/d/1-ju5efGK6vvuYjxj1pKhfFcpgGlornuxZYvOC-nsbjs/edit), the router receives a client query request, decrypts it, and then re-encrypts it for each Fog View Store that it knows about. This PR implements the view enclave API that performs this decryption and re-encryption step. In the next PR, I'll implement the logic that takes the output of this API and creates a MultiViewStoreQuery (see the design doc for the definition of this message)

Addresses the second item in #2028 

### Future Work
- Use the output of the API to create the MultiViewStoreQuery and send it to the Fog View Load Balancers

### DIFF
--- a/fog/view/enclave/api/src/lib.rs
+++ b/fog/view/enclave/api/src/lib.rs
@@ -83,6 +83,9 @@ pub enum ViewEnclaveRequest {
     Query(EnclaveMessage<ClientSession>, UntrustedQueryResponse),
     /// Request from untrusted to add encrypted tx out records to ORAM
     AddRecords(Vec<ETxOutRecord>),
+    /// Takes an encrypted fog_types::view::QueryRequest and returns a list of
+    /// fog_types::view::QueryRequest.
+    CreateMultiViewStoreQuery(EnclaveMessage<ClientSession>),
 }
 
 /// The parameters needed to initialize the view enclave
@@ -134,6 +137,15 @@ pub trait ViewEnclaveApi: ReportableEnclave {
     /// Add encrypted tx out records from the fog recovery db to the view
     /// enclave's ORAM
     fn add_records(&self, records: Vec<ETxOutRecord>) -> Result<()>;
+
+    /// Transforms a client query request into a list of query request data.
+    ///
+    /// The returned list is meant to be used to construct the
+    /// MultiViewStoreQuery, which is sent to each shard.
+    fn create_multi_view_store_query_data(
+        &self,
+        client_query: EnclaveMessage<ClientSession>,
+    ) -> Result<Vec<EnclaveMessage<ClientSession>>>;
 }
 
 /// Helper trait which reduces boiler-plate in untrusted side
@@ -187,6 +199,8 @@ pub enum Error {
     Poison,
     /// Enclave not initialized
     EnclaveNotInitialized,
+    /// Cipher encryption failed
+    Cipher,
 }
 
 impl From<SgxError> for Error {
@@ -240,5 +254,11 @@ impl From<AttestEnclaveError> for Error {
 impl From<AddRecordsError> for Error {
     fn from(src: AddRecordsError) -> Self {
         Error::AddRecords(src)
+    }
+}
+
+impl From<mc_crypto_noise::CipherError> for Error {
+    fn from(_: mc_crypto_noise::CipherError) -> Self {
+        Error::Cipher
     }
 }
--- a/fog/view/enclave/api/src/lib.rs
+++ b/fog/view/enclave/api/src/lib.rs
@@ -18,6 +18,7 @@ use mc_attest_enclave_api::{
 };
 use mc_common::ResponderId;
 use mc_crypto_keys::X25519Public;
+use mc_crypto_noise::CipherError;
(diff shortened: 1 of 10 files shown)


fog/view/enclave/impl/src/lib.rs
```diff
@@ -182,6 +193,29 @@ where
...
+    /// Takes in a client's query request and returns a list of query requests
+    /// to be sent off to each Fog View Store shard.
+    fn create_multi_view_store_query_data(
+        &self,
+        client_query: EnclaveMessage<ClientSession>,
+    ) -> Result<Vec<EnclaveMessage<ClientSession>>> {
+        let client_query_bytes = self.ake.client_decrypt(client_query.clone())?;
+
+        let mut encryptors = self.store_encryptors.lock()?;
+        let mut results = Vec::new();
```
remoun, line 209: Let's reserve the expected size
```suggestion
        let mut results = Vec::with_capactiy(encryptors.len());
```

fog/view/enclave/api/src/lib.rs
```diff
@@ -187,6 +199,8 @@ pub enum Error {
     Poison,
     /// Enclave not initialized
     EnclaveNotInitialized,
+    /// Cipher encryption failed
+    Cipher,
```
remoun, line 203: It seems useful to include the actual `CipherError`

```suggestion
    /// Cipher encryption failed: {0}
    Cipher(CipherError),
```
(also `use mc_crypto_noise::CipherError`)

fog/view/enclave/api/src/lib.rs
```diff
@@ -242,3 +256,9 @@ impl From<AddRecordsError> for Error {
         Error::AddRecords(src)
     }
 }
+
+impl From<mc_crypto_noise::CipherError> for Error {
+    fn from(_: mc_crypto_noise::CipherError) -> Self {
+        Error::Cipher
```
remoun, line 262:
```suggestion
    fn from(src: CipherError) -> Self {
        Error::Cipher(src)
```

fog/view/enclave/impl/src/lib.rs
```diff
@@ -182,6 +193,29 @@ where
...
+        for store_encryptor in encryptors.deref_mut() {
+            let data = store_encryptor.encrypt(&client_query.aad, &client_query_bytes)?;
+            results.push(EnclaveMessage {
+                aad: client_query.clone().aad,
+                channel_id: client_query.clone().channel_id,
+                data,
+            });
```
remoun, line 216: Let's clone the specific fields we want, rather than cloning the query twice.
```suggestion
            let aad = client_query.aad.clone();
            let data = store_encryptor.encrypt(&aad, &client_query_bytes)?;
            let channel_id = client_query.channel_id.clone();
            results.push(EnclaveMessage { aad, channel_id, data });
```

fog/view/enclave/impl/src/lib.rs
samdealy: Done.

fog/view/enclave/api/src/lib.rs
samdealy: Done.

fog/view/enclave/impl/src/lib.rs
samdealy: Done.

fog/view/enclave/api/src/lib.rs
samdealy: Done.

remoun (approved):

fog/view/enclave/impl/src/lib.rs
```diff
@@ -39,6 +42,13 @@ where
 
     /// Logger object
     logger: Logger,
+
+    /// Encrypts a QueryRequest for each individual Fog View Store.
+    /// TODO: Use a BTreeMap<FogViewShardLoadBalancerID,
+    /// BTreeMap<FogViewStoreId, Ready<...>>>  when implement the cursoring
+    /// optimization. For right now, it's fine to leave as a Vec because a
+    /// follow up PR will implement cursoring.
+    store_encryptors: Mutex<Vec<Ready<Aes256Gcm>>>,
```
iamalwaysuncomfortable, line 51: Would this need a Mutex? These carry the the ciphers, but from my read - I don't think the underlying data behind the cipher config changes much and there's not really a reason to lock it?

iamalwaysuncomfortable: Also, would different FOG View Stores be using different AES implementations?

samdealy: This needs to be a Mutex because the Cipher's `encrypt` method is mutable, and we can't get a mutable reference to `self` because the enclave can't provide this. Therefore we have to use a`Mutex` to enable interior mutability, which allows us to mutate a struct without having a mutable reference.

iamalwaysuncomfortable (approved):

PR #2116, opened 2022-06-09
Feature/v1.2.2 ledger from archive

jgreat:
### Motivation

- Add ledger-from-archive logic to restore ledger when there isn't a local copy. Needed for persistent environments.
- Make standalone node_hw container operations consistent with helm charts.
- Enable standalone operators to configure node via ENV var or files (.env import in entrypoint script).
- Make helm and container vars consent with applications (minimize remapping of values).
- Re-enable filebeat log shipping for standalone nodes.
- Include ledger restore, migration and bootstrapping operations as container startup.

### Changes

- Move supervisord configs from chart/configMap injection into the node_hw container.
- Add wrapper scripts to ledger-distribution and consensus-service to enforce start up order and keep from overwriting existing data.
- Add entrypoint script to handle ledger bootstrapping operations.
- Update charts to use application env vars in the same way as standalone container operations.
- Generate and default to json configs (tokens, network) for easier tooling and less string templating.


joekottke (approved):
After discussion, this looks good.  Lots of config changes, so keep eyes peeled for typos, etc.  Can do a quick PR turn-around as necessary.

joekottke:
One more thought: When `ledger-from-archive` does a full successful download, its last attempt is to try to download the BLOCK_INDEX + 1 block, and will correctly fail to download that block.  I'm not sure if it will then `exit 0` , or `exit 1`.  Shouldn't really matter, but it would mean that an `initContainer` failed, and the pod might be rescheduled.  Which at that point the `test -f ledger.mdb` would succeed, and it would then `exit 0`

.internal-ci/helm/consensus-node-config/values.yaml
```diff
@@ -13,6 +13,9 @@ node:
   peer:
     hostname: ''
 
+  ### S3 bucket for thins nodes ledger
```
remoun, line 16: Nit: typo
```suggestion
  ### S3 bucket for this node's ledger
```

.internal-ci/helm/consensus-node/values.yaml
```diff
@@ -102,35 +108,34 @@ node:
...
-    image: '{{ .Values.node.image.org | default .Values.image.org }}/{{ .Values.node.image.name }}:{{ .Values.image.tag | default .Chart.AppVersion }}'
-    imagePullPolicy: Always
-    command: [ '/bin/bash' ]
-    args:
-    - -c
-    - |
-      set -e
-      if [[ -f "/ledger/data.mdb" ]]
-      then
+        echo "Ledger /ledger/data.mdb already exists migrate to latest version"
```
remoun, line 138:
```suggestion
        echo "Ledger /ledger/data.mdb already exists, migrating to latest version"
```

.internal-ci/helm/consensus-node/values.yaml
```diff
@@ -102,35 +108,34 @@ node:
...
           cp /tmp/sample_data/ledger/data.mdb /ledger
-        elif [ -f /var/lib/mobilecoin/origin_data/data.mdb ]; then
+          exit 0
+        fi
+
+        echo "Get ledger from S3 - ledger-from-archive"
+        # Check to see if archive exists
+        curl -I -f "${TX_SOURCE_URL}/00/00/00/00/00/00/00/0000000000000000.pb"
+        /usr/bin/ledger-from-archive --ledger-db /ledger --tx-source-url "${TX_SOURCE_URL}"
+        exit 0
```
remoun, line 126: This unconditional exit doesn't look right.
```suggestion
        if curl -I -f "${TX_SOURCE_URL}/00/00/00/00/00/00/00/0000000000000000.pb"
        then
          /usr/bin/ledger-from-archive --ledger-db /ledger --tx-source-url "${TX_SOURCE_URL}"
          exit 0
        fi
```

.internal-ci/helm/mc-core-dev-env-setup/values.yaml
```diff
@@ -110,6 +114,7 @@ consensusNodeConfig5:
       hostname: '{{ printf "node5.%s.development.mobilecoin.com" .Release.Namespace }}'
     peer:
       hostname: '{{ printf "peer5.%s.development.mobilecoin.com" .Release.Namespace }}'
+    txSourceUrl: '{{ with .Values.global.node.ledgerDistribution }}{{ printf "https://s3-%s.amazonaws.com/%s/%s/" .awsRegion .s3Bucket (printf "node5.%s.development.mobilecoin.com" $.Release.Namespace) }}{{ end }}'
```
remoun, line 117: This pattern seems like a good candidate for a helper.

.internal-ci/helm/consensus-node-config/templates/node-ledger-distribution-secret.yaml
```diff
@@ -12,6 +12,6 @@ stringData:
   AWS_SECRET_ACCESS_KEY: {{ .awsSecretAccessKey | quote }}
   AWS_REGION: {{ .awsRegion | quote }}
   LEDGER_DISTRIBUTION_S3_BUCKET: {{ .s3Bucket | quote }}
-  LEDGER_DISTRIBUTION_START_FROM: {{ .startFrom | quote }}
-  LEDGER_DISTRIBUTION_S3_PATH: {{ tpl .awsPath $ | quote }}
+  MC_START_FROM: {{ .startFrom | quote }}
```
jgreat, line 15: pull this out so we don't get confused, the ledger-distribution wrapper sets this

joekottke (changes requested):
Had a walk-through with @jgreat , and it all looks good.  He knows what he needs to fix, will approve when additional commits come in

.internal-ci/docker/support/node_hw/supervisor/conf.d/ledger-distribution.conf
```diff
@@ -0,0 +1,15 @@
+; Copyright (c) 2018-2022 The MobileCoin Foundation
+[program:ledger-distribution]
+priority=50
+; if we don't start in 20 seconds go to fatal
+startsecs=20
+command=/usr/local/bin/wrapper-ledger-distribution.sh 2>&1
```
remoun, line 6: Do we need `2>&1` here, since we're configuring stderr logging below?

.internal-ci/docker/support/node_hw/bin/wrapper-consensus-service.sh
```diff
@@ -0,0 +1,82 @@
...
+        ((aesm++))
+    else
+        echo "mc.app:wrapper-consensus-service aesm-service is not RUNNING or ${aesm_socket} not found - reset counter"
+        aesm=0
+    fi
+
+    # is ledger-distribution running?
+    if [[ "${ledger_distribution_status}" == "RUNNING" ]]
+    then
+        echo "mc.app:wrapper-consensus-service ledger-distribution is RUNNING - success ${ledger}"
```
remoun, line 74: Optional/Equivalent: Add a helper function for logging this `mc.app:wrapper-consensus-service ` prefix.

.internal-ci/docker/support/node_hw/bin/wrapper-consensus-service.sh
```diff
@@ -0,0 +1,82 @@
...
+        ((ledger++))
+    else
+        echo "mc.app:wrapper-consensus-service ledger-distribution is not RUNNING - reset counter"
```
remoun, line 77: IIUC we should reset `ledger=0` here
```suggestion
        echo "mc.app:wrapper-consensus-service ledger-distribution is not RUNNING - reset counter"
        ledger=0
```

.internal-ci/docker/support/node_hw/bin/wrapper-consensus-service.sh
```diff
@@ -0,0 +1,82 @@
...
+# Optional vars - sentry/logging
+# CONSENSUS_SERVICE_SENTRY_DSN
+# MC_LOG_UDP_JSON
+
+# Check to see if aesm service is ready.
+aesm_socket=/var/run/aesmd/aesm.socket
+aesm=0
+ledger=0
+echo "mc.app:wrapper-consensus-service Wait for ledger-distribution and aesm service to become ready"
+while [[ "${aesm}" -le 1 ]] && [[ "${ledger}" -le 1 ]]
```
remoun, line 54: IIUC we want to check for at least one of each, so this should use `-lt` and `||`; we can also use `||` inside `[[`.
```suggestion
while [[ "${aesm}" -lt 1 || "${ledger}" -lt 1 ]]
```

.internal-ci/docker/support/node_hw/bin/wrapper-ledger-distribution.sh
```diff
@@ -0,0 +1,41 @@
...
+# Default vars
+export MC_LEDGER_PATH=${MC_LEDGER_PATH:-"/ledger"}
+export MC_STATE_FILE=${MC_STATE_FILE:-"/ledger/.distribution-state"}
+export MC_SENTRY_DSN=${LEDGER_DISTRIBUTION_SENTRY_DSN}
+
+if [[ -f "/ledger/.distribution-state" ]]
+then
+    export MC_START_FROM=last
+else
+    export MC_START_FROM=next
```
remoun, line 38: How about we propagate existing values here?
```suggestion
    export MC_START_FROM=${MC_START_FROM:-next}
```

remoun (commented):
LGTM overall, with some minor comments

.internal-ci/helm/mc-core-dev-env-setup/values.yaml
jgreat: That chart is basically one big helper.   The template resolution is done in the child charts.

.internal-ci/docker/support/node_hw/bin/wrapper-consensus-service.sh
jgreat: Yeah, that's on the list for future revisions.   We need to make all logging json so we can properly parse it.

.internal-ci/docker/support/node_hw/bin/wrapper-ledger-distribution.sh
jgreat: We do not want an existing value.  This start-from needs additional logic or we can end up overwriting blocks in s3.  This is a work around until we get to fixing the logic in ledger-distribution.

.internal-ci/docker/support/node_hw/supervisor/conf.d/ledger-distribution.conf
jgreat: fixed.

.internal-ci/docker/support/node_hw/bin/wrapper-consensus-service.sh
jgreat: fixed

.internal-ci/docker/support/node_hw/bin/wrapper-consensus-service.sh
jgreat: yeah caught that, logic is fixed

joekottke (approved):
LGTM

remoun (approved):

PR #2120, opened 2022-06-10
Add release versioning api to mobilecoind andmobilecoind json

### LINKED ISSUE #682: Add an API to get the release version of mobilecoind & mobilecoind-json
Didn't find any version info in 
https://github.com/mobilecoinfoundation/mobilecoin/blob/master/mobilecoind-json/src/bin/main.rs#L698

pouneh:
### Motivation

Allow callers to get what version of the MobileCoin Daemon is running via API

Resolves #682


mobilecoind/api/proto/mobilecoind_api.proto
```diff
@@ -844,3 +847,9 @@ message SetDbPasswordRequest {
 message UnlockDbRequest {
     bytes password = 1;
 }
+
+// Get the mobilecoind version
+// empty request
+message MobilecoindVersionResponse {
+    string version = 1;
+}
```
remoun, line 855: Nit: Add newline
```suggestion
}

```

mobilecoind-json/src/bin/main.rs
```diff
@@ -79,6 +79,16 @@ fn unlock_db(
     Ok(Json(JsonUnlockDbResponse { success: true }))
 }
 
+/// Gets current mobilecoindversion
```
remoun, line 82: Nit: Add space
```suggestion
/// Gets current mobilecoind version
```

mobilecoind-json/src/data_types.rs
```diff
@@ -1547,3 +1547,16 @@ mod test {
...
+
+#[derive(Serialize, Default, Debug)]
+pub struct JsonMobilecoindVersionResponse {
+    pub version: String,
+}
+
+impl From<&mc_mobilecoind_api::MobilecoindVersionResponse> for JsonMobilecoindVersionResponse {
+    fn from(src: &mc_mobilecoind_api::MobilecoindVersionResponse) -> Self {
+        Self {
+            version: src.get_version().to_string(),
```
remoun, line 1559: IIUC we can just reference the proto field
```suggestion
            version: src.version,
```

mobilecoind/src/service.rs
```diff
@@ -5977,4 +5988,29 @@ mod test {
...
+                &mut rng,
+            );
+
+        let expected_version = String("1.3.0-pre0");
+
+        // Send request.
+        let response = client
+            .get_release_version()
+            .expect("Incorrect version returned");
+        assert_eq!(expected_version, response.get_version());
```
remoun, line 6014: It seems more robust to not hardcode a version.

```suggestion
        // Send request.
        let response = client
            .get_release_version()
            .expect("Incorrect version returned");
        assert!(!response.version.is_empty());
```

mobilecoind/api/proto/mobilecoind_api.proto
```diff
@@ -68,6 +68,9 @@ service MobilecoindAPI {
     // Database encryption
     rpc SetDbPassword (SetDbPasswordRequest) returns (google.protobuf.Empty) {}
     rpc UnlockDb (UnlockDbRequest) returns (google.protobuf.Empty) {}
+
+    // Versioning
+    rpc GetReleaseVersion(google.protobuf.Empty) returns (MobilecoindVersionResponse) {}
```
remoun, line 73: What does 'Release' version mean? Might `GetVersion` work well enough?

mobilecoind/src/service.rs
```diff
@@ -5977,4 +5988,29 @@ mod test {
...
+                logger.clone(),
+                &mut rng,
+            );
+
+        let expected_version = String("1.3.0-pre0");
+
+        // Send request.
+        let response = client
+            .get_release_version()
+            .expect("Incorrect version returned");
```
nick-mobilecoin, line 6013: Builds off @remoun's suggestion of not hardcoding the version string. 
It feels a bit like it's testing a re-implementation of the code.
It just happens that `CARGO_PKG_VERSOIN` is the common way to get the version in code and one is testing that that is the value being returned.
```suggestion
        let expected_version = env!("CARGO_PKG_VERSION");

        // Send request.
        let response = client
            .get_release_version(&mc_util_grpc::Empty::new())
            .expect("Failed to get version");
```

mobilecoind-json/src/data_types.rs
```diff
@@ -1547,3 +1547,16 @@ mod test {
         assert_eq!(proto_proposal, proto2);
     }
 }
+
+#[derive(Serialize, Default, Debug)]
```
nick-mobilecoin, line 1551: Suggest putting this before the `mod test` block

mobilecoind-json/src/data_types.rs
pouneh: Doesn't seem to work since version is a String, and the string type does implement the Copy trait ( see below for cargo build error message)

And I guess get_version() returns `&str` so I have to `to_string()` it ... is there something else I should try?
```
error[E0507]: cannot move out of `src.version` which is behind a shared reference
    --> mobilecoind-json/src/data_types.rs:1559:22
     |
1559 |             version: src.version,
     |                      ^^^^^^^^^^^ move occurs because `src.version` has type `std::string::String`, which does not implement the `Copy` trait
```

mobilecoind/api/proto/mobilecoind_api.proto
pouneh: Yea I guess 'Release' doesn't add any more information at this point. I'll change that thanks!

mobilecoind/src/service.rs
pouneh: Yea! That was my thoughts initially, so that's why I hard-coded the version string :/

remoun: This seems OK. I would also verify that the string is not empty.

mobilecoind-json/src/data_types.rs
remoun: Ah OK. You could write `src.version.clone()`, which is equivalent, so we might as well keep it as-is.

mobilecoind/src/service.rs
```diff
@@ -2103,7 +2112,9 @@ build_api! {
 
     // Database encryption
     set_db_password SetDbPasswordRequest Empty set_db_password_impl,
-    unlock_db UnlockDbRequest Empty unlock_db_impl
+    unlock_db UnlockDbRequest Empty unlock_db_impl,
+
+    get_version Empty MobilecoindVersionResponse get_version_impl
```
remoun, line 2117: IIUC we can add a trailing comma so future additions don't need to modify this line.

```suggestion
    get_version Empty MobilecoindVersionResponse get_version_impl,
```

remoun (approved):

iamalwaysuncomfortable (approved):
This is good, thanks!

nick-mobilecoin (approved):

PR #2122, opened 2022-06-10
Update all crate versions to match the current version

remoun:
`mc-util-cli` and `go-grpc-gateway-testing` depend on `mc-*` crates, so use the same version.

Found via the following command: `find . -iname "*.toml" | xargs grep '^version *=' | grep -v 'version = "1.3.0-pre0"'`
With this change, this command no longer yields any results.

### DIFF
--- a/util/cli/Cargo.toml
+++ b/util/cli/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mc-util-cli"
-version = "0.1.0"
+version = "1.3.0-pre0"
 authors = ["MobileCoin"]
 edition = "2021"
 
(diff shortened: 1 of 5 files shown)


util/cli/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "mc-util-cli"
-version = "0.1.0"
+version = "1.3.0-pre0"
```
remoun, line 3: Alternatively, we could lower the version of `mc-util-build-info`, since it does not have any other `mc-*` deps.

iamalwaysuncomfortable: Tracking the release version seems a viable strategy

iamalwaysuncomfortable (approved):

nick-mobilecoin (approved):

Issue #2104, opened 2022-06-07
AVR-on-Chain in 1.3


jcape:
- [x] MCIP for BlockVersion 3 requiring `BlockMetadata`: https://github.com/mobilecoinfoundation/mcips/pull/43
- [x] #1403
  - [x] #2067
  - [x] #2058 
  - [x] #2052
  - [x] #2205
  - [x] #2216

remoun:
This could be helpful for collecting SCP signing keys from consensus nodes: https://github.com/wiberlin/mc-crawler

PR #2131, opened 2022-06-14
Add a `MetadataValidator` for validating block metadata

remoun:
This PR adds validation for metadata signing keys; #2136 adds AVR validation.

This config can be used to revoke metadata signing keys, when `LedgerDB` calls this as part of its validation before appending blocks.

See [MCIP 43](https://github.com/mobilecoinfoundation/mcips/blob/block-metadata/text/0043-block-metadata.md#signing-key-history) and #2104 for more details.

### DIFF
--- a/blockchain/validators/src/metadata/key_range/config.rs
+++ b/blockchain/validators/src/metadata/key_range/config.rs
@@ -1,0 +1,442 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
+//! Configuration for the key range validator.
+
+use crate::ParseError;
+use hex::ToHex;
+use mc_blockchain_types::BlockIndex;
+use mc_crypto_keys::{DistinguishedEncoding, Ed25519Public};
+use serde::{Deserialize, Serialize};
+use std::{
+    collections::BTreeMap,
+    ffi::OsStr,
+    fs,
+    ops::RangeInclusive,
+    path::{Path, PathBuf},
+};
+
+/// Container for keys with a range of block indexes that are valid per key.
+pub type KeyValidityMap = BTreeMap<Ed25519Public, Vec<RangeInclusive<BlockIndex>>>;
+
+const PUBLIC_KEY_TAG: &str = "PUBLIC KEY";
+
+/// Container for key validity configs.
+/// Supports parsing a `metadata-signers.toml` as specified in [MCIP #43](https://github.com/mobilecoinfoundation/mcips/pull/43).
+#[derive(Clone, Debug, Deserialize, Eq, Ord, PartialEq, PartialOrd)]
+pub struct Config {
+    /// The key validity ranges.
+    #[serde(alias = "node")] // alias for parsing
+    pub configs: Vec<KeyValidity>,
+
+    /// Optional base path for parsing relative PEM file paths.
+    #[serde(skip)]
+    pub base_path: Option<PathBuf>,
+}
+
+impl Config {
+    /// Load the config as TOML or JSON from the given file path.
+    pub fn load(path: impl AsRef<Path>) -> Result<Self, ParseError> {
+        let path = path.as_ref();
+        let bytes = fs::read(path)?;
+        let mut config: Config = match path.extension().and_then(OsStr::to_str) {
+            Some("toml") => Ok(toml::from_slice(&bytes)?),
+            Some("json") => Ok(serde_json::from_slice(&bytes)?),
+            _ => Err(ParseError::UnrecognizedExtension(path.into())),
+        }?;
+        config.base_path = path.parent().map(Into::into);
+        Ok(config)
+    }
+
+    /// Get a [KeyValidityMap] from this config.
+    pub fn to_validity_map(&self) -> Result<KeyValidityMap, ParseError> {
+        let mut map = KeyValidityMap::new();
+        for config in &self.configs {
+            let key = parse_pem_or_hex(&config.pub_key, self.base_path.clone())?;
+            map.entry(key).or_default().push(config.to_range());
+        }
+        Ok(map)
+    }
+}
+
+/// A declaration that a public key is valid for a specified range of block
+/// indexes.
+#[derive(Clone, Debug, Deserialize, Ord, PartialOrd, Serialize)]
+pub struct KeyValidity {
+    /// The pub key, as a PEM file path or 64 hexadecimal characters.
+    /// File paths are parsed relative to the config file.
+    #[serde(alias = "message_signing_pub_key")] // alias for parsing
+    pub pub_key: String,
+    /// The first block index that used this key.
+    pub first_block_index: BlockIndex,
+    /// The last block index that used this key. Can be `None` for active keys.
+    pub last_block_index: Option<BlockIndex>,
+}
+
+impl KeyValidity {
+    /// Instantiate an instance with the given serialized value and index range.
+    pub fn new(
+        pub_key: String,
+        first_block_index: BlockIndex,
+        // Allows passing `N`, `Some(N)` or `None`
+        last_block_index: impl Into<Option<BlockIndex>>,
+    ) -> Self {
+        Self {
+            pub_key,
+            first_block_index,
+            last_block_index: last_block_index.into(),
+        }
+    }
+
+    /// Instantiate an instance with the given pub key and index range.
+    pub fn with_key(
+        key: Ed25519Public,
+        first_block_index: BlockIndex,
+        // Allows passing `N`, `Some(N)` or `None`
+        last_block_index: impl Into<Option<BlockIndex>>,
+    ) -> Self {
+        Self::new(key.encode_hex(), first_block_index, last_block_index)
+    }
+
+    /// Get a range of indexes from this config.
+    pub fn to_range(&self) -> RangeInclusive<BlockIndex> {
+        self.first_block_index..=self.last_block_index.unwrap_or(BlockIndex::MAX)
+    }
+}
+
+impl PartialEq for KeyValidity {
+    fn eq(&self, other: &Self) -> bool {
+        self.first_block_index == other.first_block_index
+            && self.last_block_index == other.last_block_index
+            && self.pub_key.trim() == other.pub_key.trim()
+    }
+}
+impl Eq for KeyValidity {}
+
+/// A helper method for deserializing an Ed25519Public from a PEM string, PEM
+/// file path, or hexadecimal string.
+fn parse_pem_or_hex(
+    value: &str,
+    base_path: impl Into<Option<PathBuf>>,
+) -> Result<Ed25519Public, ParseError> {
+    let value = value.trim();
+
+    let base_path = base_path.into().or_else(|| std::env::current_dir().ok());
+    let path = if let Some(base) = base_path {
+        base.join(&value)
+    } else {
+        value.into()
+    };
+
+    Ok(if path.exists() {
+        let pem = pem::parse(fs::read(path)?)?;
+        if pem.tag == PUBLIC_KEY_TAG {
+            Ed25519Public::try_from_der(&pem.contents)?
+        } else {
+            Err(ParseError::InvalidPemTag(pem.tag))?
+        }
+    } else if let Ok(pem) = pem::parse(value) {
+        if pem.tag == PUBLIC_KEY_TAG {
+            Ed25519Public::try_from_der(&pem.contents)?
+        } else {
+            Err(ParseError::InvalidPemTag(pem.tag))?
+        }
+    } else if value.len() == 64 {
+        Ed25519Public::try_from(hex::decode(value)?)?
+    } else {
+        Err(ParseError::InvalidPubKeyValue(value.to_owned()))?
+    })
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use tempfile::TempDir;
+
+    static INPUT_TOML: &str = r#"
+        # Key A is always valid.
+        [[node]]
+        pub_key = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+        first_block_index = 0
+
+        # Key B is only valid for the first 10 blocks.
+        [[node]]
+        pub_key = "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb"
+        first_block_index = 0
+        last_block_index = 10
+
+        # Key C is valid for all blocks except 11..=19
+        [[node]]
+        pub_key = "cccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc"
+        first_block_index = 0
+        last_block_index = 10
+
+        [[node]]
+        # Can also use `message_signing_pub_key`
+        message_signing_pub_key = "cccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc"
+        first_block_index = 20
+    "#;
+
+    static INPUT_JSON: &str = r#"
+    {
+        "node": [
+            {
+                "pub_key": "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
+                "first_block_index": 0
+            },
+            {
+                "pub_key": "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb",
+                "first_block_index": 0,
+                "last_block_index": 10
+            },
+            {
+                "pub_key": "cccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc",
+                "first_block_index": 0,
+                "last_block_index": 10
+            },
+            {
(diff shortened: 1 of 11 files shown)


blockchain/validators/src/config.rs
```diff
@@ -0,0 +1,212 @@
...
+
+        let result = KeyValidityConfig::load_from_path(&file);
+        assert!(matches!(result, Err(Error::MissingExtension)));
+    }
+
+    #[test]
+    fn config_with_pem_path() {
+        let dir = tempdir().unwrap();
+        let config_path = dir.path().join("testing.toml");
+        let pem_path = dir.path().join("test.pem");
```
nick-mobilecoin, line 176: Definitely not needed as part of this pr and maybe never needed.
Should this support files relative to the toml file?
right now this will most likely treat relative files as relative to the current working directory.

blockchain/validators/src/metadata_key.rs
```diff
@@ -0,0 +1,131 @@
+// Copyright (c) 2018-20&&22 The MobileCoin Foundation
```
nick-mobilecoin, line 1:
```suggestion
// Copyright (c) 2018-2022 The MobileCoin Foundation
```

blockchain/validators/src/metadata_key.rs
```diff
@@ -0,0 +1,131 @@
...
+            keys.entry(config.message_signing_pub_key)
+                .or_default()
+                .insert(config.clone());
+        }
+
+        Self { keys }
+    }
+
+    /// Check whether the given block metadata is valid at the given index.
+    pub fn verify(&self, metadata: &BlockMetadata, block_index: BlockIndex) -> Result<(), Error> {
```
nick-mobilecoin, line 30: I don't think I have the full picture and maybe it depends on how the key -> ranges are built up and what trust needs to be there.

Let us say that Key A is valid from blocks 20-30, so this function gets in index 25 with key A.  This method would return true.  But what if A was only used for block 20 and block 30, but nothing inbetween?

It could be that "verify" and "valid" might need to be explained more for this method.  I.e. is it that this is a potential _candidate_ and more looking needs to be done to ensure it really is correct for this block

blockchain/validators/src/metadata_key.rs
```diff
@@ -0,0 +1,131 @@
...
+        let signer = Ed25519Pair::from_random(&mut rng);
+
+        BlockMetadata::from_contents_and_keypair(
+            make_block_metadata_contents(make_block_id(&mut rng), &mut rng),
+            &signer,
+        )
+        .expect("BlockMetadata::from_contents_and_keypair")
+    }
+
+    fn make_verifier() -> MetadataKeyValidator {
```
nick-mobilecoin, line 73: nit: should this be named `make_validator`?

blockchain/validators/src/metadata_key.rs
```diff
@@ -0,0 +1,131 @@
...
+        let metadata = make_metadata(2);
+        assert_eq!(verifier.verify(&metadata, 11), Err(Error::InvalidPubKey));
+        assert_eq!(verifier.verify(&metadata, 10), Ok(()));
+    }
+
+    #[test]
+    fn premature_key() {
+        let verifier = make_verifier();
+        let metadata = make_metadata(3);
+        assert_eq!(verifier.verify(&metadata, 1), Err(Error::InvalidPubKey));
```
nick-mobilecoin, line 110: Should test right at the boundary to avoid off by ones
```suggestion
        assert_eq!(verifier.verify(&metadata, 9), Err(Error::InvalidPubKey));
```

nick-mobilecoin (approved):
minor suggestions, LGTM

blockchain/validators/src/config.rs
```diff
@@ -0,0 +1,212 @@
...
+    /// Check whether a given block index falls within the range of this
+    /// [NodeKeyValidity].
+    pub fn index_in_range(&self, block_index: BlockIndex) -> bool {
+        block_index >= self.first_block_index
+        // If there is a last block, enforce that it's no greater than that.
+            && !self
+                .last_block_index
+                .map(|last_index| block_index>last_index)
+                .unwrap_or_default()
+    }
```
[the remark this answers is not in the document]
remoun: Good idea! Done.

blockchain/validators/src/config.rs
remoun: Good idea! I'll make it relative to the config file.

blockchain/validators/src/metadata_key.rs
remoun: Good catch! Left over from a rename.

blockchain/validators/src/config.rs
remoun: I couldn't find a not-too-hacky way to get the path from `load_from_path` into the custom deserialize method `deserialize_pem_or_hex`. I'm guessing operators will use absolute paths by default, and it seems OK to require absolute paths (or/and document that the paths are treated relative to CWD).

blockchain/validators/src/config.rs
```diff
@@ -0,0 +1,217 @@
...
+            last_block_index: last_block_index.into(),
+        }
+    }
+
+    /// Check whether a given block index falls within the range of this
+    /// [NodeKeyValidity].
+    pub fn index_in_range(&self, block_index: BlockIndex) -> bool {
+        block_index >= self.first_block_index &&
+        // If there is a last block, enforce that it's no greater than that.
+        self.last_block_index.map(|last_index| block_index <= last_index).unwrap_or(true)
```
nick-mobilecoin, line 44: I think the intermediate comment is confusing cargo fmt, or perhaps my previous suggestion here was malformated
```suggestion
        // If there is a last block, enforce that it's no greater than that.
        block_index >= self.first_block_index
            && self
                .last_block_index
                .map(|last_index| block_index <= last_index)
                .unwrap_or(true)
```

blockchain/validators/src/config.rs
```diff
@@ -0,0 +1,217 @@
...
+use mc_blockchain_types::BlockIndex;
+use mc_crypto_keys::Ed25519Public;
+use serde::{Deserialize, Deserializer, Serialize};
+use std::{ffi::OsStr, fs, path::Path};
+
+/// A declaration that a pub key is valid for a specified range of block
+/// indexes.
+#[derive(Clone, Debug, Deserialize, Eq, Ord, PartialEq, PartialOrd, Serialize)]
+pub struct NodeKeyValidity {
+    /// The pub key, as a PEM file path or 64 hexadecimal characters.
```
nick-mobilecoin, line 15: Would it be good to call out the path behaviour here?
Will the toml have any stand alone documentation?  
```suggestion
    /// The pub key, as a PEM file path or 64 hexadecimal characters.
    /// When providing a PEM file path, the path should be absolute or
    /// relative to the current working directory.
```

nick-mobilecoin (approved):

blockchain/validators/src/config.rs
nick-mobilecoin: Sounds good.

Again not saying to do this now, just more curious if you feel this falls into the "hacky" camp.  In the past in Python I've use a context manager to change the current working directory to the path of the config file.  Looks like there is https://docs.rs/chwd/0.2.0/chwd/struct.ChangeWorkingDirectory.html which does something similar.

remoun: Interesting! I ended up using a different approach, where I defer parsing as `Ed25519Public` and track the dir passed to `load()`

blockchain/validators/src/metadata_key/config.rs
```diff
@@ -0,0 +1,268 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
+//! Config for validators.
```
iamalwaysuncomfortable, line 3: My intuition is for objects that validate - I feel like we should organize it in a submodule like:

blockchain/validators/src/types

And then we'd have
blockchain/validators/src/types/metadata_key.rs
blockchain/validators/src/types/avr_records.rs

Thoughts?

blockchain/validators/src/metadata_key/config.rs
```diff
@@ -0,0 +1,268 @@
...
+        }?;
+        config.base_path = path.parent().map(Into::into);
+        Ok(config)
+    }
+
+    /// Get a [KeyValidityMap] from this config.
+    pub fn to_validity_map(&self) -> Result<KeyValidityMap, ParseError> {
+        let mut map = KeyValidityMap::new();
+        for config in &self.configs {
+            let key = parse_pem_or_hex(&config.message_signing_pub_key, self.base_path.clone())?;
```
iamalwaysuncomfortable, line 50: Is there a way go from Ed25519Public to hex yet?

blockchain/validators/src/metadata_key/validator.rs
```diff
@@ -0,0 +1,147 @@
...
+    /// Check whether the given block metadata is valid at the given index.
+    pub fn validate(
+        &self,
+        metadata: &BlockMetadata,
+        block_index: BlockIndex,
+    ) -> Result<(), ValidationError> {
+        let ranges = self
+            .config
+            .get(metadata.node_key())
+            .ok_or(ValidationError::UnknownPubKey)?;
```
iamalwaysuncomfortable, line 29: What happens in the missing data case? Definitely for AVRs there is missing data where a key isn't existent in the records and I'm guessing there might be this case for historical data or in some odd case in the future.

There should be a case here for handling the situation where there are no historical records. Either that or we should check if it's okay to interpolate them.

The exception is if there's no missing data in these records.

iamalwaysuncomfortable (approved):
A few comments about the organization and also what happens in the case of missing records

blockchain/validators/src/config.rs
nick-mobilecoin: IDK I just talked with @jcape he mentioned there is some concerns about changing working directory in app.
it looks like it's not thread safe.

The delayed parsing is probably the safest, but definitely feels like more than needed at this time

blockchain/validators/src/metadata_key.rs
remoun: Fair point. I elaborated in the method docs.

FWIW the intent here is that `LedgerDB` will call this as part of its validation before appending blocks, and this config can be used to revoke signing keys.

blockchain/validators/src/metadata_key/config.rs
remoun: It's unclear to me where the combined metadata validator would go in that scheme.

I just restructured to: having a `mc_blockchain_validators::MetadataValidator` that composes `mc_blockchain_types::BlockMetadata::verify()`, `mc_blockchain_validators::metadata::KeyRangeValidator` and soon an `mc_blockchain_validators::metadata::AvrHistoryValidator`. How does that sound?

blockchain/validators/src/metadata_key/validator.rs
remoun: It seems easier to enumerate the SCP message signing keys (than historical AVRs) since they're configured by the node operators, and we don't have anybody generating metadata yet. This PR is in part a key revocation mechanism.

To answer your immediate question: if we're missing a key, we won't be able to validate metadata signed by that key.

nick-mobilecoin (approved):

eranrund (changes requested):
Nice and easy to follow. Mostly LGTM, have some questions and also comments about configuration formats (JSON support, PEM stuff).

blockchain/validators/src/metadata/key_range/config.rs
```diff
@@ -0,0 +1,274 @@
...
+    pub base_path: Option<PathBuf>,
+}
+
+impl Config {
+    /// Load the config TOML from the given file path.
+    pub fn load(path: impl AsRef<Path>) -> Result<Self, ParseError> {
+        let path = path.as_ref();
+        let bytes = fs::read(path)?;
+        let mut config: Config = match path.extension().and_then(OsStr::to_str) {
+            Some("toml") => Ok(toml::from_slice(&bytes)?),
```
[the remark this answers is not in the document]
remoun: Good call. On it.

blockchain/validators/src/metadata/mod.rs
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    /// Validate that the given metadata is valid at the given block index.
+    pub fn validate(
+        &self,
+        metadata: &BlockMetadata,
+        block_index: BlockIndex,
+    ) -> Result<(), ValidationError> {
+        self.key_range.validate(metadata.node_key(), block_index)?;
+        // TODO: Validate AVR.
```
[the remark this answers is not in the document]
remoun: Yes it is! #2136 starts on that; this TODO is a placeholder for it.

blockchain/validators/src/metadata_key/config.rs
iamalwaysuncomfortable: Ah yes, context is good. What I meant was do we have a nice existing way to serialize and deserialize to hex for readibility in print & serialization to disk?

blockchain/validators/src/metadata_key/validator.rs
iamalwaysuncomfortable: I suppose the larger philisophical question is, there's always likely to be a window of time when a node might not be participating in consensus, and so how do we indicate that gap and what do we do about it? 

In terms of representation bit:
I think in this case, indicating what ranges the message signing key is valid for works well.

In terms of action:
For non-consensus nodes syncing from a particular consensus peer or proxy representing that peer - what happens when the metadata can't be validated because that node didn't form blocks in that range? That's a subtle yet import question with a few potential answers. 

I think for unknown keys/sigs that can't be validated/keys at invalid ranges, the answer is "validator throws hard error". In the case that the data is simply missing due to a node not having formed any blocks in those ranges the answer is either record "None" as a valid answer in the client node's local block ledger or try to find a peer that does have metadata records for that range.

iamalwaysuncomfortable (commented):
Some musing on serialization, otherwise looking great

blockchain/validators/src/metadata/key_range/config.rs
```diff
@@ -0,0 +1,442 @@
...
+            Some("json") => Ok(serde_json::from_slice(&bytes)?),
+            _ => Err(ParseError::UnrecognizedExtension(path.into())),
+        }?;
+        config.base_path = path.parent().map(Into::into);
+        Ok(config)
+    }
+
+    /// Get a [KeyValidityMap] from this config.
+    pub fn to_validity_map(&self) -> Result<KeyValidityMap, ParseError> {
```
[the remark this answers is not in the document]
remoun: This `struct Config` is the piece that leverages automatic serde, with a structure matching the config file format, and this method maps to the internal representation.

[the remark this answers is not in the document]
remoun: I initially had `parse_pem_or_hex` in a `serde` annotation, but I moved it to this method so we can treat the relative paths  as relative to the TOML file rather than CWD.

eranrund:
I'm ready to approve once I understand the plan for this getting to `master` - just want to make sure there is a good plan for what the plan is for metadata in the ledger sync code. Thank you!

blockchain/validators/src/metadata_key/config.rs
remoun: The `Debug` and `Display` impls do use hex encoding. This logic parses as inline hex, inline PEM, or a PEM file path.

blockchain/validators/src/metadata/key_range/config.rs
```diff
@@ -0,0 +1,442 @@
...
+use std::{
+    collections::BTreeMap,
+    ffi::OsStr,
+    fs,
+    ops::RangeInclusive,
+    path::{Path, PathBuf},
+};
+
+/// Container for keys with a range of block indexes that are valid per key.
+pub type KeyValidityMap = BTreeMap<Ed25519Public, Vec<RangeInclusive<BlockIndex>>>;
```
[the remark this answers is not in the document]
remoun: Fair point. Done.

ledger/db/src/ledger_db.rs
```diff
@@ -741,6 +773,12 @@ impl LedgerDB {
             }
         }
 
+        let block_version = BlockVersion::try_from(block.version)
+            .or(Err(Error::InvalidBlockVersion(block.version)))?;
+        if block_version.require_block_metadata() && metadata.is_none() {
```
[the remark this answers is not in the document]
remoun: Good question. I removed https://github.com/mobilecoinfoundation/mobilecoin/pull/2052 as the base, so let's continue this discussion there.

remoun:
> I'm ready to approve once I understand the plan for this getting to `master` - just want to make sure there is a good plan for what the plan is for metadata in the ledger sync code. Thank you!

This did not actually need to build on #2052. Now that #2058 is merged, I changed this PR to merge into `master`, with no changes to `LedgerDB` yet.

eranrund (approved):

blockchain/validators/src/metadata_key/config.rs
remoun: I can see us adding other blockchain validators in the near future. If not, we can inline the `mod metadata` later.

blockchain/validators/src/metadata_key/validator.rs
remoun: Got it, I think. Yeah, we still have work to do around catch-up mode, so I just removed the check that requires block metadata from #2052 until we have a good answer there.

If we want to prohibit a node's key from appearing when it wasn't participating in consensus, we can split the range that includes this disconnect into two, one leading up to the disconnect and one after reconnecting.

PR #2144, opened 2022-06-15
Fix OMAP settings and adjust probe timings.

jgreat:
### Motivation

- The OMAP_CAPACITY env changed to MC_OMAP_CAPACITY in 1.2.x. 
- Update health check probe timings for long startup times in testnet/mainnet
- make ledger and ingress instances come up in parallel. Ordered startup is unecessary.


.internal-ci/helm/fog-ingest/templates/fog-ingest-statefulset.yaml
```diff
@@ -6,6 +6,7 @@ metadata:
   labels:
     {{- include "fogIngest.labels" . | nindent 4 }}
 spec:
+  podManagementPolicy: Parallel
```
joekottke, line 9: _NIIIICE_

joekottke (approved):
LGTM

hummingalpaca (approved):
lgtm

remoun (approved):

garbageslam (approved):
sorry we didn't communicate this change more clearly to ops

.internal-ci/helm/fog-ingest/templates/fog-ingest-statefulset.yaml
MCrank: Small nit.  This was defined in values.yaml for fog-ledger. Do we need/want to do for ingest?  Won't really matter in the long run unless we think we might need to override this in the future.

MCrank (approved):
Small nit on fog ingest `podManagementPolicy`

jgreat:
> Small nit on fog ingest `podManagementPolicy`

Since we do a blue/green (separate chart release) we git different StatefulSets. The reason I needed to make it variable for fog-ledger is that we are doing a rolling update.  The 1.1.3-dev chart defaults podManagementPolicy to orderedReady.  I guess you can't change this field in an existing StatefulSet.

PR #2145, opened 2022-06-15
Replace StdRng in test_hashing_is_consistent_block_version_one with get_seeded_rng() which uses Hc128Rng

### LINKED ISSUE #2042: Replace `StdRng` in determinism tests with `Hc128Rng` from `mc-util-test-helper` crate
In `transaction/core/src/blockchain/block.rs`, we have tests that hash sample blockchain data and assert that the hash has a particular value. This test can help catch compatibility-breaking changes, because if we accidentally change the way things are hashed, that will lead to a hash mismatch at this unit test.

Unfortunately, this test is implemented using `rand::StdRng`. This is bad because the `rand` crate has, in the past, changed which RNG algorithm is considered the "default" and published as `StdRng`, causing a lot of our determinism tests to break.

We should prefer to use a fixed RNG algorithm such as `rand_hc`. We are currently exporting `Hc128Rng` as `RngType` from the `mc-util-test-helper` crate, and most tests are using, and should continue to use this. We should replace usages of `StdRng` with this, particularly around the hash consistency tests, otherwise one day it will randomly break and someone will have to deal with it.

pouneh:
Per #2042 , the `rand` crate may change what Rng algorithm is used by default, which would cause this test to misleadingly fail. We change this test to use `mc-util-test-helper::get_seeded-rng()` which uses a fixed RNG algorithm. We now no longer depend on the default Rng algorithm in `rand` to be the same in the future.

Resolves #2042


blockchain/types/src/block.rs
```diff
@@ -384,30 +384,24 @@ mod block_tests {
...
     /// calculation of old blocks.
     fn test_hashing_is_consistent_block_version_one() {
-        // FIXME (#2042): This shouuld not use StdRng, because the rust standard library
-        // sometimes updates what RNG algorithm this point to. We should
-        // rerun this with Hc128 RNG (from mc_util_test_helper) and update the
-        // test vectors.
-        use rand::rngs::StdRng;
-
-        let mut rng = StdRng::from_seed([1u8; 32]);
+        let mut rng = get_seeded_rng();
```
jcape, line 387: Nit: https://doc.rust-lang.org/1.0.0/style/style/imports.html#prefer-fully-importing-types/traits-while-module-qualifying-functions.

jcape (commented):
Looks good, one nitpick

blockchain/types/src/block.rs
pouneh: Per that style-guide, importing the function without module-qualification is appropriate as a convenience in test sub-modules

wjuan-mob (approved):

PR #2149, opened 2022-06-16
Add bindings for CRC computation

dolanbernard:
Added Android bindings for commitment CRC computation

### Motivation

This PR is needed in order to verify the reconstructed TxOutRecord commitment data

### DIFF
--- a/android-bindings/Cargo.toml
+++ b/android-bindings/Cargo.toml
@@ -11,6 +11,9 @@ name = "mobilecoin_android"
 crate-type = ["cdylib"]
 
 [dependencies]
+# External dependencies
+crc = "3.0.0"
+
 # fog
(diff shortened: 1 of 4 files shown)


android-bindings/Cargo.toml
```diff
@@ -11,6 +11,9 @@ name = "mobilecoin_android"
 crate-type = ["cdylib"]
 
 [dependencies]
+# External dependencies
+crc = "3.0.0"
```
remoun, line 15: Nit: Move to `third-party` section below

remoun (approved):
I was wondering about feature parity across SDKs, but I see we already have an analog in `libmobilecoin`, namely `mc_tx_out_commitment_crc32`, so this makes sense.

voloshyn (approved):

PR #2150, opened 2022-06-16
Kusco/infra dev

hummingalpaca:
This is deploy a new cluster to infra-dev.

- Updated rancher_cluster secret 
- Changed development specific urls to infra-dev
- Updated s3 bucket location


.github/workflows/mobilecoin-dispatch-dev-delete.yaml
```diff
@@ -18,7 +18,7 @@ jobs:
     with:
       namespace: ${{ github.event.inputs.namespace }}
     secrets:
-      RANCHER_CLUSTER: ${{ secrets.RANCHER_CLUSTER }}
+      RANCHER_CLUSTER: ${{ secrets.RANCHER_CLUSTER_INFRA_DEV }}
```
jgreat, line 21: Please use INFRA_DEV_RANCHER_CLUSTER as the pattern here.  Secrets are already namespaces as DEV_<var>

PR #2155, opened 2022-06-16
Longer startup, topology and affinity rules 

jgreat:
### Motivation

Fog-view may or may not complete in the 2 hour start up time. Bump the startup wait time to 4 hours.

In addition if we were to loose all the view pods in an envrionment we would be looking at several hours of downtime while view reloads.  To minimize the chance of a single node or availability zone taking down the entire view service, add topology  and anti-affinity rules.

- Anti-affinity rules for view, ingest and consensus to make sure each pod ends up on their own node. Report is stateless and fast to come up. We don't need to constrain it to specific nodes.
- Topology rules to enforce balanced distribution across availability zones.
- LabelsSelectors should enforce rules across the app *and* helm chart version.  This way when we have a new helm release version we don't share the constraint and block pod deployment.
- double checked testnet/mainnet and we have enough nodes to support consensus "required" rules for anti-affinity.
- adjust strategy on view to allow pods half the pods to be terminated during a rolling upgrade. This should allow view to be rotated even if there is only 4 nodes.

Bonus, `concurrency` rule on GHA to keep builds from overlapping on a branch. Simultaneous but separate branches are ok, just not to the same branch at the same time.

### DIFF
--- a/.internal-ci/helm/consensus-node/templates/node-deployment.yaml
+++ b/.internal-ci/helm/consensus-node/templates/node-deployment.yaml
@@ -28,20 +28,28 @@ spec:
         client-load-balanced: 'true'
         {{- end }}
     spec:
-      nodeSelector:
-        {{- toYaml .Values.node.nodeSelector | nindent 8 }}
+      # Try to balance pods across zones
+      topologySpreadConstraints:
+      - topologyKey: topology.kubernetes.io/zone
+        maxSkew: 1
+        # Wait until we have nodes
+        whenUnsatisfiable: DoNotSchedule
+        labelSelector:
+          matchLabels:
+            # match app and helm chart version
+            app: consensus-node
+            helm.sh/chart: {{ include "consensusNode.chart" . }}
       affinity:
         podAntiAffinity:
-          preferredDuringSchedulingIgnoredDuringExecution:
-          - weight: 100
-            podAffinityTerm:
-              labelSelector:
-                matchExpressions:
-                - key: app
-                  operator: In
-                  values:
-                  - consensus-node
-              topologyKey: 'kubernetes.io/hostname'
+          requireDuringSchedulingIgnoredDuringExecution:
+          - topologyKey: kubernetes.io/hostname
+            labelSelector:
+              matchLabels:
+                # match all consensus-nodes deployed by the same chart version
+                app: consensus-node
+                helm.sh/chart: {{ include "consensusNode.chart" . }}
+      nodeSelector:
+        {{- toYaml .Values.node.nodeSelector | nindent 8 }}
       tolerations:
       {{- toYaml .Values.node.tolerations | nindent 6 }}
       imagePullSecrets:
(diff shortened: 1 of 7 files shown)


.internal-ci/helm/consensus-node/templates/node-deployment.yaml
```diff
@@ -28,20 +28,28 @@ spec:
...
+      topologySpreadConstraints:
+      - topologyKey: topology.kubernetes.io/zone
+        maxSkew: 1
+        # Wait until we have nodes
+        whenUnsatisfiable: DoNotSchedule
+        labelSelector:
+          matchLabels:
+            # match app and helm chart version
+            app: consensus-node
+            helm.sh/chart: {{ include "consensusNode.chart" . }}
```
remoun, line 41: The additional labels are used in other configs but not this one.
```suggestion
            helm.sh/chart: {{ include "consensusNode.chart" . }}
            {{- include "consensusNode.selectorLabels" . | nindent 6 }}
```

.internal-ci/helm/fog-ingest/templates/fog-ingest-statefulset.yaml
```diff
@@ -21,6 +21,28 @@ spec:
...
+      - topologyKey: topology.kubernetes.io/zone
+        maxSkew: 1
+        # Wait until we have nodes
+        whenUnsatisfiable: DoNotSchedule
+        labelSelector:
+          matchLabels:
+            # match on this helm chart install
+            app: fog-ingest
+            helm.sh/chart: {{ include "fogIngest.chart" . }}
+            {{- include "fogIngest.selectorLabels" . | nindent 6 }}
```
remoun, line 35: Is this the correct indentation? Looks like it should be 12 or 14

.internal-ci/helm/consensus-node/templates/node-deployment.yaml
```diff
@@ -28,20 +28,28 @@ spec:
...
-            podAffinityTerm:
-              labelSelector:
-                matchExpressions:
-                - key: app
-                  operator: In
-                  values:
-                  - consensus-node
-              topologyKey: 'kubernetes.io/hostname'
+          requireDuringSchedulingIgnoredDuringExecution:
+          - topologyKey: kubernetes.io/hostname
```
remoun, line 45:
```suggestion
          # Require pods to be on separate nodes.
          - topologyKey: kubernetes.io/hostname
```

.internal-ci/helm/consensus-node/templates/node-deployment.yaml
jgreat: This is intentional, consensus gets deployed in multiple helm "releases". So we want don't want to restrict matching on release name, which using selectorLabels would do.

.internal-ci/helm/fog-ingest/templates/fog-ingest-statefulset.yaml
jgreat: nope, good catch, I'm sure the build is just about to fail on that.

.internal-ci/helm/consensus-node/templates/node-deployment.yaml
jgreat: Added that comment.

remoun (approved):

PR #2180, opened 2022-06-21
fix mob docker command arg

david1alvarez:
adds a `--` before the `cmd` arg to enable parser to work correctly
changes args from `REMAINDER` to `+` for consistency


remoun (approved):

remoun (approved):
I was hoping to avoid passing `--cmd`, but perhaps that's not feasible.

christian-oudard (approved):

PR #2178, opened 2022-06-21
Pick sgx bumps

garbageslam:
This cherry-picks the SGX bumps to master.

I am not sure if this is preferable to merging release/v1.2.2 to master but in the meantime I need to get a passing deploy.

PR #2192, opened 2022-06-24
SGX bumps + CI fix

remoun:


.circleci/config.yml
```diff
@@ -258,11 +259,11 @@ commands:
           name: Install CI dependencies
           command: |
             cd ..
-            command -v sccache >/dev/null || rustup run --install stable cargo install sccache
-            command -v cargo-cache >/dev/null || rustup run --install stable cargo install cargo-cache
-            command -v cargo2junit >/dev/null || rustup run --install stable cargo install cargo2junit
+            command -v sccache >/dev/null || cargo install sccache
```
nick-mobilecoin, line 262: Is there any advantage to using sccache on CLI?
Not suggesting any change as part of this PR just a question.

.circleci/config.yml
```diff
@@ -764,8 +777,20 @@ jobs:
...
+            command -v go >/dev/null || (
+              wget https://go.dev/dl/go1.18.3.linux-amd64.tar.gz
+              tar -C /usr/local -xzf go1.18.3.linux-amd64.tar.gz
+              rm go1.18.3.linux-amd64.tar.gz
+
+              echo 'export GOPATH=/usr/local/go' >> $BASH_ENV
+              echo 'export PATH=$PATH:$GOPATH/bin' >> $BASH_ENV
+            )
+            . $BASH_ENV
+            go version
```
nick-mobilecoin, line 793: Is this `go version` a debug thing left over?
I could also see wanting to have the version info in the CI output

nick-mobilecoin (approved):
LGTM
How does this relate to, #2178?  Guessing this is more or less #2178 rebased on the current mainline.

.circleci/config.yml
remoun: It's actually all invoked via the CLI; this is just a direct invocation. sccache is integrated into the Cargo build by setting `RUSTC_WRAPPER=sccache`, which this config exports in line 93.

.circleci/config.yml
remoun: It was useful for verifying go is installed after that check, as well as printing the version. I can remove it if you prefer.

jcape (changes requested):
Last I'd heard, SGX 2.17 doesn't actually work in a deployment, we should continue to use 2.16 in CI until that's been fixed.

garbageslam:
Lets abandon #2178 in favor of this

remoun:
Alas, the test deploy did not succeed.

Key logs:
```
[init ../../../psw/ae/aesm_service/source/core/ipc/UnixCommunicationSocket.cpp:225] Failed to connect to socket /var/run/aesmd/aesm.socket
OH NO, WE CRASHED :( thread main on /usr/bin/consensus-service
Details: panicked at 'Could not create consensus enclave: SGX_ERROR_SERVICE_UNAVAILABLE', consensus/enclave/src/lib.rs:68:10
```

PR #2195, opened 2022-06-24
Add max token id argument to local network bootstrap script.

christian-oudard:
.

### DIFF
--- a/util/generate-sample-ledger/src/lib.rs
+++ b/util/generate-sample-ledger/src/lib.rs
@@ -70,9 +70,10 @@ pub fn bootstrap_ledger(
     log::info!(logger, "recipients: {}", recipients.len());
     log::info!(
         logger,
-        "Making {:?} outputs of {:?} picoMOB.",
+        "Making {:?} outputs of {:?} picoMOB, up to token ID {:?}.",
         num_outputs,
-        picomob_per_output
+        picomob_per_output,
+        max_token_id,
     );
 
     let mut blocks_and_contents: Vec<(Block, BlockContents)> = Vec::new();
@@ -85,6 +86,7 @@ pub fn bootstrap_ledger(
     } else {
         BLOCK_VERSION
     };
+    log::info!(logger, "Using block version {}", block_version);
 
     for block_index in 0..num_blocks as u64 {
         log::info!(logger, "Creating block {} of {}.", block_index, num_blocks);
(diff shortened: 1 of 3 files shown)


util/generate-sample-ledger/src/lib.rs
```diff
@@ -131,23 +138,35 @@ pub fn bootstrap_ledger(
     }
 
     for (block, block_contents) in blocks_and_contents {
-        db.append_block(&block, &block_contents, None).unwrap();
+        db.append_block(&block, &block_contents, None).expect("Duplicate key, because of re-using old sample ledger. Delete target/sample_data/ledger  to start from scratch.");
```
remoun, line 141: Nit: Log the actual path
```suggestion
        db.append_block(&block, &block_contents, None).unwrap_or_else(|err| {
            panic!(
                "Duplicate key, probably because of re-using old sample ledger. Delete {:?} to start from scratch.\nError: {}",
                path,
                err
            )
        });
```
(needs formatting)

util/generate-sample-ledger/src/lib.rs
```diff
@@ -70,9 +70,10 @@ pub fn bootstrap_ledger(
     log::info!(logger, "recipients: {}", recipients.len());
     log::info!(
         logger,
-        "Making {:?} outputs of {:?} picoMOB.",
+        "Making {:?} outputs of {:?} picoMOB, up to token ID {:?}.",
         num_outputs,
-        picomob_per_output
+        picomob_per_output,
+        max_token_id,
```
garbageslam, line 76: so, there's an issue here which is, we added `max_token_id` thinking it would work, but in fact it doesn't work. I'm not sure exactly why but I've never been able to get alternate tokens that are spendable by fog distro.

I think we should actually just delete the `--max-token-id` thing, we've already founder other solutions where the custom tokens are minted normally

util/generate-sample-ledger/src/lib.rs
```diff
@@ -131,23 +138,35 @@ pub fn bootstrap_ledger(
...
+        &mut file,
+        r##"
+{{
+    "NUM_KEYS": {},
+    "NUM_UTXOS_PER_ACCOUNT": {},
+    "NUM_BLOCKS": {},
+    "NUM_EXTRA_KEY_IMAGES_PER_BLOCK": {},
+    "MAX_TOKEN_ID": {},
+    "GIT_COMMIT": "{}"
+ }}"##,
```
remoun, line 157: Nit: Extra space
```suggestion
}}"##,
```

garbageslam (changes requested):
i think we should redevelop this --

we should actually remove the max token id argument, instead of using it in more places, since it doesn't actually work

PR #2196, opened 2022-06-24
Change heading level in README

jcape:
- Make "Transactions" H3 instead of H5

### Motivation

Test GHA from forks.


remoun:
It works!

PR #2199, opened 2022-06-27
View Account Key

BrianCorbin:
Adding ViewAccountKey struct which stores the view_private_key and spend_public_key, which can be used to generate PublicAddresses using a different derivation path than the current implementation in AccountKey

Current:
173395537-4f3c246f-8962-43e5-8cad-9fc81ae3b9c9

New:
173395410-bb207166-9fd8-45d4-b3b9-855d881cbcfa

### DIFF
--- a/account-keys/src/account_keys.rs
+++ b/account-keys/src/account_keys.rs
@@ -22,7 +22,7 @@ use core::{
     fmt,
     hash::{Hash, Hasher},
 };
-use curve25519_dalek::scalar::Scalar;
+use curve25519_dalek::{ristretto::RistrettoPoint, scalar::Scalar};
 use mc_crypto_digestible::Digestible;
 use mc_crypto_hashes::{Blake2b512, Digest};
 use mc_crypto_keys::{RistrettoPrivate, RistrettoPublic};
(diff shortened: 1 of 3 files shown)


account-keys/src/account_keys.rs
```diff
@@ -455,6 +455,165 @@ impl AccountKey {
...
+impl Eq for ViewAccountKey {}
+
+impl PartialEq for ViewAccountKey {
+    fn eq(&self, other: &Self) -> bool {
+        self.default_subaddress().eq(&other.default_subaddress())
+    }
+}
+
+impl PartialOrd for ViewAccountKey {
+    fn partial_cmp(&self, other: &ViewAccountKey) -> Option<Ordering> {
```
garbageslam, line 487: I think it may be marginally better to implement partial ord by calling ord, since if partial ord and ord disagree it's a bug

https://doc.rust-lang.org/std/cmp/trait.PartialOrd.html#how-can-i-implement-partialord

account-keys/src/account_keys.rs
```diff
@@ -605,4 +764,20 @@ mod account_key_tests {
...
+        let mut rng: StdRng = SeedableRng::from_seed([42u8, 32]);
+        let view_private = RistrettoPrivate::from_random(&mut rng);
+        let spend_private = RistrettoPrivate::from_random(&mut rng);
+        let account_key = AccountKey::new(&spend_private, &view_private);
+        let view_account_key = ViewAccountKey::from(&account_key);
+
+        assert_eq!(
+            account_key.default_subaddress(),
+            view_account_key.default_subaddress()
+        )
```
garbageslam, line 781: may be worth testing one of the subaddresses of index other than 0 also?

garbageslam (approved):
LGTM except for nits, thanks for doing this

account-keys/src/account_keys.rs
BrianCorbin: Updated this to use cmp, and also updated for AccountKey PartialOrd

jcape:
This needs to go to master.

account-keys/src/account_keys.rs
```diff
@@ -605,4 +762,27 @@ mod account_key_tests {
...
+            account_key.default_subaddress(),
+            view_account_key.default_subaddress()
+        );
+
+        assert_eq!(
+            account_key.change_subaddress(),
+            view_account_key.change_subaddress()
+        );
+
+        assert_eq!(account_key.subaddress(500), view_account_key(500));
```
remoun, line 786: Seems to be missing the method name.
```suggestion
        assert_eq!(account_key.subaddress(500), view_account_key.subaddress(500));
```

remoun (changes requested):

account-keys/src/account_keys.rs
```diff
@@ -455,6 +455,165 @@ impl AccountKey {
...
+
+// Note: Hash, Ord is implemented in terms of default_subaddress() because
+// we don't want comparisons to leak private key details over side-channels.
+impl Hash for ViewAccountKey {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.default_subaddress().hash(state)
+    }
+}
+
+impl Eq for ViewAccountKey {}
```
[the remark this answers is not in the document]
garbageslam: I guess I disagree: we are comparing the Public address which contains all the data, and the point of doing this at all is to be compatible with hash sets and btree sets etc which require such traits

garbageslam:
Otherwise lgtm still

garbageslam (approved):
LGTM

account-keys/src/account_keys.rs
```diff
@@ -475,6 +475,180 @@ impl AccountKey {
...
+        &self.view_private_key
+    }
+
+    /// Get the spend public key.
+    pub fn spend_public_key(&self) -> &RistrettoPublic {
+        &self.spend_public_key
+    }
+
+    /// Create a view account key with random keys
+    pub fn random<T: RngCore + CryptoRng>(rng: &mut T) -> Self {
```
remoun, line 553: Optional/Equivalent: Use `impl` instead of generics
```suggestion
    pub fn random(rng: &mut (impl RngCore + CryptoRng)) -> Self {
```

remoun (approved):

PR #2200, opened 2022-06-27
Implement enclave APIs that facilitate Fog View Router dynamic store discovery

samdealy:
### Motivation
The Fog View Router is responsible for taking incoming client requests, forwarding them to Fog View Shards, and then returning these received responses to the client. This PR adds three ake enclave methods (and corresponding view enclave methods) that facilitate this process. 

The first method, `backend_init`, allows an enclave to create a Noise auth request to another enclave that serves as a backend.A backend connection is one in which the current enclave is a client to another enclave. It is not the same as a peer connection, in which the current enclave sends peer messages to another enclave, and it's distinct from a client connection, in which the current enclave services client requests. The `backend_connect` method finishes the Noise connection process, and the `reencrypt_client_message_for_backends` takes a client message and makes a message for each connected backend.

Addresses the  `Implement shard component discovery in Fog Router" item in #2028. This item will be marked complete once the next PR, which uses these enclave methods to perform the dynamic store discovery, is merged.

This PR implements enclave methods described in the design doc's "Option 1: 'On The Fly' Authentication" [section](https://docs.google.com/document/d/1-ju5efGK6vvuYjxj1pKhfFcpgGlornuxZYvOC-nsbjs/edit#bookmark=id.d11a9antj10l). The one slight difference is the shape of the `Map` and that it's stored in the ake enclave rather than the Fog View enclave. (It went from being `Map<LoadBalancerHostPort, Map<StoreHostPort, Ready<_>>` to `Mutex<LruCache<ResponderId, Ready<Aes256Gcm>>>`.

### Future Work
* FogViewRouterService uses the `reencrypt_client_message_for_backends` method to create query requests for each Fog View Store
* FogViewRouterService uses the auth enclave methods to authenticate and connect to Fog View Stores via the "dynamic discovery" method.

### DIFF
--- a/crypto/ake/enclave/src/lib.rs
+++ b/crypto/ake/enclave/src/lib.rs
@@ -182,6 +182,10 @@ impl<EI: EnclaveIdentity> AkeEnclaveState<EI> {
     }
 
     /// Constructs a ClientAuthRequest to be sent to an enclave backend.
+    ///
+    /// Differs from peer_init in that this enclave does not establish a peer
+    /// connection to the enclave described by `backend_id`. Rather, this
+    /// enclave serves as a client to this other backend enclave.
     pub fn backend_init(&self, backend_id: ResponderId) -> Result<ClientAuthRequest> {
         let mut csprng = McRng::default();
 
@@ -195,6 +199,10 @@ impl<EI: EnclaveIdentity> AkeEnclaveState<EI> {
     }
 
     /// Connect to an enclave backend as a client.
+    ///
+    /// This establishes the client to backend enclave connection, see
+    /// `backend_init` for more details on how this differs from a peer
+    /// connection.
     pub fn backend_connect(
         &self,
         backend_id: ResponderId,
@@ -429,6 +437,36 @@ impl<EI: EnclaveIdentity> AkeEnclaveState<EI> {
         })
     }
 
+    /// Transforms an incoming client message, i.e. a message sent from a client
+    /// to the current enclave, into a list of outbound  messages for
+    /// other enclaves that serve as backends to the current enclave.
+    ///
+    ///                               / --> Backend Enclave 1
+    ///   Client -> Current Enclave ---> Backend Enclave 2
+    ///                              \ --> Backend Enclave N
+    pub fn reencrypt_client_message_for_backends(
+        &self,
+        incoming_client_message: EnclaveMessage<ClientSession>,
+    ) -> Result<Vec<EnclaveMessage<ClientSession>>> {
+        let client_query_bytes: Vec<u8> = self.client_decrypt(incoming_client_message.clone())?;
+        let mut backends = self.backends.lock()?;
+        let backend_messages = backends
+            .iter_mut()
+            .map(|(_, encryptor)| {
+                let aad = incoming_client_message.aad.clone();
+                let data = encryptor.encrypt(&aad, &client_query_bytes)?;
+                let channel_id = incoming_client_message.channel_id.clone();
+                Ok(EnclaveMessage {
+                    aad,
+                    channel_id,
+                    data,
+                })
+            })
+            .collect::<Result<_>>()?;
+
+        Ok(backend_messages)
+    }
+
     //
     // IAS related
     //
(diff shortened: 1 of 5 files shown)


crypto/ake/enclave/src/lib.rs
```diff
@@ -165,6 +181,45 @@ impl<EI: EnclaveIdentity> AkeEnclaveState<EI> {
...
+
+        let init_input = ClientInitiate::<X25519, Aes256Gcm, Sha512>::default();
+        let (initiator, auth_request_output) = initiator.try_next(&mut csprng, init_input)?;
+        self.backend_auth_pending.lock()?.put(backend_id, initiator);
+        let client_auth_request_data: Vec<u8> = auth_request_output.into();
+        Ok(client_auth_request_data.into())
+    }
+
+    /// Connect to an enclave backend as a client.
+    pub fn backend_connect(
```
remoun, line 198: Should we add a `backend_close` method a la `peer_close` and `client_close`?

remoun (commented):
Looking good!

crypto/ake/enclave/src/lib.rs
```diff
@@ -507,6 +562,35 @@ impl<EI: EnclaveIdentity> AkeEnclaveState<EI> {
...
+    ///                               /
+    ///   Client -> Current Enclave ---> Backend Enclave 2
+    ///                              \
+    ///                               \ --> Backend Enclave N
+    pub fn reencrypt_client_message_for_backends(
+        &self,
+        incoming_client_message: EnclaveMessage<ClientSession>,
+    ) -> Result<Vec<EnclaveMessage<ClientSession>>> {
+        let mut backends = self.backends.lock()?;
+        let client_query_bytes: Vec<u8> = self.client_decrypt(incoming_client_message.clone())?;
```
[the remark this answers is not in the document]
samdealy: Done.

fog/view/enclave/api/src/lib.rs
```diff
@@ -123,9 +130,22 @@ pub trait ViewEnclaveApi: ReportableEnclave {
     /// Accept an inbound authentication request
     fn client_accept(&self, req: ClientAuthRequest) -> Result<(ClientAuthResponse, ClientSession)>;
 
+    /// Complete the connection to a Fog View Store that our accepted our
```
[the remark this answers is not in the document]
samdealy: Done.

crypto/ake/enclave/src/lib.rs
```diff
@@ -507,6 +562,35 @@ impl<EI: EnclaveIdentity> AkeEnclaveState<EI> {
...
     }
 
+    /// Transforms an incoming client message, i.e. a message sent from a client
+    /// to the current enclave, into a list of outbound  messages for
+    /// other enclaves that serve as backends to the current enclave.
+    ///                                 --> Backend Enclave 1
+    ///                               /
+    ///   Client -> Current Enclave ---> Backend Enclave 2
+    ///                              \
+    ///                               \ --> Backend Enclave N
```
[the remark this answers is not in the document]
samdealy: Done.

crypto/ake/enclave/src/lib.rs
```diff
@@ -507,6 +562,35 @@ impl<EI: EnclaveIdentity> AkeEnclaveState<EI> {
         Ok(())
     }
 
+    /// Transforms an incoming client message, i.e. a message sent from a client
```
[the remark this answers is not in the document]
samdealy: Done.

crypto/ake/enclave/src/lib.rs
samdealy: IIUC, I'm not sure those methods are really used: https://github.com/mobilecoinfoundation/mobilecoin/search?q=client_close. In this case, I also can't see how it would be used effectively- I guess if a Fog View Store goes down, then we would maybe able to call `backend_close`? I'm not entirely sure, but perhaps we can address in a follow-up PR if we find that we need this method because as of right now I don't see a use case. 

@jcape any thoughts?

fog/view/enclave/api/src/lib.rs
```diff
@@ -123,6 +129,19 @@ pub trait ViewEnclaveApi: ReportableEnclave {
     /// Accept an inbound authentication request
     fn client_accept(&self, req: ClientAuthRequest) -> Result<(ClientAuthResponse, ClientSession)>;
 
+    /// Begin a connection to a Fog View Store. The enclave calling this method
```
remoun, line 132: Nit: Keep `client_accept` and `client_close` adjacent.

fog/view/enclave/src/lib.rs
```diff
@@ -155,6 +155,25 @@ impl ViewEnclaveApi for SgxViewEnclave {
         mc_util_serial::deserialize(&outbuf[..])?
     }
 
+    fn view_store_init(&self, view_store_id: ResponderId) -> Result<ClientAuthRequest> {
```
remoun, line 158: Nit: Keep `client_accept` and `client_close` next to each other.

remoun (approved):

fog/view/enclave/api/src/lib.rs
samdealy: Done

fog/view/enclave/src/lib.rs
samdealy: Done.

crypto/ake/enclave/src/lib.rs
jcape: I think it's fine to leave `backend_close` off the list. If we end up with a bunch of stale sessions in the enclave we can use Eran's LRU to keep track of them, but practically speaking the only reason a session would need to be closed is if the store behind it dies.

samdealy: Sounds good.

eranrund (approved):

PR #2220, opened 2022-07-02
Merge latest changes into `feature/block-streaming`

remoun:
Merge from `block-metadata-catchup`, to include changes from #2216 as well as #2205 and #2052.


ledger/streaming/api/src/error.rs
```diff
@@ -36,7 +36,7 @@ pub enum Error {
     BlockValidation(String),
 
     /// SCP Consensus Behind: last externalized block {0} - highest block {1}
-    ConsensusBlocked(SlotIndex, SlotIndex),
+    ConsensusBlocked(BlockIndex, BlockIndex),
```
nick-mobilecoin, line 39: Is there an issue to come back and disambiguate _Slot_ versus _Block_?
Looking at the usage of `SlotIndex` and `consensus/scp/src/slot.rx` there is:
```rust
//! A unit of time during which the nodes agree on transactions.
//!
//! The transactions validated in this slot determine the values to include in
//! the next block appended to the ledger.
```
Which seems to imply that Slot is a unit of the consensus protocol, looking at `scp_validators.rs` where the `ConsensusBlocked` error is created there is:
```rust
                    return future::ready(Some(Ready::Ready(Err(Error::ConsensusBlocked(
                        scp_state.highest_externalized_slot(),
                        scp_state.highest_block_received(),
```
The `highest_externalized_slot()` is raising some concerns as to what the type should really be.
is this supposed to be `ConsensusBlocked(SlotIndex, BlockIndex)` or is `SlotIndex` really an alias for `BlockIndex`?

ledger/streaming/client/src/http_fetcher.rs
```diff
@@ -1,16 +1,18 @@
 // Copyright (c) 2018-2022 The MobileCoin Foundation
 
-//! A [Fetcher] that downloads [BlockData] from the given URI.
+//! A [Fetcher] that downloads [ArchiveBlocks] from the given URI.
```
nick-mobilecoin, line 3: Since this was updated should the doc comment on the `HttpBlockFetcher` be updated as well?

Also is `URI` correct here? It looks like `URL` is used in `new()` and `from_url()`

ledger/streaming/client/src/scp_validator.rs
```diff
@@ -180,17 +182,18 @@ impl<ID: GenericNodeId + Send + Clone> SCPValidationState<ID> {
...
-                    QuorumSetMember::Node(node_id) => {
+                    Some(QuorumSetMember::Node(node_id)) => {
                         if let Some(block_data) = tracked_blocks.get(node_id) {
                             // Record a vote for the block externalized
                             ballot_map
                                 .entry(&block_data.block().id)
-                                .and_modify(|vec| vec.push(member))
-                                .or_insert_with(|| vec![member]);
+                                .or_default()
+                                .push(member.unwrap());
```
nick-mobilecoin, line 196: While I understand that `member.unwrap()` shouldn't fail here and there are already some `unwrap()` calls in this module, it still feels off to have an `unwrap()`.
Perhaps something like this:
```suggestion
                    Some(node @ QuorumSetMember::Node(node_id)) => {
                        if let Some(block_data) = tracked_blocks.get(node_id) {
                            // Record a vote for the block externalized
                            ballot_map
                                .entry(&block_data.block().id)
                                .or_default()
                                .push(node);
```

ledger/streaming/client/src/scp_validator.rs
```diff
@@ -206,13 +209,13 @@ impl<ID: GenericNodeId + Send + Clone> SCPValidationState<ID> {
...
-                    QuorumSetMember::InnerSet(qs) => {
+                    Some(QuorumSetMember::InnerSet(qs)) => {
                         // If internal slice reached quorum, record the block voted for
                         if let Some(block_id) = self.find_quorum(qs, index) {
                             ballot_map
                                 .entry(block_id)
-                                .and_modify(|vec| vec.push(member))
-                                .or_insert_with(|| vec![member]);
+                                .or_default()
+                                .push(member.unwrap());
```
nick-mobilecoin, line 218: Similar to above, consider omitting the `unwrap()` in favor of:
```suggestion
                    Some(inner_set @ QuorumSetMember::InnerSet(qs)) => {
                        // If internal slice reached quorum, record the block voted for
                        if let Some(block_id) = self.find_quorum(qs, index) {
                            ballot_map
                                .entry(block_id)
                                .or_default()
                                .push(inner_set);
```

nick-mobilecoin (changes requested):
The blocker is the ambiguity between `SlotIndex` and `BlockIndex`.  Need some clarity to approve this change, and perhaps a follow on issue to further address the ambiguity.

ledger/streaming/api/src/error.rs
remoun: ... Except `mc_consensus_scp_types::SlotIndex` exists, so I'll switch back to that and add a TODO

nick-mobilecoin (approved):

PR #2236, opened 2022-07-07
Publish protobufs to buf.build

### LINKED ISSUE #1704: Universal Protobuf Repository
We should create a new repository which is released on it's own cadence, and contains all the protobufs we care about.

While we're doing this we should create a better directory structure to more easily allow us to bifurcate what exists in `mc_api` into more specific crates (i.e. we should be able to create an `mc_crypto_keys_api` crate that contains the `RistrettoPublic` message).

For now this issue is mainly intended to be a place for discussion, which we may want to put into an MCIP.

remoun:
Add configs to host our protobufs in https://buf.build/mobilecoin, and a helper script to update all of them.

Mostly resolves #1704 -- the remaining questions are around the process aspect, e.g. when to publish updates


jgreat:
What are our triggers to push updates? Do we need to add a release step to publish new proto updates to buf.build?

api/proto/buf.yaml
```diff
@@ -0,0 +1,8 @@
+version: v1
+name: buf.build/mobilecoin/api
+lint:
+  use:
+    - DEFAULT
```
nick-mobilecoin, line 5: nit: most likely should be a follow on if any change is desired.

Guessing these files came form `buf mod init` and then stripping out the comments.
having lint here implies that linting is meant to happen, but it looks like all the proto files have lint warnings.  While removing the lint key doesn't disable linting, perhaps it would indicate to maintainers that these don't fully support linting yet.

nick-mobilecoin (approved):

fog/api/proto/buf.yaml
```diff
@@ -0,0 +1,11 @@
+version: v1
+name: buf.build/mobilecoin/fog-api
+deps:
+  - buf.build/mobilecoin/attest-api
+  - buf.build/mobilecoin/api
```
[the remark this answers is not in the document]
remoun: `googleapis` does not include the Well-Known Types so I removed it.

WKT are a standard part of proto3. The closest thing I could find in the buf docs is "The file `google/protobuf/empty.proto` is part of the [Well-Known Types](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf), and can be directly included in any Protobuf schema."  ([link](https://docs.buf.build/lint/configuration#rpc_allow_))

tools/buf_push.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+# Copyright (c) 2018-2022 The MobileCoin Foundation
```
garbageslam, line 3: can you add some example usage for this script?

garbageslam:
> Good question. Invoking the update script as a release step SGTM.

I think this won't work -- the way our release process works, we release servers and SDKs concurrently. So the proto have to be published before SDK work can begin on the next release.

I would suggest we should push new protos whenever we tag a commit for SDK teams to work on.

garbageslam (changes requested):
scripts that take arguments should have example usage somewhere documented

sugargoat (approved):
Cool! Thanks, @remoun !

garbageslam (approved):
thank you

PR #2240, opened 2022-07-11
Basic AVR config file information

pouneh:
Plumbing to read in an AVR config file (ie. the historical AVR Bootstrap as described in MCIP43)

### Motivation
BlockMetadata for Block V3 and further will contain AVRs. However, we still want to be able to provide AVRs for previous block versions that don't support it. Per MCIP 43, the solution we will use is to create an AVR boostrap file that will provide this information for older blocks

### Future Work
- Validator for AVR
- Using Validator for AVR in config::try_from_file to better live RAII


blockchain/validators/src/metadata/avr/config.rs
```diff
@@ -0,0 +1,59 @@
...
+#[derive(Clone, Debug, Deserialize, Eq, Ord, PartialEq, PartialOrd, Serialize)]
+pub struct AvrHistoryRecord {
+    /// Uri of the consensus node
+    pub responder_id: ResponderId,
+
+    /// Block the AVR Report for the signing key becomes valid
+    pub first_block_index: BlockIndex,
+
+    /// Final block the AVR Report for the signing key is valid
+    pub last_block_index: BlockIndex,
```
remoun, line 31: Let's make this optional
```suggestion
    pub last_block_index: Option<BlockIndex>,
```

blockchain/validators/src/metadata/avr/config.rs
```diff
@@ -0,0 +1,59 @@
...
+    pub avr: Option<VerificationReport>,
+}
+
+impl AvrHistoryConfig {
+    /// Load the [AvrHistoryConfig] from a .json or .toml file
+    ///
+    /// # Arguments
+    ///
+    /// * `path` - Path to the configuration file containing
+    /// the history of AVRs generated MobileCoin consensus node
```
remoun, line 44: Nit: Missing word
```suggestion
    /// the history of AVRs generated by MobileCoin consensus node
```

blockchain/validators/src/metadata/avr/config.rs
```diff
@@ -0,0 +1,59 @@
...
+    /// enclaves
+    pub fn try_from_file(path: impl AsRef<Path>) -> Result<AvrHistoryConfig, ParseError> {
```
remoun, line 46: How about we name this method `load`? I think the `try_` prefix is useful when there's a variant that panics on failure, which isn't the case here.

remoun (commented):
Structure looks good so far; please add tests with sample inputs and outputs.

blockchain/validators/src/error.rs
```diff
@@ -58,6 +58,9 @@ pub enum ParseError {
 
     /// Failed to parse JSON: {0}
     Json(String),
+
+    /// Failed to parse supported type
+    UnsuportedFileFormat,
```
nick-mobilecoin, line 63: nit: spelling
```suggestion
    UnsupportedFileFormat,
```

blockchain/validators/src/metadata/avr/config.rs
```diff
@@ -0,0 +1,153 @@
...
+
+impl AvrHistoryConfig {
+    /// Load the [AvrHistoryConfig] from a .json or .toml file
+    ///
+    /// # Arguments
+    ///
+    /// * `path` - Path to the configuration file containing
+    /// the history of AVRs generated by MobileCoin consensus node
+    /// enclaves
+    pub fn load(path: impl AsRef<Path>) -> Result<AvrHistoryConfig, ParseError> {
```
nick-mobilecoin, line 49: nit: one can use `Self` to minimize type name changes
```suggestion
    pub fn load(path: impl AsRef<Path>) -> Result<Self, ParseError> {
```

blockchain/validators/src/metadata/avr/config.rs
```diff
@@ -0,0 +1,153 @@
...
+        let data = fs::read_to_string(path)?;
+
+        if let Ok(mut config) = serde_json::from_str(&data): Result<AvrHistoryConfig, _> {
+            config.node.sort();
+            Ok(config)
+        } else if let Ok(mut config) = toml::from_str(&data): Result<AvrHistoryConfig, _> {
+            config.node.sort();
+            Ok(config)
+        } else {
+            Err(ParseError::UnsuportedFileFormat)
```
nick-mobilecoin, line 59: This error lacks symmetry with `Config::load()` from `blockchain/validators/src/metadata/key_range/config.rs`.
This one returns a new unsupported file format, while `key_range` returns `UnrecognizedExtension`

blockchain/validators/src/metadata/avr/config.rs
```diff
@@ -0,0 +1,153 @@
...
+    #[serde_as(as = "hex::Hex")]
+    pub sig: VerificationSignature,
+
+    /// Attestation Report Signing Certificate Chain, as an array of
+    /// DER-formatted bytes, from the X-IASReport-Signing-Certificate HTTP
+    /// header.
+    #[serde_as(as = "Vec<hex::Hex>")]
+    pub chain: Vec<Vec<u8>>,
+
+    /// The raw report body JSON, as a byte sequence
```
nick-mobilecoin, line 80: nit: This may have been copied form elsewhere, but `byte sequence` seems to contradict the underlying `String` type.
```suggestion
    /// The raw report body JSON
```

blockchain/validators/src/metadata/avr/config.rs
```diff
@@ -0,0 +1,153 @@
...
+        // Get a manually constructed AVR history config to act as control data
+
+        let control_avr_history = get_avr_history_config();
+
+        // Serialize the config to JSON & TOML
+        let toml_str = toml::to_string_pretty(&control_avr_history).unwrap();
+        let json_str = serde_json::to_string_pretty(&control_avr_history).unwrap();
+
+        let history_from_toml: AvrHistoryConfig = toml::from_str(toml_str.as_str()).unwrap();
+        let history_from_json: AvrHistoryConfig = serde_json::from_str(json_str.as_str()).unwrap();
```
nick-mobilecoin, line 125: nit: can use `&` instead of `as_str()`
```suggestion
        let history_from_toml: AvrHistoryConfig = toml::from_str(&toml_str).unwrap();
        let history_from_json: AvrHistoryConfig = serde_json::from_str(&json_str).unwrap();
```

blockchain/validators/src/metadata/avr/config.rs
```diff
@@ -0,0 +1,153 @@
...
+
+        // Assert that deserialization from JSON and TOML is the same as the original
+        // config
+        assert_eq!(control_avr_history, history_from_toml);
+        assert_eq!(control_avr_history, history_from_json);
+    }
+
+    #[test]
+    fn test_avr_history_load_from_disk() {
```
nick-mobilecoin, line 134: suggest adding a test case for the error case from `load()`

blockchain/validators/src/metadata/avr/config.rs
```diff
@@ -0,0 +1,153 @@
...
+    };
+    use serde_json;
+    use std::fs;
+    use tempfile::TempDir;
+
+    #[test]
+    fn test_avr_history_serialization_roundtrip_works() {
+        // Get a manually constructed AVR history config to act as control data
+
+        let control_avr_history = get_avr_history_config();
```
nick-mobilecoin, line 118: nit: comment seems to float away from the thing it's commenting
```suggestion
        // Get a manually constructed AVR history config to act as control data
        let control_avr_history = get_avr_history_config();
```

blockchain/validators/src/test_utils.rs
```diff
@@ -26,3 +26,131 @@ pub fn make_metadata(seed: u64) -> BlockMetadata {
     )
     .expect("BlockMetadata::from_contents_and_keypair")
 }
+
+use crate::metadata::avr::{AvrHistoryConfig, AvrHistoryRecord};
+use mc_common::ResponderId;
+use std::str::FromStr;
```
nick-mobilecoin, line 32: suggest moving up to the top of the file next to the other use statements

blockchain/validators/src/test_utils.rs
```diff
@@ -26,3 +26,131 @@ pub fn make_metadata(seed: u64) -> BlockMetadata {
...
+
+/// Get first two actual Intel Attestation Verification Reports for
+/// node1.prod.mobilecoinww.com in order to build control data
+/// for tests
+pub fn get_ias_reports() -> (VerificationReport, VerificationReport) {
```
nick-mobilecoin, line 37: does this need to be `pub`?
```suggestion
fn get_ias_reports() -> (VerificationReport, VerificationReport) {
```

blockchain/validators/src/test_utils.rs
```diff
@@ -26,3 +26,131 @@ pub fn make_metadata(seed: u64) -> BlockMetadata {
...
+        responder_id: responder_id_1.clone(),
+        first_block_index: 0,
+        last_block_index: Some(480),
+        avr: None,
+    };
+    let rec2 = AvrHistoryRecord {
+        responder_id: responder_id_1.clone(),
+        first_block_index: 481,
+        last_block_index: Some(10399),
+        avr: Some(avr1),
```
nick-mobilecoin, line 71: It looks like there is a a bit of dependency between `get_ias_reports()`, `get_avr_history_config()`, `SAMPLE_AVR_HISTORY_TOML` and `SAMPLE_AVR_HISTORY_JSON`.

Suggest creating a containing struct to group this logic, might consider having functions like
```rust
SampleAvrHistory::as_json();
SampleAvrHistory::as_toml();
SampleAvrHistory::as_config();
```
(not sold on the "as_x" names)

nick-mobilecoin (changes requested):
Mostly nits, the only blocker is in the error handling of failure to load being inconsistent across the API.

remoun (changes requested):

blockchain/validators/src/metadata/key_range/config.rs
```diff
@@ -38,13 +37,17 @@ impl Config {
...
+        if let Ok(mut config) = serde_json::from_slice(&bytes): Result<Self, _> {
+            config.base_path = path.parent().map(Into::into);
+            Ok(config)
+        } else if let Ok(mut config) = toml::from_slice(&bytes): Result<Self, _> {
+            config.base_path = path.parent().map(Into::into);
+            Ok(config)
+        } else {
+            Err(ParseError::UnsupportedFileFormat(
+                path.to_string_lossy().into_owned(),
+            ))
```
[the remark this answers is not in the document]
remoun: This approach would fail with only an "Unsupported file" error if we get an otherwise valid TOML or JSON, e.g. trailing comma in JSON, mismatched quotes or other delimiter in TOML or JSON

blockchain/validators/src/test_utils.rs
```diff
@@ -26,3 +26,141 @@ pub fn make_metadata(seed: u64) -> BlockMetadata {
...
+        ],
+        http_body: String::from("{\"nonce\":\"0edca460aa5c7c9a952526ee904111f6\",\"id\":\"239437120075880885026599322830514941911\",\"timestamp\":\"2021-03-08T16:32:15.337612\",\"version\":4,\"epidPseudonym\":\"gKH0dexEpYfuyaGgaKKWmH4VJ8r0L3af1W//p6ya+WaN9BAlSW1Gj3NOWvrQIEAyLCof3fwS9pkLnZrYk3CXQFlpkTmn56hip16QlU876R7o0WpMfIpTY0VVyFj0kM3pAoSIeurvSGDNLrkq1qdKin3mPC2DwzsL1TdEs2ECbkI=\",\"advisoryURL\":\"https://security-center.intel.com\",\"advisoryIDs\":[\"INTEL-SA-00334\"],\"isvEnclaveQuoteStatus\":\"SW_HARDENING_NEEDED\",\"isvEnclaveQuoteBody\":\"AgABAP4LAAALAAoAAAAAAL3lg34HPOkq5u/y0l94xuMAAAAAAAAAAAAAAAAAAAAAEREDBf+ABgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABQAAAAAAAAAHAAAAAAAAAOZts4uKQ6M/bBYQ0zWjYZY7srMeBWrw3AqJWsbIV8q5AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAsGlYcSrZMvAS/pEXN977Zsq1vawTTjTE382IrKf2zDgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAEAAQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAACCQbFoCTirZ6Uvkspay6i0N3AKG+RG15miHkmNrloKRafPBVg+aovhB0Yxr5DBnb+p0GM2A6Wmm1IOXiOma5os\"}"),
```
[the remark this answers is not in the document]
remoun: I personally prefer `into()`, but it's a minor preference :-)

remoun (approved):

PR #2250, opened 2022-07-12
Collate shard response in constant time (obliviously)

samdealy:
### Motivation
This PR uses the oblivious programming methods implemented in #2252 to obliviously collate shard responses for  Fog Router.

PR #2252, opened 2022-07-12
Create helper methods to obliviously collate shard responses

samdealy:
### Motivation

#2189 implemented the core logic for routing queries from the Fog View Router to the Fog Shards and authenticating with Fog View Stores. That PR deferred on collating the shards' resposnes into one response for the client.

This PR sets up the next PR to implement this collation. Specifically, this PR adds an oblivious, constant time method that takes all the `TxOutSearchResult` objects returned by all shards and uses them to fill out the `TxOutSearchResults` that go into the client's `QueryResponse`.

Note that I made the assumption that all ciphertexts will have a length of `237`, and there's an `assert` statement plus a test that enforces this. We may consider making the client ciphertext length `255` and then setting one byte that describes how much less than 255 the portion of the ciphertext that corresponds to the payload is. I could use feedback on this point in particular so please comment your thoughts on these options below.

### Future Work
- Use these methods to do the response collation in the Fog View Router's business logic.


fog/view/enclave/impl/src/oblivious_utils.rs
```diff
@@ -0,0 +1,508 @@
...
+    shard_tx_out_search_result: &TxOutSearchResult,
+) {
+    let should_overwrite_tx_out_search_result = should_overwrite_tx_out_search_result(
+        client_tx_out_search_result,
+        shard_tx_out_search_result,
+    );
+    let shard_ciphertext_length = shard_tx_out_search_result.ciphertext.len();
+    assert_eq!(
+        shard_ciphertext_length, CIPHERTEXT_LENGTH,
+        "Shard ciphertext has unexpected length"
```
garbageslam, line 62: i think this isn't going to be true in prod, because we've already changed the ciphertext length in prod several times:

* we added memos
* we added token ids

so there are currently shorter ciphertexts than this in the production database

So we have to figure out a way to factor out this assumption

samdealy: As discussed offline, I modified the client ciphertext as follows: the first byte is the `delta` between 255 and the shard ciphertext length, and the rest of the bits are filled with the shard ciphertext data until it runs out (and the rest is zero padded).

fog/view/enclave/impl/src/lib.rs
```diff
@@ -2,11 +2,13 @@
 
 //! View Enclave Implementation
 
-#![no_std]
+#![cfg_attr(not(test), no_std)]
```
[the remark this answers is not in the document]
samdealy: Used `extern crate` under `mod tests`. Done.

samdealy:
* feature/fog-view-router
  * **PR #2252** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2252" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2252?utm_source=stack-comment).

NotGyro (approved):
This looks good to me. You should take my review with a grain of salt, of course, because I do not have much experience in writing constant-time algorithms, but I did not see anything here that would leak information by the timing channel, as long as aligned-cmov's guarantees remain true (i.e. LLVM doesn't try to optimize this until it's no longer constant-time). I also like how thorough these tests are.

fog/view/enclave/impl/src/oblivious_utils.rs
```diff
@@ -0,0 +1,628 @@
...
+        client_tx_out_search_result_code.ct_eq(&(TxOutSearchResultCode::NotFound as u32));
+
+    let shard_code_is_found: Choice =
+        shard_tx_out_search_result_code.ct_eq(&(TxOutSearchResultCode::Found as u32));
+
+    let shard_code_is_retryable_error =
+        is_code_retryable_error(shard_tx_out_search_result.result_code);
+    let shard_code_is_bad_search_key =
+        shard_tx_out_search_result_code.ct_eq(&(TxOutSearchResultCode::BadSearchKey as u32));
+
```
garbageslam, line 107: it would be nice to have a comment just like trying to explain at a high level what we want to happen:

* we make the same query to several shards and get several responses
* we want to prioritize "found" responses, and then beyond that, the "bad search key", which means the argument was invalid.
* after that the other two errors are like "retriable" errors that the client will probably just wait a while and retry
* "not found" response gets overwritten by  any other response

fog/view/enclave/impl/src/oblivious_utils.rs
```diff
@@ -0,0 +1,628 @@
...
+    client_tx_out_search_result: &mut TxOutSearchResult,
+    shard_tx_out_search_result: &TxOutSearchResult,
+) {
+    let should_overwrite_tx_out_search_result = should_overwrite_tx_out_search_result(
+        client_tx_out_search_result,
+        shard_tx_out_search_result,
+    );
+    let shard_ciphertext_length = shard_tx_out_search_result.ciphertext.len();
+
+    let shard_cipher_text_length_delta = (CLIENT_CIPHERTEXT_LENGTH - shard_ciphertext_length) as u8;
```
garbageslam, line 60: it might be good to change this from `as u8` to like `u8::try_from`, and expect with e.g. `"shard_ciphertext_length exceeds bounds"` just to help debugging if this happens

garbageslam (approved):
This all looks correct to me, thank you!

I'm going to make a ticket to memorialize some follow on items

fog/view/enclave/impl/src/oblivious_utils.rs
samdealy: Done.

fog/view/enclave/impl/src/oblivious_utils.rs
samdealy: done.

PR #2257, opened 2022-07-13
Fix warning in CI

remoun:
Fixes ```warning: please specify `--format-version` flag explicitly to avoid compatibility problems``` in CI


nick-mobilecoin (approved):

jgreat (approved):
LGTM

This is a very silly warning since the documentation says `Currently 1 is the only possible value.` :frowning_face: 

https://doc.rust-lang.org/cargo/commands/cargo-metadata.html

PR #2258, opened 2022-07-13
Replace `should_panic` with explicit assertions

remoun:


nick-mobilecoin (approved):
LGTM, love the explicit asserts!

jcape (approved):

PR #2275, opened 2022-07-19
Branch and prep 2.0.0 release

jcape:
- Re-introduce SGX 2.17
- Update Changelog
- Bump all versions to 2.0.0, make long-form deps parsimonious

### Motivation

This is the start of the 2.0.0 branch, with desired code changes on top of the 1.2.2. branch.

### Future Work
- [x] Test on SGX 2.17 machine


jgreat (approved):
The Docker, CD versions look good.

attest/verifier/src/status.rs
```diff
@@ -451,6 +665,26 @@ mod test {
...
+        let verifier = MrEnclaveVerifier {
+            mr_enclave: MrEnclave::from(&MR_ENCLAVE),
+            config_ids: vec!["INTEL-SA-00334".to_owned()],
+            sw_ids: vec!["INTEL-SA-00334".to_owned()],
+        };
+
+        let report = VerificationReport {
+            sig: Default::default(),
+            chain: vec![],
+            http_body: IAS_CONFIG_SW.trim().to_owned(),
```
nick-mobilecoin, line 681: Based on the function name I'm guessing this is meant to be testing for a non excluded advisory for both config and software
```suggestion
    /// Ensure a CONFIGURATION_AND_SW_HARDENING_NEEDED result with the expected
    /// MRENCLAVE but an insufficient sw and config advisory allow-listing 
    /// fails.
    #[test]
    fn mrenclave_multi_config_sw_fail_short() {
        let verifier = MrEnclaveVerifier {
            mr_enclave: MrEnclave::from(&MR_ENCLAVE),
            config_ids: vec!["INTEL-SA-00334".to_owned()],
            sw_ids: vec!["INTEL-SA-00334".to_owned()],
        };

        let report = VerificationReport {
            sig: Default::default(),
            chain: vec![],
            http_body: IAS_CONFIG_SW_334_615.trim().to_owned(),
```

attest/verifier/src/status.rs
```diff
@@ -647,19 +927,111 @@ mod test {
...
+            product_id: 0,
+            minimum_svn: 0,
+            config_ids: vec!["INTEL-SA-00334".to_owned(), "INTEL-SA-00615".to_owned()],
+            sw_ids: vec![],
+        };
+
+        let report = VerificationReport {
+            sig: Default::default(),
+            chain: vec![],
+            http_body: IAS_CONFIG_SW.trim().to_owned(),
```
nick-mobilecoin, line 949:
```suggestion
            http_body: IAS_CONFIG_SW_334_615.trim().to_owned(),
```

attest/verifier/src/status.rs
```diff
@@ -647,19 +927,111 @@ mod test {
...
+        };
+
+        let data = VerificationReportData::try_from(&report).expect("Could not parse IAS result");
+        assert!(!verifier.verify(&data))
+    }
+
```
nick-mobilecoin, line 978: I'm not sure this test is needed, based on the setup this seems equivalent to the testcase on line 1007.  Perhaps this was meant to test having a config advisory but the advisory in the report is different.

nick-mobilecoin (changes requested):
Main blocker `android-bindings/lib-wrapper/android-bindings/publish.gradle` needs to get its version updated.
The test comments aren't show stoppers and if this is a pull over of the changes from another branch I could understand a follow on PR to correct the tests.

attest/verifier/src/status.rs
jcape: This is testing what happens when config is empty, the other is testing when config has one element but needs two.

nick-mobilecoin (approved):

PR #2280, opened 2022-07-20
working on no_std for hardware support

ryankurte:
propagating feature gates for `alloc` and `serde` to allow `no_std` use of `ring-signer`... with a few API alterations to avoid the need for `alloc` where possible.

cc. @garbageslam @jcape


crypto/keys/src/traits.rs
```diff
@@ -48,60 +70,89 @@ pub trait DistinguishedEncoding: Sized {
...
+
+/// A trait indicating that a fingerprint can be generated for an object.
 pub trait Fingerprintable {
-    fn fingerprint<D: Digest>(&self) -> Result<String, KeyError>;
+    /// Generate fingerprint
+    fn fingerprint<D: Digest>(&self) -> Fingerprint<D>;
+}
+
+/// A fingerprint object, generic over digest output size
+pub struct Fingerprint<D: digest::OutputSizeUser> {
```
ryankurte, line 97: swapping to a fingerprint object means we can easily impl `Debug` and `Display` without requiring `alloc`

crypto/keys/src/x25519.rs
```diff
@@ -45,22 +50,25 @@ impl AsRef<[u8]> for X25519Secret {
...
 /// the secret itself
-impl Debug for X25519Secret {
-    fn fmt(&self, f: &mut Formatter) -> FmtResult {
+impl core::fmt::Debug for X25519Secret {
+    fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
         let mut hasher = Sha256::new();
         hasher.update(self.as_ref());
         let hash_results = hasher.finalize();
 
-        let output = hex::decode(&hash_results).map_err(|_e| FmtError)?;
```
ryankurte, line 54: it looks to me like this was backwards..? assuming the intent was to hex _encode_ the hash

crypto/keys/src/traits.rs
```diff
@@ -2,19 +2,41 @@
...
+#[cfg(feature = "alloc")]
+use alloc::{vec::Vec};
+
+#[cfg(feature = "serde")]
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
 
+
+/// Marker trait for serialization when `serde` feature is enabled
+#[cfg(feature = "serde")]
+pub trait MaybeSerde: DeserializeOwned + Serialize {}
```
garbageslam, line 22: this is very clever, kudos

ryankurte: you may be able to tell i have experienced this problem before 🤣

ryankurte:
per discussion i'll split this up into more manageable chunks _^

PR #2281, opened 2022-07-20
Make fog sample paykit support (at least, scan) historical change subaddress

garbageslam:
This should fix a bunch of warnings appearing in the test client when
running on testnet, because we ran it for months before the change subaddres
was changed from 1 to u64::MAX - 1.


jgreat (approved):
LGTM Thanks!

PR #2288, opened 2022-07-21
Add FeeMap to MockBlockchainConnection

BrianCorbin:
Adding the fee map to support MockBlockchainConnections having fees for more than just the MOB token, specified by the test setup.

### DIFF
--- a/connection/test-utils/src/blockchain.rs
+++ b/connection/test-utils/src/blockchain.rs
@@ -39,18 +39,13 @@ pub struct MockBlockchainConnection<L: Ledger + Sync> {
 }
 
 impl<L: Ledger + Sync> MockBlockchainConnection<L> {
-    pub fn new(
-        uri: ConsensusClientUri,
-        ledger: L,
-        latency_millis: u64,
-        fee_map: Option<FeeMap>,
-    ) -> Self {
+    pub fn new(uri: ConsensusClientUri, ledger: L, latency_millis: u64, fee_map: FeeMap) -> Self {
         Self {
             uri,
             ledger,
             latency_millis,
             proposed_txs: Vec::new(),
-            fee_map: fee_map.unwrap_or_default(),
+            fee_map,
         }
     }
 }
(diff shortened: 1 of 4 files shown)


connection/test-utils/src/blockchain.rs
```diff
@@ -33,15 +33,24 @@ pub struct MockBlockchainConnection<L: Ledger + Sync> {
...
+        ledger: L,
+        latency_millis: u64,
+        fee_map: Option<FeeMap>,
+    ) -> Self {
         Self {
             uri,
             ledger,
             latency_millis,
             proposed_txs: Vec::new(),
+            fee_map: fee_map.unwrap_or_default(),
```
remoun, line 53: Since `FeeMap` implements `Default`, it seems fine to require it as a non-optional argument, and callers can pass `FeeMap::default()`
```suggestion
        fee_map: FeeMap,
    ) -> Self {
        Self {
            uri,
            ledger,
            latency_millis,
            proposed_txs: Vec::new(),
            fee_map,
```

connection/test-utils/src/blockchain.rs
```diff
@@ -118,7 +127,7 @@ impl<L: Ledger + Sync> BlockchainConnection for MockBlockchainConnection<L> {
     fn fetch_block_info(&mut self) -> ConnectionResult<BlockInfo> {
         Ok(BlockInfo {
             block_index: self.ledger.num_blocks().unwrap() - 1,
-            minimum_fees: FeeMap::default_map(),
+            minimum_fees: self.fee_map.as_ref().clone(),
```
remoun, line 130: `as_ref` seems unnecessary
```suggestion
            minimum_fees: self.fee_map.clone(),
```

connection/test-utils/src/blockchain.rs
BrianCorbin: Good point! thanks :)

connection/test-utils/src/blockchain.rs
BrianCorbin: rust amirite?

BrianCorbin: actually, it is necessary to do that. as_ref() turns it into a &BTreeMap

remoun (approved):

PR #2297, opened 2022-07-23
Mc crypto ring signature no alloc improvements

garbageslam:
This implements two changes discussed in a meeting on Wednesday.
This helps to move us closer towards making `mc-crypto-ring-signature` not depends on the `alloc` crate, so that it can be used more easily on a Trezor without requiring the rust standard allocator.

### DIFF
--- a/account-keys/types/src/traits.rs
+++ b/account-keys/types/src/traits.rs
@@ -3,7 +3,7 @@ use mc_crypto_keys::RistrettoPublic;
 /// An object which has represents a subaddress, and has RingCT-style
 /// view and spend public keys.
 pub trait RingCtAddress {
-    /// Get the subaddress' view public key 
+    /// Get the subaddress' view public key
     fn view_public_key(&self) -> &RistrettoPublic;
     /// Get the subaddress' spend public key
     fn spend_public_key(&self) -> &RistrettoPublic;
--- a/account-keys/types/src/traits.rs
+++ b/account-keys/types/src/traits.rs
@@ -1,6 +1,6 @@
 use mc_crypto_keys::RistrettoPublic;
 
-/// An object which has represents a subaddress, and has RingCT-style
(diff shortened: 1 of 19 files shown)


account-keys/types/src/traits.rs
```diff
@@ -0,0 +1,10 @@
+use mc_crypto_keys::RistrettoPublic;
+
+/// An object which has represents a subaddress, and has RingCT-style
```
sugargoat, line 3:
```suggestion
/// An object which represents a subaddress, and has RingCT-style
```

crypto/ring-signature/src/onetime_keys.rs
```diff
@@ -98,7 +98,7 @@ fn hash_to_scalar(point: RistrettoPoint) -> Scalar {
 /// * `recipient` - The recipient subaddress `(C,D)`.
 pub fn create_tx_out_target_key(
     tx_private_key: &RistrettoPrivate,
-    recipient: &PublicAddress,
+    recipient: &impl RingCtAddress,
```
sugargoat, line 101: Does this become a breaking change for clients? 

I see it being used in full-service (tests only)

https://github.com/mobilecoinofficial/full-service/blob/f1e58e60e6915710bc45210643dc088dde379a84/full-service/src/test_utils.rs#L459

sugargoat (approved):
LGTM!

crypto/ring-signature/src/onetime_keys.rs
garbageslam: I think it's okay because the current object they are passing implements this trait

nick-mobilecoin (approved):
LGTM

Still spinning up on github, I scratched my head for a bit wondering how the use statements for the `GeneratorCache` worked before.
glossed over the rename in the UI
image

PR #2298, opened 2022-07-23
[Fixes] CI/CD - Update GHA CD and metadata script to support new release branch patterns

jgreat:
### Motivation

GHA
- move to release/* and semver compatible tags
- Remove the redundant "branch" output and change to using "namespace" in the GHA CD.

metadata.sh
- Rework how the metadata script to support the move to new `release/v2` branch pattern. 
- Clean up script for clearer logic flow.
- Truncate `really-super-extra-there-s-no-way-you-need-this-way-too-long` branch names so we don't exceed the 63 character limit in a lot of K8s identifier values.

### Future Work

We are starting to use similar logic for dynamic testing environments in some of the other MobileCoin projects. Break this functionality out in to it's own discrete GitHub Action for reuse on other projects.

### DIFF
--- a/.github/workflows/mobilecoin-dev-cd.yaml
+++ b/.github/workflows/mobilecoin-dev-cd.yaml
@@ -12,9 +12,10 @@ env:
 on:
   push:
     branches:
-    - release/v*
     - feature/*
-    - develop
+    - release/*
+    tags:
+      - v[0-9]+*
 
 # don't run more than one at a time for a branch/tag
 concurrency: mobilecoin-dev-cd-${{ github.ref }}
@@ -27,7 +28,6 @@ jobs:
     name: 👾 Environment Info 👾
     runs-on: [self-hosted, Linux, small]
     outputs:
-      branch: ${{ steps.meta.outputs.branch }}
       namespace: ${{ steps.meta.outputs.namespace }}
       tag: ${{ steps.meta.outputs.tag }}
       docker_tag: ${{ steps.meta.outputs.docker_tag }}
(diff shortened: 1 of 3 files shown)


MCrank (approved):
LGTM

.github/workflows/mobilecoin-dev-cd.yaml
```diff
@@ -12,9 +12,10 @@ env:
 on:
   push:
     branches:
-    - release/v*
     - feature/*
-    - develop
+    - release/*
```
nick-mobilecoin, line 16: `develop` got removed here, we're currently using `develop` in the [sgx](https://github.com/mobilecoinfoundation/sgx) repo.  I'm guessing this may be CD versus CI usage.

.internal-ci/util/metadata.sh
```diff
@@ -25,66 +17,107 @@ is_set()
   fi
 }
 
+filter()
```
nick-mobilecoin, line 20: nit: (pretty nitty, I would probably ignore for this PR and if there is anything of value consider in the future.)
1. It may be my notoriously poor vocabulary or general lack of domain knowledge, but I think of `filter` as "removing" and/or "allowing" certain items to proceed.   Like an air filter or oil filter.  Would `normalize` be a better term here?  I know normalize can be a bit loaded and general at times.
2. often in programing languages filter is accompanied by a function or functor, would it be better to more explicitly name it `filter_git_ref` or similar?

again these are pretty nitty so ignore as you see fit

nick-mobilecoin (approved):
LGTM

.github/workflows/mobilecoin-dev-cd.yaml
jgreat: We're not using develop in this repo, so pulling this out.

.internal-ci/util/metadata.sh
jgreat: Good points, naming things is hard :grin:

PR #2314, opened 2022-07-27
Auditor frontend

david1alvarez:
Adds a front end to display: a list of mints/deposits, a list of burns/withdrawals, a linked gnosis safe's value, and the mint/burn totals

the front end is a stripped-down app using react/typescript as the framework, webpack for transpilation, materialUI for rendering, and husky for hooks to lint and format the code on git commits.

Screen Shot 2022-07-22 at 17 06 15

Adds a back end to serve total minted and burned tx amounts

### Motivation

We currently don't have a user-friendly way of visualizing the mint auditor's results

### Future Work
refactor to allow for switching LPs, and move the constants in to a constructed .env file for compatibility in both local dev environments and deployed production environments


david1alvarez:
closing, moving PR over here: https://github.com/mobilecoinofficial/reserve-auditor/pull/1

PR #2123, opened 2022-06-10
Remove CircleCI config

remoun:
We have stopped CircleCI from building this repo, and removed the required check status.

Closes #1687.

PR #2315, opened 2022-07-28
add missing go build and tests back to ci

garbageslam:
this commit puts the go-grpc-gateway build and tests back into ci.

this was lost accidentally when we moved from circle ci to github
actions

https://github.com/mobilecoinfoundation/mobilecoin/pull/2123/files#r931675109


.github/workflows/ci.yml
```diff
@@ -126,11 +126,24 @@ jobs:
             v1-go-${{ runner.os }}
       - name: Set up Go
         uses: actions/setup-go@v3
+      - name: Build go
+        working-directory: go-grpc-gateway
+        run: ./install_tools.sh && ./build.sh
```
wjuan-mob, line 131: I think you need to install protoc judging from the error in gha?

remoun: I couldn't find a Docker image with `golang + protoc + cargo` (all of which this job requires). It might be best to add golang to our builder image: https://github.com/mobilecoinofficial/docker-rust-sgx-base

remoun: Alternatively, we could try https://github.com/arduino/setup-protoc

garbageslam:
i tried to fix this by bumping the repo to track 0.0.16 of the builder-install image

garbageslam:
It looks that the go tests are fixed in this PR, but bumping the image has caused us to build against SGX 2.17, and none of the enclave are happy with that right now.

Will circle back on this when we are ready to use SGX 2.17 in master

sugargoat (approved):

.github/workflows/ci.yml
```diff
@@ -126,15 +126,31 @@ jobs:
...
+      - name: Check dirty git
+        uses: ./.github/actions/check-dirty-git
```
nick-mobilecoin, line 133: do we need to check after each step or is at the end sufficient?
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -126,15 +126,31 @@ jobs:
...
       - name: Lint Go code
         working-directory: go-grpc-gateway
         run: ./lint.sh
       - name: Check dirty git
         uses: ./.github/actions/check-dirty-git
```
nick-mobilecoin, line 138: do we need to check after each step or is at the end sufficient?
```suggestion
```

nick-mobilecoin (approved):
LGTM
I only diffed back to ea36df793057a405f374c5507cb8c1692ad0ca6e since, #2362 should capture those changes.

.github/workflows/ci.yml
garbageslam: idk, the original ci had this at both steps, it seems like it doesn't hurt to do it multiple times

nick-mobilecoin: I agree it's only calling git status. Linux with few if any submodules should be a sub-second operation. 
(Windows with submodules == :puke:)

PR #2335, opened 2022-08-02
Propagate BlockVersion creation error up through the FFI

the-real-adammork:
Previously, an "invalid" (too new)  BlockVersion would cause a rust panic due to the `.unwrap()` call in `mc_transaction_builder_create`. The correct way to deal with invalid BlockVersions is too propagate the error up through the FFI in the form of an `LibMcError:InvalidInput`

### Motivation

@garbageslam pointed out that the conditions for this panic could become common in the future and should be fixed asap. 

### Future Work
* Merge change into master/2.1.x but probably after we finish cherry picking 1.2.x changes.


jcape:
This should go against the v2 branch, 1.2.3 is closed.

PR #2350, opened 2022-08-04
Implement epoch sharding via trait

samdealy:
### Motivation
As desribed in the [design doc](https://docs.google.com/document/d/1-ju5efGK6vvuYjxj1pKhfFcpgGlornuxZYvOC-nsbjs/edit), each Fog View Store instance is responsible for serving a "shard", i.e. a subset, of the total set of TxOuts. This PR introduces the `trait` that will be used to divide the TxOuts into the subsets. 

### Future Work
- Create an `EpochShardingStrategy` that implements the trait for the epoch sharding strategy.

### DIFF
--- a/fog/view/server/src/sharding_strategy.rs
+++ b/fog/view/server/src/sharding_strategy.rs
@@ -1,0 +1,128 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
+//! Enables a Fog View Store to know for which blocks to process TxOuts.
+//!
+//! By determining which TxOuts to process, we are able to "shard" the set of
+//! TxOuts across Fog View Store instances.
+
+use mc_blockchain_types::BlockIndex;
+use mc_fog_types::common::BlockRange;
+
+/// Tells a Fog View Store for which blocks it should process TxOuts.
+pub trait ShardingStrategy {
+    /// Returns true if the Fog View Store should process this block.
+    fn should_process_block(&self, block_index: BlockIndex) -> bool;
+}
+
+/// Determines whether or not to process a block's TxOuts based on the "epoch"
+/// sharding strategy, in which a block is processed IFF it falls within the
+/// contiguous range of blocks.
+///
+/// In practice, the set of Fog View Shards will contain overlapping
+/// [epoch_block_ranges] in order to obfuscate which shard processed the TxOuts.
+pub struct EpochShardingStrategy {
+    /// If a block falls within this range, then the Fog View Store should
+    /// process its TxOuts.
+    epoch_block_range: BlockRange,
+}
+
+impl ShardingStrategy for EpochShardingStrategy {
+    fn should_process_block(&self, block_index: BlockIndex) -> bool {
+        self.epoch_block_range.contains(block_index)
+    }
+}
+
+impl Default for EpochShardingStrategy {
+    fn default() -> Self {
+        Self {
+            epoch_block_range: BlockRange::new(0, u64::MAX),
+        }
+    }
+}
+
+impl EpochShardingStrategy {
+    #[allow(dead_code)]
+    pub fn new(epoch_block_range: BlockRange) -> Self {
+        Self { epoch_block_range }
+    }
+}
+
+#[cfg(test)]
+mod epoch_sharding_strategy_tests {
+    use super::*;
+
+    #[test]
+    fn should_process_block_block_index_is_before_epoch_start_returns_false() {
+        const START_BLOCK: BlockIndex = 50;
+        const END_BLOCK_EXCLUSIVE: BlockIndex = 100;
+        let epoch_block_range = BlockRange::new(START_BLOCK, END_BLOCK_EXCLUSIVE);
+        let epoch_sharding_strategy = EpochShardingStrategy::new(epoch_block_range);
+
+        let should_process_block = epoch_sharding_strategy.should_process_block(START_BLOCK - 1);
+
+        assert!(!should_process_block)
+    }
+
+    #[test]
+    fn should_process_block_block_index_is_epoch_start_returns_true() {
+        const START_BLOCK: BlockIndex = 50;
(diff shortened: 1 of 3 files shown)


samdealy:
* feature/fog-view-router
    * **PR #2350** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2350" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2350?utm_source=stack-comment).

fog/view/server/src/sharding_strategy.rs
```diff
@@ -0,0 +1,117 @@
...
+}
+
+#[cfg(test)]
+mod epoch_sharding_strategy_tests {
+    use super::*;
+
+    #[test]
+    fn should_process_block_block_index_is_before_epoch_start_returns_false() {
+        const START_BLOCK: BlockIndex = 50;
+        const END_BLOCK_NON_INCLUSIVE: BlockIndex = 100;
```
nick-mobilecoin, line 57: nit: I don't think it matters too much, curious if `END_BLOCK_EXCLUSIVE` was intentionally avoided as a name.

fog/view/server/src/sharding_strategy.rs
```diff
@@ -0,0 +1,117 @@
...
+        let included_block_index = ((END_BLOCK_NON_INCLUSIVE - START_BLOCK) / 2) + START_BLOCK;
+        let epoch_block_range = BlockRange::new(START_BLOCK, END_BLOCK_NON_INCLUSIVE);
+        let epoch_sharding_strategy = EpochShardingStrategy::new(epoch_block_range);
+
+        let should_process_block =
+            epoch_sharding_strategy.should_process_block(included_block_index);
+
+        assert!(should_process_block)
+    }
+
```
nick-mobilecoin, line 91: It depends on one's testing approach, i.e. is this meant to be an opaque box test where you don't know `BlockRange::contains()` is utilized.  If that is the intent of this series of tests then this is missing the boundary condition of `END_BLOCK_NON_INCLUSIVE - 1` returns true.
```suggestion

    #[test]
    fn should_process_block_block_index_is_one_before_epoch_end_block_range_returns_true() {
        const START_BLOCK: BlockIndex = 50;
        const END_BLOCK_NON_INCLUSIVE: BlockIndex = 100;
        let epoch_block_range = BlockRange::new(START_BLOCK, END_BLOCK_NON_INCLUSIVE);
        let epoch_sharding_strategy = EpochShardingStrategy::new(epoch_block_range);

        let should_process_block =
            epoch_sharding_strategy.should_process_block(END_BLOCK_NON_INCLUSIVE - 1);

        assert!(should_process_block)
    }

```

nick-mobilecoin (approved):

fog/view/server/src/sharding_strategy.rs
```diff
@@ -0,0 +1,117 @@
...
+impl Default for EpochShardingStrategy {
+    fn default() -> Self {
+        Self {
+            epoch_block_range: BlockRange::new(0, u64::MAX),
+        }
+    }
+}
+
+impl EpochShardingStrategy {
+    #[allow(dead_code)]
```
nick-mobilecoin, line 44: Is this necessary?  I think `pub` methods shouldn't set off dead code warnings
```suggestion
```

fog/view/server/src/sharding_strategy.rs
samdealy: Done.

fog/view/server/src/sharding_strategy.rs
samdealy: Done.

fog/view/server/src/sharding_strategy.rs
```diff
@@ -0,0 +1,117 @@
...
+        Self { epoch_block_range }
+    }
+}
+
+#[cfg(test)]
+mod epoch_sharding_strategy_tests {
+    use super::*;
+
+    #[test]
+    fn should_process_block_block_index_is_before_epoch_start_returns_false() {
```
jcape, line 55: Can these names be shorter, e.g. the name of this in the log is going to be `sharding_strategy::epoch_sharding_strategy_tests::should_process_block_block_index_is_before_epoch_start_returns_false`?

Like, `sharding::epoch::tests::no_block_index_before_epoch_start` is probably long enough, right?

samdealy: I follow the coneventions established [here](https://testing.googleblog.com/2007/02/tott-naming-unit-tests-responsibly.html) when naming unit tests. I always write the method under test, parameters / set up that makes the test unique, and then the expected result. While it may be shorter logs, it's less descriptive and the failure is harder to decipher. That said I don't feel particularly strongly about this, so would be happy to change it.

fog/view/server/src/sharding_strategy.rs
samdealy: The package doesn't build if this is omitted.

fog/view/server/src/sharding_strategy.rs
jcape: I appreciate the intent of those conventions, but I think there's a couple problems with it.

On the one hand, it doesn't really help actual practice: when a test fails, I'm going to read the test anyways, so I can just read the test and parameters. It's just a missed opportunity for self-documentation, in the proper "comments should describe why the code is more than what it is" sense.

The second issue is that it's also just not very rustic, and the really long identifiers seem out of place with the rest of the ecosystem---e.g. it's `core::mem::ptr`, not `rust_core_library::memory_handling::pointer`.

jcape (approved):

fog/view/server/src/sharding_strategy.rs
samdealy: As discussed offline, we'll address this in a follow-up PR and take some time to discuss team conventions around naming unit tests.

PR #2351, opened 2022-08-04
Use sharding strategy to tell view what blocks to process

samdealy:
### Motivation
The previous PR (#2350) created the `EpochShardingStrategy`. This PR uses this concrete sharding strategy to tell Fog View which TxOuts its responsible for. 

### Future Work
- The next PR (#2352) allows for the sharding strategy to be passed as a CLI argument.

PR #2352, opened 2022-08-04
Add config option for sharding strategy

samdealy:
### Motivation
This PR allows for the `EpochShardingStrategy` parameters to be passed as CLI Arguments. 

### Future Work
- Implement "readiness" for sharding strategy.

### DIFF
--- a/fog/types/src/common.rs
+++ b/fog/types/src/common.rs
@@ -1,5 +1,7 @@
 // Copyright (c) 2018-2022 The MobileCoin Foundation
 
+use alloc::{format, string::String, vec::Vec};
+use core::str::FromStr;
 use prost::Message;
 use serde::{Deserialize, Serialize};
 
@@ -45,9 +47,31 @@ impl core::fmt::Display for BlockRange {
     }
 }
 
+impl FromStr for BlockRange {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let block_indices: Vec<u64> = s
+            .split(',')
+            .map(|index_str| index_str.trim().parse())
+            .collect::<Result<Vec<_>, _>>()
+            .map_err(|_| "BlockRange index is not a number.")?;
+        if block_indices.len() != 2 {
+            return Err(format!(
+                "Block range is composed of two indices, found {} indices",
+                block_indices.len()
+            ));
+        }
+        let result = BlockRange::new(block_indices[0], block_indices[1]);
+
+        Ok(result)
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
+
     #[test]
     fn test_contains() {
         let range = BlockRange::new(10, 13);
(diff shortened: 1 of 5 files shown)


samdealy:
* feature/fog-view-router
        * **PR #2352** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2352" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2352?utm_source=stack-comment).

fog/types/src/common.rs
```diff
@@ -45,6 +47,30 @@ impl core::fmt::Display for BlockRange {
...
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let block_indices: Vec<u64> = s
+            .split(',')
+            .map(|index_str| {
+                index_str
+                    .parse()
+                    .expect("BlockRange index is not a number.")
+            })
+            .collect();
```
nick-mobilecoin, line 61: This shows up in the excellent test coverage below.
There are 2 different ways to fail:

    One panics causing the process to stop
    The other returns back an error string saying what happened, letting the caller decide if panicking is appropriate.

Unless there is good reason it seems like the failure behavior should be the same for the 2 cases.

I'm not sure where these strings are coming from yet, but generally I lean toward returning an Error over panicking if the api allows it.

Perhaps something like this would make them more consistent:
```suggestion
        let block_indices = s
            .split(',')
            .map(|index_str| {
                index_str
                    .parse()
            })
            .collect::<Result<Vec<_>, _>>().map_err(|_| "BlockRange index is not a number.")?;
```

fog/types/src/lib.rs
```diff
@@ -1,6 +1,6 @@
 // Copyright (c) 2018-2022 The MobileCoin Foundation
 
-#![no_std]
+#![cfg_attr(not(test), no_std)]
```
nick-mobilecoin, line 3: not sure if this `not(test)` snuck in here or if this was one of those things going to be fixed in a later PR
```suggestion
#![no_std]

File: fog/view/server/src/bin/main.rs [hunk 3]
```diff
@@ -60,15 +58,15 @@ fn main() {
...
     let mut server = ViewServer::new(
         config.clone(),
         sgx_enclave,
         recovery_db,
         ias_client,
         SystemTimeProvider::default(),
-        // TODO: Change the EpochShardingStrategy to incorporate config values that specify
-        //  start and end block indices. See PR #2352
-        EpochShardingStrategy::default(),
+        sharding_strategy,
```
nick-mobilecoin, line 69: A couple of things:
* The config is being cloned, when it seems like only the sharding strategy needs to be cloned
* There is currently only one strategy, but it seems off to be forcing this to be that one all the time. 
```suggestion
    let mut server = ViewServer::new(
        config.clone(),
        sgx_enclave,
        recovery_db,
        ias_client,
        SystemTimeProvider::default(),
        config.sharding_strategy.clone(),
```

fog/view/server/src/config.rs
```diff
@@ -67,6 +67,30 @@ pub struct MobileAcctViewConfig {
     /// Postgres config
     #[clap(flatten)]
     pub postgres_config: SqlRecoveryDbConnectionConfig,
+
+    /// Determines which group of TxOuts the Fog View Store instance will
+    /// process.
+    #[clap(long)]
```
nick-mobilecoin, line 73: Did we want to support environment variables for this argument?

fog/view/server/src/sharding_strategy.rs
```diff
@@ -48,6 +50,18 @@ impl EpochShardingStrategy {
...
 
+impl FromStr for EpochShardingStrategy {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if let Ok(block_range) = BlockRange::from_str(s) {
+            return Ok(Self::new(block_range));
+        }
+
+        Err("Invalid epoch sharding strategy.".to_string())
```
nick-mobilecoin, line 61: Would it be better to propagate the block range parsing error?
```suggestion
        let block_range = BlockRange::from_str(s)?;
        Ok(Self::new(block_range))
```

fog/types/src/common.rs
```diff
@@ -87,4 +113,41 @@ mod tests {
         assert!(!range.overlaps(&BlockRange::new(0, 10)));
         assert!(!range.overlaps(&BlockRange::new(13, 100)));
     }
+
+    #[test]
+    fn from_string_well_formatted_creates_block_range() {
+        let start_block = 0;
+        let end_block = 10;
+        let block_range_str = format!("{},{}", start_block, end_block);
```
nick-mobilecoin, line 121: I think if desired it could be a follow on PR/Issue, but `3, 5` will fail to parse.  My hunch is hat most users will want to put a space after the comma

nick-mobilecoin (changes requested):
The main blocker is the inconsistent error handling in `from_str` for `BlockRange`

fog/types/src/common.rs
samdealy: The strings are coming from the command-line, so I think returning an Error is fine. Changed it (and the test) accordingly.

fog/view/server/src/bin/main.rs
samdealy: Changed the clone. And yes there's one sharding strategy right now, EpochShardingStrategy, and the default version of it is actually what Fog View does currently.

fog/types/src/common.rs
samdealy: This probably won't be an issue because these ranges will be passed as CLI args (and CLI convention is to not use whitespace between arguments). But I implemented it and added a test.

fog/view/server/src/config.rs
samdealy: For right now, I don't because I want to wait until we meet with Ops to discuss their needs / use cases. I could see them wanting a different way to store sharding strategy etc.

fog/types/src/lib.rs
samdealy: Done

fog/view/server/src/sharding_strategy.rs
samdealy: This suggestion only works when there's one sharding srategy. I'll leave it as is because this makes it easier to add a new strategy in the future (just add a new if block).

nick-mobilecoin (approved):

jcape (approved):

jcape:
(approved pending a CI pass)

PR #2353, opened 2022-08-04
Add readiness API to ShardingStrategy

samdealy:
### Motivation

This PR implements  the "readiness" check for the `EpochShardingStrategy` (and also adds this functionality to the `ShardingStrategy` trait). This check is needed because a Fog View Store may not be ready to serve TxOuts given a certain sharding strategy. In the case of a  Fog View Store that employs the `EpochShardingStrategy, that store is not ready to serve TxOuts unless 50% of its TxOuts have been loaded.

### DIFF
--- a/fog/view/server/src/fog_view_service.rs
+++ b/fog/view/server/src/fog_view_service.rs
@@ -1,6 +1,8 @@
 // Copyright (c) 2018-2022 The MobileCoin Foundation
 
-use crate::{config::ClientListenUri, server::DbPollSharedState};
+use crate::{
+    config::ClientListenUri, server::DbPollSharedState, sharding_strategy::ShardingStrategy,
+};
 use grpcio::{RpcContext, RpcStatus, RpcStatusCode, UnarySink};
 use mc_attest_api::attest;
 use mc_common::logger::{log, Logger};
@@ -22,7 +24,12 @@ use mc_util_telemetry::{tracer, Tracer};
 use std::sync::{Arc, Mutex};
 
 #[derive(Clone)]
-pub struct FogViewService<E: ViewEnclaveProxy, DB: RecoveryDb + Send + Sync> {
+pub struct FogViewService<E, DB, SS>
+where
+    E: ViewEnclaveProxy,
+    DB: RecoveryDb + Send + Sync,
+    SS: ShardingStrategy,
+{
     /// Enclave providing access to the Recovery DB
     enclave: E,
 
@@ -40,9 +47,17 @@ pub struct FogViewService<E: ViewEnclaveProxy, DB: RecoveryDb + Send + Sync> {
 
     /// Slog logger object
     logger: Logger,
+
+    /// Dictates what blocks to process.
+    sharding_strategy: SS,
 }
 
-impl<E: ViewEnclaveProxy, DB: RecoveryDb + Send + Sync> FogViewService<E, DB> {
+impl<E, DB, SS> FogViewService<E, DB, SS>
+where
+    E: ViewEnclaveProxy,
+    DB: RecoveryDb + Send + Sync,
+    SS: ShardingStrategy,
+{
     /// Creates a new fog-view-service node (but does not create sockets and
     /// start it etc.)
     pub fn new(
@@ -51,6 +66,7 @@ impl<E: ViewEnclaveProxy, DB: RecoveryDb + Send + Sync> FogViewService<E, DB> {
         db_poll_shared_state: Arc<Mutex<DbPollSharedState>>,
         authenticator: Arc<dyn Authenticator + Send + Sync>,
         client_listen_uri: ClientListenUri,
+        sharding_strategy: SS,
         logger: Logger,
     ) -> Self {
         Self {
@@ -59,6 +75,7 @@ impl<E: ViewEnclaveProxy, DB: RecoveryDb + Send + Sync> FogViewService<E, DB> {
             db_poll_shared_state,
             authenticator,
             client_listen_uri,
+            sharding_strategy,
             logger,
         }
     }
@@ -169,7 +186,12 @@ impl<E: ViewEnclaveProxy, DB: RecoveryDb + Send + Sync> FogViewService<E, DB> {
 }
 
 // Implement grpc trait
-impl<E: ViewEnclaveProxy, DB: RecoveryDb + Send + Sync> FogViewApi for FogViewService<E, DB> {
+impl<E, DB, SS> FogViewApi for FogViewService<E, DB, SS>
+where
+    E: ViewEnclaveProxy,
+    DB: RecoveryDb + Send + Sync,
+    SS: ShardingStrategy,
+{
     fn auth(
         &mut self,
         ctx: RpcContext,
@@ -204,7 +226,12 @@ impl<E: ViewEnclaveProxy, DB: RecoveryDb + Send + Sync> FogViewApi for FogViewSe
 }
 
 /// Implement the FogViewStoreService gRPC trait.
-impl<E: ViewEnclaveProxy, DB: RecoveryDb + Send + Sync> FogViewStoreApi for FogViewService<E, DB> {
+impl<E, DB, SS> FogViewStoreApi for FogViewService<E, DB, SS>
+where
+    E: ViewEnclaveProxy,
+    DB: RecoveryDb + Send + Sync,
+    SS: ShardingStrategy,
+{
     fn auth(
         &mut self,
         ctx: RpcContext,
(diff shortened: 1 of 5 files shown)


samdealy:
* feature/fog-view-router
          * **PR #2353** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2353" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2353?utm_source=stack-comment).

fog/view/server/src/sharding_strategy.rs
```diff
@@ -48,6 +59,22 @@ impl EpochShardingStrategy {
...
+        }
+
+        let epoch_block_range_length =
+            self.epoch_block_range.end_block - self.epoch_block_range.start_block;
+        let minimum_processed_block_count = epoch_block_range_length / 2;
+
+        u64::from(processed_block_count) >= minimum_processed_block_count
+    }
+
+    fn is_first_epoch(&self) -> bool {
```
nick-mobilecoin, line 75: Like the functionalization here.  Even though it's only one line, the function name more clearly communicates intent where it's used.

fog/view/server/src/sharding_strategy.rs
```diff
@@ -140,4 +167,80 @@ mod epoch_sharding_strategy_tests {
 
         assert!(!should_process_block)
     }
+
+    #[test]
+    fn is_ready_to_serve_tx_outs_first_epoch_processed_block_count_is_zero_less_than_50_percent_processed_returns_true(
+    ) {
+        // The first epoch has a start block == 0.
+        const START_BLOCK: BlockIndex = 0;
+        const END_BLOCK_NON_INCLUSIVE: BlockIndex = 100;
```
nick-mobilecoin, line 176: Looks like #2350 changed to `END_BLOCK_EXCLUSIVE` might consider it here as well since they show up in the test cases above

fog/view/server/src/sharding_strategy.rs
```diff
@@ -140,4 +167,80 @@ mod epoch_sharding_strategy_tests {
...
+        let is_ready_to_serve_txouts = epoch_sharding_strategy.is_ready_to_serve_tx_outs(70.into());
+
+        assert!(is_ready_to_serve_txouts)
+    }
+
+    #[test]
+    fn is_ready_to_serve_tx_outs_not_first_epoch_processed_block_count_less_than_50_percent_of_epoch_range_length_returns_false(
+    ) {
+        const START_BLOCK: BlockIndex = 100;
+        const END_BLOCK_NON_INCLUSIVE: BlockIndex = 110;
```
nick-mobilecoin, line 203: Since this is integer division for halving suggest using an odd number
```suggestion
        const END_BLOCK_NON_INCLUSIVE: BlockIndex = 111;
```

fog/view/server/src/sharding_strategy.rs
```diff
@@ -140,4 +167,80 @@ mod epoch_sharding_strategy_tests {
...
+            .is_ready_to_serve_tx_outs((minimum_processed_block_count - 1).into());
+
+        assert!(!is_ready_to_serve_txouts)
+    }
+
+    #[test]
+    fn is_ready_to_serve_tx_outs_not_first_epoch_processed_block_count_equals_50_percent_of_epoch_range_length_caught_up_returns_true(
+    ) {
+        const START_BLOCK: BlockIndex = 100;
+        const END_BLOCK_NON_INCLUSIVE: BlockIndex = 110;
```
nick-mobilecoin, line 219: Since this is integer division for halving suggest using an odd number
```suggestion
        const END_BLOCK_NON_INCLUSIVE: BlockIndex = 111;
```

fog/view/server/src/fog_view_service.rs
```diff
@@ -151,6 +170,35 @@ impl<E: ViewEnclaveProxy, DB: RecoveryDb + Send + Sync> FogViewService<E, DB> {
...
+        fog_view_store_uri: FogViewStoreUri,
+        queries: Vec<attest::Message>,
+    ) -> MultiViewStoreQueryResponse {
+        let mut response = MultiViewStoreQueryResponse::new();
+        response.set_fog_view_store_uri(fog_view_store_uri.url().to_string());
+        for query in queries.into_iter() {
+            let result = self.query_impl(query);
+            if let Ok(attested_message) = result {
+                {
+                    let shared_state = self.db_poll_shared_state.lock().expect("mutex poisoned");
```
nick-mobilecoin, line 184: *s there a reason to more narrowly scope the `db_polled_shared_state.lock()`?  it seems like unlocking at the brace or after `return response` shouldn't matter much.  It may be a convention I'm not aware of.  I see https://github.com/mobilecoinfoundation/mobilecoin/blob/master/CONTRIBUTING.md#avoid-manual-drops, but it says to use scopes over manual drops.  I don't think the intent is to always use scopes for lock guards

fog/view/server/src/fog_view_service.rs
```diff
@@ -151,6 +170,35 @@ impl<E: ViewEnclaveProxy, DB: RecoveryDb + Send + Sync> FogViewService<E, DB> {
...
+                    if !self
+                        .sharding_strategy
+                        .is_ready_to_serve_tx_outs(shared_state.processed_block_count.into())
+                    {
+                        response.set_error(MultiViewStoreQueryResponseError::NOT_READY);
+                    } else {
+                        response.set_query_response(attested_message);
+                    }
+                }
+                return response;
```
nick-mobilecoin, line 194: It was like this before so maybe I missed it or I'm just not recollecting the full behavior.  This seems to `return response` on the first response it's able to authenticate.  Is there something that prevents multiple serviceable queries coming to this shard?.  i.e. what if there were 2 queries this shard could handle it would return with the first one and not answer the second based on the code here.

nick-mobilecoin (approved):
LGTM other than my understanding of how many responses each shard is expected to handle per request

fog/view/server/src/sharding_strategy.rs
samdealy: Done.

fog/view/server/src/sharding_strategy.rs
samdealy: Done

fog/view/server/src/sharding_strategy.rs
samdealy: Done.

fog/view/server/src/fog_view_service.rs
samdealy: I only want to unlock the state if the `attested_message` result is OK.

fog/view/server/src/fog_view_service.rs
samdealy: Yes this is the correct behavior. There should only be one response that the shard can decrypt and fulfill. There should not be any second response that it can decrypt.

fog/view/server/src/sharding_strategy.rs
```diff
@@ -140,4 +167,80 @@ mod epoch_sharding_strategy_tests {
 
         assert!(!should_process_block)
     }
+
+    #[test]
+    fn is_ready_to_serve_tx_outs_first_epoch_processed_block_count_is_zero_less_than_50_percent_processed_returns_true(
+    ) {
+        // The first epoch has a start block == 0.
+        const START_BLOCK: BlockIndex = 0;
+        const END_BLOCK_END_BLOCK_EXCLUSIVE: BlockIndex = 100;
```
nick-mobilecoin, line 176: nit looks like this got double tagged, rest of tests cases too:
```suggestion
        const END_BLOCK_EXCLUSIVE: BlockIndex = 100;
```

samdealy: Done.

fog/view/server/src/fog_view_service.rs
jcape: Could you add a comment to this effect somewhere (i.e. only one of the query messages in an MVSQ is intended for this store)?

fog/view/server/src/sharding_strategy.rs
jcape: Is this fixed in a later PR?

fog/view/server/src/sharding_strategy.rs
```diff
@@ -140,4 +167,80 @@ mod epoch_sharding_strategy_tests {
 
         assert!(!should_process_block)
     }
+
+    #[test]
+    fn is_ready_to_serve_tx_outs_first_epoch_processed_block_count_is_zero_less_than_50_percent_processed_returns_true(
```
jcape, line 172: What is this even testing? That zero is less than half of 100?

jcape (changes requested):
It's harder to understand `is_ready_to_serve_tx_outs_first_epoch_processed_block_count_is_zero_less_than_50_percent_processed_returns_true` than `allow_zero_in_shard_of_0_to_100`.

fog/view/server/src/sharding_strategy.rs
samdealy: No, that the first block returns a ready status even if < 50% of blocks are processed.

jcape (approved):

PR #2354, opened 2022-08-04
Implement shard management API sans auth

samdealy:
### Motivation

Adds the "shard management" API ,which allows for Fog View Shards to be added to a Fog View Router on the fly while it's running. This allow us to spin up new shards without having to restart Fog View Router instances.

The next piece of this project is to implement and authorization scheme that only allows authorized clients to add shards to FVR. As discussed on discord, we'll accomplish this by configuring Kubernetes to only expose the API, which is running on it's own host / port, to authorized users. 

RFC: I used a `RwLock` to manage thread access to the shards, but we may consider using a `Mutex`. Let me know what you think.

### Future Work
- During deployment, configure kubernetes to only expose the endpoint to authorized users.

### DIFF
--- a/fog/view/server/src/bin/router.rs
+++ b/fog/view/server/src/bin/router.rs
@@ -13,7 +13,12 @@ use mc_fog_view_server::{
 };
 use mc_util_cli::ParserWithBuildInfo;
 use mc_util_grpc::ConnectionUriGrpcioChannel;
-use std::{env, str::FromStr, sync::Arc};
+use std::{
+    collections::HashMap,
+    env,
+    str::FromStr,
+    sync::{Arc, RwLock},
+};
 
 fn main() {
     mc_common::setup_panic_handler();
@@ -38,7 +43,7 @@ fn main() {
     );
 
     // TODO: Remove and get from a config.
-    let mut fog_view_store_grpc_clients = Vec::new();
+    let mut fog_view_store_grpc_clients = HashMap::new();
     let grpc_env = Arc::new(
         grpcio::EnvBuilder::new()
             .name_prefix("Main-RPC".to_string())
@@ -54,8 +59,9 @@ fn main() {
             ChannelBuilder::default_channel_builder(grpc_env.clone())
                 .connect_to_uri(&shard_uri, &logger),
         );
-        fog_view_store_grpc_clients.push(fog_view_store_grpc_client);
+        fog_view_store_grpc_clients.insert(shard_uri, Arc::new(fog_view_store_grpc_client));
     }
+    let fog_view_store_grpc_clients = Arc::new(RwLock::new(fog_view_store_grpc_clients));
 
     let ias_client = Client::new(&config.ias_api_key).expect("Could not create IAS client");
     let mut router_server = FogViewRouterServer::new(
--- a/fog/view/server/src/fog_view_router_admin_service.rs
+++ b/fog/view/server/src/fog_view_router_admin_service.rs
@@ -1,0 +1,84 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
+use grpcio::{ChannelBuilder, RpcContext, RpcStatus, UnarySink};
+use itertools::Itertools;
+use mc_common::logger::{log, Logger};
+use mc_fog_api::{
+    view::AddShardRequest,
+    view_grpc::{FogViewRouterAdminApi, FogViewStoreApiClient},
+};
+use mc_fog_uri::FogViewStoreUri;
+use mc_util_grpc::{
+    rpc_invalid_arg_error, rpc_logger, rpc_precondition_error, send_result,
+    ConnectionUriGrpcioChannel, Empty,
+};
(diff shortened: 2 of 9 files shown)


samdealy:
* sam/oblivious
          * **PR #2353** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2353" target="_blank">Graphite</a> 
            * **PR #2354** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2354" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2354?utm_source=stack-comment).

fog/view/server/src/fog_view_router_admin_service.rs
```diff
@@ -0,0 +1,84 @@
...
+        })?;
+        let mut shard_clients = self.shard_clients.write().expect("RwLock Poisoned");
+        if shard_clients.keys().contains(&view_store_uri) {
+            let error = rpc_precondition_error(
+                "add_shard",
+                format!("Shard uri {} already exists in the shard list", shard_uri),
+                logger,
+            );
+            return Err(error);
+        }
```
nick-mobilecoin, line 55: :question: What happens if a shard is reset?  Will we be stuck in an error state? 
Not sure if stale shard removal is coming in a later PR to address this possibility.

fog/view/server/src/bin/router.rs
```diff
@@ -38,7 +43,7 @@ fn main() {
     );
 
     // TODO: Remove and get from a config.
-    let mut fog_view_store_grpc_clients = Vec::new();
+    let mut fog_view_store_grpc_clients = HashMap::new();
     let grpc_env = Arc::new(
         grpcio::EnvBuilder::new()
             .name_prefix("Main-RPC".to_string())
```
nick-mobilecoin, line 49: :wrench: Perhaps this is in the scope of the TODO and not sure if people may look at the threads to differentiate which shards were _original_ and which ones were later added.  
These initial shards are created with the `Main-RPC` thread prefix and the new ones are added with `add-shard`.  
It probably depends if the thread names are used for logging and health, but it may be good to use similar names

nick-mobilecoin (approved):

fog/view/server/src/fog_view_router_admin_service.rs
samdealy: What do you mean by reset? This API is only meant to add active shards to an already running FVR cluster. I imagine if a shard is "reset" (what I take to mean "deactivated") then we'd do so in a cluster upgrade rather than on the fly with an API.

fog/view/server/src/bin/router.rs
samdealy: I don't think we'll need to differentiate the threads- they should be treated the same. If we need to differentiate shards then we can do so with ResponderId.

fog/view/server/src/fog_view_router_admin_service.rs
nick-mobilecoin: I'm thinking shard X goes down for some reason and comes back up trying to re-register itself, it will get an error saying it's already registered.
Maybe that's handled by clusters, but I'm pretty ignorant of how clusters work

samdealy: To clarify, this API is going to be called by a human operator when they want to add a new shard to the FVR's list of shards. It won't be called by a shard. 

In the failure case you're describing, Kubernetes, i.e the cluster manager, will try to reboot the shard and since the shard is already in the FVR's list of known shards it won't need to take additional action.

fog/view/server/src/lib.rs
```diff
@@ -13,5 +13,6 @@ pub mod sharding_strategy;
 mod block_tracker;
 mod counters;
 mod db_fetcher;
+mod fog_view_router_admin_service;
```
jcape, line 16: So, the full path to this thing is:

`mc_fog_view_server::fog_view_router_admin_service;`

Can this just be `router_admin_service`?

jcape (approved):
Couple nitpicks, otherwise LGTM.

fog/view/server/src/fog_view_router_admin_service.rs
```diff
@@ -0,0 +1,84 @@
...
+        shard_clients: Arc<RwLock<HashMap<FogViewStoreUri, Arc<FogViewStoreApiClient>>>>,
+        logger: Logger,
+    ) -> Self {
+        Self {
+            shard_clients,
+            logger,
+        }
+    }
+
+    pub fn add_shard_impl(&mut self, shard_uri: &str, logger: &Logger) -> Result<Empty, RpcStatus> {
```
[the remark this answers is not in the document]
samdealy: Nope, done.

fog/view/server/src/lib.rs
samdealy: Yes, done.

PR #2362, opened 2022-08-06
Cherry pick sgx 2.17 to master

garbageslam:
This cherry-picked two commits from release/v2

https://github.com/mobilecoinfoundation/mobilecoin/commit/344bc424307a21b7ad523676b3bbb9c4561e2ea6

and

https://github.com/mobilecoinfoundation/mobilecoin/commit/5ccb18921aac5e329e1b9db8c32431d7ec18eb77

---

There were some conflicts in the attest-verifier code which I resolved to the best of my ability


account-keys/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "mc-account-keys"
-version = "1.3.0-pre0"
+version = "2.0.0"
```
remoun, line 3: IIUC these should be `3.0.0-pre0`

garbageslam: I don't think so, we should do that when we start releasing 3.0.0, at least that's how I understand the process right now

nick-mobilecoin (approved):

garbageslam:
This was merged in the go build and tests commit

PR #2366, opened 2022-08-08
Propagate BlockVersion creation error up through the FFI, if a user t…

the-real-adammork:
Previously, an "invalid" (too new)  BlockVersion would cause a rust panic due to the `.unwrap()` call in `mc_transaction_builder_create`. The correct way to deal with invalid BlockVersions is too propagate the error up through the FFI in the form of an `LibMcError:InvalidInput`

### Motivation

@garbageslam pointed out that the conditions for this panic could become common in the future and should be fixed asap. 

### Future Work
* Merge change into master but probably after we finish cherry picking 1.2.x changes.

### DIFF
--- a/libmobilecoin/include/transaction.h
+++ b/libmobilecoin/include/transaction.h
@@ -179,13 +179,18 @@ MC_ATTRIBUTE_NONNULL(1, 2, 3);
 
 /* ==== McTransactionBuilder ==== */
 
+///
+/// # Errors
+///
+/// * `LibMcError::InvalidInput`
 McTransactionBuilder* MC_NULLABLE mc_transaction_builder_create(
   uint64_t fee,
   uint64_t token_id,
   uint64_t tombstone_block,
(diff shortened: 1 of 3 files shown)


remoun (approved):
LGTM

garbageslam (approved):

libmobilecoin/include/transaction.h
```diff
@@ -179,13 +179,18 @@ MC_ATTRIBUTE_NONNULL(1, 2, 3);
 
 /* ==== McTransactionBuilder ==== */
 
+///
+/// # Errors
+///
+/// * `LibMcError::InvalidInput`
 McTransactionBuilder* MC_NULLABLE mc_transaction_builder_create(
```
garbageslam, line 186: It may be good to document something like:

"SDK consumers may wish to handle this error in part by checking if a software update is available"

the-real-adammork: good idea, updated

PR #2375, opened 2022-08-09
Create Fog View Router integration test

samdealy:
### Motivation
Creates the base set of integration tests for the Fog View Router.

PR #2361, opened 2022-08-06
Create FogViewRouterAdminUri

samdealy:
### Motivation

This PR creates a special URI for the admin server, which will allow us to have an admin server and a router server serving gRPC requests from the same gRPC server. 

### Future Work
* Implement the gRPC admin API

PR #2360, opened 2022-08-06
Create gRPC FogViewRouterAdminAPI

samdealy:
### Motivation
The FVR [design](https://docs.google.com/document/d/1-ju5efGK6vvuYjxj1pKhfFcpgGlornuxZYvOC-nsbjs/edit) requires an API that allows us to add shards to FVRs list of known shards on the fly, i.e. without having to restart the router server. This PR adds the gRPC service defintion that will be implemented in the next PR.
 
### Future Work
* Implement this API

PR #2387, opened 2022-08-10
Return minimum highest processed block to client

samdealy:
### Motivation
Following @garbageslam's comments in #2351, we need to figure out a way to return the highest block count for which the Fog Router's returned TxOuts are valid. I.e. the client needs to be able to tie their balance to a certain block index.

To accomplish this, I (a) refactored the `blocks_processed` terminology to `blocks_observed` and  (b) we mark a block as observed whenever the db_fetcher sees a block (and doesn't necessarily process it, meaning adds its TxOuts to ORAM). This means that each shard will report that it saw a block even when it might be done processing.

This PR also modifies the enclave response collation code to return the minimum highest observed block index to the client. 

### Future Work
* We may consider adding a "is shard done processing it's blocks" method to the `ShardingStrategy` trait. We'd then only use a shard's highest observed block index if it is not done processing.

### DIFF
--- a/fog/view/enclave/api/src/lib.rs
+++ b/fog/view/enclave/api/src/lib.rs
@@ -235,7 +235,7 @@ pub enum Error {
     EnclaveNotInitialized,
     /// Cipher encryption failed: {0}
     Cipher(CipherError),
-    /// Fog View Shard query response collation error
+    /// Fog View Shard query response collation error.
     QueryResponseCollation,
 }
 
--- a/fog/view/enclave/impl/src/lib.rs
+++ b/fog/view/enclave/impl/src/lib.rs
@@ -266,8 +266,22 @@ where
                 Error::ProstDecode
             })?;
 
-        shard_query_response.tx_out_search_results =
-            self.get_collated_tx_out_search_results(client_query_request, shard_query_responses)?;
+        let shard_query_responses = shard_query_responses
+            .into_iter()
+            .map(|(responder_id, enclave_message)| {
+                let plaintext_bytes = self.ake.backend_decrypt(responder_id, enclave_message)?;
+                let query_response: QueryResponse = mc_util_serial::decode(&plaintext_bytes)?;
+
+                Ok(query_response)
+            })
+            .collect::<Result<Vec<QueryResponse>>>()?;
+
+        shard_query_response.tx_out_search_results = self.get_collated_tx_out_search_results(
+            client_query_request,
+            shard_query_responses.clone(),
+        )?;
+        shard_query_response.highest_processed_block_count =
+            self.get_minimum_highest_processed_block_count(shard_query_responses);
 
         Ok(shard_query_response)
     }
@@ -275,25 +289,27 @@ where
     fn get_collated_tx_out_search_results(
         &self,
         client_query_request: QueryRequest,
-        shard_query_responses: BTreeMap<ResponderId, EnclaveMessage<ClientSession>>,
+        shard_query_responses: Vec<QueryResponse>,
     ) -> Result<Vec<TxOutSearchResult>> {
-        let plaintext_shard_tx_out_search_results: Vec<TxOutSearchResult> = shard_query_responses
+        let plaintext_search_results = shard_query_responses
             .into_iter()
-            .map(|(responder_id, enclave_message)| {
-                let plaintext_bytes = self.ake.backend_decrypt(responder_id, enclave_message)?;
-                let plaintext_response: QueryResponse = mc_util_serial::decode(&plaintext_bytes)?;
-
-                Ok(plaintext_response.tx_out_search_results)
-            })
-            .collect::<Result<Vec<Vec<TxOutSearchResult>>>>()?
-            .iter()
-            .flat_map(|array| array.iter())
-            .cloned()
-            .collect();
+            .flat_map(|response| response.tx_out_search_results)
+            .collect::<Vec<TxOutSearchResult>>();
 
         oblivious_utils::collate_shard_tx_out_search_results(
             client_query_request.get_txos,
-            plaintext_shard_tx_out_search_results,
+            plaintext_search_results,
         )
+    }
+
+    fn get_minimum_highest_processed_block_count(
+        &self,
+        shard_query_responses: Vec<QueryResponse>,
+    ) -> u64 {
+        shard_query_responses
+            .into_iter()
+            .map(|query_response| query_response.highest_processed_block_count)
+            .min()
+            .unwrap_or_default()
     }
 }


samdealy:
* feature/fog-view-router
        * **PR #2387** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2387" target="_blank">Graphite</a>  👈
            * **PR #2353** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2353" target="_blank">Graphite</a> 
              * **PR #2360** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2360" target="_blank">Graphite</a> 
                * **PR #2361** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2361" target="_blank">Graphite</a> 
                  * **PR #2354** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2354" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2387?utm_source=stack-comment).

fog/view/enclave/impl/src/lib.rs
```diff
@@ -296,4 +305,19 @@ where
...
+    ) -> Result<u64> {
+        shard_query_responses
+            .into_iter()
+            .map(|query_response| query_response.highest_processed_block_count)
+            .min()
+            .ok_or_else(|| {
+                Error::QueryResponseCollation(
+                    "No minimum highest processed block count found.".to_string(),
+                )
+            })
```
nick-mobilecoin, line 321: It looks like the only time one would get an option is when the vector is empty, perhaps something like
```suggestion
    ) -> u64 {
        shard_query_responses
            .into_iter()
            .map(|query_response| query_response.highest_processed_block_count)
            .min()
            .unwrap_or_default()
```
The reasons to suggest this:
* I think most people would understand 0 as a minimum for an empty container.
* The `QueryResponseCollation` had to be expanded for this.  It looks like it was actually unused prior
* There is a check for having more than one response in `create_client_query_response`, which means this is an error that would never propagate to the user, yet they still may have to acknowledge it.

fog/view/server/src/block_tracker.rs
```diff
@@ -506,20 +507,17 @@ mod tests {
         assert_eq!(block_tracker.next_blocks(&[rec1, rec2]), expected_state);
     }
 
-    // highest_fully_processed_block_count behaves as expected
+    // highest_fully_observed_block_count behaves as expected
     #[test_with_logger]
-    fn highest_fully_processed_block_count_all_empty(logger: Logger) {
+    fn highest_fully_observed_block_count_all_empty(logger: Logger) {
```
nick-mobilecoin, line 512:
```suggestion
    // highest_observed_block_count behaves as expected
    #[test_with_logger]
    fn highest_observed_block_count_all_empty(logger: Logger) {
```

fog/view/server/src/block_tracker.rs
```diff
@@ -506,20 +507,17 @@ mod tests {
...
-            (0, None)
-        );
+        assert_eq!(block_tracker.highest_observed_block_count(&[]), (0, None));
     }
 
-    // Check with a key that hasn't yet processed anything.
+    // Check with a key that hasn't yet observed anything.
     #[test_with_logger]
-    fn highest_fully_processed_block_missing_blocks_nothing_processed1(logger: Logger) {
+    fn highest_fully_observed_block_missing_blocks_nothing_observed1(logger: Logger) {
```
nick-mobilecoin, line 520:
```suggestion
    fn highest_observed_block_missing_blocks_nothing_observed1(logger: Logger) {
```

fog/view/server/src/block_tracker.rs
```diff
@@ -534,15 +532,15 @@ mod tests {
...
             (0, None),
         );
     }
 
     // A block tracker with a single ingestable range tracks it properly as blocks
-    // are processed when the start block is 0.
+    // are observed when the start block is 0.
     #[test_with_logger]
-    fn highest_fully_processed_block_tracks_block_processed1(logger: Logger) {
+    fn highest_fully_observed_block_tracks_block_observed1(logger: Logger) {
```
nick-mobilecoin, line 543:
```suggestion
    fn highest_observed_block_tracks_block_observed1(logger: Logger) {
```

fog/view/server/src/block_tracker.rs
```diff
@@ -577,7 +575,7 @@ mod tests {
     // A block tracker with a single ingestable range makes progress appropriately
     // when the start block is greater than zero
     #[test_with_logger]
-    fn highest_fully_processed_block_tracks_block_processed2(logger: Logger) {
+    fn highest_fully_observed_block_tracks_block_observed2(logger: Logger) {
```
nick-mobilecoin, line 578:
```suggestion
    fn highest_observed_block_tracks_block_observed2(logger: Logger) {
```

fog/view/server/src/block_tracker.rs
```diff
@@ -593,27 +591,27 @@ mod tests {
...
         }
     }
 
     // A block tracker with a single ingestable range makes progress appropriately
-    // when the start block is greater than zero, and some blocks are processed,
+    // when the start block is greater than zero, and some blocks are observed,
     // then the key is reported lost
     #[test_with_logger]
-    fn highest_fully_processed_block_tracks_block_processed3(logger: Logger) {
+    fn highest_fully_observed_block_tracks_block_observed3(logger: Logger) {
```
nick-mobilecoin, line 614:
```suggestion
    fn highest_observed_block_tracks_block_observed3(logger: Logger) {
```

fog/view/server/src/block_tracker.rs
```diff
@@ -629,38 +627,38 @@ mod tests {
...
-        // processed = highest known.
+        // observed = highest known.
         assert_eq!(block_tracker.highest_known_block_count(), 15);
     }
 
     // A block tracker with a multiple ingestable ranges waits for both of them.
     // When the slow one is marked lost, that unblocks progress.
     #[test_with_logger]
-    fn highest_fully_processed_block_tracks_multiple_recs(logger: Logger) {
+    fn highest_fully_observed_block_tracks_multiple_recs(logger: Logger) {
```
nick-mobilecoin, line 661:
```suggestion
    fn highest_observed_block_tracks_multiple_recs(logger: Logger) {
```

fog/view/server/src/block_tracker.rs
```diff
@@ -759,7 +757,7 @@ mod tests {
     // and if a key is reported lost, it still blocks up until last-scanned for that
     // key is loaded
     #[test_with_logger]
-    fn highest_fully_processed_block_tracks_multiple_recs_some_lost2(logger: Logger) {
+    fn highest_fully_observed_block_tracks_multiple_recs_some_lost2(logger: Logger) {
```
nick-mobilecoin, line 760:
```suggestion
    fn highest_observed_block_tracks_multiple_recs_some_lost2(logger: Logger) {
```

fog/view/server/src/block_tracker.rs
```diff
@@ -837,45 +835,45 @@ mod tests {
...
+            block_tracker.highest_observed_block_count(&[rec1, rec2]),
             expected,
         );
     }
 
     /// A block tracker with a retired key, followed by a gap, followed by a new
     /// key, makes progress
     #[test_with_logger]
-    fn highest_fully_processed_block_tracks_retired_key_followed_by_gap(logger: Logger) {
+    fn highest_fully_observed_block_tracks_retired_key_followed_by_gap(logger: Logger) {
```
nick-mobilecoin, line 876:
```suggestion
    fn highest_observed_block_tracks_retired_key_followed_by_gap(logger: Logger) {
```

fog/view/server/src/block_tracker.rs
```diff
@@ -974,7 +972,7 @@ mod tests {
     /// This is expected to correspond with how we do ingest enclave upgrades,
     /// when everything works.
     #[test_with_logger]
-    fn highest_fully_processed_block_tracks_retired_key_concurrent_with_active(logger: Logger) {
+    fn highest_fully_observed_block_tracks_retired_key_concurrent_with_active(logger: Logger) {
```
nick-mobilecoin, line 975:
```suggestion
    fn highest_observed_block_tracks_retired_key_concurrent_with_active(logger: Logger) {
```

fog/view/server/src/block_tracker.rs
```diff
@@ -1120,7 +1118,7 @@ mod tests {
     /// This is expected to correspond to, doing an ingest enclave upgrade, but
     /// then everything blows up and we have to restart the service later.
     #[test_with_logger]
-    fn highest_fully_processed_block_tracks_retired_key_concurrent_with_active_both_lost(
+    fn highest_fully_observed_block_tracks_retired_key_concurrent_with_active_both_lost(
```
nick-mobilecoin, line 1121:
```suggestion
    fn highest_observed_block_tracks_retired_key_concurrent_with_active_both_lost(
```

fog/view/server/src/db_fetcher.rs
```diff
@@ -296,6 +296,11 @@ where
         );
 
         for (ingress_key, block_index) in next_block_index_per_ingress_key.into_iter() {
+            // Mark that this block has been observed. Do this whether or not we add the
+            // block's TxOuts to ORAM, i.e. "process" the block. This allows us
+            // to tell the client an accurate balance per block height.
+            self.block_tracker.block_observed(ingress_key, block_index);
```
nick-mobilecoin, line 302: It looks like where this is at, if `self.db.get_tx_outs_by_block_and_key(ingress_key, block_index)` fails then the `block_index` will have been observed, but not actually "processed".

Based on this comment
> This allows us to tell the client an accurate balance per block height.

If that failed block happens to correspond to the client then the balance would not be correct.

IFF I'm correct, suggest reverting the "process" -> "observed" rename and make a call to `block_{observerd/processed}` inside of the `should_process_block()` check.

nick-mobilecoin (changes requested):
The comment in `fog/view/server/src/db_fetcher.rs` should probably be looked at first.  I may just need better clarification on the intended behavior

fog/view/server/src/db_fetcher.rs
jcape: Agreed.

jcape (commented):
I'm going to echo some of nick's questions, but also note that if we're going to rename stuff, we should aim for brevity (e.g. `last_highest_observed_block_count` vs. `max_block_height`).

fog/types/src/view.rs
```diff
@@ -51,6 +51,7 @@ pub struct QueryRequest {
 pub struct QueryResponse {
     /// The number of blocks processed by this view enclave at the time that the
     /// request was evaluated.
+    ///  TODO: Rename to highest_observed_block_count
```
garbageslam, line 54: I think it would be better to keep this named as highest processed block count at the API boundary, because the client needs to know basically, what blocks are we telling them we have fully processed, or alternatively, what is the index of the first block we haven't fully processed

For the internal block tracker object it may be a good idea to rename but in the proto offered to the client, they have a different field `last_known_block_count` which is the size of the blockchain right now, and it may become confusing how that is different from `highest_observed_block_count`. The previous name was good because we're telling the client what we have actually processed

fog/view/server/src/db_fetcher.rs
```diff
@@ -328,9 +333,6 @@ where
                     // next block on the next loop iteration.
                     may_have_more_work = true;
 
-                    // Mark that we are done fetching data for this block.
-                    self.block_tracker.block_processed(ingress_key, block_index);
-
```
jcape, line 333: sharding strategy continue goes here

fog/view/enclave/impl/src/lib.rs
samdealy: Done

fog/view/server/src/block_tracker.rs
samdealy: I changed back to processed so this is moot.

fog/view/server/src/db_fetcher.rs
samdealy: Resolved offline. The fix is in the previous PR

fog/view/enclave/impl/src/lib.rs
```diff
@@ -266,26 +266,35 @@ where
                 Error::ProstDecode
             })?;
 
-        shard_query_response.tx_out_search_results =
-            self.get_collated_tx_out_search_results(client_query_request, shard_query_responses)?;
+        let shard_query_responses: Vec<QueryResponse> = shard_query_responses
```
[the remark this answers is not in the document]

fog/view/server/src/block_tracker.rs
```diff
@@ -5,47 +5,48 @@ use mc_crypto_keys::CompressedRistrettoPublic;
...
 /// - Finding out what is the highest block index we have encountered so far.
-/// - Finding out for which block index have we processed data for all ingress
-///   keys, while taking into account ingress keys reported lost
+/// - Finding out for which block index have we observed for all ingress keys,
+///   while taking into account ingress keys reported lost
 pub struct BlockTracker {
-    processed_block_per_ingress_key: HashMap<CompressedRistrettoPublic, u64>,
-    last_highest_processed_block_count: u64,
+    observed_block_per_ingress_key: HashMap<CompressedRistrettoPublic, u64>,
+    last_highest_observed_block_count: u64,
```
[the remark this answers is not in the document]
samdealy: Moot now that we reverted the naming change.

fog/view/enclave/impl/src/lib.rs
```diff
@@ -266,26 +266,35 @@ where
...
             .into_iter()
-            .map(|(responder_id, enclave_message)| {
-                let plaintext_bytes = self.ake.backend_decrypt(responder_id, enclave_message)?;
-                let plaintext_response: QueryResponse = mc_util_serial::decode(&plaintext_bytes)?;
-
-                Ok(plaintext_response.tx_out_search_results)
-            })
-            .collect::<Result<Vec<Vec<TxOutSearchResult>>>>()?
+            .map(|query_response| query_response.tx_out_search_results)
+            .collect::<Vec<Vec<TxOutSearchResult>>>()
```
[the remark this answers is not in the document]

fog/view/enclave/api/src/lib.rs
```diff
@@ -8,7 +8,7 @@
 
 extern crate alloc;
 
-use alloc::{collections::BTreeMap, vec::Vec};
+use alloc::{collections::BTreeMap, string::String, vec::Vec};
```
nick-mobilecoin, line 11: I don't think this change is necessary
```suggestion
use alloc::{collections::BTreeMap, vec::Vec};
```

fog/view/enclave/api/src/lib.rs
```diff
@@ -235,8 +235,8 @@ pub enum Error {
     EnclaveNotInitialized,
     /// Cipher encryption failed: {0}
     Cipher(CipherError),
-    /// Fog View Shard query response collation error
-    QueryResponseCollation,
+    /// Fog View Shard query response collation error: {0}
+    QueryResponseCollation(String),
```
nick-mobilecoin, line 239: this doesn't look to be used, suggest reverting and or removing since the error isn't used
```suggestion
```

nick-mobilecoin (approved):
LGTM

fog/view/enclave/api/src/lib.rs
samdealy: Done

fog/view/enclave/api/src/lib.rs
```diff
@@ -235,8 +235,8 @@ pub enum Error {
     EnclaveNotInitialized,
     /// Cipher encryption failed: {0}
     Cipher(CipherError),
-    /// Fog View Shard query response collation error
-    QueryResponseCollation,
+    /// Fog View Shard query response collation error.
+    QueryResponseCollation(),
```
jcape, line 239: Why do we need trailing parens here?

jcape (approved):
LGTM save the error w/ empty parens...

fog/view/enclave/api/src/lib.rs
samdealy: Typo

PR #2391, opened 2022-08-12
Add support for `digestible(transparent)` on rust enums

garbageslam:
The intention is that transparent enums are not incorporated into
the hash, and digestible tries to jump directly to the data in the
enum.

The point of this is to allow for schema evolutions where a struct
member is replaced with an enum, one of the options for which is
the previous type. This is a valid way to evolve protobufs, but
it's not currently compatible with digestible, until this commit.

This is being pulled out from the masked-amount-v2 PR to ease review.
We also added a bunch of unit tests and schema evolution tests in
this commit.


remoun (approved):
Nicely done!

crypto/digestible/derive/test/tests/enums.rs
```diff
@@ -0,0 +1,364 @@
...
+            101, 107, 59, 55, 249, 37, 221, 203, 13, 36, 193, 85, 104, 62, 192, 193, 243, 144, 224,
+            171, 85, 200, 48, 29, 36, 71, 187, 89, 102, 228, 109, 87
+        ]
+    );
+}
+
+// Test that a C instance of ExampleEnum is being mapped to AST and hashed as
+// expected
+#[test]
+fn trans_enum_c() {
```
nick-mobilecoin, line 196: If I add an empty `E` variant to the enum, this passes
```rust
    let arg = TransEnum::C;
    let expected_ast = ASTNode::from(ASTNone { context: b"C_test" });
    digestible_test_case_ast("C_test", &arg, expected_ast);
    assert_eq!(
        arg.digest32::<MerlinTranscript>(b"C_test"),
        TransEnum::E.digest32::<MerlinTranscript>(b"C_test")
    );
```
I'm not quite clear on when/where the context changes.
Initial thought was that `C` should be part of the digest, but then that would go against the transparent.  I'm not clear on the use for digesting these types, but perhaps there needs to be a warning or check against having more than one `Unit` variant in the same Enum.
Thinking more I'm also getting the impression that multiple variants with the same type or size of unnamed fields would collide.

nick-mobilecoin (changes requested):
I think I need to understand the case for or against `TransEnum::E` == `TransEnum::C` in order to approve

crypto/digestible/README.md
```diff
@@ -503,6 +503,47 @@ struct Thing {
 This feature makes it easy to add fields in a backwards-compatible way without having to wrap them in an `Option<>`.
 For example, for integer fields this can remove the ambiguity of `None` vs `Some(0)`. It mimics the Protobuf behavior, where fields set to zero are omitted.
 
+Transparent Digestible
+----------------------
+
+Sometimes it is necessary to put a wrapper around a type, for instance, to support serialization somehow, or to implement a trait on it.
+Normally, this is a breaking change, because digestible will want to faithfully represent the wrapper in the hash, inserting the type name etc.
+
+To make wrappers that don't cause a breaking change in this manner, simply use the `digestible(transparent)` directive:
```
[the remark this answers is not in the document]
garbageslam: thank you, this is a great improvement

crypto/digestible/derive/test/tests/enums.rs
garbageslam: I'm fine with removing support for the unit variant in `digestible(transparent)`.

It seemed like a good idea when I was making this PR, but I don't think it's actually needed in the PR that depends on this. Rust's protobuf crate uses `Option<Enum>` to represent `oneof` fields.

It may be that the support for unit variants is YAGNI

crypto/digestible/derive/src/lib.rs
```diff
@@ -545,6 +547,101 @@ fn try_digestible_enum(
     Ok(expanded.into())
 }
 
+fn try_digestible_enum_transparent(
+    ident: &Ident,
+    generics: &Generics,
+    variant_data: &DataEnum,
+) -> Result<TokenStream, &'static str> {
```
jcape, line 554: I think there maybe should be some extra checks here around things like all variants must be non-unit and contain distinct, non-transparent, inner types.

jcape (commented):
I feel like this opens something of a hole in how digestible works---like, you can do:

```
pub enun Constraint {
    NotBefore(BlockIndex),
    NotAfter(BlockIndex)
}
```

And get the same hash if the internal variants are the same. Yes, it's opt-in, but we should probably have some kind of safety on this footgun.

crypto/digestible/derive/src/lib.rs
garbageslam: That's an interesting idea

I'm not sure that it is technically possible to check if types are transparent in a proc macro. Proc macros can only see the code for item that the proc macro is tagging directly, it can't go look at the code for another structure.

We might be able to do something where we check if the token streams for the types are all different.
Or, we could like, add a `pub const _IGNORED: _ = { match .. }` to the code gen which will fail to compile if the types are not all different because the match statement would be ambiguous or something

garbageslam:
> I feel like this opens something of a hole in how digestible works

I don't think it's a "hole", I mean, the requirement here is to make this schema evolution possible. I think it's just another tool that should be used with care.

I'm happy to try to reduce foot guns but I'm not sure it's possible to do everything described.

garbageslam:
Another example:

```
#[derive(Digestible)]
struct Foo {
   a: u16,
   b: u16,
}

mod something {
    #[derive(Digestible)]
    struct Foo {
       a: u16,
       b: u16,
    }
}
```

It's already the case that distinct types can have the same hash, even without using the special directives. We don't hash module paths or crate names into the digest. Moreover, it would be a bad thing to do so -- that would make it impossible to refactor our code.

So, I don't see how this is opening up a hole that's more than what currently exists. I think there's just a misconception about what the guarantees of the tool are -- it is not actually a goal that "distinct types always have distinct hashes", and we aren't interested in trying to enforce that somehow.

Using the tool correctly requires thought from the engineers, and as long as they don't do something weird and messy then it should generally work, because it has safe and sane defaults. So I think transparent enums is inline with that design philosophy.

crypto/digestible/derive/src/lib.rs
```diff
@@ -545,6 +547,101 @@ fn try_digestible_enum(
...
+                        })
+                    } else {
+                        panic!("Exactly zero or one unnamed fields may be present when using digestible(transparent) on an enum");
+                    }
+                }
+
+                // For an enum variant that has named fields (e.g. SomeEnum::MyVariant { a: u64, b: u64 },
+                // we cannot digest it transparently.
+                Fields::Named(FieldsNamed { .. }) => {
+                    panic!("Named fields cannot be used when using digestible(transparent) on an enum");
```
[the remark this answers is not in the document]
garbageslam: Right now the tool guarantees that, if a field has a name, then the name makes it into the hash.
If we support transparent when the field is named, then we would be creating an exception to that.

Is there a use-case for this functionality?

crypto/digestible/derive/test/tests/enums.rs
```diff
@@ -0,0 +1,364 @@
...
+        data: vec![3u8, 0u8],
+    });
+    digestible_test_case_ast("A_test", &arg, expected_ast);
+    assert_eq!(
+        arg.digest32::<MerlinTranscript>(b"A_test"),
+        [
+            196, 66, 116, 1, 87, 75, 116, 241, 250, 78, 20, 22, 208, 227, 89, 118, 121, 77, 109,
+            255, 15, 184, 217, 249, 111, 181, 66, 141, 37, 23, 204, 243
+        ]
+    );
```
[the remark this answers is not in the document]
garbageslam: Yeah that seems fine to me

crypto/digestible/derive/src/lib.rs
garbageslam: Okay we have now banned unit variants, and we check if the inner types are the same

crypto/digestible/derive/test/tests/enums.rs
garbageslam: Okay we have now removed unit variants

garbageslam:
@jcape okay, I implemented and tested a fix where, in this scenario, you get an error message:

```
pub enum Constraint {
    NotBefore(BlockIndex),
    NotAfter(BlockIndex)
}
```

```
error: proc-macro derive panicked
  --> crypto/digestible/derive/test/tests/enums.rs:42:10
   |
42 | #[derive(Digestible)]
   |          ^^^^^^^^^^
   |
   = help: message: The types of the 0'th and 1'th variants of Constraint appear to be the same. When using digestible(transparent), this is highly suspect
```

Note that this version is not an error:

```
#[derive(Digestible)]
#[digestible(transparent)]
pub enum Constraint {
    NotBefore(u64),
    NotAfter(BlockIndex)
}
```

Because the code is just comparing tokens and not resolving type aliases etc. I think this should be good enough though.

crypto/digestible/derive/test/tests/enums.rs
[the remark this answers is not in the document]
garbageslam: fixed now

crypto/digestible/derive/src/lib.rs
```diff
@@ -545,6 +547,112 @@ fn try_digestible_enum(
...
+        .variants
+        .iter()
+        .map(|variant| {
+            let variant_ident = &variant.ident;
+
+            // Our behavior differs based on whether the enum variant is a unit (has no data
+            // associated with it), or named (has struct data associated with it), or unnamed (has
+            // tuple data associated with it).
+            //
+            // We can only support unit variants and unnamed tuples of size 1, in which case we skip right to digesting that.
```
nick-mobilecoin, line 565: not sure if there is a good comment to put here, but the "only support unit variants and ..." is now wrong.
```suggestion
```

nick-mobilecoin (approved):

PR #2416, opened 2022-08-22
libmobilecoin changes for MobileCoin-Swift SDK transaction idempotence

carybakker:
### NOTE: 
 * This PR replaces PR #2383 (this is rebased off master and includes PR feedback updates from #2383
 * I still need to validate latest commits with MobileCoin-Swift

### Changes to libmobilecoin/

 * expose ChaCha20Rng for MobileCoin-Swift

### Motivation

To support re-trying transaction submission without violating idempotency, Moby (and other clients) need the ability to re-create identical transaction data via the `prepareTransaction` family of methods on MobileCoinClient.

### In this PR
* Added ability to provide a deterministic RNG to prepareTransaction methods, so the client can re-create the rng state prior to the call, if needed, in order to deterministically re-create identical transaction data.

Pivotal Tracker: [[iOS] Moby Transaction Idempotence](https://www.pivotaltracker.com/story/show/182820695)

### DIFF
--- a/libmobilecoin/include/chacha20_rng.h
+++ b/libmobilecoin/include/chacha20_rng.h
@@ -50,7 +50,7 @@ MC_ATTRIBUTE_NONNULL(1);
 /// # Arguments
 ///
 /// * `chacha20_rng` - must be a valid ChaCha20Rng
-/// * `out_word_pos` - pointer to buffer of 128 bytes where the current
+/// * `out_word_pos` - pointer to buffer of 16 bytes where the current
 ///   chacha20_rng wordpos will be returned
 ///
 /// # Errors
--- a/libmobilecoin/include/chacha20_rng.h
+++ b/libmobilecoin/include/chacha20_rng.h
@@ -54,7 +54,7 @@ MC_ATTRIBUTE_NONNULL(1);
 /// # Errors
 ///
 /// * `LibMcError::Poison`
-void mc_chacha20_rng_get_word_pos(
+bool mc_chacha20_rng_get_word_pos(
     ChaCha20Rng* MC_NULLABLE chacha20_rng,
     const McBuffer* MC_NONNULL out_word_pos,
     McError* MC_NULLABLE * MC_NULLABLE out_error
@@ -72,7 +72,7 @@ MC_ATTRIBUTE_NONNULL(1,2);
 /// # Errors
 ///
 /// * `LibMcError::Poison`
-void mc_chacha20_set_word_pos(
+bool mc_chacha20_rng_set_word_pos(
     ChaCha20Rng* MC_NULLABLE chacha20_rng,
     const McBuffer* MC_NONNULL bytes,
     McError* MC_NULLABLE * MC_NULLABLE out_error
@@ -94,7 +94,7 @@ uint64_t mc_chacha20_rng_next_long(
 )
 MC_ATTRIBUTE_NONNULL(1);
 
-/// frees the ChaCha20Rng
+/// Frees the ChaCha20Rng
 ///
 /// # Preconditions
 /// 
@@ -105,7 +105,10 @@ MC_ATTRIBUTE_NONNULL(1);
 /// * `chacha20_rng` - must be a valid ChaCha20Rng
 ///
 /// * `chacha20_rng` - must be a valid ChaCha20Rng
-void mc_chacha20_rng_free(ChaCha20Rng* MC_NULLABLE chacha20_rng)
+bool mc_chacha20_rng_free(
+    ChaCha20Rng* MC_NULLABLE chacha20_rng,
+    McError* MC_NULLABLE * MC_NULLABLE out_error
+)
 MC_ATTRIBUTE_NONNULL(1);
 
 #ifdef __cplusplus
--- a/libmobilecoin/include/chacha20_rng.h
+++ b/libmobilecoin/include/chacha20_rng.h
@@ -105,11 +105,8 @@ MC_ATTRIBUTE_NONNULL(1);
 /// # Arguments
 ///
 /// * `chacha20_rng` - must be a valid ChaCha20Rng
-///
-/// * `chacha20_rng` - must be a valid ChaCha20Rng
 bool mc_chacha20_rng_free(
-    ChaCha20Rng* MC_NONNULL chacha20_rng,
-    McError* MC_NULLABLE * MC_NULLABLE out_error
+    ChaCha20Rng* MC_NONNULL chacha20_rng
 )
 MC_ATTRIBUTE_NONNULL(1);
 
--- a/libmobilecoin/include/chacha20_rng.h
+++ b/libmobilecoin/include/chacha20_rng.h
@@ -11,18 +11,35 @@ extern "C" {
 
 typedef struct _ChaCha20Rng ChaCha20Rng;
 
-ChaCha20Rng* MC_NULLABLE mc_chacha20_rng_create_with_long(uint64_t value);
+ChaCha20Rng* MC_NULLABLE mc_chacha20_rng_create_with_long(
+    uint64_t value,
+    McError* MC_NULLABLE * MC_NULLABLE out_error
+);
 
-ChaCha20Rng* MC_NULLABLE mc_chacha20_rng_create_with_bytes(const McBuffer* MC_NONNULL bytes)
+ChaCha20Rng* MC_NULLABLE mc_chacha20_rng_create_with_bytes(
+    const McBuffer* MC_NONNULL bytes,
+    McError* MC_NULLABLE * MC_NULLABLE out_error
+)
 MC_ATTRIBUTE_NONNULL(1);
 
-void mc_chacha20_rng_get_word_pos(ChaCha20Rng* MC_NULLABLE chacha20_rng, const McBuffer* MC_NONNULL out_word_pos)
+void mc_chacha20_rng_get_word_pos(
+    ChaCha20Rng* MC_NULLABLE chacha20_rng,
+    const McBuffer* MC_NONNULL out_word_pos,
+    McError* MC_NULLABLE * MC_NULLABLE out_error
+)
 MC_ATTRIBUTE_NONNULL(1,2);
 
-void mc_chacha20_set_word_pos(ChaCha20Rng* MC_NULLABLE chacha20_rng, const McBuffer* MC_NONNULL bytes)
+void mc_chacha20_set_word_pos(
+    ChaCha20Rng* MC_NULLABLE chacha20_rng,
+    const McBuffer* MC_NONNULL bytes,
+    McError* MC_NULLABLE * MC_NULLABLE out_error
+)
 MC_ATTRIBUTE_NONNULL(1,2);
 
-uint64_t mc_chacha20_rng_next_long(ChaCha20Rng* MC_NULLABLE chacha20_rng)
+uint64_t mc_chacha20_rng_next_long(
+    ChaCha20Rng* MC_NULLABLE chacha20_rng,
+    McError* MC_NULLABLE * MC_NULLABLE out_error
+)
 MC_ATTRIBUTE_NONNULL(1);
 
 void mc_chacha20_rng_free(ChaCha20Rng* MC_NULLABLE chacha20_rng)
--- a/libmobilecoin/include/chacha20_rng.h
+++ b/libmobilecoin/include/chacha20_rng.h
@@ -1,0 +1,35 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
+#ifndef MC_CHACHA20_RNG_H_
+#define MC_CHACHA20_RNG_H_
+
+#include "common.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef struct _ChaCha20Rng ChaCha20Rng;
+
+ChaCha20Rng* MC_NULLABLE mc_chacha20_rng_create_with_long(uint64_t value);
+
+ChaCha20Rng* MC_NULLABLE mc_chacha20_rng_create_with_bytes(const McBuffer* MC_NONNULL bytes)
+MC_ATTRIBUTE_NONNULL(1);
+
+void mc_chacha20_rng_get_word_pos(ChaCha20Rng* MC_NULLABLE chacha20_rng, const McBuffer* MC_NONNULL out_word_pos)
+MC_ATTRIBUTE_NONNULL(1,2);
+
+void mc_chacha20_set_word_pos(ChaCha20Rng* MC_NULLABLE chacha20_rng, const McBuffer* MC_NONNULL bytes)
+MC_ATTRIBUTE_NONNULL(1,2);
+
+uint64_t mc_chacha20_rng_next_long(ChaCha20Rng* MC_NULLABLE chacha20_rng)
+MC_ATTRIBUTE_NONNULL(1);
+
+void mc_chacha20_rng_free(ChaCha20Rng* MC_NULLABLE chacha20_rng)
+MC_ATTRIBUTE_NONNULL(1);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* MC_CHACHA20_RNG_H_ */
--- a/libmobilecoin/include/chacha20_rng.h
+++ b/libmobilecoin/include/chacha20_rng.h
@@ -11,17 +11,49 @@ extern "C" {
 
 typedef struct _ChaCha20Rng ChaCha20Rng;
 
+/// Returns a new ChaCha20Rng instance initialized with the
+/// seed value provided by the u64 long_val parameter
+///
+/// # Arguments
+///
+/// * `long_val` - an unsigned 64 bit value to use as the rng seed
+///
+/// # Errors
+///
+/// * `LibMcError::Poison`
 ChaCha20Rng* MC_NULLABLE mc_chacha20_rng_create_with_long(
     uint64_t value,
     McError* MC_NULLABLE * MC_NULLABLE out_error
 );
 
+/// Returns a new ChaCha20Rng instance initialized with the
+/// seed value provided by the bytes data, which must be at
+/// least 32 bytes (only the first 32 bytes will be used)
+///
+/// # Arguments
+///
+/// * `bytes` - 32 bytes of data to use as the rng seed
+///
+/// # Errors
+///
+/// * `LibMcError::Poison`
 ChaCha20Rng* MC_NULLABLE mc_chacha20_rng_create_with_bytes(
     const McBuffer* MC_NONNULL bytes,
     McError* MC_NULLABLE * MC_NULLABLE out_error
 )
 MC_ATTRIBUTE_NONNULL(1);
 
+/// Returns the current word_pos of the ChaCha20Rng instance
+///
+/// # Arguments
+///
+/// * `chacha20_rng` - must be a valid ChaCha20Rng
+/// * `out_word_pos` - pointer to buffer of 128 bytes where the current
+///   chacha20_rng wordpos will be returned
+///
+/// # Errors
+///
+/// * `LibMcError::Poison`
 void mc_chacha20_rng_get_word_pos(
     ChaCha20Rng* MC_NULLABLE chacha20_rng,
     const McBuffer* MC_NONNULL out_word_pos,
@@ -29,6 +61,17 @@ void mc_chacha20_rng_get_word_pos(
 )
 MC_ATTRIBUTE_NONNULL(1,2);
 
+/// Sets the current word_pos of the ChaCha20Rng instance
+///
+/// /// # Arguments
+///
+/// * `chacha20_rng` - must be a valid ChaCha20Rng
+/// * `out_word_pos` - pointer to buffer of 128 bytes where the current
+///   chacha20_rng wordpos will be returned
+///
+/// # Errors
+///
+/// * `LibMcError::Poison`
 void mc_chacha20_set_word_pos(
     ChaCha20Rng* MC_NULLABLE chacha20_rng,
     const McBuffer* MC_NONNULL bytes,
@@ -36,12 +79,32 @@ void mc_chacha20_set_word_pos(
 )
 MC_ATTRIBUTE_NONNULL(1,2);
 
+/// Returns the next random u64 value from the ChaCha20Rng
+///
+/// /// # Arguments
+///
+/// * `chacha20_rng` - must be a valid ChaCha20Rng
+///
+/// # Errors
+///
+/// * `LibMcError::Poison`
 uint64_t mc_chacha20_rng_next_long(
     ChaCha20Rng* MC_NULLABLE chacha20_rng,
     McError* MC_NULLABLE * MC_NULLABLE out_error
 )
 MC_ATTRIBUTE_NONNULL(1);
 
+/// frees the ChaCha20Rng
+///
+/// # Preconditions
+/// 
+/// * The ChaCha20Rng is no longer in use
+/// 
+/// # Arguments
+///
+/// * `chacha20_rng` - must be a valid ChaCha20Rng
+///
+/// * `chacha20_rng` - must be a valid ChaCha20Rng
 void mc_chacha20_rng_free(ChaCha20Rng* MC_NULLABLE chacha20_rng)
 MC_ATTRIBUTE_NONNULL(1);
 
--- a/libmobilecoin/include/chacha20_rng.h
+++ b/libmobilecoin/include/chacha20_rng.h
@@ -106,9 +106,8 @@ MC_ATTRIBUTE_NONNULL(1);
 ///
 /// * `chacha20_rng` - must be a valid ChaCha20Rng
 bool mc_chacha20_rng_free(
-    ChaCha20Rng* MC_NONNULL chacha20_rng
-)
-MC_ATTRIBUTE_NONNULL(1);
+    ChaCha20Rng* MC_NULLABLE chacha20_rng
+);
 
 #ifdef __cplusplus
 }
--- a/libmobilecoin/include/chacha20_rng.h
+++ b/libmobilecoin/include/chacha20_rng.h
@@ -5,6 +5,8 @@
 
 #include "common.h"
 
+/* ==================== ChaCha20Rng ==================== */
+
 #ifdef __cplusplus
 extern "C" {
 #endif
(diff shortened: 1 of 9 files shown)


libmobilecoin/include/chacha20_rng.h
```diff
@@ -0,0 +1,35 @@
...
+
+#include "common.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef struct _ChaCha20Rng ChaCha20Rng;
+
+ChaCha20Rng* MC_NULLABLE mc_chacha20_rng_create_with_long(uint64_t value);
```
the-real-adammork, line 14: missing function documentation comments, also formatting looks a lil different from other header files should be adjusted to match style in other files.

libmobilecoin/src/chacha20_rng.rs
```diff
@@ -0,0 +1,184 @@
...
+
+/// frees the ChaCha20Rng
+///
+/// # Preconditions
+/// 
+/// * The ChaCha20Rng is no longer in use
+/// 
+/// # Arguments
+///
+/// * `chacha20_rng` - must be a valid ChaCha20Rng
```
the-real-adammork, line 176: duplicate

libmobilecoin/include/chacha20_rng.h
```diff
@@ -0,0 +1,118 @@
...
+///
+/// * `chacha20_rng` - must be a valid ChaCha20Rng
+/// * `out_word_pos` - pointer to buffer of 128 bytes where the current
+///   chacha20_rng wordpos will be returned
+///
+/// # Errors
+///
+/// * `LibMcError::Poison`
+bool mc_chacha20_rng_get_word_pos(
+    ChaCha20Rng* MC_NULLABLE chacha20_rng,
```
remoun, line 58: Sounds like this is not nullable
```suggestion
    ChaCha20Rng* MC_NONNULL chacha20_rng,
```
(throughout this PR)

libmobilecoin/src/chacha20_rng.rs
```diff
@@ -0,0 +1,188 @@
...
+#[no_mangle]
+pub extern "C" fn mc_chacha20_rng_create_with_bytes(
+    bytes: FfiRefPtr<McBuffer>,
+    out_error: FfiOptMutPtr<FfiOptOwnedPtr<McError>>,
+) -> FfiOptOwnedPtr<Mutex<McChaCha20Rng>> {
+    ffi_boundary_with_error(out_error, || {
+        let bytes: [u8; 32] = bytes
+            .as_slice()
+            .try_into()
+            .expect("seed size must be 32 bytes");
```
remoun, line 85: I'm not sure panicking/`expect` is the right thing to do here. `ffi_boundary_with_error` does handle panics, but I'm not sure what the result would be, let alone whether it's reasonable when a caller passes < 32 bytes

libmobilecoin/src/chacha20_rng.rs
```diff
@@ -0,0 +1,188 @@
...
+
+pub type McChaCha20Rng = ChaCha20Rng;
+
+// McU128 facilitates conversion between u128 and 16 bytes of u8
+pub struct McU128 {
+    pub bytes: [u8; 16],
+}
+
+impl McU128 {
+    pub fn from_u128(val: u128) -> McU128 {
```
remoun, line 17: Consider adding `new`/`from_bytes`

```suggestion
    pub fn new(bytes: [u8; 16]) -> McU128 {
        McU128 { bytes }
    }

    pub fn from_u128(val: u128) -> McU128 {
```

libmobilecoin/include/chacha20_rng.h
```diff
@@ -0,0 +1,118 @@
...
+/// 
+/// * The ChaCha20Rng is no longer in use
+/// 
+/// # Arguments
+///
+/// * `chacha20_rng` - must be a valid ChaCha20Rng
+///
+/// * `chacha20_rng` - must be a valid ChaCha20Rng
+bool mc_chacha20_rng_free(
+    ChaCha20Rng* MC_NULLABLE chacha20_rng,
```
jcape, line 109: I'm less certain about this... in standards-compliant C, `free(NULL)` is a no-op, but if this is our convention elsewhere in `libmobilecoin` we should keep it:

```suggestion
    ChaCha20Rng* MC_NONNULL chacha20_rng,
```

libmobilecoin/src/common/buffer.rs
```diff
@@ -182,6 +182,26 @@ impl AsMut<[u8]> for McMutableBuffer<'_> {
...
+}
+
+impl<'a> TryFromFfi<&McBuffer<'a>> for [u8; 16] {
+    type Error = LibMcError;
+
+    #[inline]
+    fn try_from_ffi(src: &McBuffer<'a>) -> Result<Self, LibMcError> {
+        Ok(*<&'a [u8; 16]>::try_from_ffi(src)?)
+    }
+}
```
remoun, line 203: Optional/Equivalent: Replace all the variants (incl. the ones below) with const generics
```suggestion
impl<'a, const N: usize> TryFromFfi<&McBuffer<'a>> for &'a [u8; N] {
    type Error = LibMcError;

    #[inline]
    fn try_from_ffi(src: &McBuffer<'a>) -> Result<Self, LibMcError> {
        let src = src.as_slice_of_len(N)?;
        // SAFETY: ok to unwrap because we just checked length
        Ok(<&[u8; N]>::try_from(src).unwrap())
    }
}

impl<'a, const N: usize> TryFromFfi<&McBuffer<'a>> for [u8; N] {
    type Error = LibMcError;

    #[inline]
    fn try_from_ffi(src: &McBuffer<'a>) -> Result<Self, LibMcError> {
        Ok(*<&'a [u8; N]>::try_from_ffi(src)?)
    }
}
```

libmobilecoin/src/chacha20_rng.rs
```diff
@@ -0,0 +1,188 @@
...
+        Ok(Mutex::new(McChaCha20Rng::from_seed(bytes)))
+    })
+}
+
+/// Returns the current word_pos of the ChaCha20Rng instance
+///
+/// # Arguments
+///
+/// * `chacha20_rng` - must be a valid ChaCha20Rng
+/// * `out_word_pos` - pointer to buffer of 128 bytes where the current
```
jcape, line 95:
```suggestion
/// * `out_word_pos` - pointer to buffer of 16 byteswhere the current
```

remoun (approved):

libmobilecoin/src/chacha20_rng.rs
```diff
@@ -0,0 +1,188 @@
...
+        }
+    }
+
+    #[inline]
+    fn into_ffi(self) -> McU128 {
+        self
+    }
+}
+
+impl_into_ffi!(FfiOwnedPtr<McU128>);
```
jcape, line 42: Is this actually necessary? Can this just be Rust's u128?

libmobilecoin/src/chacha20_rng.rs
```diff
@@ -0,0 +1,188 @@
...
+    out_word_pos: FfiMutPtr<McMutableBuffer>,
+    out_error: FfiOptMutPtr<FfiOptOwnedPtr<McError>>,
+) -> bool {
+    ffi_boundary_with_error(out_error, || {
+        let word_pos = chacha20_rng.lock()?.get_word_pos();
+        let mc_u128 = McU128::from_u128(word_pos);
+
+        let out_word_pos = out_word_pos.into_mut().as_slice_mut_of_len(16)?;
+
+        out_word_pos.copy_from_slice(&mc_u128.bytes);
```
jcape, line 113:
```suggestion
        let out_word_pos = out_word_pos.into_mut().as_slice_mut_of_len(16)?;
        out_word_pos.copy_from_slice(&word_pos.to_be_bytes());
```

libmobilecoin/src/chacha20_rng.rs
```diff
@@ -0,0 +1,188 @@
...
+        let mc_u128 = McU128 {
+            bytes: bytes
+                .as_slice()
+                .try_into()
+                .expect("word_pos length is not exaclty 16 bytes"),
+        };
+        let word_pos = mc_u128.to_u128();
+
+        chacha20_rng.lock()?.set_word_pos(word_pos);
+
```
jcape, line 146: This may not compile/pass lint as-is, but you don't actually need the intermediate `McU128` type here, you can just use the u128 type directly, e.g.:

```suggestion
        chacha20_rng
            .lock()?
            .set_word_pos(u128::from_be_bytes(bytes
                .as_slice()
                .try_into()
                .expect("word_pos length is not exaclty 16 bytes"),
            );
```

jcape:
Lint failure is due to the `[dependencies]` not being sorted:
```
error: Dependencies for libmobilecoin are not sorted
```

libmobilecoin/src/common/buffer.rs
jcape: Agreed

libmobilecoin/include/chacha20_rng.h
carybakker: Added comments in a subsequent commit, and formatting was updated in a subsequent commit as well.
Adding the /* ======= NAME ========= */ comment line under #include "common.h" (last change not committed yet b/c want to avoid having to cherry-pick too much... so going to bundle the review updates together).

libmobilecoin/src/chacha20_rng.rs
carybakker: got it

libmobilecoin/include/chacha20_rng.h
carybakker: Yes - thank you!

libmobilecoin/include/chacha20_rng.h
```diff
@@ -0,0 +1,35 @@
...
+MC_ATTRIBUTE_NONNULL(1,2);
+
+void mc_chacha20_set_word_pos(ChaCha20Rng* MC_NULLABLE chacha20_rng, const McBuffer* MC_NONNULL bytes)
+MC_ATTRIBUTE_NONNULL(1,2);
+
+uint64_t mc_chacha20_rng_next_long(ChaCha20Rng* MC_NULLABLE chacha20_rng)
+MC_ATTRIBUTE_NONNULL(1);
+
+void mc_chacha20_rng_free(ChaCha20Rng* MC_NULLABLE chacha20_rng)
+MC_ATTRIBUTE_NONNULL(1);
```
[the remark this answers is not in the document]
carybakker: Yes - the ChaCha20Rng should be NONNULL... Reymoun flagged that earlier, so that's been updated.  Also, you're right about the out_error param - I realized that after the initial submission and added in a subsequent commit.  Will have the NONNULL change pushed in a bit - want to go through all the comments first so I don't have to cherry pick for release/v2 too much.  Thanks!

libmobilecoin/include/chacha20_rng.h
```diff
@@ -0,0 +1,115 @@
...
+///
+/// * `chacha20_rng` - must be a valid ChaCha20Rng
+/// * `out_word_pos` - pointer to buffer of 128 bytes where the current
+///   chacha20_rng wordpos will be returned
+///
+/// # Errors
+///
+/// * `LibMcError::Poison`
+void mc_chacha20_rng_get_word_pos(
+    ChaCha20Rng* MC_NULLABLE chacha20_rng,
```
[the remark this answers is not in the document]
carybakker: Yup - Reymoun saw that too.

libmobilecoin/src/chacha20_rng.rs
carybakker: Yup - meant to write 'bits', but I think 16 bytes makes more sense.

libmobilecoin/src/chacha20_rng.rs
carybakker: I was thinking the same thing and was going to try to clean this out.  Originally I had used McU128 as a parameter type, but started pulling it out to get rid of it - but still had it here to convert to/from [u8; 16]

libmobilecoin/src/chacha20_rng.rs
carybakker: will update.

libmobilecoin/src/chacha20_rng.rs
carybakker: Updating

libmobilecoin/src/chacha20_rng.rs
carybakker: McU128 isn't really necessary - removed it after making the change below from one of James' review comments removing the dependency.

libmobilecoin/src/chacha20_rng.rs
carybakker: Removed - thanks!

libmobilecoin/src/common/buffer.rs
carybakker: Nice! Thanks Reymoun!

libmobilecoin/src/chacha20_rng.rs
```diff
@@ -0,0 +1,144 @@
...
+#[no_mangle]
+pub extern "C" fn mc_chacha20_rng_create_with_bytes(
+    bytes: FfiRefPtr<McBuffer>,
+    out_error: FfiOptMutPtr<FfiOptOwnedPtr<McError>>,
+) -> FfiOptOwnedPtr<Mutex<McChaCha20Rng>> {
+    ffi_boundary_with_error(out_error, || {
+        let bytes: [u8; 32] = bytes
+            .as_slice()
+            .try_into()
+            .expect("seed size must be 32 bytes");
```
iamalwaysuncomfortable, line 51: `bytes.as_slice().try_into()` is very general and spans beyond the case of rngs, so I understand why the expect was used here, but if you'd want to go for a more informative error for this specialized case, it could look something like: 

```suggestion
    ffi_boundary_with_error(out_error, || {
        let bytes: [u8; 32] = bytes
            .as_slice()
            .try_into()
            .map_err(|_| LibMcError::RngInitialization("seed size must be 32 bytes"))?
```

iamalwaysuncomfortable (commented):
Some comments on the panic vs. error questoin

libmobilecoin/src/chacha20_rng.rs
```diff
@@ -0,0 +1,144 @@
...
+    chacha20_rng: FfiMutPtr<Mutex<McChaCha20Rng>>,
+    bytes: FfiRefPtr<McBuffer>,
+    out_error: FfiOptMutPtr<FfiOptOwnedPtr<McError>>,
+) -> bool {
+    ffi_boundary_with_error(out_error, || {
+        chacha20_rng.lock()?.set_word_pos(u128::from_be_bytes(
+            bytes
+                .as_slice()
+                .try_into()
+                .expect("word_pos length is not exaclty 16 bytes"),
```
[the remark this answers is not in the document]
iamalwaysuncomfortable:
```suggestion
            bytes
                .as_slice()
                .try_into()
                .map_err(|_| LibMcError::RngError("word_pos length is not exactly 16 bytes"))?;
```

libmobilecoin/include/chacha20_rng.h
```diff
@@ -0,0 +1,120 @@
...
+    McError* MC_NULLABLE * MC_NULLABLE out_error
+)
+MC_ATTRIBUTE_NONNULL(1);
+
+/// Returns the current word_pos of the ChaCha20Rng instance
+///
+/// # Arguments
+///
+/// * `chacha20_rng` - must be a valid ChaCha20Rng
+/// * `out_word_pos` - pointer to buffer of 128 bytes where the current
```
remoun, line 53:
```suggestion
/// * `out_word_pos` - pointer to buffer of 16 bytes where the current
```

libmobilecoin/src/chacha20_rng.rs
```diff
@@ -0,0 +1,144 @@
+use crate::common::*;
+use mc_util_ffi::*;
```
remoun, line 2: MC style guide prohibits glob imports. Please pull specific deps.

libmobilecoin/src/chacha20_rng.rs
```diff
@@ -0,0 +1,144 @@
...
+use rand_chacha::ChaCha20Rng;
+use rand_core::{RngCore, SeedableRng};
+use std::{convert::TryInto, sync::Mutex};
```
remoun, line 5: Nit: `TryInto` is now part of the standard preamble for Rust edition 2021.
```suggestion
use std::sync::Mutex;
```

libmobilecoin/src/chacha20_rng.rs
```diff
@@ -0,0 +1,144 @@
...
+/// # Preconditions
+///
+/// * The ChaCha20Rng is no longer in use
+///
+/// # Arguments
+///
+/// * `chacha20_rng` - must be a valid ChaCha20Rng
+#[no_mangle]
+pub extern "C" fn mc_chacha20_rng_free(
+    chacha20_rng: FfiOptOwnedPtr<Mutex<McChaCha20Rng>>,
```
remoun, line 137: Why is this type different from the other methods? The corresponding C/++ signature looks like its peers.
```suggestion
    chacha20_rng: FfiMutPtr<Mutex<McChaCha20Rng>>,
```

libmobilecoin/src/chacha20_rng.rs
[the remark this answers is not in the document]
carybakker: I tried the map_err here, and got a compilation error from type mismatch that was giving me some troubht... so I just added a basic length check for now... The current code will pre-empt the panic and return the error as expected, but could use some help rustifying the code a bit

libmobilecoin/src/chacha20_rng.rs
```diff
@@ -0,0 +1,144 @@
...
+///
+/// # Errors
+///
+/// * `LibMcError::Poison`
+#[no_mangle]
+pub extern "C" fn mc_chacha20_rng_next_long(
+    chacha20_rng: FfiMutPtr<Mutex<McChaCha20Rng>>,
+    out_error: FfiOptMutPtr<FfiOptOwnedPtr<McError>>,
+) -> u64 {
+    ffi_boundary_with_error(out_error, || Ok(chacha20_rng.lock()?.next_u64()))
```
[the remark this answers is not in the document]
carybakker: I think this one was from James :)

libmobilecoin/include/chacha20_rng.h
carybakker: thanks!

libmobilecoin/src/chacha20_rng.rs
carybakker: thanks - didn't know..will update for these changes...a lot of the other libmobilecoin files have glob imports too so I was modeling after them... can try to clean the others up a little later too.

libmobilecoin/src/chacha20_rng.rs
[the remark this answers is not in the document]
carybakker: ah, I think I got it...will update shortly.

libmobilecoin/src/chacha20_rng.rs
carybakker: I believe this may be because of the rust ownership - to free the reference.  Does that make sense to you as well?  I was modeling this code after another class where the free was implemented with FfiOptOwnedPtr for the param, and assumed it was for the ownership.

carybakker: But the C func does need to match...will update that.

libmobilecoin/src/chacha20_rng.rs
carybakker: updated.

libmobilecoin/src/chacha20_rng.rs
carybakker: updated...

libmobilecoin/src/chacha20_rng.rs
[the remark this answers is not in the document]
carybakker: updated.

libmobilecoin/src/chacha20_rng.rs
carybakker: Confirmed w/Remoun that this looks correct.  Thanks @remoun

libmobilecoin/include/chacha20_rng.h
```diff
@@ -0,0 +1,120 @@
...
+/// # Arguments
+///
+/// * `chacha20_rng` - must be a valid ChaCha20Rng
+///
+/// * `chacha20_rng` - must be a valid ChaCha20Rng
+bool mc_chacha20_rng_free(
+    ChaCha20Rng* MC_NONNULL chacha20_rng,
+    McError* MC_NULLABLE * MC_NULLABLE out_error
+)
+MC_ATTRIBUTE_NONNULL(1);
```
remoun, line 114: It seems the convention for `_free` methods is to take one nullable pointer
```suggestion
bool mc_chacha20_rng_free(
    ChaCha20Rng* MC_NULLABLE chacha20_rng,
);
```

libmobilecoin/libmobilecoin_cbindgen.h
```diff
@@ -381,6 +385,100 @@ ssize_t mc_bip39_entropy_from_mnemonic(FfiStr mnemonic,
...
+ * # Preconditions
+ *
+ * * The ChaCha20Rng is no longer in use
+ *
+ * # Arguments
+ *
+ * * `chacha20_rng` - must be a valid ChaCha20Rng
+ */
+bool mc_chacha20_rng_free(FfiOptOwnedPtr<Mutex<McChaCha20Rng>> chacha20_rng,
+                          FfiOptMutPtr<FfiOptOwnedPtr<McError>> out_error);
```
remoun, line 480:
```suggestion
void mc_chacha20_rng_free(FfiOptOwnedPtr<Mutex<McChaCha20Rng>> chacha20_rng);
```

libmobilecoin/src/chacha20_rng.rs
```diff
@@ -0,0 +1,149 @@
...
+/// * `chacha20_rng` - must be a valid ChaCha20Rng
+#[no_mangle]
+pub extern "C" fn mc_chacha20_rng_free(
+    chacha20_rng: FfiOptOwnedPtr<Mutex<McChaCha20Rng>>,
+    out_error: FfiOptMutPtr<FfiOptOwnedPtr<McError>>,
+) -> bool {
+    ffi_boundary_with_error(out_error, || {
+        let _ = chacha20_rng;
+        Ok(())
+    })
```
remoun, line 148:
```suggestion
) {
    ffi_boundary(|| {
        let _ = chacha20_rng;
    })
```

libmobilecoin/libmobilecoin_cbindgen.h
```diff
@@ -381,6 +385,100 @@ ssize_t mc_bip39_entropy_from_mnemonic(FfiStr mnemonic,
...
+ *
+ * # Arguments
+ *
+ * * `long_val` - an unsigned 64 bit value to use as the rng seed
+ *
+ * # Errors
+ *
+ * * `LibMcError::Poison`
+ */
+FfiOptOwnedPtr<Mutex<McChaCha20Rng>> mc_chacha20_rng_create_with_long(uint64_t long_val,
```
remoun, line 400: I notice none of the other bindings use `Mutex`, including ones that are definitely not thread-safe. Perhaps the intent is that the thread synchronizations happen on the Java/caller side?

carybakker: In our case, we definitely want this to be thread safe... Not sure of the other C APIs that expose rust instances - it's possible they are all explicitly not threadsafe b/c all the processing that uses the rust instances are within a single thread, or possibly they SHOULD be threadsafe w/a mutex...  In this case, I know we need to be thread-safe.

Resolving this after our Slack chat :)

libmobilecoin/libmobilecoin_cbindgen.h
carybakker: Removing error param for free 👍

libmobilecoin/src/chacha20_rng.rs
carybakker: Updating as above - no need for the Ok(()) w/out error param (existing transaction.rs also does not rely on error param for its free call)

libmobilecoin/include/chacha20_rng.h
carybakker: Updating to match existing free convention of just having the pointer.... however, keeping it non-null b/c I think the others should too :)  All the other funcs with object params have them non-null, not sure why free would be different. 🤷‍♂️

libmobilecoin/include/chacha20_rng.h
```diff
@@ -0,0 +1,117 @@
...
+/// 
+/// * The ChaCha20Rng is no longer in use
+/// 
+/// # Arguments
+///
+/// * `chacha20_rng` - must be a valid ChaCha20Rng
+bool mc_chacha20_rng_free(
+    ChaCha20Rng* MC_NONNULL chacha20_rng
+)
+MC_ATTRIBUTE_NONNULL(1);
```
remoun, line 111: The convention seems to be to take one nullable arg and return void; this matches the cbindgen-generated declaration

```suggestion
void mc_chacha20_rng_free(ChaCha20Rng* MC_NULLABLE chacha20_rng;
```

remoun (approved):

libmobilecoin/include/chacha20_rng.h
remoun: Oops, just noticed your reply. I think it's nullable since `_create` returns a nullable pointer

libmobilecoin/include/chacha20_rng.h
carybakker: I mean... it can be changed to be nullable, but does that really make sense?  I think the others should actually be changed to be non-null....

@the-real-adammork  thoughts?

carybakker: The cbindgen doesn't seem to specify null/not null... just has FfiOptOwnedPtr.

carybakker: After Slack convo, going to stick with non-null as there's no valid use of free with a nullable value :)   Thanks for all the help Remoun 👍

the-real-adammork (approved):

libmobilecoin/include/chacha20_rng.h
jcape: I mentioned in my review, I think that having `free()` take `NULL` is the generally expected behavior in C, at least on modern platforms/compilers.

In terms of what cbindgen is specifying, I assume the `Opt` in `FfiOptOwnedPtr` refers to `Option<Box<T>>`. Like, `Box<T>` can never be `NULL`, but `Option` is defined as:

```rust
pub enum Option<B> {
    None,
    Some(B),
}
```

Recent Rust is smart enough to know that `Option<Box<T>>::None` is a way to represent a nullable pointer without giving anyone the ability to dereference it when it's `NULL`, so that's how it puts it together.

PR #2417, opened 2022-08-22
libmobilecoin changes for MobileCoin-Swift SDK transaction idempotence

carybakker:
### Changes to libmobilecoin/

 * expose ChaCha20Rng for MobileCoin-Swift

### Motivation

To support re-trying transaction submission without violating idempotency, Moby (and other clients) need the ability to re-create identical transaction data via the `prepareTransaction` family of methods on MobileCoinClient.

### In this PR
* Added ability to provide a deterministic RNG to prepareTransaction methods, so the client can re-create the rng state prior to the call, if needed, in order to deterministically re-create identical transaction data.

Pivotal Tracker: [[iOS] Moby Transaction Idempotence](https://www.pivotaltracker.com/story/show/182820695)

### DIFF
--- a/libmobilecoin/include/chacha20_rng.h
+++ b/libmobilecoin/include/chacha20_rng.h
@@ -50,7 +50,7 @@ MC_ATTRIBUTE_NONNULL(1);
 /// # Arguments
 ///
 /// * `chacha20_rng` - must be a valid ChaCha20Rng
-/// * `out_word_pos` - pointer to buffer of 128 bytes where the current
+/// * `out_word_pos` - pointer to buffer of 16 bytes where the current
 ///   chacha20_rng wordpos will be returned
 ///
 /// # Errors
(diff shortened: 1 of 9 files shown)


libmobilecoin/include/chacha20_rng.h
```diff
@@ -0,0 +1,116 @@
...
+/// Frees the ChaCha20Rng
+///
+/// # Preconditions
+/// 
+/// * The ChaCha20Rng is no longer in use
+/// 
+/// # Arguments
+///
+/// * `chacha20_rng` - must be a valid ChaCha20Rng
+bool mc_chacha20_rng_free(
```
remoun, line 108: This should return `void`
```suggestion
void mc_chacha20_rng_free(
```

remoun (approved):

the-real-adammork (approved):

PR #2432, opened 2022-08-24
[Added] CI/CD - GHA CD for master part 2

jgreat:
Add 1.1.3 bv0 -> 2.0.0 bv0 -> 2.0.0 bv2 -> current bv2 -> current bv3 upgrade testing.

- update job names to match upgrade path.
- create block version 3 set of tests
- reduce redundant tests between upgrade stages
- enhance fog-ingest activate/retire to use test_client to advance chain for old ingest retirement (mostly in gha-k8s-toolbox)
- minor fix in wrapper-ledger-distribution.sh to reduce the chance of confusing the state file path (thanks Joe!).
- smarter minting config tx test - wait for chain to advance before moving on.

### DIFF
--- a/.github/workflows/mobilecoin-workflow-dev-test.yaml
+++ b/.github/workflows/mobilecoin-workflow-dev-test.yaml
@@ -67,6 +67,11 @@ on:
         type: string
         required: false
         default: 'true'
+      testing_block_v3:
+        description: "Run block v3 tests"
+        type: string
+        required: false
+        default: 'true'
     secrets:
       RANCHER_CLUSTER:
         description: "Rancher cluster name"
@@ -98,6 +103,7 @@ jobs:
         echo fog_distribution ${{ inputs.fog_distribution }}
         echo testing_block_v0 ${{ inputs.testing_block_v0 }}
         echo testing_block_v2 ${{ inputs.testing_block_v2 }}
+        echo testing_block_v3 ${{ inputs.testing_block_v3 }}
 
   setup-environment:
     runs-on: [self-hosted, Linux, small]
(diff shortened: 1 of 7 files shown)


.github/workflows/mobilecoin-workflow-dev-test.yaml
```diff
@@ -345,3 +351,139 @@ jobs:
...
+        namespace: ${{ inputs.namespace }}
+        rancher_cluster: ${{ secrets.RANCHER_CLUSTER }}
+        rancher_url: ${{ secrets.RANCHER_URL }}
+        rancher_token: ${{ secrets.RANCHER_TOKEN }}
+        command: |
+          /test/mint-auditor-test.sh \
+            --token-id 1
+
+    - name: Test - use drain_accounts to transfer id 1 balances to fog keys
+      if: inputs.testing_block_v3 == 'true'
```
remoun, line 460: Optional/Equivalent: Move this condition to the job level

remoun (approved):

joekottke (approved):
I like what @remoun suggested, if its possible: moving the check if_block_v3 to the job level, otherwise its just repeated.

.github/workflows/mobilecoin-workflow-dev-test.yaml
jgreat: I believe we've gone over this one before. Job level conditionals to make parts of the workflow optional don't work very well in GHA.  The problem is dependency management.  For each job that depends on the job you make conditional, you have to do a nasty set of `if` statements to check if it was a failed job or skipped job.  If you have multiple dependencies, you need to do checks on all of the dependencies in the chain.  

Like just about any CI/CD system I've worked with, easy things are easy... any kind of complexity makes it 10x harder to maintain.

jgreat:
> I like what @remoun suggested, if its possible: moving the check if_block_v3 to the job level, otherwise its just repeated.

See the response on remoun's comment.  Short answer is this is way cleaner than trying to do job level conditionals.

PR #2449, opened 2022-08-29
Sealed messages for clients with backends

awygle:
Add a method to convert a client query into a sealed query and modify `reencrypt_for_backends` to accept a sealed query.

### Motivation

This avoids the issue of repeated decryptions of the client query failing due to the Noise nonce advancing in the decryptor.

### Future Work
Update the fog view router to use the new method(s) in the creation of the `MultiViewStoreQuery` messages.

[Soundtrack of this PR](https://www.youtube.com/watch?v=Pgum6OT_VH8)

### DIFF
--- a/crypto/ake/enclave/src/lib.rs
+++ b/crypto/ake/enclave/src/lib.rs
@@ -25,8 +25,11 @@ use mc_crypto_keys::{X25519Private, X25519Public, X25519};
 use mc_crypto_rand::McRng;
 use mc_sgx_compat::sync::Mutex;
 use mc_util_from_random::FromRandom;
+use serde::{Deserialize, Serialize};
 use sha2::{Sha256, Sha512};
 
+/// An EnclaveMessage<ClientSession> sealed for the current enclave
+#[derive(Clone, Debug, Deserialize, Serialize)]
 pub struct SealedClientMessage {
     pub aad: Vec<u8>,
     pub channel_id: ClientSession,
--- a/crypto/ake/enclave/src/lib.rs
+++ b/crypto/ake/enclave/src/lib.rs
@@ -27,7 +27,11 @@ use mc_sgx_compat::sync::Mutex;
 use mc_util_from_random::FromRandom;
 use sha2::{Sha256, Sha512};
 
-pub type SealedClientMessage = EnclaveMessage<ClientSession>;
+pub struct SealedClientMessage {
+    pub aad: Vec<u8>,
+    pub channel_id: ClientSession,
+    pub data: IntelSealed,
+}
 
 /// Max number of pending quotes.
 const MAX_PENDING_QUOTES: usize = 64;
@@ -453,10 +457,9 @@ impl<EI: EnclaveIdentity> AkeEnclaveState<EI> {
         let aad = incoming_client_message.aad.clone();
         let channel_id = incoming_client_message.channel_id.clone();
         let client_query_bytes = self.client_decrypt(incoming_client_message)?;
-        let sealed_data =
-            IntelSealed::seal_raw(&client_query_bytes, &[]).map(|x| x.as_ref().to_vec())?;
+        let sealed_data = IntelSealed::seal_raw(&client_query_bytes, &[])?;
 
-        Ok(EnclaveMessage {
+        Ok(SealedClientMessage {
             channel_id,
             aad,
             data: sealed_data,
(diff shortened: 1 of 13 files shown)


crypto/ake/enclave/src/lib.rs
```diff
@@ -27,6 +27,8 @@ use mc_sgx_compat::sync::Mutex;
 use mc_util_from_random::FromRandom;
 use sha2::{Sha256, Sha512};
 
+pub type SealedClientMessage = EnclaveMessage<ClientSession>;
+
```
jcape, line 31: This should be a real type, even if it's identical to `EnclaveMessage<ClientSession>`---it's always an error to try use an `attest::Message`/`EnclaveMessage` directly as  a `SealedClientMessage`.

jcape (changes requested):
This is nearly there, but `SealedClientMessage` should be it's own type, for footgun protection reasons if nothing else.

I'm not familiar with what (if any) support `IntelSealed` has for serde, but it would be nice to just use that directly (rather than converting to bytes).

crypto/ake/enclave/src/lib.rs
```diff
@@ -438,24 +440,43 @@ impl<EI: EnclaveIdentity> AkeEnclaveState<EI> {
     }
 
     /// Transforms an incoming client message, i.e. a message sent from a client
-    /// to the current enclave, into a list of outbound  messages for
-    /// other enclaves that serve as backends to the current enclave.
-    ///
-    ///                               / --> Backend Enclave 1
-    ///   Client -> Current Enclave ---> Backend Enclave 2
-    ///                              \ --> Backend Enclave N
```
samdealy, line 446: Can you re-add this in the new `reencrypt_sealed_message_for_backends` method? The visual aid promotes understanding more effectively IMO.

crypto/ake/enclave/src/lib.rs
```diff
@@ -438,24 +440,43 @@ impl<EI: EnclaveIdentity> AkeEnclaveState<EI> {
...
+    /// later for use by this enclave without advancing the Noise nonce.
+    pub fn decrypt_client_message_for_enclave(
         &self,
         incoming_client_message: EnclaveMessage<ClientSession>,
+    ) -> Result<SealedClientMessage> {
+        let aad = incoming_client_message.aad.clone();
+        let channel_id = incoming_client_message.channel_id.clone();
+        let client_query_bytes = self.client_decrypt(incoming_client_message)?;
+        let sealed_data =
+            IntelSealed::seal_raw(&client_query_bytes, &[]).map(|x| x.as_ref().to_vec())?;
```
samdealy, line 453: IIUC, this method not only converts the `client_query_bytes` into a seal message, but also stores the sealed message in the enclave? If this is accurate, can you update the methods comment to reflect this?

If so, does the `sealed_data` represent a "handle" or a "pointer" to the actual data that's being sealed? If so, can we rename to something like `sealed_data_handle`? I'd want to make it clear that the thing that's being passed back across the enclave boundary isn't the data itself but some sort of identifier.

samdealy, line 453: Nit: Can you use something more descriptive than `x` as the variable name?

fog/view/enclave/impl/src/lib.rs
```diff
@@ -196,11 +196,11 @@ where
     /// to be sent off to each Fog View Store shard.
     fn create_multi_view_store_query_data(
         &self,
-        client_query: EnclaveMessage<ClientSession>,
+        sealed_query: SealedClientMessage,
```
samdealy, line 199: Echoing the above comment, if this is a handle to the sealed data, I'd recommend renaming to something like "sealed_query_handle" etc.

samdealy (approved):

crypto/ake/enclave/src/lib.rs
awygle: Fixed. I thought Rust type aliases wouldn't allow you to just pass the aliased type, whoops.

crypto/ake/enclave/src/lib.rs
awygle: My understanding is that the sealed data doesn't live in the enclave, instead the sealing is done by encrypting with a key stored in the enclave and returning the ciphertext. I believe `IntelSealed` contains the full ciphertext. Maybe @jcape can clear this up for us.

awygle: This map call is overcome by events.

awygle:
> I'm not familiar with what (if any) support `IntelSealed` has for serde, but it would be nice to just use that directly (rather than converting to bytes).

This is done now, the `SealedClientMessage` type stores an `IntelSealed` directly.

samdealy (approved):

crypto/ake/enclave/src/lib.rs
awygle: I mean that since James asked me to return IntelSealed directly, there's no more map call, so the 'x' name is gone.

jcape (changes requested):
There are a *lot* of seemingly unrelated updates in the `Cargo.lock`s, could you restore those files then re-run `cargo build`.

jcape (approved):

jcape:
@awygle Are you able to rebase this against the new changes (sync of feature branch from master)?

PR #2452, opened 2022-08-29
Feature/fog resolver refactor

BrianCorbin:
Split the mc-fog-report-validation crate into multiple crates, separating FogResolver, FogPubkeyResolver, and IngestReportVerifier.

Would appreciate feedback/suggestions on name changes for both the basic trait implementation of FogResolver defined within the crate itself, and the FogResolver implementation that's defined in its separate crate.

### Motivation

We needed this to remove the mc-attest-verifier dependency that was being pulled into transaction builder (which was unnecessary) causing build issues on hardware wallets.

[Soundtrack of this PR]()


fog/report/validation/src/lib.rs
```diff
@@ -10,131 +10,103 @@
 //! mobilecoin transactions with fog recipients on an embedded device like a
 //! tiny hardware wallet, that doesn't have threads and won't have rust std.
 
-#![deny(missing_docs)]
```
BrianCorbin, line 13: Not sure if there's a specific reason this is needed. It was causing trouble when I was building and just removed it for the time being. Happy to re-add it if it's important.

fog/report/validation/src/lib.rs
```diff
@@ -10,131 +10,103 @@
...
 
+impl<A: Debug + Display, R: Debug + Display> From<FogSigError<A, R>> for FogPubkeyError {
+    fn from(src: FogSigError<A, R>) -> Self {
+        FogPubkeyError::Authority(src.to_string())
+    }
+}
+
+pub struct FogResolver(HashMap<PublicAddress, FullyValidatedFogPubkey>);
+
 impl FogResolver {
```
BrianCorbin, line 96: Do we want to give this a specific name?

fog/report/resolver/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+/// A collection of unvalidated fog reports, together with an IAS verifier.
+/// This object is passed to the TransactionBuilder object.
+/// When fog is not involved, it can simply be defaulted.
+///
+/// Once constructed, this object can get validated fog pubkeys to build fog
+/// hints for transactions, without talking to the internet, and so is
+/// compatible with offline transactions to fog recipients. Only getting the
+/// FogReportResponses requires an online connection.
+#[derive(Default, Clone, Debug, Serialize, Deserialize)]
+pub struct FogResolver {
```
BrianCorbin, line 41: Do we want to rename this?

fog/report/validation/src/lib.rs
garbageslam: Its important to force developers to add doc strings, can we add doc strings wherever they are missing so it will build with this?

fog/report/validation/src/lib.rs
```diff
@@ -10,131 +10,103 @@
...
+    }
 }
 
+impl<A: Debug + Display, R: Debug + Display> From<FogSigError<A, R>> for FogPubkeyError {
+    fn from(src: FogSigError<A, R>) -> Self {
+        FogPubkeyError::Authority(src.to_string())
+    }
+}
+
+pub struct FogResolver(HashMap<PublicAddress, FullyValidatedFogPubkey>);
```
garbageslam, line 94: Can we add a doc string for this?

fog/report/validation/src/lib.rs
BrianCorbin: figured that's what it was, but it didn't seem consistent with what was already there. I'll look into it and add doc updates where necessary :)

BrianCorbin: aaah okay, it applies this rule to the entire file, duh

fog/report/resolver/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+                        });
+                    }
+                }
+                Err(FogPubkeyError::NoMatchingReportId(url, report_id))
+            } else {
+                Err(FogPubkeyError::NoMatchingReportResponse(url))
+            }
+        } else {
+            Err(FogPubkeyError::NoFogReportUrl)
+        }
```
remoun, line 99: Optional/Equivalent: Use [`ok_or`](https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or) to avoid 2 levels of nesting
```suggestion
        let url = recipient.fog_report_url().ok_or(FogPubkeyError::NoFogReportUrl)?;
        // Normalize the string to URL before lookup
        let url = FogUri::from_str(url)?.to_string();
        let result = self.responses.get(&url).ok_or_else(|| FogPubkeyError::NoMatchingReportResponse(url))?;
        // Verify the authority signature chain
        recipient.verify_fog_sig(result)?;
        // Get the report corresponding to our ID
        let report_id = recipient.fog_report_id().unwrap_or("").to_string();
        for report in result.reports.iter() {
            if report_id == report.fog_report_id {
                let pubkey = self
                    .verifier
                    .validate_ingest_ias_report(report.report.clone())
                    .map_err(|e| FogPubkeyError::IngestReport(e.to_string()))?;
                return Ok(FullyValidatedFogPubkey {
                    pubkey,
                    pubkey_expiry: report.pubkey_expiry,
                });
            }
        }
        Err(FogPubkeyError::NoMatchingReportId(url, report_id))
```

remoun (approved):
LGTM

garbageslam (approved):
LGTM (but Remoun's suggestions also seem good to me)

BrianCorbin:
Will implement suggestions and fix CI stuff 😊

fog/report/resolver/src/lib.rs
```diff
@@ -0,0 +1,101 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
+//! Logic for representing fog public keys from the fog-report server
+//! that have been fully validated, and the associated metadata.
+//!
+//! Note: Ideally this crate would be no_std compatible, but that is
+//! aspirational. The ReportResponse object is not no_std right now, and neither
+//! is x509 stuff. This is tracked in FOG-334.
```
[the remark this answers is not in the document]
BrianCorbin: Not sure, @garbageslam?

garbageslam:
I think the lint failure is this:

```
Checking resolver...
error: Dependencies for resolver are not sorted
```

fog/report/resolver/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+[features]
+default = []
+automock = ["mockall"]
+
+[dependencies]
+mc-account-keys = { path = "../../../account-keys" }
+mc-attest-verifier = { path = "../../../attest/verifier", default-features = false }
+mc-fog-report-types = { path = "../types" }
+mc-fog-report-validation = { path = "../validation" }
+mc-fog-ingest-report = { path = "../../ingest/report" }
```
garbageslam, line 16:
```suggestion
mc-fog-ingest-report = { path = "../../ingest/report" }
mc-fog-report-types = { path = "../types" }
mc-fog-report-validation = { path = "../validation" }
```

I think the linter just wants these packages to be sorted alphabetically

fog/report/resolver/src/lib.rs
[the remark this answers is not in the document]
garbageslam: It's not still accurate, I think you're essentially fixing this in this PR. and FOG-334 is probably a reference to a JIRA ticket from 2 years ago that no longer exists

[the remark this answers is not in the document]
BrianCorbin: Ah I see, yeah that makes sense. Thanks!

PR #2461, opened 2022-08-30
Messages for noise protocol exchanges with explicit nonces

jcape:
- Add attest::NonceMessage
- Add EnclaveNonceMessage
- Add Ready::decrypt_with_nonce, Ready::encrypt_with_nonce
- Add NoncePlaintext, NonceCiphertext events
- Add relevant transitions

### Motivation

There are some internal use cases for enclave messages with explicit nonces.

### Future Work
- Use these objects in MultiViewStoreQuery

### DIFF
--- a/attest/api/proto/attest.proto
+++ b/attest/api/proto/attest.proto
@@ -48,3 +48,20 @@ message Message {
         /// for use in the enclave.
         bytes data = 3;
 }
+
+/// An AEAD message with an explicit nonce.
+///
+/// This message is technically compatible with [`Message`], but exists to
+// ensure generated code doesn't use Message.
+message NonceMessage {
+        /// A byte array containing plaintext authenticated data.
+        bytes aad = 1;
+        /// An byte array containing the channel ID this message is
+        /// associated with. A zero-length channel ID is not valid.
+        bytes channel_id = 2;
+        /// A potentially encrypted bytestream containing opaque data intended
+        /// for use in the enclave.
+        bytes data = 3;
+        /// The explicit nonce.
+        fixed64 nonce = 4;
+}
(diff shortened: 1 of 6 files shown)


jcape:
* feature/fog-view-router
  * **PR #2461** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2461" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2461?utm_source=stack-comment).

attest/api/proto/attest.proto
```diff
@@ -48,3 +48,20 @@ message Message {
         /// for use in the enclave.
         bytes data = 3;
 }
+
+/// An AEAD message with an explicit nonce.
+///
+/// This message is technically compatible with [`Message`], but exists to
```
samdealy, line 54: What does it mean that a `NonceMessage` is compatible with `Message`? And why would we want the generated code to not use `Message` if it's compatible?

samdealy (approved):

attest/api/proto/attest.proto
jcape: It's theoretically possible for a gRPC participant to send a `Message` on one side, and have it get decoded as a `NonceMessage` on the other, because the all fields are optional, and a field with a defaulted value is omitted.

So a `NonceMessage` with `nonce = 0` looks identical to a `Message` when encoded.

The reason to split the types up is so that you know one of these should have a nonce and one of them shouldn't _in the code_, even if you can't tell from the bytes.

samdealy: Ah thanks for explaining. I totally agree that we should have explicit type for `NonceMessage` to ensure its bytes aren't confused for a `Message`.

awygle (approved):

PR #2484, opened 2022-09-07
add partial fill rules for swaps

garbageslam:
This implements new input rules for partial fill swaps, building on the masked-amount-v2 work, and extends the transaction builder and SCI builder to support building this kind of swap.

(See MCIP 42)


transaction/std/src/signed_contingent_input_builder.rs
```diff
@@ -282,6 +294,173 @@ impl<FPR: FogPubkeyResolver> SignedContingentInputBuilder<FPR> {
...
+            self.block_version,
+            amount,
+            recipient,
+            hint,
+            memo_fn,
+            rng,
+        )?;
+        self.impose_tombstone_block_limit(pubkey_expiry);
+
+        let amount_shared_secret = MaskedAmountV2::compute_amount_shared_secret(&shared_secret);
```
garbageslam, line 438: This is possibly an abstraction leakage, maybe it should be like

```
tx_out.masked_amount.compute_amount_shared_secret(...)
```

or something?

eranrund (commented):
Did an initial pass on this, cool stuff. This is going to require some serious unit test coverage...

transaction/core/src/revealed_tx_out.rs
```diff
@@ -0,0 +1,69 @@
...
+/// An error that can occur when attempting to reveal a TxOut using its Amount
+/// shared secret
+#[derive(Clone, Debug, Display, Ord, PartialOrd, Hash, Eq, PartialEq, Serialize, Deserialize)]
+pub enum RevealedTxOutError {
+    /// Invalid Amount Shared Secret
+    InvalidAmountSharedSecret,
+    /// TxOut conversion: {0}
+    TxOutConversion(String),
+    /// Amount: {0}
+    Amount(String),
```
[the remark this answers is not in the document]
garbageslam: The problem is that this is used ultimately in `InputRuleError` which becomes part of `TransactionValidationError` which has to convert to `TxProposalResult`, and that has to be convertible back from an error code, so anything that is a member of any of these enums has to be default constructible. See for instance here:

https://github.com/mobilecoinfoundation/mobilecoin/blob/f142750ae329fd94bb0cbc480d4d8019bf182763/consensus/api/src/conversions.rs#L51

https://github.com/mobilecoinfoundation/mobilecoin/blob/f142750ae329fd94bb0cbc480d4d8019bf182763/consensus/api/src/conversions.rs#L107

I think that is the reason that "ledger error" is a String here:

https://github.com/mobilecoinfoundation/mobilecoin/blob/f142750ae329fd94bb0cbc480d4d8019bf182763/transaction/core/src/validation/error.rs#L116

There is a lot of stuff that is crappy about that conversions module -- for instance, if we want to convert every error into an error code, maybe we can include the text in the body of the response? And we could try to convert an error code + response body back into this error I guess, so that the conversion is not lossy?

I suppose that instead of putting `String` here we could give `TxOutConversionError` and `AmountError` some default states which are "unknown" or something, but this also seems crappy.

If you have ideas I'm happy to try to improve it in another PR. This part of the diff was actually already merged in the masked_amount_v2 PR and I haven't yet rebased this PR on master. But the review is helpful and I'm going to make a PR to address the follow on issues.

transaction/core/src/input_rules.rs
```diff
@@ -53,6 +69,97 @@ impl InputRules {
...
+            // Compute fill-fraction num and denom. This is the fraction of the maximum
+            // possible trade that could have occurred which did occur. This
+            // calculation is used later when checking if real output amounts respected the
+            // fill fraction.
+            let fill_fraction_num =
+                (fractional_change_amount.value - real_change_amount.value) as u128;
+            let fill_fraction_denom = fractional_change_amount.value as u128;
+
+            // Verify fractional_outputs
+            for fractional_output in self.fractional_outputs.iter() {
```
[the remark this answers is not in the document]
garbageslam: Yeah they just end up screwing themselves

This is a silly thing to do but I'm not sure if it should be an error.

I'm trying to keep in mind that we might add more stuff to `InputRules` later that interacts with existing fields, so it's possible that at a future revision it's okay for `self.fractional_outputs` to be empty because some other list is not empty.

Idk, maybe that's not good, maybe it should just be an error. I've tried to make a lot of forms of misuse be an error.

transaction/core/src/revealed_tx_out.rs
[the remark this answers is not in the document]
eranrund: I see, yes, that is annoying. Thanks for explaining it.
I made an attempt [here](https://github.com/mobilecoinfoundation/mobilecoin/blob/master/consensus/api/proto/consensus_client.proto#L32) to try and support a bit less lossy error conversions (also see [this](https://github.com/mobilecoinfoundation/mobilecoin/blob/master/consensus/api/src/conversions.rs#L206) for example of how that stuff got used). Not sure that is ideal either... I think losing the error details is going to be the thing that we get upset with the moment we run into these errors and are left wondering what happened, but this is a problem that is outside the scope of this PR.

transaction/core/src/input_rules.rs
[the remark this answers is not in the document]
eranrund: I don't feel strongly - mostly wanted to test my understanding of that edgecase.

transaction/std/src/transaction_builder.rs
```diff
@@ -212,12 +216,178 @@ impl<FPR: FogPubkeyResolver> TransactionBuilder<FPR> {
...
+            return Err(SignedContingentInputError::ChangeExceededOffer);
+        }
+        if rules.max_allowed_change_value != 0
+            && rules.max_allowed_change_value < sci_change_amount.value
+        {
+            return Err(SignedContingentInputError::ChangeLimitExceeded);
+        }
+
+        let fill_fraction_num = (fractional_change_amount.value - sci_change_amount.value) as u128;
+        let fill_fraction_denom = fractional_change_amount.value as u128;
```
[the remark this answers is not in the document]
garbageslam: will think about it

transaction/std/src/signed_contingent_input_builder.rs
```diff
@@ -310,13 +489,19 @@ impl<FPR: FogPubkeyResolver> SignedContingentInputBuilder<FPR> {
         let (outputs, output_secrets): (Vec<TxOut>, Vec<_>) =
             self.required_outputs_and_secrets.drain(..).unzip();
 
+        self.fractional_outputs
```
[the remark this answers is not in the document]
garbageslam: maybe? will think about it

transaction/core/src/revealed_tx_out.rs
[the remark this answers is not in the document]
garbageslam: It looks that we only use this conversion in `mc-connection` here:

https://github.com/mobilecoinfoundation/mobilecoin/blob/bae5a20a59e54c5a288a1772971684b3b37a351a/connection/src/error.rs#L92

And we are only marshalling back to `TransactionValidationError` to satisfy this:

https://github.com/mobilecoinfoundation/mobilecoin/blob/bae5a20a59e54c5a288a1772971684b3b37a351a/connection/src/error.rs#L35

I wonder if we can just change that to `ProposeTxResult` and be done with it?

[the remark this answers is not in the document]
eranrund: I think  the reason the conversion from our custom error codes to a Rust `Result::Err` variant take place is in order to essentially obfuscate the fact that we are doing an RPC call, and make the caller experience be similar to had they called propose_tx directly. It's possible this can all be simplified if https://github.com/mobilecoinfoundation/mobilecoin/blob/bae5a20a59e54c5a288a1772971684b3b37a351a/connection/src/thick.rs#L403 is changed to return the full RPC result (https://github.com/mobilecoinfoundation/mobilecoin/blob/bae5a20a59e54c5a288a1772971684b3b37a351a/consensus/api/proto/consensus_common.proto#L99) and have callers handle it appropriately. I think the original thinking was that the callers would have a better experience if they got a proper `Result::Err` when the transaction was declined (but the GRPC call succeeded).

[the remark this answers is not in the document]
garbageslam: So like, what I was thinking was, if that enum option was just:

```
 TransactionValidation(TxProposeResult), 
```

then we don't need to do the conversion at all and that saves us a headache? And the caller still has a "normal" rust error I think, so it shouldn't be that disruptive.

I can understand that like, maybe the error strings for `ProposeTxResponse` are worse than for `TransactionValidationError` if you log it? But maybe we can simply improve them if that's the issue?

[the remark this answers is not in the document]
garbageslam: I opened a PR showing how I think this should work and why: https://github.com/mobilecoinfoundation/mobilecoin/pull/2501

garbageslam:
I'm not quite done developing this, there are still some comments to address, and some more tests I can write.

However, I am basically convinced at this point that the functionality is working or pretty close, so I think it's ready for review.

It's already 2k lines, and about 1000 of that is tests, so if we add significantly more tests it's going to get pretty large.

api/src/convert/tx_in.rs
```diff
@@ -72,6 +72,19 @@ impl From<&InputRules> for external::InputRules {
...
 
+        let fractional_outputs = source
+            .fractional_outputs
+            .iter()
+            .map(external::RevealedTxOut::from)
+            .collect();
+        input_rules.set_fractional_outputs(fractional_outputs);
+
+        if let Some(fractional_change) = source.fractional_change.as_ref() {
+            input_rules.set_fractional_change(external::RevealedTxOut::from(fractional_change));
```
remoun, line 83: Optional/Equivalent: Use `into()`
```suggestion
            input_rules.set_fractional_change(fractional_change.into());
```

api/src/convert/tx_in.rs
```diff
@@ -87,9 +100,48 @@ impl TryFrom<&external::InputRules> for InputRules {
...
+    fn from(source: &RevealedTxOut) -> Self {
+        let mut result = external::RevealedTxOut::new();
+        result.set_tx_out(external::TxOut::from(&source.tx_out));
+        result.set_amount_shared_secret(source.amount_shared_secret.to_vec());
+        result
+    }
+}
+
+/// Convert external::RevealedTxOut --> RevealedTxOut
+impl TryFrom<&external::RevealedTxOut> for RevealedTxOut {
```
remoun, line 135: Please add some tests exercising these conversions, e.g. round-trip conversion tests are useful for catching missing field mappings

remoun (dismissed):

transaction/std/src/signed_contingent_input_builder.rs
```diff
@@ -282,6 +291,170 @@ impl<FPR: FogPubkeyResolver> SignedContingentInputBuilder<FPR> {
...
+    /// impose a minimum fill requirement on the counterparty (to prevent
+    /// griefing).
+    ///
+    /// # Arguments
+    /// * `value` - The u64 value which the fractional change output in the
+    ///   filled order must not exceed. This is denominated in the token id of
+    ///   the fractional change output.
+    pub fn set_maximum_allowed_change_value(&mut self, value: u64) -> u64 {
+        self.maximum_allowed_change_value = value;
+        self.maximum_allowed_change_value
```
remoun, line 455: Why return the value from this setter?

garbageslam: maybe no good reason, will remove

consensus/api/src/conversions.rs
```diff
@@ -123,6 +153,39 @@ impl TryInto<Error> for ProposeTxResult {
...
+                InputRuleError::RealChangeOutputAmountExceedsLimit,
+            )),
+            Self::InputRuleInvalidAmountSharedSecret => Ok(Error::InputRule(
+                InputRuleError::RevealedTxOut(RevealedTxOutError::InvalidAmountSharedSecret),
+            )),
+            Self::InputRuleTxOutConversion => Ok(Error::InputRule(InputRuleError::RevealedTxOut(
+                RevealedTxOutError::TxOutConversion(String::default()),
+            ))),
+            Self::InputRuleAmount => Ok(Error::InputRule(InputRuleError::RevealedTxOut(
+                RevealedTxOutError::Amount(String::default()),
```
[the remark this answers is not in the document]
garbageslam: now fixed

transaction/std/src/transaction_builder.rs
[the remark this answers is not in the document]
garbageslam: we're going to pull `fill_to_fractional_output_at` to a new PR to try to make this one smaller, I'll revisit this issue at this time

transaction/std/src/signed_contingent_input_builder.rs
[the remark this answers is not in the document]
garbageslam: I don't see right now that it is necessary, if validation just sorts the final outputs, that seems fine?

transaction/core/tests/input_rules.rs
```diff
@@ -63,25 +72,288 @@ fn test_input_rules_verify_max_tombstone() {
...
+        get_first_rules(&tx).verify(block_version, &tx),
+        Err(InputRuleError::RealChangeOutputAmountExceedsLimit)
+    );
+
+    // Remove limit altogether, we should be valid again
+    get_first_rules_mut(&mut tx).max_allowed_change_value = 0;
+    get_first_rules(&tx).verify(block_version, &tx).unwrap();
+
+    // Change the frational change output to be 3000. This means the implied fill
+    // fraction is now 2/3, since the real change output only returns
```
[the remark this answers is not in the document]
garbageslam: yes that's exactly right :)

garbageslam:
i got deployment timeouts 3x in a row, decided to rebase on master (maybe there were some deployment fixes that were merged?)

transaction/core/src/input_rules.rs
```diff
@@ -53,6 +69,103 @@ impl InputRules {
...
+                // Try to unblind the fractional output amount.
+                let fractional_output_amount = fractional_output.reveal_amount()?;
+                // Let's check if there is a corresponding real output.
+                let real_output = tx
+                    .prefix
+                    .outputs
+                    .iter()
+                    .find(|x| {
+                        x.public_key == fractional_output.tx_out.public_key
+                            && x.target_key == fractional_output.tx_out.target_key
```
garbageslam, line 122: Coding it this way is a bit janky, it would be better if we had a special function on `TxOut` which means "check if a and b are equal EXCEPT in the masked_amount portion". That way if we add future fields to TxOut's they will still be enforced to match at this part

garbageslam: Will fix shortly

transaction/core/src/input_rules.rs
```diff
@@ -53,6 +69,103 @@ impl InputRules {
...
+            // There is a fractional change output. Let's try to unblind its amount.
+            let fractional_change_amount = fractional_change.reveal_amount()?;
+            // Let's check if there is a corresponding real change output.
+            let real_change = tx
+                .prefix
+                .outputs
+                .iter()
+                .find(|x| {
+                    x.public_key == fractional_change.tx_out.public_key
+                        && x.target_key == fractional_change.tx_out.target_key
```
garbageslam, line 83: Same comment as below -- we really should have a function on `TxOut`'s that checks if they are equal everywhere excepting the masked amount, and call that function here.

eranrund: You can have a helper function that clones one of the sides of the equation and override its masked amount with the other side, that ensures everything gets compared other than the masked amount.
A bit slower but probably doesn't matter

transaction/core/tests/input_rules.rs
```diff
@@ -63,25 +72,290 @@ fn test_input_rules_verify_max_tombstone() {
...
+    // 1/3 of this. The tx should then be invalid because we are only filling 1/2 of
+    // the fractional output that was required and not 2/3 of it.
```
nick-mobilecoin, line 225: :x: I think this is worded incorrectly and the logic may possibly be backwards
Alice has put up 3000 of token X.  Bob is taking 1000 of token X (txout_0)
Alice requested 2000 of token Y, Bob is providing 1000 of token Y (txout_1)

While this is disadvantageous for Bob, in the end he's the one making this overall TX

There should also be a test for the boundary conditions, there is quite a bit of room between 1/3 and 1/2

transaction/core/tests/input_rules.rs
```diff
@@ -63,25 +72,290 @@ fn test_input_rules_verify_max_tombstone() {
...
+    get_first_rules_mut(&mut tx).fractional_change = Some(change_committed_amount(
+        &revealed_tx_outs[0],
+        Amount::new(3000, 0.into()),
```
nick-mobilecoin, line 228: :thought_balloon: By the time I got down to here it's hard to recall what the overall state of the TX is.  I think this would be better to have a common test set up method and then individual test cases that focused on what they want the state to be over this recipe/workflow style test

transaction/core/tests/input_rules.rs
```diff
@@ -63,25 +72,290 @@ fn test_input_rules_verify_max_tombstone() {
...
+    get_first_rules_mut(&mut tx).fractional_change = Some(change_committed_amount(
+        &revealed_tx_outs[0],
+        Amount::new(500, 0.into()),
+    ));
+    assert_matches!(
+        get_first_rules(&tx).verify(block_version, &tx),
+        Err(InputRuleError::RealOutputAmountExceedsFractional)
+    );
+
+    // Change the frational change output to be 1500. This means the implied fill
```
nick-mobilecoin, line 246: :thought_balloon: again it may just be me, but I think  the use of "fractional" and "change" my be misleading, throughout this pr.

All of the "fractional" values are proposed or idealized values, and the final txouts are the partial or fractional values.

Getting here I think I finally understand that txout_0 is Alice's "change" from the money she fronted

nick-mobilecoin (commented):
Got a few questions and thoughts so far, still need to dig more into this PR to fully review

transaction/core/tests/input_rules.rs
```diff
@@ -63,25 +72,290 @@ fn test_input_rules_verify_max_tombstone() {
...
+        &revealed_tx_outs[0],
+        Amount::new(2000, 0.into()),
+    ));
+
+    // Check that the Tx is following input rules
+    get_first_rules(&tx).verify(block_version, &tx).unwrap();
+
+    // Lets try imposing a limit on the change value
+    // This is larger than 1000 so we should still be valid.
+    get_first_rules_mut(&mut tx).max_allowed_change_value = 1010;
```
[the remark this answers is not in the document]
garbageslam: thanks!

transaction/core/tests/input_rules.rs
```diff
@@ -63,25 +72,290 @@ fn test_input_rules_verify_max_tombstone() {
...
+            Amount::new(2000, 0.into()),
+        ));
+    assert_matches!(
+        get_first_rules(&tx).verify(block_version, &tx),
+        Err(InputRuleError::FractionalOutputsNotExpected)
+    );
+
+    // Add a fractional input, also by doubling one of the revealed tx outs.
+    // This means the fill fraction is 1/2, which we are satisfying.
+    get_first_rules_mut(&mut tx).fractional_change = Some(change_committed_amount(
```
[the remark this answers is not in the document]
garbageslam: I think the way this term arose is the following:

* We have well established concepts of "outputs" and "change outputs" (at least since RTH, these are the names of specific methods on `TransactionBuilder`)
* In MCIP 31 we added the concept of "required outputs", which appear in rules of SCIs. We naturally then had the concepts of "required outputs" and "required change outputs", which are the names of specific methods on `SignedContingentInputBuilder`. `add_required_output` is called when you want a required output with the Authenticated Sender memo, and `add_required_change_output` is called when you want a required output with the Destination Memo, which is destined for yourself.
* In the new MCIP we have a different type of constraint that can appear in InputRules, and that's this whole partial-fill concept. So naturally we now have `fractional_outputs` which appear in the rules, have amounts which vary according to the fill fraction, and have Authenticated Sender memo, and we have `fractional_change_outputs` which appear in the rules, have amounts which vary according to the fill fraction, and have Destination memo.

I'm happy to consider alternate names, but let's try to make the names all line up if we can.

It may be that "fractional" is just not the right adjective to use here. We could say that instead of "fractional output" and "fractional change" it's just "required partial fill output" and "required partial fill change". I would prefer something a little shorter.

transaction/core/tests/input_rules.rs
garbageslam: I agree that having better test infrastructure would make the tests better -- most of the test infrastructure in transaction-core is trash.

I've also gotten a lot of feedback that these PRs are too big. I don't think we can justify introducing new test infrastructure in this PR, but perhaps in a later PR.

transaction/core/tests/input_rules.rs
garbageslam: Thanks for your thoughts, lets discuss further here: https://github.com/mobilecoinfoundation/mobilecoin/pull/2484#discussion_r979438141

The terms we use should not be a source of confusion, we should pick better terms if these names are bad.

transaction/core/tests/input_rules.rs
garbageslam: >  I think this is worded incorrectly and the logic may possibly be backwards

Can you clarify exactly what you think is wrong and what you think is backwards?

I'd prefer to use the precise terms "fractional output" and "fractional change output" rather than "put up" and "requested" because these terms are ambiguous. It's important to note that the amount Alice has "put up" is the value of the input underlying the SCI, which should, but is not technically required, to match the value of the fractional change output. And in this test, these things do not match, but it isn't important for testing the functionality we are interested in here.

> There should also be a test for the boundary conditions, there is quite a bit of room between 1/3 and 1/2

I agree, but I think it's also valuable to change one thing, test, change another thing, test, to simply get more test coverage than we would have otherwise. I can make sure we test the boundary though as well.

transaction/core/src/input_rules.rs
```diff
@@ -53,6 +69,103 @@ impl InputRules {
...
+            // If there is no fractional change output specified, then none of the related
+            // fractional output rules or rule verification data should be present.
+            if !self.fractional_outputs.is_empty() {
+                return Err(InputRuleError::FractionalOutputsNotExpected);
+            }
+            if self.max_allowed_change_value != 0 {
+                return Err(InputRuleError::MaxAllowedChangeValueNotExpected);
+            }
+        }
+
```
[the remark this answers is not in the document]
garbageslam: I would rather not break this up to this extent, I think it makes it harder to follow and not easier.

transaction/core/tests/input_rules.rs
```diff
@@ -63,25 +72,290 @@ fn test_input_rules_verify_max_tombstone() {
...
+            .0
+    );
+
+    result.tx_out.masked_amount = Some(MaskedAmount::V2(new_masked_amount));
+    result
+}
+
+// Test that input rules verification is working for fractional output rules
+#[test]
+fn test_input_rules_verify_fractional_outputs() {
```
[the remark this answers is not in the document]
garbageslam: > individual unit tests for each case and or suggested function

I don't think it is a good idea to test the suggested functions instead of `verify`. I want to ensure very strong test coverage on the `verify` API because that is the actual public API that is called by transaction validation. Testing only its internals leaves open the possibility that a logic mistake in the overall function (perhaps a call to one of the internals is skipped?) leads to a bug in transaction validation.

I'm okay with the idea of splitting the recipe test into a few smaller tests. However, I've been getting a lot of feedback that there are too many LOC in the PR. Splitting things up into more tests, and repeating all the test boilerplate, will increase the LOC of the PR. So I'm reluctant to do too much of that.

transaction/core/tests/input_rules.rs
```diff
@@ -63,25 +72,290 @@ fn test_input_rules_verify_max_tombstone() {
...
+            assert_eq!(*amount.token_id, 0);
+            RevealedTxOut {
+                tx_out: txo.clone(),
+                amount_shared_secret: amount_shared_secret.to_vec(),
+            }
+        })
+        .collect();
+
+    // Add a fractional output, by doubling one of the revealed tx outs.
+    // So this means we are filling the 1/2 of the request.
```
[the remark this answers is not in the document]
garbageslam: Starting at 0.5 is a less valuable test, but it's also easier for the reviewer to follow as a warmup. We have subsequent tests that do 2/3 and 1/3.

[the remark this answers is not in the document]
garbageslam: > I think the underlying logic is incorrect.

Can you explain what precisely you think is incorrect?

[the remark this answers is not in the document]
garbageslam: > I'm going to put the numbers back to [MICP 42[(https://github.com/garbageslam/mcips/blob/partial-fill-rules/text/0042-partial-fill-rules.md)

If you mainly want to discuss the example from the MCIP, that MCIP is also open for review, and we could discuss there rather than in this PR review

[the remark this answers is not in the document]
garbageslam: >  : where does Alices 980 MOB "change" get accounted for? I think this is part of my hiccup on the "fractional change" term. I think of 980 being the "change" but it may be my lack of financial domain knowledge here.

It gets accounted for as the "real change output" corresponding to the "fractional change output" which appears in the input rules.

>  It was mentioned we aren't verifying the signatures so maybe that's the piece, but how do we know the 200MEOWB output is for Alice?

We can't verify this cryptographically, but we can assume that Alice is directing profits from the trade to some account that she controls. To some extent, whenever Alice is able to direct a location to which money flows, Alice "controls" that money.

>  In the example from the MCIP what prevents Bob from dropping off the 500MOB change to Alice?

This is exactly the same mechanism as what prevents Bob from dropping a required output for Alice in MCIP 31. Bob cannot tamper with the InputRules or it invalidate's Alice's MLSAG signature which signed away her money in the first place. If the InputRules are left undisturbed, Bob must supply a "real output" that corresponds to each "fractional output", and the code in this PR is called by the enclave to validate that that happens. So Bob's only choice is to provide sufficient fractional change, or the Tx will be invalid.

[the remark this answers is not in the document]
garbageslam: The other thing I will say is, we added a number of tests in `transaction/std/src/signed_contingent_input_builder.rs` that test all this functionality at a higher level (using the SCI builder and the Transaction builder), and those tests all seem to be working as expected. But if we got the low-level arithmetic relating to fractional change wrong, then all those tests should be failing.

So that leaves me doubtful that "the underlying logic is incorrect", but if you see a specific issue let's talk about it.

transaction/core/tests/input_rules.rs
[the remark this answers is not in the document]
garbageslam: from the other comment:

> All of the "fractional" values are proposed or idealized values, and the final txouts are the partial or fractional values.
>
> Getting here I think I finally understand that txout_0 is Alice's "change" from the money she fronted

Yes -- anything that appears in "rules" is not real, only `TxPrefix.outputs` are real and actually enter the blockchain. The terms used in docs and the MCIP all speak of the "real output corresponding to the fractional output" which appears in the rules.

Looking back I currently believe this is well documented, but if you think there's a place I could have documented this better or you want to propose different naming, I'm open to it.

transaction/core/tests/input_rules.rs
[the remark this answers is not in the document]
garbageslam: I'm going to assume after this comment:

> Getting here I think I finally understand that txout_0 is Alice's "change" from the money she fronted

that you actually don't believe the underlying logic is incorrect now, and what seems wrong is the terminology.

transaction/core/src/input_rules.rs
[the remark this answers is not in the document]
nick-mobilecoin: I think breaking up `verify()` to the three
```rust
        self.verify_max_tombstone_block(tx)?;
        self.verify_required_outputs(tx)?;
        self.verify_partial_fill_rules(tx)?;
        Ok(())
    }
```
Would be good.  As it currently sits it's relying on code comments to break up the larger groupings of what's being verified.  If a fourth one gets added it will likely be after the partial fill logic and be fairly easy to lose sight of

transaction/core/tests/input_rules.rs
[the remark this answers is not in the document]
nick-mobilecoin: > Testing only its internals leaves open the possibility that a logic mistake in the overall function
This is a very real concern.  I think ideally one would have _at least_ 4 tests of `verify()` that exercised:
1. Happy path
2. Failure on tombstone
2. Failure on required outputs
3. Failure on partial fill rules
One could add a couple more per failure as an extra feel good.
Then there would be full logic coverage with the unit tests of each sub function.

Often it feels like it's unit vs integration test.  I don't see it as one or the other. They both should exist and they help to support each other.  Integration tests support higher level ideas, while not needing to robustly exercise all the edge cases.  Unit tests get into the nitty gritty, helping prevent off by one errors, correct loop terminations, proper unit conversions, etc.
It can feel like wasted effort because the higher level integration tests seem to be testing the same or similar things as the unit tests, but like the `verify()` breakup suggestion, the integration test is ensuring that `verify()` delegates to the proper functionality and that those pieces of functionality work together, while the unit tests are robustly ensuring the functionality.

transaction/core/tests/input_rules.rs
nick-mobilecoin: > but I think it's also valuable to change one thing, test, change another thing, test
:100:

garbageslam:
Based on discussion with Nick, I'm considering the following renamings:

* `InputRules::fractional_outputs` and `InputRules::fractional_change` are replaced with `InputRules::partial_fill_outputs` and `InputRules::partial_fill_change`.

The main reason is, it is confusing that the "fractional" output is less in value than the corresponding "real" output. It seems reasonable to expect that something that is fractional is smaller.

* Everywhere we talk about the "real" output corresponding to a fractional output etc., we could call that the "fractional" output corresponding to a partial fill output. This mainly affects the code only in error messages and code comments. This might be a little more specific than "real".

Please thumbs up if you think this is a good rename, or comment if you think you have a better idea. I will propagate this back to MCIP 42 if there is agreement.

transaction/std/src/transaction_builder.rs
```diff
@@ -280,12 +284,205 @@ impl<FPR: FogPubkeyResolver> TransactionBuilder<FPR> {
...
+            .collect::<Result<_, _>>()?;
+
+        // Fill all fractional outputs to precisely the smallest degree required
+        // This helper function function takes a fractional output value, and returns
+        // num / denom * fractional_output_value, rounded up
+        fn division_helper(fractional_output_value: u64, num: u128, denom: u128) -> u64 {
+            let num_128 = fractional_output_value as u128 * num;
+            // Divide by fill_fraction_denom, rounding up, and truncate to u64
+            ((num_128 + (denom - 1)) / denom) as u64
+        }
```
nick-mobilecoin, line 392: :thought_balloon: This logic seems important to be correct and ensure it stays correct, however the builder tests seems to be focusing on values that easily divide and would still work if someone got rid of this and did ((output_value * numerator) / denominator)

transaction/std/src/transaction_builder.rs
```diff
@@ -280,12 +284,205 @@ impl<FPR: FogPubkeyResolver> TransactionBuilder<FPR> {
...
+            {
+                // If not, add it
+                self.outputs_and_secrets
+                    .push((required_output.clone(), unmasked_amount.clone().into()));
+            }
+        }
+        // 2. Max tombstone block
+        if rules.max_tombstone_block != 0 {
+            self.impose_tombstone_block_limit(rules.max_tombstone_block);
+        }
```
nick-mobilecoin, line 480: :wrench: This looks suspiciously like a copy and paste from `add_presigned_input()`.  Is there a reason this didn't get extracted out to a common method?  It seems likely that as more rule types get added these become out of sync.  Even now there doesn't appear to be tests to ensure this code does what it's supposed to.

nick-mobilecoin (approved):
Finish full review.
Some thoughts:
1. SCI hasn't been released yet, per some comments in the review could SCI and partial fills be better combined?  Should they be...???
I haven not read (or need to read again) MCIP 31, but it seems that partial fills are a subset of the common SCI.  Should SCI and partial fill use the same interface and the only difference being that SCI requires 100% fulfillment?
2. for future PRs suggest one for the `verify()` in the enclave and then a separate one for the transaction builder logic.
3. I think we want to consider multiple levels of testing.  Many tests in `transaction/std/src/signed_contingent_input_builder.rs` re-use the same values and use convenient values.  The tests are fairly long because their testing at such a high level.  They end up missing boundary conditions especially around not so nice of fractions.

transaction/core/src/input_rules.rs
```diff
@@ -53,6 +69,107 @@ impl InputRules {
...
+            // Check the bounds of the real change amount
+            if real_change_amount.token_id != fractional_change_amount.token_id {
+                return Err(InputRuleError::RealOutputTokenIdMismatch);
+            }
+            if real_change_amount.value > fractional_change_amount.value {
+                return Err(InputRuleError::RealOutputAmountExceedsFractional);
+            }
+            if self.max_allowed_change_value != 0
+                && real_change_amount.value > self.max_allowed_change_value
```
samdealy, line 106: This logic enforces minimum order volume, but the reader of the code has to perform one additional mental step to understand this- i.e subtract the `real_change_amount_value` from the `fractional_change_amount_value`. This point was discussed in the PR review video, but I'd like to +1 that point and suggest that we consider using a `minimum_order_volume_price` instead of `max_allowed_change_value` and add an extra line that does the subtraction. 

I take the point that this simplifies the logic for the transaction builder (and saves the line of code that performs the subtraction), but it's less readable IMO. That said, I know you've thought about this a lot and other people seem to not have a problem with it, so I'm fine to leave it as is. If that's the case though, I'd recommend adding a comment that includes an Alice / Bob scenario and describes the subtraction and how it computes minimum order volume.

garbageslam:
> Finish full review. Some thoughts:
> 

Thanks for these thoughts

>     1. SCI hasn't been released yet, per some comments in the review could SCI and partial fills be better combined?  Should they be...???
>        I haven not read (or need to read again) MCIP 31, but it seems that partial fills are a subset of the common SCI.  Should SCI and partial fill use the same interface and the only difference being that SCI requires 100% fulfillment?
> 

So here's the reasons why I didn't steer in that route:

a. MCIP 31 SCIs have a slightly different privacy model from the MCIP 42 partial fill SCIs. The difference is that to verify the partial fill stuff, you have to reveal the amount shared secret in the enclave. That's a departure from the normal MobileCoin transaction validation model -- normally we never reveal amounts or token ids in the enclave. We can only really justify that because to propose a trade at all, the originator has to reveal something to the counterparty, and we are planning to use this to launch like a decentralized quoting service, so the originator reveals this to all the counter parties.

However, there might be some other use-case we haven't envisioned yet where it really is preferable for some reason not to have to reveal the amounts to the enclave.

b. I think at some point it may be interesting to have SCIs that mix both features, like, you have a fractional output and fractional change, but you also have a required_output which is like a flat fee which doesn't depend on the fill fraction, which represents a fee which is paid to the quoting service operators, or some other third party who facilitates the interaction.

>     2. for future PRs suggest one for the `verify()` in the enclave and then a separate one for the transaction builder logic.
> 

Yeah so, I think that would have been possible for this PR in retrospect, because eventually we did figure out a way to test `InputRules.verify` without using the SCI builder and the TransactionBuilder.

But when I started, I wasn't sure that there was going to be any way to generate test data that essentially uses the SCI builder and the transaction builder.

Most of the transaction-core tests actually do use the transaction-builder under the hood, and even, mc-ledger-db, which is frankly wierd and unfortunate. All of our existing testing relies on instantiating an lmdb database to produce valid transactions with valid merkle proofs, and there are no like "simple" test harnesses for transaction-core that don't rely on concepts from higher-level crates.

I think that sucks, but it also doesn't seem like a high enough priority to go back and fix it. I've generally gotten used to that idea that sometimes, to test new features of transactions, you have to add something to the TransactionBuilder that will enable you to build transactions like that and test them, because that's kind of the status quo for most of the testing.

It's true though that after I'd worked through the whole PR, I did eventually figure out a reasonable way to test the InputRules::verify without having to exercise the SCI builder. At that point I could have split the PR if I thought about it.

>     3. I think we want to consider multiple levels of testing.  Many tests in `transaction/std/src/signed_contingent_input_builder.rs` re-use the same values and use convenient values.  The tests are fairly long because their testing at such a high level.  They end up missing boundary conditions especially around not so nice of fractions.

I'm happy to try to improve those tests also and make sure they hit the boundary conditions, thanks for pointing this out

transaction/core/src/input_rules.rs
```diff
@@ -53,6 +69,107 @@ impl InputRules {
...
+    fn verify_partial_fill_rules(
+        &self,
+        _block_version: BlockVersion,
+        tx: &Tx,
+    ) -> Result<(), InputRuleError> {
+        if let Some(fractional_change) = self.fractional_change.as_ref() {
+            // There is a fractional change output. Let's try to unblind its amount.
+            let fractional_change_amount = fractional_change.reveal_amount()?;
+            // Let's check if there is a corresponding real change output.
+            let real_change = tx
```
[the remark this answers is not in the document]
garbageslam: there should be a value of 0, because if the originator put a change memo or something, there has to be an output to carry it, even if the value is 0

transaction/core/src/input_rules.rs
```diff
@@ -53,6 +69,107 @@ impl InputRules {
...
+        Ok(())
+    }
+
+    // Partial-Fill rules verification (MCIP #42)
+    fn verify_partial_fill_rules(
+        &self,
+        _block_version: BlockVersion,
+        tx: &Tx,
+    ) -> Result<(), InputRuleError> {
+        if let Some(fractional_change) = self.fractional_change.as_ref() {
```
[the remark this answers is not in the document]
garbageslam: Yeah I think that's right

transaction/core/src/input_rules.rs
```diff
@@ -34,13 +35,28 @@ pub struct InputRules {
...
+    pub fractional_outputs: Vec<RevealedTxOut>,
+
+    /// A change output for any leftover from this input, if a fractional fill
+    /// occurs (MCIP #42)
+    #[prost(message, tag = "4")]
+    pub fractional_change: Option<RevealedTxOut>,
+
+    /// A maximum allowed value for the real change output. (MCIP #42)
+    /// This can be used to impose a minimum traded amount on the counterparty
+    /// to an SCI. This limit is ignored if set to zero.
```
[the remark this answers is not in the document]
garbageslam: I'm okay to change this to minimum_fill, this is a good point. thank you

transaction/core/src/input_rules.rs
garbageslam: I think it's a good point and Nick had a similar comment, I think you guys are right and we should change this

transaction/core/src/input_rules.rs
```diff
@@ -53,6 +69,107 @@ impl InputRules {
...
+                .ok_or(InputRuleError::MissingRealChangeOutput)?;
+            // Let's try to unblind its amount.
+            let real_change_amount =
+                try_reveal_amount(real_change, fractional_change.amount_shared_secret.as_ref())?;
+
+            // Check the bounds of the real change amount
+            if real_change_amount.token_id != fractional_change_amount.token_id {
+                return Err(InputRuleError::RealOutputTokenIdMismatch);
+            }
+            if real_change_amount.value > fractional_change_amount.value {
```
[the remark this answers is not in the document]
garbageslam: It could happen I think if the `counterparty` used a bad `TransactionBuilder` instead of the real one? Nothing in the enclave verification code, except what we put in this function, prevents them from changing the value of that TxOut to anything they want when the make the corresponding "real" output

transaction/core/src/input_rules.rs
```diff
@@ -53,6 +69,107 @@ impl InputRules {
...
+                    fractional_output.amount_shared_secret.as_ref(),
+                )?;
+
+                // Check the bounds of the real change amount
+                if real_output_amount.token_id != fractional_output_amount.token_id {
+                    return Err(InputRuleError::RealOutputTokenIdMismatch);
+                }
+                if real_output_amount.value > fractional_output_amount.value {
+                    return Err(InputRuleError::RealOutputAmountExceedsFractional);
+                }
```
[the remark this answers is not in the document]
garbageslam: Yeah maybe this check is unnecessary, maybe we only actually need to do this check on the fractional change.

transaction/std/src/transaction_builder.rs
garbageslam: good point, I'll make sure I add a test that hits the boundary condition

api/src/convert/tx_in.rs
```diff
@@ -89,9 +102,99 @@ impl TryFrom<&external::InputRules> for InputRules {
...
+    type Error = ConversionError;
+
+    fn try_from(source: &external::RevealedTxOut) -> Result<Self, Self::Error> {
+        let tx_out =
+            tx::TxOut::try_from(source.tx_out.as_ref().ok_or(Self::Error::ObjectMissing)?)?;
+        let amount_shared_secret = source.get_amount_shared_secret().to_vec();
+        Ok(RevealedTxOut {
+            tx_out,
+            amount_shared_secret,
+        })
```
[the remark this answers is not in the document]
garbageslam: I am not sure.

We have code elsewhere in the functions that use `RevealedTxOut` that would surface this error. So we could treat that as a business logic bug rather than a serialization bug.

In a lot of existing code I think we do flag these immediately.

But I think we have been trying to embrace a new world where we have distinct Proto types and business logic types?

I think it can't hurt to flag this immediately, and since a bunch of our existing code does that I should probably continue the pattern.

transaction/std/src/transaction_builder.rs
garbageslam: I think you are right -- we should extract a common method here.

I also think since I envision that people may use "combined" SCIs that have required and fractional outputs, I should probably have a test case like that.

api/proto/external.proto
```diff
@@ -246,6 +246,30 @@ message InputRules {
     //
     // A value of zero here means no limit is enforced
     fixed64 max_tombstone_block = 2;
+
+    // Outputs permitted to be filled fractionally instead of fully (MCIP #42)
+    repeated RevealedTxOut fractional_outputs = 3;
+
+    // A change output for any leftover from this input, if a fractional fill occurs (MCIP #42)
```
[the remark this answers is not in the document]
garbageslam: The signed input is the TxOut that the originator "puts up" to the counterparty
This should typically be equal in value to the "fractional change output" (not yet renamed in this diff).

It's conceivable that it isn't, it's conceivable that instead some "required output" + "fractional change output" are equal to the value of the signed input.

The meaning of the "fractional change output" is that it is an idealized TxOut that the counterparty must provide a "real" version of in the Tx for it to be valid, and whatever amount is in the real version will be compared with the value of the idealized version to infer the "fill fraction". This fill fraction then determines the minimum amounts for all TxOut's corresponding each the "fractional output" in the Tx, and if any of those is insufficient, then the Tx is invalid.

nick-mobilecoin:
> you have a fractional output and fractional change, but you also have a required_output which is like a flat fee which doesn't depend on the fill fraction

I could see that as a potential use case so only allow partial might require odd rule logic if put into the same "levers"

> I think those tests are useful 

Sorry I failed to say that I'm not opposed to the current tests and their level.  I just feel that there should be more tests and these should probably be lower level tests closer to the logic, so as not to drown future maintainers with all of the higher level setup.  The higher level tests are needed, it's when one uses higher level tests for all the behavior, the signal quickly gets drowned out by the noise of the tests themselves.

transaction/core/src/input_rules.rs
```diff
@@ -53,6 +69,112 @@ impl InputRules {
...
+            for partial_fill_output in self.partial_fill_outputs.iter() {
+                // Try to unblind the partial fill output amount.
+                let partial_fill_output_amount = partial_fill_output.reveal_amount()?;
+                // Let's check if there is a corresponding fractional output.
+                let fractional_output = tx
+                    .prefix
+                    .outputs
+                    .iter()
+                    .find(|x| partial_fill_output.tx_out.matches_ignoring_amount(x))
+                    .ok_or(InputRuleError::MissingFractionalOutput)?;
```
nick-mobilecoin, line 132: @garbageslam,
One question that came up talking with @samdealy yesterday
&iquest; What happens if the counterparty provides 2 outputs that match.  I.e. 20MOB back to themselves for the swap, copying that txout for another 20MOB back to themselves?

I know there was mention that certain keys could ever show up once in the blockchain, is that these txout keys?

garbageslam: yeah that's exactly right :)

at the protocol level, consensus prevents two TxOut's from having the same TxOut public key, and a Tx is invalid if it has two TxOut's like that

garbageslam: But I guess the other part is, the required_output and fractional_output generally don't go to the counterparty, they are created by the originator and go back to the originator. The part that the counterparty gets is the input underlying the SCI. So copying required or fractional outputs would just be giving more money to the originator, so the counterparty would have no reason to do that

transaction/core/src/tx.rs
```diff
@@ -481,6 +481,17 @@ impl TxOut {
             .as_mut()
             .ok_or(TxOutConversionError::UnknownMaskedAmountVersion)
     }
+
+    /// Check if a TxOut matches another TxOut, except possibly in the
+    /// masked_amount. This is used in MCIP #42 partial fills rules
+    /// verification.
+    pub fn matches_ignoring_amount(&self, other: &TxOut) -> bool {
```
eranrund, line 488: Nit: rename to `eq_ignoring_amount`, I think `matches` is ambiguous in this context.

eranrund (approved):

api/proto/external.proto
```diff
@@ -246,6 +246,45 @@ message InputRules {
     //
     // A value of zero here means no limit is enforced
     fixed64 max_tombstone_block = 2;
+
+    // Outputs required to appear in the TxPrefix, but which are permitted to be
```
samdealy, line 250: Great comment, thanks for updating.

transaction/core/src/input_rules.rs
```diff
@@ -53,6 +69,116 @@ impl InputRules {
...
+        &self,
+        _block_version: BlockVersion,
+        tx: &Tx,
+    ) -> Result<(), InputRuleError> {
+        if let Some(partial_fill_change) = self.partial_fill_change.as_ref() {
+            // There is a partial fill change output. Let's try to unblind its amount.
+            let (partial_fill_change_amount, _) = partial_fill_change.reveal_amount()?;
+            // If the min fill value exceeds the fractional change, then the SCI is
+            // ill-formed
+            if self.min_partial_fill_value > partial_fill_change_amount.value {
```
samdealy, line 89: nit: I think it's slightly easier to understand that the `partial_fill_change_amount.value < min_partial_fill_value` and would recommend rewriting as such. Just my opinion however.

samdealy (approved):

transaction/core/src/input_rules.rs
```diff
@@ -53,6 +69,116 @@ impl InputRules {
...
+                //
+                // where n/d is the fill fraction. The fill fraction is 1 if
+                // fractional_change_amount = 0, and proportionally less as more of
+                // the input is returned as change.
+                // However, to avoid numerical issues, we multiply this out and
+                // verify this as a comparison of u128 values:
+                //
+                // fractional_output_value * d >= n * partial_fill_output_value
+                //
+                // Note: This should be constant time ideally
```
wjuan-mob, line 165: Just to confirm, but is this currently constant time?

transaction/core/tests/input_rules.rs
```diff
@@ -63,25 +72,486 @@ fn test_input_rules_verify_max_tombstone() {
...
+    get_first_rules(&tx).verify(block_version, &tx).unwrap();
+
+    // Add a partial fill output, by doubling one of the revealed tx outs.
+    get_first_rules_mut(&mut tx)
+        .partial_fill_outputs
+        .push(change_committed_amount(
+            &revealed_tx_outs[1],
+            Amount::new(2000, 0.into()),
+        ));
+    // Add a parital fill change, also by doubling one of the revealed tx outs.
```
wjuan-mob, line 295:
```suggestion
    // Add a partial fill change, also by doubling one of the revealed tx outs.
```

transaction/std/src/transaction_builder.rs
```diff
@@ -246,42 +247,212 @@ impl<FPR: FogPubkeyResolver> TransactionBuilder<FPR> {
...
+        if partial_fill_change_amount.value == 0 {
+            return Err(SignedContingentInputError::ZeroPartialFillChange);
+        }
+        if rules.min_partial_fill_value > partial_fill_change_amount.value {
+            return Err(SignedContingentInputError::MinPartialFillValueExceedsPartialChange);
+        }
+        if partial_fill_change_amount.token_id != sci_change_amount.token_id {
+            return Err(SignedContingentInputError::TokenIdMismatch);
+        }
+        if partial_fill_change_amount.value - rules.min_partial_fill_value < sci_change_amount.value
```
wjuan-mob, line 322: Sorry, could you clarify what this check is about? Why are we verifying that the sci_change_amount is greater than the difference between the partial_fill_change_amount and the minimum partial fill value?

wjuan-mob (commented):
I still have some confusion about the test for input rules, but everything else seems pretty good. Had some nits/typos etc.

transaction/core/src/input_rules.rs
```diff
@@ -34,13 +35,28 @@ pub struct InputRules {
...
+    /// Outputs permitted to be filled fractionally instead of fully (MCIP #42)
+    #[prost(message, repeated, tag = "3")]
+    pub partial_fill_outputs: Vec<RevealedTxOut>,
+
+    /// A change output for any leftover from this input, if a fractional fill
+    /// occurs (MCIP #42)
+    #[prost(message, tag = "4")]
+    pub partial_fill_change: Option<RevealedTxOut>,
+
+    /// A maximum allowed value for the real change output. (MCIP #42)
```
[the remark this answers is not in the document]
garbageslam: thank you!

transaction/core/src/input_rules.rs
garbageslam: i am not entirely sure honestly

it depends on what assembly `u128 < u128` is lowered to, I am not sure if that can done in one operation on x86-64 or not.

it is honestly probably fine either way, but we could make an issue to investigate.

transaction/core/tests/input_rules.rs
```diff
@@ -63,25 +72,486 @@ fn test_input_rules_verify_max_tombstone() {
...
+            &revealed_tx_outs[1],
+            Amount::new(2000, 0.into()),
+        ));
+    get_first_rules_mut(&mut tx).partial_fill_change = Some(change_committed_amount(
+        &revealed_tx_outs[0],
+        Amount::new(2000, 0.into()),
+    ));
+    get_first_rules(&tx).verify(block_version, &tx).unwrap();
+
+    // Change that partial fill output to 2000, so everything should be good again.
```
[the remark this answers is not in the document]
garbageslam: I think this is a junk comment, sorry. Will fix

transaction/core/tests/input_rules.rs
```diff
@@ -63,25 +72,486 @@ fn test_input_rules_verify_max_tombstone() {
...
+    // So this means we are filling the 1/3 of the request.
+    get_first_rules_mut(&mut tx)
+        .partial_fill_outputs
+        .push(change_committed_amount(
+            &revealed_tx_outs[1],
+            Amount::new(3000, 0.into()),
+        ));
+
+    // Set the partial fill change output to be 1500. This means the implied fill
+    // fraction is now 1/3. This should be valid.
```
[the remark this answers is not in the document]
garbageslam: we're not giving 3000, what's happening here is:

* All actual TxOut's in the transaction have a value of 1000
* We've now specified a partial_fill_change output by taking one of those and making it's amount be 1500

When validation runs, it will see the partial fill change rule for 1500, and find a corresponding output for 1000. So we gave back 2/3, and we took 1/3 of the partial_fill_change. So the fill fraction is 1/3

transaction/std/src/transaction_builder.rs
garbageslam: We have to enforce the `min_partial_fill_value` constraint, or the Tx will be invalid.

The amount that we are attempting to take, out of the partial fill change, is `partial_fill_change_amount.value - sci_change_amount.value`.

The minimum that we must take is `rules.min_partial_fill_value`.

If `partial_fill_change_amount.value - sci_change_amount.value < rules.min_partial_fill_value`, then we broke the rule and that's supposed to be an error. The error status is currently called "change limit exceeded".

This is the same as what we are checking in the code, it's just that we re-balance the equation slightly. That's mainly because when we subtract we need to do a bounds check for negative numbers or we can panic. We earlier tested that min_partial_fill_value <= partial_fill_change_amount, so that's the expression I want to use to do the test.

transaction/core/tests/input_rules.rs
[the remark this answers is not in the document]
wjuan-mob: For the sake of convenience I'm going to refer to the stuff Originator wants to receive as a result of this transaction as token2, and the stuff that the originator is putting up as token1 even though I think in this case they're all token0?

By setting partial fill amount to 3000, we are saying that the originator wants 3000 of token2.
By setting partial fill change to 1500, we are saying that the originator is providing up to 1500 of token1. Thus an exchange rate of 2 to 1.
By default of get_input_rules_test_tx the originator is receiving 1000 of token 2 as output (1/3 of their request), and 1000 of token1 as change (2/3 of their original stake)

This should verify, because 1/3 of their stake is being spent, and 1/3 of their request is being given to them.

If we set their requested output to 3001, then we are giving them less than 1/3 of their request, but spending 1/3 of their stake, so that should reject.

If we set the partial fill change to 1501, then because we are sending back 1000, we are effectively taking 501 of token1, and only giving back 1000/3000 of token2.

if the originator wants 2997 token2, and is putting up a stake of 1501 token 1, then giving them back 1000 token1 means we are taking 501/1501 token1. Which means we need to be giving back slightly more than 1000 token2 in order to make this even. So only giving 1000 should reject.

if the originator wants 2996 token2, and is putting up a stake of 1501 token 1, then giving them back 1000 token1 means we are taking 501/1501 token1. Which means we need to be giving back slightly more than 999.99 token2 in order to make this even. So only giving 1000 should accept..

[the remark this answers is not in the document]
wjuan-mob: I think rubber ducking here has explained it to me. I was just confusing myself.

api/src/convert/tx.rs
```diff
@@ -231,4 +231,136 @@ mod tests {
...
+                    &ReservedSubaddresses::from(&charlie),
+                    &mut rng,
+                )
+                .unwrap();
+            let mut sci = sci_builder.build(&NoKeysRingSigner {}, &mut rng).unwrap();
+
+            // Alice adds proofs to the SCI
+            sci.tx_in.proofs = proofs;
+
+            // Alice sends this token2 amount to Bob from Charlie, paying Charlie 1 MOB
```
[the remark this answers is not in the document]
garbageslam: you're right, this comment is wrong

transaction/core/tests/input_rules.rs
```diff
@@ -63,25 +72,486 @@ fn test_input_rules_verify_max_tombstone() {
...
+    assert_matches!(
+        get_first_rules(&tx).verify(block_version, &tx),
+        Err(InputRuleError::FractionalOutputAmountDoesNotRespectFillFraction)
+    );
+    get_first_rules_mut(&mut tx).partial_fill_outputs[0] =
+        change_committed_amount(&revealed_tx_outs[1], Amount::new(6025, 0.into()));
+    get_first_rules(&tx).verify(block_version, &tx).unwrap();
+
+    // Try tweaking the required output
+    get_first_rules_mut(&mut tx).required_outputs[0] =
```
wjuan-mob, line 545: Am I correct in thinking that the reason this fails is because even though there is an output of 1000, the required output requires an output of EXACTLY 987 and thus 1000 doesn't match?

As a corollary, if there is a partial fill ratio you want that would result in giving 1001 to originator, that would actually result in you needing to give 1000 + 1001 to originator in order to match the required output, but if you could get the partial fill ratio to exactly 1000, you only give them 1000 and the sci would be satisfied?

transaction/core/src/input_rules.rs
garbageslam: I'm going to remove the code comment

transaction/std/src/transaction_builder.rs
garbageslam: added a comment

transaction/core/src/amount/error.rs
```diff
@@ -26,4 +26,7 @@ pub enum AmountError {
 
     /// Token Id is not supported at this block version
     TokenIdNotSupportedAtBlockVersion,
+
+    /// Amount version is too old to have amount shared secret
+    AmountVersionTooOldForAmountSharedSecret,
```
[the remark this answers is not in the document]
garbageslam: I think this version is more clear right now, because of the way this error actually arises

transaction/core/src/input_rules.rs
```diff
@@ -36,32 +36,33 @@ pub struct InputRules {
...
-    //
-    // This field must be present whenever partial fills are used, because the
-    // comparison of this "idealized" output and the corresponding "fractional"
-    // change which appears in the TxPrefix is what determines the "fill fraction",
-    // that is, the degree to which a counterparty is obliged to fill every
-    // partial fill output in these rules.
-    //
-    // It is an error to use any of the partial fill options without also
-    // setting this.
+    /// A change output for any leftover from this input, which may occur during
```
nick-mobilecoin, line 46: :thought_balloon: Thoughts on `partial_fill_input` as a counter to `partial_fill_outputs`

garbageslam: my thinking is that, usually in our system an Input means:

* A specific UTXO that you are adding to a transaction
* An input ring (containing said UTXO and some mixins), which becomes a TxIn in the prefix

The `partial_fill_change` really has nothing to do with that UTXO (except that it might have the same value) -- it is a TxOut object, which is required to appear (possibly in a modified form) in the tx.prefix.outputs list.

nick-mobilecoin: I was thinking about TxOuts the other day.  

Part of my hiccups have been, I see a function and internally ask what are the inputs and outputs.  Then I see `TxOut` and instinctively want to mark it an output. As you say, it's often an input in the code. It is named appropriately as it's a data structure which is an _output_ of the final transaction.

garbageslam: Yeah I remember discussing this a lot with Kyle, it's a stumbling block for people that TxOut is an input to a transaction.

I don't know if it's actionable though -- the idea of the UTXO blockchain model is that every input to a transaction is the output of an earlier transaction, and the industry term for these objects is TXO / UTXO. If we depart too much from this terminology it's going to be confusing to developers from other blockchains, and to users who are familiar with other blockchains.

garbageslam: Like ultimately, these terms come from Bitcoin

nick-mobilecoin: I agree common industry terms should win out.  I think it's just something I'll need to keep in mind

Thanks!

wjuan-mob (approved):

PR #2493, opened 2022-09-08
[Fixed] CI/CD - fix automatic CD environment delete when feature/release branch is deleted.

jgreat:
### Motivation
Fix automatic dev environment clean up action.  I tried to use the GITHUB_REF_NAME, but this is set for the repo default branch when a `on: delete` action is triggered. Needed to get the deleted branch name out of the event payload.

- Fix `on: delete` logic to use the GHA event payload instead of GHA ref_name.  
  - The ref name will always be the default branch, we need the branch that was deleted.
- Update metadata.sh script to use DELETE_REF_NAME as branch when provided. 
- Merge namespace reset/delete functionality 
- Remove redundant reset workflow_dispatch
- Move cd workflow to use workflow_call for environment reset.
- Minor fixes around `required` usage in workflow inputs for clean `actionlint` output.

Note, there's no need to cherry-pick or back port this to the v2 release.  The delete action only runs logic from the head of the repo default branch.


.github/workflows/mobilecoin-dev-cd.yaml
```diff
@@ -380,24 +380,18 @@ jobs:
 # Reset existing namespace
 #################################
   dev-reset:
-    runs-on: [self-hosted, Linux, small]
     needs:
     - generate-metadata
-    steps:
```
jgreat, line 386: Convert this from a workflow_dispatch to the workflow_call. 

This is a little bit more efficient since you don't need to consume a runner to wait on the dispatch to complete.

GHA didn't support retries on workflow_calls when we were first developing these patterns, I'll be converting the rest as we go.

joekottke (approved):
LGTM

garbageslam (approved):
I don't fully understand this but I'm willing to approve it, it seems important :)

PR #2501, opened 2022-09-09
Simplify the situation around ProposeTxResult vs. TransactionValidationError

garbageslam:
Note: This PR came out of this discussion on another PR: https://github.com/mobilecoinfoundation/mobilecoin/pull/2484#discussion_r966436661 and also discussion in discord

`TransactionValidationError` is an error type created by `mc-transaction-core` in the validation routines. This is the actual low level error returned by validation, and is marshalled across the consensus enclave boundary when something fails.

`ProposeTxResult` is a protobuf enum defined in the consensus_common.proto file. When a submission happens, but fails due to a transaction validation error, we return a grpc success, but with a nonzero "status code" in the form of `ProposeTxResult`. This is supposed to capture the details of the `TransactionValidationError`. In terms of code, the `ProposeTxResult` type is created in a crate that uses `std` and cannot be represented in the enclave.

Currently, we support converting `TransactionValidationError` to `ProposeTxResult`, which is necessary to convert errors returned by the enclaves to errors that can go into a grpc response.

But, we also support the reverse conversion, for no particularly good reason. `ThickClient` exercises the reverse conversion, so that non-zero `ProposeTxResult` get converted into a transaction-core / enclave error type when they are returned to the caller.

This conversion creates a bunch of complexity because it seems bad for it to be lossy, but making it not lossy is very complicated and imposes constraints on what kinds of data these errors can contain. It is better if the transaction-core code can do whatever it wants with it's error types, without regard to concerns about marshalling them into protobuf, which should be a higher-level concern.

The idea here is:
* Remove the conversion from `ProposeTxResult` back into `TransactionValidationError`. Only the one direction is supported. Once it is turned into a grpc error type, it doesn't get converted back into the enclave error type later.
* The consensus client api service now returns both the `ProposeTxResult` status code and an error message, which is the display-string of the original error, in the protobuf response. This is a non-breaking change.
* On the thick client side, instead of trying to reconstruct the `TransactionValidationError`, we just return the `ProposeTxResult` and the error string.

This promotes a separation of concerns -- now the transaction-core error code doesn't need to support any particular conversions to support funky deserialization strategies.

This also frees us to make the `ProposeTxResult` status code a little bit less granular if we want. Developers can always get all the details from the err_msg string to debug something, and we don't have to create more status codes for every possible transaction validation error, unless we think the client will actually handle those failure modes differently. (In a recent draft PR, we got up to 65 distinct ways that a transaction can fail at one point, but client code only actually checks for one or two of them.) It also means that we can put data in the transaction validation errors even if we don't have a good way to reconstruct it from `ProposeTxResult`.

### DIFF
--- a/connection/src/error.rs
+++ b/connection/src/error.rs
@@ -10,7 +10,6 @@ use grpcio::Error as GrpcError;
 use mc_blockchain_types::ConvertError;
 use mc_consensus_api::{consensus_common::ProposeTxResult, ConversionError};
 use mc_crypto_noise::CipherError;
-use mc_transaction_core::validation::TransactionValidationError;
 use std::{array::TryFromSliceError, result::Result as StdResult};
 
 pub type Result<T> = StdResult<T, Error>;
@@ -31,8 +30,8 @@ pub enum Error {
     Cipher(CipherError),
     /// Attestation failure: {0}
     Attestation(Box<dyn AttestationError + 'static>),
-    /// Transaction validation failure: {0}
-    TransactionValidation(TransactionValidationError),
+    /// Transaction validation failure: {0:?}: {1}
+    TransactionValidation(ProposeTxResult, String),
     /// Other error: {0}
     Other(String),
 }
@@ -86,14 +85,6 @@ impl From<ConversionError> for Error {
 impl From<TryFromSliceError> for Error {
     fn from(_src: TryFromSliceError) -> Self {
         ConversionError::ArrayCastError.into()
-    }
-}
-
-impl From<ProposeTxResult> for Error {
-    fn from(src: ProposeTxResult) -> Self {
-        src.try_into()
-            .map(Self::TransactionValidation)
-            .unwrap_or_else(|err| Error::Other(err.into()))
     }
 }
 
--- a/connection/src/thick.rs
+++ b/connection/src/thick.rs
@@ -429,7 +429,10 @@ impl<CP: CredentialsProvider> UserTxConnection for ThickClient<CP> {
         if resp.get_result() == ProposeTxResult::Ok {
             Ok(resp.get_block_count())
         } else {
-            Err(resp.get_result().into())
+            Err(Error::TransactionValidation(
+                resp.get_result(),
+                resp.get_err_msg().to_owned(),
+            ))
         }
     }
 }
--- a/consensus/api/proto/consensus_common.proto
+++ b/consensus/api/proto/consensus_common.proto
@@ -105,4 +105,7 @@ message ProposeTxResponse {
 
     /// The block version which is in effect right now
     uint32 block_version = 3;
+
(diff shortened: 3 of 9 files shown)


consensus/api/src/conversions.rs
```diff
@@ -71,63 +71,6 @@ impl From<Error> for ProposeTxResult {
     }
 }
 
-/// Convert ProposeTxResult --> TransactionValidationError.
-impl TryInto<Error> for ProposeTxResult {
```
garbageslam, line 75: Removing this conversion is the main goal of this PR, since it helps develop #2484

consensus/api/proto/consensus_common.proto
```diff
@@ -105,4 +105,7 @@ message ProposeTxResponse {
 
     /// The block version which is in effect right now
     uint32 block_version = 3;
+
+    /// Human-readable error message, in case of nonzero ProposeTxResult
+    string err_msg = 4;
```
garbageslam, line 110: This is where we store an error message with the `ProposeTxResponse`, adding this is a non-breaking change.

connection/src/thick.rs
```diff
@@ -429,7 +429,10 @@ impl<CP: CredentialsProvider> UserTxConnection for ThickClient<CP> {
         if resp.get_result() == ProposeTxResult::Ok {
             Ok(resp.get_block_count())
         } else {
-            Err(resp.get_result().into())
+            Err(Error::TransactionValidation(
+                resp.get_result(),
+                resp.get_err_msg().to_owned(),
+            ))
```
garbageslam, line 435: this is where, instead of invoking the conversion that has since been removed, we return the `ProposeTxResult` that the server gave us, and the error message the server offered us

connection/src/error.rs
```diff
@@ -31,8 +30,8 @@ pub enum Error {
     Cipher(CipherError),
     /// Attestation failure: {0}
     Attestation(Box<dyn AttestationError + 'static>),
-    /// Transaction validation failure: {0}
-    TransactionValidation(TransactionValidationError),
+    /// Transaction validation failure: {0:?}: {1}
+    TransactionValidation(ProposeTxResult, String),
```
garbageslam, line 34: This is the part of this that is a breaking change to the `ThickClient`, since, instead of returning the previous error type, it now returns this error code as a message. But, this is what the server actually gave you over grpc, instead of what you sort of hackishly reconstructed from the grpc response.

For instance, previously if the Validation error was "Ledger", you would just get the empty string, because it's payload is a string on the server side and that can't be marshalled into a proto enum. But now you would actually get the error string on the client side.

nick-mobilecoin (approved):

remoun (approved):

eranrund (approved):

PR #2505, opened 2022-09-12
expose signing data from transaction builder

APshenkin:
This will allow external services to sign and constuct Mobilecoin transactions

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5513,6 +5513,7 @@ dependencies = [
  "mc-account-keys",
  "mc-api",
  "mc-crypto-keys",
+ "mc-crypto-ring-signature-signer",
  "mc-transaction-core",
  "mc-transaction-std",
  "rand 0.8.5",
--- a/transaction/core/src/ring_ct/mod.rs
+++ b/transaction/core/src/ring_ct/mod.rs
@@ -10,6 +10,6 @@ pub use self::{
     error::Error,
     generator_cache::GeneratorCache,
     rct_bulletproofs::{
-        InputRing, OutputSecret, PresignedInputRing, SignatureRctBulletproofs, SignedInputRing,
+        InputRing, OutputSecret, PresignedInputRing, SignatureRctBulletproofs, SignedInputRing, SigningData
     },
 };
--- a/transaction/core/src/ring_ct/rct_bulletproofs.rs
+++ b/transaction/core/src/ring_ct/rct_bulletproofs.rs
@@ -45,6 +45,40 @@ pub struct PresignedInputRing {
     pub pseudo_output_secret: OutputSecret,
 }
 
+/// Signature data that is required for any signer to produce signature
+/// and construct transaction
+#[derive(Clone, Debug)]
+pub struct SigningData {
+    /// Transaction extended message
+    pub extended_message: Vec<u8>,
+
+    /// pseudo output blindings
+    pub pseudo_output_blindings: Vec<Scalar>,
+
+    /// Commitments of value equal to each real input.
+    pub pseudo_output_commitments: Vec<CompressedCommitment>,
+
+    /// Proof that all pseudo_outputs and transaction outputs are in [0, 2^64).
+    /// This contains range_proof.to_bytes(). It is stored this way so that this struct may derive
+    /// Default, which is a requirement for serializing with Prost.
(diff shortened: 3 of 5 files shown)


garbageslam:
Hi,

Right now, I don't understand the movitation for this change. (I have some idea from the offline discussions what problem you are trying to solve, but none of that is documented in this PR, and it still seems to me that this can be simpler than the code I am reading now.)

1. Why is it not sufficient to inject your own ring-signer into the transaction builder?
2. If the answer is that you need more context on what you are signing, what if we extended the `RingSigner` trait to provide that information?

The code as written has a bunch of drawbacks:

> This will allow external services to sign and constuct Mobilecoin transactions

1. If the idea is that `SigningData` becomes an externally facing API type, that's very different from our current contract with users. Currently we get to change the `SignatureRctBulletproofs` and `TxPrefix` types at will without regard to API stability, because these types are never committed to the blockchain, and the `Tx` object is deleted in the enclave. We made heavy use of this in the development of confidential token ids for instance.

So, I think there's no way we are willing to commit to any API stability around `SigningData`. If you want to use this API in an "external service", you may have to accept constant breaking changes to it.

2. We may not be comfortable with the idea that we have external services depending on APIs that aren't stable, especially if they are important partners that really cannot break.

3. The `RingSigner` trait seems extremely similar to your ultimate goal, of providing an external / offline signing facility that doesn't need to understand all the details of a MobileCoin transaction. I would like to strengthen and develop that API so that it can meet all requirements, and we don't end up with different engineers solving the same problem in multiple different ways, and then having to maintain lots of different ways of doing the same thing. We already have hardware wallet projects depending on the `RingSigner` trait, so we are committed to supporting that.

4. I don't understand why you actually need the `SigningData`. Copper has been working for years without you needing 

```
    /// pseudo output blindings
    pub pseudo_output_blindings: Vec<Scalar>,

    /// Commitments of value equal to each real input.
    pub pseudo_output_commitments: Vec<CompressedCommitment>,

    /// Proof that all pseudo_outputs and transaction outputs are in [0, 2^64).
    /// This contains range_proof.to_bytes(). It is stored this way so that this struct may derive
    /// Default, which is a requirement for serializing with Prost.
    pub range_proof_bytes: Vec<u8>,
```

And these things have existed since day 1.

What changed that caused you to need these things?

5. If you switch to the API of the `RingSigner` trait, you will get this:

https://github.com/mobilecoinfoundation/mobilecoin/blob/e4e29c63604453aafb0a60458e7765671e84a706/crypto/ring-signature/signer/src/traits.rs#L15

I believe this will enable you to compute everything you have listed in `SigningData`, with the exception of the range proofs. But again, I don't understand right now why you need to inspect the range proofs. I would rather not give that to you so that I know that if I change it, I can't possibly break you.

It seems like there may be a scope creep implicit in this PR, or a tendency to just add as much stuff as possible to this API in case you might need it later.

That is not a good way to design stable APIs -- we want to preserve the ability to evolve our transaction format without breaking copper. So let's try to sit down and figure out if you actually need this API, why you need each thing you are asking for, and why this is better than the alternatives.

For example, what if we changed the `RingSigner` trait so that:

* It didn't just sign one ring at a time, the caller passes all the rings to be signed at once. (The hardware wallets will do their own for-loop, or batch sign all the rings at once, depending on what is convenient for them.)
* It passed additional context that might not be strictly needed but might be interesting to the caller. (The hardware wallets will simply ignore that.)

Would that meet your needs? Because I think that would leave us with a simpler unified API.

I understand that copper unlimited wants to submit the Tx and not have full-service do that. But once you return from the ring signer call, it can spit out the Tx it built and give that to you, and you can submit it. And then even if we change the Tx format that the enclave recognizes, you will not be broken.

eranrund (approved):

eranrund (changes requested):
Sorry, I approved this by mistake. We're still working on reviewing it.

garbageslam:
Thanks for the continued offline discussion -- can you confirm that this version of the PR will meet your needs? We made this other version that contains a struct `UnsignedTx` that seems like it should be good for your purposes

https://github.com/mobilecoinfoundation/mobilecoin/pull/2519

APshenkin:
@garbageslam 
Looks exactly what we are expecting! Thank you

Will close this for now

PR #2514, opened 2022-09-13
Add ci.yml and actions from master

jcape:
- Copy our GHA CI from `master` to `release/v2`.

### Motivation

Right now, `release/v2` only has CI from CircleCI, and `master` only has CI from GHA. Unfortunately, turning on CircleCI for `release/v2` breaks CI on `master`, and leaving it off means no CI on `release/v2`.

### Future Work
- [ ] Turn off CircleCI for good once this is merged.

### DIFF
--- a/.github/actions/check-dirty-git/action.yml
+++ b/.github/actions/check-dirty-git/action.yml
@@ -1,0 +1,17 @@
+name: Check dirty git
+description: Check the git status of the current directory, and raise an error if there are uncommitted changes.
+
+runs:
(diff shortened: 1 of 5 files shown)


garbageslam:
@jgreat can we confirm that we would even deploy `go-grpc-gateway` from this repo? I thought at one point it was in a separate repo and i'm not sure how the release-v2 release process is working. If it's not actually being deployed from here then it seems okay to disable it's tests?

PR #2525, opened 2022-09-14
Add `mc_core` package with new no_std compatible key / account types and logic

ryankurte:
adds `mc_core` (and `mc_core_types`) to provide a `#![no_std]` compatible set of base business-logic types and interfaces for doin mobilecoin things both within and outside the mobilecoin codebase.

- adds basic account, subaddress, key types (and their derivations and conversions)
  (primary derivations are implemented over tuple-references so they're cheap to / easy reuse as required)
  - `Subaddress` trait for deriving subaddresses
  - `RingCTAddress` trait moved here (now returns objects instead of refs as this must be implementable over private key types)
- replaces `mc_account_keys_slip10` with `mc_core::slip10`
  (we need the slip10 component for the h/w wallet)

**this PR does not replace all use of keys with the new types, it is intended that updates to the wider codebase can be made in a piecewise manner to avoid merge conflict nightmares**

### Motivation

imagine if the type system could make sure your keys were in the right order...

`no_std` support requires base types that don't pull in the dependency zoo, and this is a first step towards enforcing correct business logic via the type system (which should improve the docs too).

### Future Work

- support ring signatures (to be included in streaming ring_signature refactor PR )
- propagate new types... everywhere (better to do this in a piece-wise manner)

Soundtrack: [I dreamed a Dream (of concrete business logic types)](https://www.youtube.com/watch?v=uimsKQ0mlS4)


core/types/src/keys.rs
```diff
@@ -0,0 +1,356 @@
...
+}
+
+/// [`core::fmt::LowerHex`] where `KEY: AsRef<[u8]>`
+impl<ADDR, KIND, KEY: AsRef<[u8]> + Default + Zeroize> core::fmt::LowerHex
+    for Key<ADDR, KIND, KEY>
+{
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        let data = self.key.as_ref();
+        for d in data {
+            write!(f, "{:02x}", d)?;
```
eranrund, line 86: I think we have avoided putting out private keys when doing string representations (e.g. https://github.com/mobilecoinfoundation/mobilecoin/blob/master/crypto/keys/src/ristretto.rs#L150)
That reduces the chance of the key ending in a log message somewhere.

eranrund: Maybe this is fine due to https://github.com/mobilecoinfoundation/mobilecoin/blob/328df8f3e51aa461df4232393eeb0b56a9c6d47d/core/types/src/keys.rs#L289

core/build.rs
```diff
@@ -0,0 +1,49 @@
...
+    Ok(())
+}
+
+#[cfg(feature = "protos")]
+const FIELD_ATTRS: &[(&str, &str)] = &[];
+
+#[cfg(feature = "protos")]
+const TYPE_ATTRS: &[(&str, &str)] = &[];
+
+/// Build protocol files from specified directory
```
samdealy, line 21: nit: protobuf

core/examples/slip10.rs
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
+//! SLIP-0010 example applet
+
+use bip39::Mnemonic;
+use clap::Parser;
+use mc_core::{
+    account::Account, prelude::RingCtAddress, slip10::Slip10KeyGenerator, subaddress::Subaddress,
```
samdealy, line 8: I'm curious- how do you decide whether to import a struct via the `preulde` module or, in this case, the `account` module?

core/src/prelude.rs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
+//! prelude to simplify imports for consumers of `mc_core`
+
+pub use mc_core_types::{
+    account::{Account, PublicSubaddress, RingCtAddress, SpendSubaddress, ViewSubaddress},
```
samdealy, line 6: See question above about when to use `Account` as `use mc_core_types::Account` or `use mc_core_types::prelude::Account`.

core/src/slip10/mod.rs
```diff
@@ -0,0 +1,241 @@
...
+
+//! MobileCoin SLIP-0010 / BIP39 Based Key Derivation
+//!
+//! This provides utilities to handle SLIP-0010 key bytes and their relation to
+//! the MobileCoin [`Account`](mc_core::Account) structure, which contains a
+//! pair of Ristretto255 view/spend private scalars.
+// As well as providing traits to create a Slip10Key from entropy and path,
+// along with the canonical method of converting a BIP-39
+// [`Mnemonic`](tiny_bip32::Mnemonic) with a given BIP-32 path into a
+// [`Slip10Key`](Slip10Key) usable within MobileCoin.
```
samdealy, line 11: nit: Should these lines be prepended with `//!`?

core/types/src/lib.rs
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
+//! Mobilecoin core type definitions
+
+#![no_std]
+
+pub mod keys;
+
+pub mod account;
```
samdealy, line 9: nit: 
`
pub mod account; 
pub mod keys;
`

util/keyfile/src/mnemonic_acct.rs
```diff
@@ -1,3 +1,5 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
 // Copyright 2018-2022 The MobileCoin Foundation
```
samdealy, line 3: <i class='graphite__hidden'>[Re: lines 3 to 3]<br/></i>

nit: remove<span class='graphite__hidden'><br/><br/>See this comment inline on <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2525?utm_source=unchanged-line-comment">Graphite</a>.</span>

core/Cargo.toml
```diff
@@ -0,0 +1,60 @@
+[package]
+name = "mc-core"
+version = "2.0.0"
+authors = ["MobileCoin"]
+edition = "2021"
+description = "MobileCoin Core Library"
+readme = "README.md"
+
+[features]
+bip39 = ["tiny-bip39", "slip10_ed25519"]
```
samdealy, line 10: optional: consider using `dep:tiny-bip39` and `dep:slip10_ed25519` to emphasize that those are dependencies and not features. I understand that you're using `dep:serde` out of neccesity to disambiguate with the feature also called `serde`, but I also find `dep:` useful in that it distinguishes dependencies from features.

https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies

core/Cargo.toml
```diff
@@ -0,0 +1,60 @@
...
+internals = [ ]
+protos = ["prost", "prost-build"]
```
samdealy, line 12: optional: see above, `dep:prost`, `dep:prost-build`

account-keys/src/account_keys.rs
```diff
@@ -24,6 +24,10 @@ use core::{
 };
 use curve25519_dalek::{ristretto::RistrettoPoint, scalar::Scalar};
 use mc_account_keys_types::RingCtAddress;
+use mc_core::{
+    prelude::{RootSpendPrivate, RootViewPrivate},
+    slip10::Slip10Key,
+};
 use mc_crypto_digestible::Digestible;
 use mc_crypto_hashes::{Blake2b512, Digest};
 use mc_crypto_keys::{RistrettoPrivate, RistrettoPublic};
```
samdealy, line 33: Do you plan to eventually replace these `RistrettoPublic` types with `RootViewPublic` and `RootSpendPublic`? (Curious if that's in the works and you just didn't want to do it in this PR because it'd make it too large.)

samdealy (approved):
Can you add a description to `core/README.md`?

core/types/src/account.rs
```diff
@@ -0,0 +1,163 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
+//! Mobilecoin Account and Subaddress objects
```
nick-mobilecoin, line 3: :upside_down_face: (and other places in this PR) 
it looks like it's usually title case *MobileCoin*
```suggestion
//! MobileCoin Account and Subaddress objects
```

core/types/src/account.rs
```diff
@@ -0,0 +1,163 @@
...
+
+    /// Fetch account spend private key
+    pub fn spend_private_key(&self) -> &RootSpendPrivate {
+        &self.spend_private
+    }
+}
+
+/// Mobilecoin private subaddress object
+#[derive(Clone, Debug, PartialEq)]
+pub struct SpendSubaddress {
```
nick-mobilecoin, line 65: :wrench: I may not be up on the terms, but the description here says "private subaddress object" and there is a `Publicsubaddress`, should this be `PrivateSubaddress`?

core/types/src/account.rs
```diff
@@ -0,0 +1,163 @@
...
+//! Mobilecoin Account and Subaddress objects
+
+use zeroize::Zeroize;
+
+use crate::keys::{
+    RootSpendPrivate, RootSpendPublic, RootViewPrivate, RootViewPublic, SubaddrSpendPrivate,
+    SubaddrSpendPublic, SubaddrViewPrivate, SubaddrViewPublic,
+};
+
+/// An object which represents a subaddress, and has RingCT-style
```
nick-mobilecoin, line 12: :question: Is the same trait in `mc-account-keys` going to go away in a subsequent PR?

core/types/src/account.rs
```diff
@@ -0,0 +1,163 @@
...
+        &self.view_private
+    }
+}
+
+/// Mobilecoin public subaddress object
+#[derive(Clone, Debug, PartialEq)]
+pub struct PublicSubaddress {
+    /// Public address view public key
+    pub view_public: SubaddrViewPublic,
+    /// Public address spend public key
```
nick-mobilecoin, line 129: :upside_down_face: the "public <blah> public" reads odd
```suggestion
    /// Subaddress view public key
    pub view_public: SubaddrViewPublic,
    /// Subaddress spend public key
```

core/types/src/keys.rs
```diff
@@ -0,0 +1,356 @@
...
+
+use curve25519_dalek::{ristretto::RistrettoPoint, scalar::Scalar};
+use zeroize::Zeroize;
+
+use mc_crypto_keys::{KeyError, RistrettoPrivate, RistrettoPublic};
+
+// Exported key types
+
+/// Subaddress view private key
+pub type SubaddrViewPrivate = Key<Subaddr, View, RistrettoPrivate>;
```
nick-mobilecoin, line 18: :wrench: The use of `Subaddr` here  is asymetrical with that of `Subaddress` in `accounts.rs` and `subaddress.rs` suggest spelling it out here as well.

core/types/src/keys.rs
```diff
@@ -0,0 +1,356 @@
...
+}
+
+/// PartialEq for backwards compatibility with private key objects
+impl<ADDR, KIND> PartialEq<Key<ADDR, KIND, RistrettoPrivate>> for RistrettoPrivate {
+    fn eq(&self, other: &Key<ADDR, KIND, RistrettoPrivate>) -> bool {
+        RistrettoPublic::from(self) == RistrettoPublic::from(&other.key)
+    }
+}
+
+/// [`core::fmt::Display`] for public key objects
```
nick-mobilecoin, line 286: :upside_down_face: 
Maybe the intent was to document the output is converted to a public key.
```suggestion
/// [`core::fmt::Display`] for private key objects
```

core/src/prelude.rs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
+//! prelude to simplify imports for consumers of `mc_core`
+
+pub use mc_core_types::{
```
nick-mobilecoin, line 5: :question: Not quite understanding the prelude here.  It looks like everywhere it's currently used the types are used directly.
Do we forsee any future where we want to expose only Public keys, then having the privates in the prelude may not be desired.

nick-mobilecoin (changes requested):
I think the behaviour looks good, mainly comments and questions suggesting alternatives or minimizing some of these changes.

Especially since this is going to be in a longer lived feature branch I think we should keep future PRs more focused on smaller changes so that it's easier to grab those changes into the mainline and or resolve conflicts syncing with mainline changes.

account-keys/src/account_keys.rs
```diff
@@ -318,6 +332,30 @@ impl AccountKey {
...
+    /// * `fog_report_id` - The id labelling the report to use, from among the
+    ///   several reports which might be served by the fog report server.
+    /// * `fog_authority` - The DER-encoded subjectPublicKeyInfo of the fog
+    ///   authority, which will be signed by the user when constructing the
+    ///   public address.
+    pub fn with_fog(
+        self,
+        fog_report_url: &str,
+        fog_report_id: &str,
+        fog_authority_spki: &[u8],
```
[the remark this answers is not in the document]
ryankurte: ohh neat idea, but i can't depend on [`alloc`](https://doc.rust-lang.org/stable/alloc/string/trait.ToString.html) here

core/src/slip10/mod.rs
```diff
@@ -0,0 +1,241 @@
...
+}
+
+#[cfg(feature = "internals")]
+impl Slip10Key {
+    /// Create a SLIP-0010 key from raw Ed25519 private key value
+    pub fn from_raw(raw: [u8; 32]) -> Self {
+        Self(raw)
+    }
+}
+
```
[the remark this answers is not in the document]
ryankurte: this is necessary for [hw support](https://github.com/mobilecoinofficial/ledger-mob/blob/main/core/src/lib.rs#L30), i'll have a look at whether `From<Scalar>` can be equivalent

account-keys/src/account_keys.rs
```diff
@@ -318,6 +332,30 @@ impl AccountKey {
...
+    ) -> Self {
+        AccountKey::new_with_fog(
```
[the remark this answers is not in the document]
ryankurte: i've been told the `AccountKeys` object (and crate) cannot be changed without causing breakages due to cursed dependencies on the field layouts / encodings? fortunately with the `mc_core` types i don't (at this stage) need anything from `mc_account_keys`...

my Prime Directive is hardware wallet support, so i'm aiming to make improvements where i need them and PR 'em back to minimise the maintenance / review burden on the hardware side, but (at least at the moment) only as much as is needed for the hardware wallets.

core/Cargo.toml
```diff
@@ -0,0 +1,60 @@
...
+serde = ["dep:serde", "mc-core-types/serde"]
+
+default = ["bip39", "heapless/default"]
+
+[dependencies]
+base64 = { version = "0.13.0", default-features = false }
+curve25519-dalek = { version = "4.0.0-pre.2", default-features = false }
```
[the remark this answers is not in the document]
ryankurte: you mean the feature flags for dalek backends? these create a Big Problem when you're trying to cross compile so should only be enabled at the very top level to avoid multiple backends being enabled when you don't intend it...

core/src/subaddress.rs
```diff
@@ -0,0 +1,162 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
+//! Mobilecoin Subaddress Derivations
+
+use curve25519_dalek::{ristretto::RistrettoPoint, scalar::Scalar};
```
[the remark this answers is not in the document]
ryankurte: nope, just that `mc_core` will provide the types / basics / canonical methods for derivation of addresses etc. to be used from other places (in a no_std and alloc free way)

core/src/subaddress.rs
```diff
@@ -0,0 +1,162 @@
...
+    /// Subaddress type
+    type Output: core::fmt::Debug;
+
+    /// Generate the subaddress for the corresponding index
+    fn subaddress(&self, index: u64) -> Self::Output;
+}
+
+/// Generate subadress private keys from root private keys
+impl<'a> Subaddress for (&'a RootViewPrivate, &'a RootSpendPrivate) {
+    type Output = (SubaddrViewPrivate, SubaddrSpendPrivate);
```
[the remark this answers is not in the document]
ryankurte: so this is an outcome of removing duplicated derivation logic... the idea was the tuple-by-reference versions are the One True Impl, then implementations on higher level types can call down without needing to construct a viable higher level object (or copy the keys into this stack frame).

it wouldn't really matter if you -did- need to construct an `Account`, it's just nice if you already have your keys hanging around not to need to create a new object.

core/types/src/account.rs
ryankurte: ahh, i could improve the comments here. afaik:

- `PublicSubaddress` has public view and spend keys and is a target for transactions
- `ViewSubaddress` has the public spend and private view keys so you can _view_ a transaction
- `SpendSubaddress` as private spend and view keys so you can _spend_ a transaction.

so i don't think `PrivateSubaddress` makes sense in this context. this is based on nomenclature used in hw support discussions, i am more than happy to be corrected if there are better descriptors.

core/types/src/account.rs
ryankurte: ahh yes, in my branch the `mc-account-keys` version is just a re-export of this but i seem to have missed pulling it in / might as well just drop it i suspect.

core/types/src/keys.rs
ryankurte: for the type aliases to work the label types have to be public, so it's actually sort of deliberate that the labels are abbreviated so they don't collide with other names.

another approach i have used in the past is to be more explicit in the labels so it's like `SubaddressTy`, could also do that if it's preferred?

core/src/prelude.rs
ryankurte: ooh yeah, good catch, given the prelude i probably shouldn't be re-exporting that without a module.

core/src/prelude.rs
ryankurte: i view the prelude as a nicety for external consumers (like the ledger-mob app!), though we could also use it internally i guess. and if you'd prefer i could drop the `keys` re-exports given -most- users probably wouldn't need em.

core/src/prelude.rs
ryankurte: right, so this is an exercise in dependency-pickle-avoidance... my concept for `mc_core` is that eventually it'd provide all the basics you need to Do MobileCoin Stuff; so basic types and re-exporting functionality like ring signatures / transactions / memos (while maintaining the no_std and no alloc reqs) for internal and external use. 
(again in the context of hardware support, this is the crate other apps will depend on)

the rules as they stand:
- internal packages **SHOULD** depend on `mc_core_types` _unless_ they require functionality in `mc_core` (`RingCTAddress` or `Subaddress` being two particular examples).
- anything `mc_core` needs to re-export (ie. `mc-crypto-ring-signature` and children) **MUST** depend on `mc_core_types` for the avoidance of dependency loops.

which i'll add to the readme / crate docs

core/src/lib.rs
```diff
@@ -0,0 +1,27 @@
...
+//! Mobilecoin core library.
+//! This provides base types and common functions for mobilecoin implementers /
+//! consumers.
+
+#![no_std]
+#![warn(missing_docs)]
+#![deny(unsafe_code)]
+#![allow(non_snake_case)]
+
+#[cfg(feature = "alloc")]
```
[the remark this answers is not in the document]
ryankurte: because i'm extracting things from my ~~fanfic~~ fork there's a bit of cludge around like this, there shouldn't be an `alloc` feature at the moment but in my alternate universe there is. good catch / sorry for the volume of these!

account-keys/src/account_keys.rs
ryankurte: sorta! the idea is that eventually all instances of `RistrettoX` would be replaced by the types in `mc_core_types::keys` to reduce opportunities for misuse (like the classic swapping spend and view keys misadventure). 

i need compatible types for the hardware wallet work, and will propagate these in the areas i'm working on (primarily to remove dependencies on `mc-account-keys`), but i'm hoping that once we have the types available other folks will pick up and run with this in other areas of the codebase.

account-keys/src/account_keys.rs
```diff
@@ -318,6 +332,30 @@ impl AccountKey {
         }
     }
 
+    /// Attach fog fields to an `AccountKey` instance
```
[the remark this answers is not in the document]
ryankurte: it is a new instance, but it's consuming `self`, how would you prefer this described?

core/build.rs
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
```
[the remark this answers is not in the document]
ryankurte: `mc-api` is not `no_std` or alloc free, i think further along it'd be useful to create the types here then consume them elsewhere but can omit this for now

core/src/slip10/mod.rs
```diff
@@ -0,0 +1,241 @@
...
+pub use bip39::{Language, Mnemonic};
+
+use crate::{
+    account::Account,
+    consts::{COINTYPE_MOBILECOIN, USAGE_BIP44},
+    keys::{RootSpendPrivate, RootViewPrivate},
+};
+
+/// Fetch the BIP39 path for a given account index
+pub const fn wallet_path(account_index: u32) -> [u32; 3] {
```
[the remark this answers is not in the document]
ryankurte: it's non-critical but useful to be able to grab a canonical path externally (ie. used in ledger-mob) rather than recreating per-platform

core/src/slip10/mod.rs
```diff
@@ -0,0 +1,241 @@
...
+    [
+        0x80000000 | USAGE_BIP44,
+        0x80000000 | COINTYPE_MOBILECOIN,
+        0x80000000 | (account_index & !0x80000000),
```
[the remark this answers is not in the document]
ryankurte: you're correct it's superfluous, the intent was to communicate to a reader that the `account_index` MSB is not used (though the SECURE flag is set), given the compiler is going to optimise it anyway.

core/src/slip10/testvectors.py
```diff
@@ -143,5 +143,4 @@ def show_testvectors(name, curvenames, seedhex, derivationpath):
 
 show_testvectors("Real", ['ed25519'],
                  '5eb00bbddcf069084889a8ab9155568165f5c453ccb85e70811aaed6f6da5fc19a5ac40b389cd370d086206dec8aa6c43daea6690f20ad3d8d48b2d2ce9e38e4',
-                 [privdev + 44, privdev + 866, privdev + 0])
-
+                 [privdev + 44, privdev + 866, privdev + 0])
```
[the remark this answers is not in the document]
ryankurte: huh, the pre-commit hooks were upset with me for -having- an empty line here

account-keys/src/account_keys.rs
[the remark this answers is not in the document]
nick-mobilecoin: While I understand it might be exposing the abstraction.  The signature takes a `self` not a `mut self`.  If one were only to look at the signature and the description it might make one question practices, i.e. is this implementation going into `unsafe` to make the item mutable.
```suggestion
    /// Copy to a new instance using with provided fog fields
```
Not sure I like the "copy to"...  Originally I had "Create a new..." but then it didn't feel clear enough how the keys were handled.

core/src/slip10/mod.rs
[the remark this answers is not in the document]
nick-mobilecoin: The "hw support" link doesn't seem to work.  I'm not sure there is currently a strong practice of documenting features, so don't have a good suggestion on where to place this for future maintainers.

I'm guessing "internals" was discussed already.  For normal language it seems to be overly generic.  It may be a domain term I'm not familiar with.

core/src/slip10/mod.rs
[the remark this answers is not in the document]
nick-mobilecoin: different perspectives then :)

If someone isn't adept enough to understand that `0x80000000 | x` will always set the MSB, then I wonder how easily they'll understand the bitwise not.  I wonder if this is any better or just a different way of thinking.
```suggestion
        0x80000000 | (account_index & 0x7FFFFFFF),
```

core/src/slip10/testvectors.py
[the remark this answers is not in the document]
nick-mobilecoin: Bad wording on my end I said "missing empty line" when really it should have been "missing newline"

I think it's the difference between a newline and a blankline.  The pre-commit hook will bark on an empty line, two consecutive newlines, while github UI puts a kind of :heavy_minus_sign: when there is no newline.  Thinking we probably need to expand the pre-commit hook to check for newlines at the end of files.

core/build.rs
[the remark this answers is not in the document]
nick-mobilecoin: I would omit it for now, even Friday there was some internal talk about wanting to clean up mc-api, maybe those two efforts should be brought together.

core/src/subaddress.rs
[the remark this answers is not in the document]
nick-mobilecoin: I could see the by reference argument for HW. Do we have any measurements to know if this provides any performance benefit.

As for higher level types not needing to construct a higher level object.  I would hope if an object had need for a private view and private spend key, the object would have an account, but there are probably some use cases I'm not aware of.

core/types/src/account.rs
nick-mobilecoin: you probably have more insight to the naming here. 
I would suggest updating the comment to something like `/// MobileCoin spend subaddress object`

core/types/src/keys.rs
nick-mobilecoin: blarg....

It may be that I'm bringing baggage of naming gone bad.  My hunch is someone will cargo cult the `Subaddr` and not realize that it was OK because it was a marker trait.

I was able to:
- create a new module `markers.rs`
-  move the marker types into `markers.rs`
- add `mod markers` to `lib.rs`
- add `use crate::markers::{Subaddr, View, Spend, Root};` to `keys.rs`
- re-run the subaddress tests in `mc-core`

Not sure if there is another use case needed here where the marker traits are desired outside the `mc-core-types` crate

core/src/prelude.rs
nick-mobilecoin: at this point my vote would be to omit the prelude until it's deemed necessary.  
My understanding is the name and behavior is a convention so having this here at this time doesn't provide us any compile time or clippy checks ensuring that future changes are working toward a "good prelude"

account-keys/src/account_keys.rs
samdealy: Got it, thanks!

core/src/prelude.rs
samdealy: I'm not sure this is an answer to the question (I may be misunderstanding), but I'm asking when do you include `prelude` vs when do you omit `prelude` when bringing a type into scope?

core/src/subaddress.rs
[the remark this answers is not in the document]
ryankurte: on reflection construction is not that big of a deal / it really comes down to how we'd like to express the type relationships. i could just as easily build an `mc_core::Account` or `mc_core::ViewAccount`` to then call `.subaddress()` on... see [here](https://github.com/mobilecoinfoundation/mobilecoin/pull/2525/files#diff-d1e7a41f50ed7d6f5dd04bba9479fa5f52a0b7e2436f73cb5a561b210ea7111fR511) for an example of how i'm currently using the key-based traits, but nbd really to change this.

> I could see the by reference argument for HW. Do we have any measurements to know if this provides any performance benefit.

at the moment i'd estimate we have < 1k free of stack on the ledger (approximately the size of the of a 11 element ring, because we could fit 10 entries but not 11 before reworking the ring signatures for streaming computation). next ledger firmware update should relax that substantially, but i'm perhaps a little over sensitive to memory use this week 😂

core/src/subaddress.rs
```diff
@@ -0,0 +1,162 @@
...
+            SubaddrSpendPrivate::from(RistrettoPrivate::from(Hs + b)),
+        )
+    }
+}
+
+/// Generate subaddress public keys from root view private and spend public keys
+impl<'a> Subaddress for (&'a RootViewPrivate, &'a RootSpendPublic) {
+    type Output = (SubaddrViewPublic, SubaddrSpendPublic);
+
+    fn subaddress(&self, index: u64) -> Self::Output {
```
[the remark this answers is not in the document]
ryankurte: ahh, i had missed pulling in the `account-keys` changes related to this, `account-keys` now [calls](https://github.com/mobilecoinfoundation/mobilecoin/pull/2525/files#diff-d1e7a41f50ed7d6f5dd04bba9479fa5f52a0b7e2436f73cb5a561b210ea7111fR511) [back](https://github.com/mobilecoinfoundation/mobilecoin/pull/2525/files#diff-d1e7a41f50ed7d6f5dd04bba9479fa5f52a0b7e2436f73cb5a561b210ea7111fR623) to the `core` implementation (though there _is_ performance benefit to deriving both keys in one pass which `account-keys` does not take advantage of atm)

core/src/slip10/mod.rs
```diff
@@ -0,0 +1,245 @@
...
+#[cfg(feature = "bip39")]
+pub use bip39::{Language, Mnemonic};
+
+use crate::{
+    account::Account,
+    consts::{COINTYPE_MOBILECOIN, USAGE_BIP44},
+    keys::{RootSpendPrivate, RootViewPrivate},
+};
+
+/// [Hardened derivation](https://github.com/bitcoin/bips/blob/master/bip-0043.mediawiki#Security) flag for path components
```
nick-mobilecoin, line 30: :+1: Like the use of this define

core/types/src/keys.rs
```diff
@@ -0,0 +1,338 @@
...
+/// Root spend private key
+pub type RootSpendPrivate = Key<Root, Spend, RistrettoPrivate>;
+
+/// Root view public key
+pub type RootViewPublic = Key<Root, View, RistrettoPublic>;
+/// Root spend public key
+pub type RootSpendPublic = Key<Root, Spend, RistrettoPublic>;
+
+/// Markers for key types
+
```
nick-mobilecoin, line 40: :upside_down_face: Think this is left over from migrations
```suggestion
```

nick-mobilecoin (approved):
LGTM thanks!

core/types/src/keys.rs
```diff
@@ -0,0 +1,336 @@
...
+        &self.key == other
+    }
+}
+
+/// PartialEq for backwards compatibility with public key objects
+impl<ADDR, KIND> PartialEq<Key<ADDR, KIND, RistrettoPublic>> for RistrettoPublic {
+    fn eq(&self, other: &Key<ADDR, KIND, RistrettoPublic>) -> bool {
+        self == &other.key
+    }
+}
```
isis-mc, line 146: Potentially less important for public keys, but is `==` here using `CtEq` or just the stdlib `Eq`?

core/types/src/keys.rs
```diff
@@ -0,0 +1,336 @@
...
+        RistrettoPublic::from(&self.key) == RistrettoPublic::from(other)
+    }
+}
+
+/// PartialEq for backwards compatibility with private key objects
+impl<ADDR, KIND> PartialEq<Key<ADDR, KIND, RistrettoPrivate>> for RistrettoPrivate {
+    fn eq(&self, other: &Key<ADDR, KIND, RistrettoPrivate>) -> bool {
+        RistrettoPublic::from(self) == RistrettoPublic::from(&other.key)
+    }
+}
```
isis-mc, line 264: Same as above, `CtEq` would be important for private key comparisons.

isis-mc (approved):

core/types/src/keys.rs
isis-mc: It's `CtEq`.

core/types/src/keys.rs
isis-mc: It's using `CtEq`.

isis-mc (approved):

PR #2528, opened 2022-09-14
fix some messed up comments

garbageslam:
These appear to be wrong comments left in from the chain-id PR


eranrund (approved):

nick-mobilecoin (approved):
LGTM
though I did find some more instances. 
```shell

watcher/src/verification_reports_collector.rs
78:            // TODO: Supply a network-id to watcher?

fog/view/load-test/src/main.rs
143:    // TODO: Supply network-id to the load-test binary?
```

garbageslam:
thanks @nick-mobilecoin will fix

PR #2532, opened 2022-09-15
Update Receipt proto schema for MaskedAmountV2 changes

### LINKED ISSUE #2530: Follow up on Masked Amount changes for Receipt proto
In the `external.proto` we have the `Receipt` type which is used by some SDKs

```
// Given to the recipient of a transaction output by the sender so that the
// recipient may verify that the other party is indeed the sender.
//
// Often given to the recipient before the transaction is finalized so that
// the recipient may know to anticipate the arrival of a transaction output,
// as well as know who it's from, when to consider it as having surpassed
// the tombstone block, and the expected amount of the output.
message Receipt {
    // Public key of the TxOut.
    CompressedRistretto public_key = 1;

    // Confirmation number of the TxOut.
    TxOutConfirmationNumber confirmation = 2;

    // Tombstone block of the Tx that produced the TxOut.
    // Note: This value is self-reported by the sender and is unverifiable.
    uint64 tombstone_block = 3;

    // Amount of the TxOut.
    // Note: This value is self-reported by the sender and is unverifiable.
    MaskedAmount masked_amount = 4;
}
```

However, in the masked amount v2 change, we changed masked amount to generally be a one-of with versioning. In TxOut it looks like this now:

```
    // Masked Amount.
    // The versioning indicates which shared secret derivation we are using.
    // For v1, the TxOut shared secret is used directly.
    // For v2, an intermediate secret called amount shared secret is used.
    oneof masked_amount {
        MaskedAmount masked_amount_v1 = 1;
        MaskedAmount masked_amount_v2 = 6;
    };
```

We need to update `Receipt` also and help the callers to update, or they will have amount decryption errors.

garbageslam:
After masked amount v2, MaskedAmounts are always stored in a proto oneof, which their versioning information is derived from. It's generally going to be a bug if a MaskedAmount is not stored in a oneof this way, because the recipient will not be able to infer which version it is.

Brian ran into this when updating full-service. It likely will impact SDKs also.

Fixes #2530

### DIFF
--- a/api/proto/external.proto
+++ b/api/proto/external.proto
@@ -340,7 +340,10 @@ message Receipt {
 
     // Amount of the TxOut.
     // Note: This value is self-reported by the sender and is unverifiable.
-    MaskedAmount masked_amount = 4;
+    oneof masked_amount {
+        MaskedAmount masked_amount_v1 = 4;
+        MaskedAmount masked_amount_v2 = 5;
+    };
 }
 
 /// The signature over an IAS JSON reponse, created by Intel


eranrund (approved):

voloshyn:
How does backward compatibility work here? 
If an updated client generates a receipt and sends that to an old client, the old client will crash.
Unless we update all clients to handle the new format, then wait 90 days for the users to update their clients.

garbageslam:
Good question.

I think the short answer is, the MaskedAmountV2 is gated on block version 3.

Any client that supports block version 3 has to support the MaskedAmountV2 or it won't be able to use TxOut at all anymore.

There will not be any MaskedAmountV2 anywhere before block version 3.

And this proto format change is not breaking.

So I think
* we could start using this version of the proto whenever we like, knowing that before block version 3, V1 will always be selected
* in a client that supports block version 3, we have to be ready to start using the V2 field here, and we would use it when the V2 field is used in the TxOut that produced this MaskedAmount.

Brian ran into trouble trying to update because he has to be able to generate Receipts (which was easy because both versions convert to `external::MaskedAmount`) but then the conversion back was hard, because just having a `MaskedAmount`, you don't know which oneof it was. That was when I realized I forgot to make this a oneof also.

voloshyn (approved):

PR #2542, opened 2022-09-16
Add a missing conversion for Receipt_one_of_masked_amount

garbageslam:
In the previous commit we added a oneof to the Receipt proto. However we didn't add a conversion similar TxOut_one_of_masked_amount.

This is needed because in the proto codegen, each oneof corresponds to a different type tied to the struct it appears in.

### DIFF
--- a/api/src/convert/amount.rs
+++ b/api/src/convert/amount.rs
@@ -91,3 +91,18 @@ impl TryFrom<&external::TxOut_oneof_masked_amount> for MaskedAmount {
         }
     }
 }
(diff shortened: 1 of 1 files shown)


eranrund (approved):
Might be worth adding the conversion for the other direction (`From<&MaskedAmount>  for external::Receipt_oneof_masked_amount`)

BrianCorbin (approved):

BrianCorbin:
Thank you for this!

PR #2545, opened 2022-09-16
Adding serdr_str to several structs

BrianCorbin:
This enables (de)serialization of u64s to/from strings, to prevent precision loss during transit


BrianCorbin:
Ya know what, never mind, this is not the right way to do this.

PR #2563, opened 2022-09-16
Adding several missing proto and conversion

BrianCorbin:
Todo:

Conversions
- [x] UnsignedTx
- [x] SigningData
- [x] InputRing
- [x] OutputSecret
- [x] InputSecret
- [x] Amount
- [x] Scalar

Docs
- [x] UnsignedTx
- [x] SigningData
- [x] InputRing
- [x] OutputSecret
- [x] InputSecret
- [x] Amount
- [x] Scalar

Tests
- [x] UnsignedTx
- [x] SigningData
- [x] InputRing
- [x] OutputSecret
- [x] InputSecret
- [x] Amount
- [x] Scalar


api/proto/external.proto
```diff
@@ -473,3 +473,31 @@ message SignedContingentInput {
...
+    // The value of the amount
+    fixed64 value = 1;
+
+    // The token_id of the amount
+    fixed64 token_id = 2;
+}
+
+message Scalar {
+    bytes data = 1;
+}
```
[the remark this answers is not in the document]
BrianCorbin: CurveScalar does

api/src/convert/error.rs
```diff
@@ -10,17 +10,18 @@ use std::{
...
     IndexOutOfBounds,
-    ObjectMissing,
-    InvalidSignature,
     InvalidContents,
+    InvalidSignature,
+    Key(KeyError),
+    KeyCastError,
+    MissingField(String),
+    NarrowingCastError,
+    ObjectMissing,
```
garbageslam, line 24: thanks for this

api/proto/external.proto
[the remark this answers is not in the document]
garbageslam: Yeah see line 143 above

garbageslam:
@BrianCorbin let's talk about how to serialize `block_version`, I think in the SCI stuff I just made it a u32:

```
// A pre-signed transaction input with associated rules, as described in MCIP #31
message SignedContingentInput {
    // The block version rules used when making this signature
    uint32 block_version = 1;
```

I'm not sure it's needed to add a Message for this since it's just a number, I would advocate just using `uint32` in the `UnsignedTx` also.

garbageslam (approved):
Code looks good to me, doc strings for new protos would be nice

Thanks!

eranrund (commented):
LGTM but I think the tests should be included as part of this, since it's possible to miss a field or have a field be added in the future and not realize the conversion is actually broken.

api/src/convert/curve_scalar.rs
```diff
@@ -32,3 +33,23 @@ impl TryFrom<&external::CurveScalar> for CurveScalar {
         CurveScalar::try_from(bytes).map_err(|_| ConversionError::ArrayCastError)
     }
 }
+
+impl From<&Scalar> for external::CurveScalar {
```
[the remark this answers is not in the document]
garbageslam: It's used in `OutputSecret` I think.

We can change that if you want, I don't feel strongly.
I don't think the `CurveScalar` type actually makes a ton of sense right now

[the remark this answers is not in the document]
BrianCorbin: Scalar is used in a few places

https://github.com/mobilecoinfoundation/mobilecoin/blob/fb3e8d15c5f9ab6ea7133540211126ae3ac67d9f/crypto/ring-signature/signer/src/traits.rs#L33-L40

https://github.com/mobilecoinfoundation/mobilecoin/blob/fb3e8d15c5f9ab6ea7133540211126ae3ac67d9f/transaction/core/src/ring_ct/rct_bulletproofs.rs#L75-L80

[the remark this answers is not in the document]
BrianCorbin: Should those have been CurveScalar?

api/proto/external.proto
```diff
@@ -473,3 +473,137 @@ message SignedContingentInput {
...
+    /// Before mixed transactions feature, this field is empty, and the token ids of
+    /// all pseudo-outputs are inferred from the tx.prefix.fee_token_id.
+    repeated fixed64 pseudo_output_token_ids = 6;
+
+    /// The token ids of each output. There must be one of these for each output of the Tx.
+    /// (tx.prefix.outputs).
+    /// Before mixed transactions feature, this field is empty, and the token ids of
+    /// all outputs are inferred from the tx.prefix.fee_token_id.
+    repeated fixed64 output_token_ids = 7;
+}
```
remoun, line 609: Nit: newline
```suggestion
}

```

api/src/convert/amount.rs
```diff
@@ -119,3 +121,21 @@ impl TryFrom<&external::Receipt_oneof_masked_amount> for MaskedAmount {
...
+        amount
+    }
+}
+
+impl From<&external::Amount> for Amount {
+    fn from(source: &external::Amount) -> Self {
+        Amount {
+            value: source.get_value(),
+            token_id: TokenId::from(source.get_token_id()),
+        }
```
remoun, line 139: Optional/Equivalent: Use `Amount::new` and `into()`
```suggestion
        Amount::new(source.get_value(), source.get_token_id().into())
```

api/src/convert/input_ring.rs
```diff
@@ -0,0 +1,101 @@
+use std::convert::{TryFrom, TryInto};
+
```
remoun, line 2: `std::convert::{TryFrom, TryInto}` are part of the 2021 edition prelude, so they can be omitted 
```suggestion
```

api/src/convert/output_secret.rs
```diff
@@ -0,0 +1,27 @@
+use std::convert::{TryFrom, TryInto};
+
```
remoun, line 2: Nit: Unnecessary import
```suggestion
```

api/src/convert/reduced_tx_out.rs
```diff
@@ -0,0 +1,27 @@
+use std::convert::TryFrom;
+
```
remoun, line 2: Nit: Unnecessary import
```suggestion
```

api/src/convert/curve_scalar.rs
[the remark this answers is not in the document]
eranrund: Ah, I see. Okay. Scalar it is.

api/src/convert/input_ring.rs
```diff
@@ -0,0 +1,102 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
+//! Convert to/from mc_transaction_core::ring_ct::InputRing.
+
+use crate::{external, ConversionError};
+use mc_crypto_ring_signature_signer::SignableInputRing;
+use mc_transaction_core::ring_ct::{InputRing, PresignedInputRing};
+
+impl From<&InputRing> for external::InputRing {
```
eranrund, line 9: I'm seeing `InputRing` checked for tests but I don't see the tests...

api/src/convert/input_secret.rs
```diff
@@ -0,0 +1,103 @@
...
+
+    // Test converting between external::InputSecret and
+    // mc_crypto_ring_signature_signer::InputSecret
+    #[test]
+    fn test_input_secret_conversion() {
+        let mut rng: StdRng = SeedableRng::from_seed([123u8; 32]);
+
+        let input_secret = InputSecret {
+            onetime_key_derive_data: OneTimeKeyDeriveData::SubaddressIndex(10),
+            amount: Amount::new(10000, TokenId::from(0)),
```
eranrund, line 81: Suggest using non-zero token id since zero is the default value for protobuf fields

api/src/convert/input_secret.rs
```diff
@@ -0,0 +1,103 @@
...
+            blinding: Scalar::random(&mut rng),
+        };
+
+        let input_secret_external: external::InputSecret = (&input_secret).into();
+        let input_secret_decrypted: InputSecret = (&input_secret_external).try_into().unwrap();
```
eranrund, line 86: decrypted seems wrong, there's no encryption taking place here

api/src/convert/output_secret.rs
```diff
@@ -0,0 +1,55 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::external;
+    use curve25519_dalek::scalar::Scalar;
+    use mc_transaction_core::{ring_ct::OutputSecret, Amount, TokenId};
+    use rand::{rngs::StdRng, SeedableRng};
+
+    // Test converting between external::ReducedTxOut and
```
eranrund, line 38: ReducedTxOut copypasta

api/src/convert/output_secret.rs
```diff
@@ -0,0 +1,55 @@
...
+    // mc_transaction_core::ring_signature::ReducedTxOut
+    #[test]
+    fn test_output_secret_conversion() {
+        let mut rng: StdRng = SeedableRng::from_seed([123u8; 32]);
+
+        let output_secret = OutputSecret {
+            amount: Amount::new(10000, TokenId::from(0)),
```
eranrund, line 45: Suggest using non-zero token id due to 0 being protobuf default

api/src/convert/signing_data.rs
```diff
@@ -0,0 +1,123 @@
...
+    use rand::{rngs::StdRng, SeedableRng};
+
+    // Test converting between external::SigningData and
+    // mc_transaction_core::ring_ct::SigningData
+    #[test]
+    fn test_signing_data_conversion() {
+        // Generate an UnsignedTx to test with.
+        let mut rng: StdRng = SeedableRng::from_seed([1u8; 32]);
+
+        for block_version in BlockVersion::iterator() {
```
eranrund, line 77: Nice :)

api/src/convert/signing_data.rs
```diff
@@ -0,0 +1,123 @@
...
+                )
+                .unwrap();
+
+            let unsigned_tx = transaction_builder
+                .build_unsigned::<StdRng, DefaultTxOutputsOrdering>()
+                .unwrap();
+
+            let signing_data = unsigned_tx.get_signing_data(&mut rng).unwrap();
+
+            // Converting mc_transaction_core::ring_ct::SigningData -> external::UnsignedTx
```
eranrund, line 112: UnsignedTx copypasta

crypto/keys/src/ristretto.rs
```diff
@@ -52,6 +52,14 @@ use zeroize::Zeroize;
 #[derive(Clone, Copy, Default, Zeroize)]
 pub struct RistrettoPrivate(pub(crate) Scalar);
 
+impl Eq for RistrettoPrivate {}
```
eranrund, line 55: I believe this was not implemented on purpose. @jcape probably knows why. Maybe we can get away by using the constant time eq here, or maybe there is a reason to just not allow this (which would mean manually `Eq` implementations for structs that need that and contain a `RistrettoPrivate`)

eranrund (changes requested):
Thanks for working on the tests! Mostly looks good, but I think one proto has tests missing and the `RistrettoPrivate` eq thing requires discussion with @jcape

api/src/convert/input_ring.rs
BrianCorbin: Oh oops forgot to commit those changes

crypto/keys/src/ristretto.rs
BrianCorbin: After discussion with @jcape, we concluded that an implementation of Eq that enforces constant time comparison is acceptable.

eranrund (approved):

PR #2585, opened 2022-09-21
Add account key abstractions

cbeck88:
In order to make progress on the hardware wallets project, we need to start adding abstractions for more of the functionality of account keys beyond just signing rings.

One of the main asks is also, the memo signing part. In this commit we also threw in the key images.

We couldn't throw in Public Address computations yet because it causes circular dependencies, unless we do more refactoring. We decided to hold off before this PR gets too large.

Without public addresses, we can't easily compute the short address hash of the source address, which is needed for the memo computations. So we can't make the transaction builder part generic over this new memo trait just yet either.

Nevertheless, committing to this API can help enable further progress on the hardware wallet side.

One thing I'm anticipating is, if the memo builder contains a reference to "handle to hardware wallet", and the ring signer is also "handle to hardware wallet", then we may want to have an `Rc`-wrapper in there to allow it to be in "two places at once". This shouldn't pose a problem on the desktop/transaction builder side.

One thing I'm unsure about is if I'm making a mistake making traits like `RingSigner` take `&self` instead of `&mut self`. We could probably easily correct course if we have to though. It seems likely that whatever handle to the hardware wallet we obtain, it may not be perfectly thread-safe, so maybe `&mut self` is better. OTOH I'd kind of like to avoid complicating the transaction builder, and using either `Mutex` or `Rc` or something to resolve the mutability issue.

### DIFF
--- a/abstract-account-keys/Cargo.toml
+++ b/abstract-account-keys/Cargo.toml
@@ -1,0 +1,19 @@
+[package]
+name = "mc-abstract-account-keys"
+version = "2.0.0"
+authors = ["MobileCoin"]
+edition = "2021"
+readme = "README.md"
+
+[dependencies]
+# External dependencies
+displaydoc = { version = "0.2", default-features = false }
+rand_core = { version = "0.6", default-features = false }
+serde = { version = "1", default-features = false }
+subtle = { version = "2", default-features = false }
+zeroize = { version = "1", default-features = false }
+
+# MobileCoin dependencies
+mc-crypto-keys = { path = "../crypto/keys", default-features = false }
+mc-crypto-ring-signature = { path = "../crypto/ring-signature", default-features = false }
+mc-transaction-types = { path = "../transaction/types", default-features = false }
--- a/abstract-account-keys/src/key_image_computer.rs
+++ b/abstract-account-keys/src/key_image_computer.rs
@@ -1,0 +1,12 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
+use crate::Error;
+use mc_crypto_keys::{CompressedRistrettoPublic};
+use mc_crypto_ring_signature::KeyImage;
+
+/// An object that can compute the key image of a TxOut, given a subset of its data and the subaddress index.
+/// This generally requires the subaddress spend private key.
+pub trait KeyImageComputer {
+    /// Compute the key image of a TxOut given its public key, and the subaddress index on which this account owns it.
+    fn compute_key_image(&self, tx_out_public_key: &CompressedRistrettoPublic, subaddress_index: u64) -> Result<KeyImage, Error>;
+}
--- a/abstract-account-keys/src/key_image_computer.rs
+++ b/abstract-account-keys/src/key_image_computer.rs
@@ -1,12 +1,18 @@
 // Copyright (c) 2018-2022 The MobileCoin Foundation
 
 use crate::Error;
-use mc_crypto_keys::{CompressedRistrettoPublic};
+use mc_crypto_keys::CompressedRistrettoPublic;
 use mc_crypto_ring_signature::KeyImage;
 
-/// An object that can compute the key image of a TxOut, given a subset of its data and the subaddress index.
-/// This generally requires the subaddress spend private key.
+/// An object that can compute the key image of a TxOut, given a subset of its
+/// data and the subaddress index. This generally requires the subaddress spend
+/// private key.
 pub trait KeyImageComputer {
-    /// Compute the key image of a TxOut given its public key, and the subaddress index on which this account owns it.
-    fn compute_key_image(&self, tx_out_public_key: &CompressedRistrettoPublic, subaddress_index: u64) -> Result<KeyImage, Error>;
+    /// Compute the key image of a TxOut given its public key, and the
+    /// subaddress index on which this account owns it.
+    fn compute_key_image(
+        &self,
+        tx_out_public_key: &CompressedRistrettoPublic,
+        subaddress_index: u64,
+    ) -> Result<KeyImage, Error>;
 }
--- a/abstract-account-keys/src/lib.rs
+++ b/abstract-account-keys/src/lib.rs
@@ -1,8 +1,9 @@
 #![no_std]
 #![deny(missing_docs)]
 
-//! This crate provides traits that either AccountKeys objects or Hardware Wallets
-//! objects can implement to support signing a Tx, checking balance, etc.
+//! This crate provides traits that either AccountKeys objects or Hardware
+//! Wallets objects can implement to support signing a Tx, checking balance,
+//! etc.
 
 // FIXME
 extern crate alloc;
@@ -13,9 +14,9 @@ mod memo_hmac_signer;
 mod ring_signer;
 
 pub use error::Error;
-pub use key_image_computer::{KeyImageComputer};
-pub use memo_hmac_signer::{MemoHmacSigner};
-pub use ring_signer::{RingSigner, SignableInputRing, OneTimeKeyDeriveData, InputSecret};
+pub use key_image_computer::KeyImageComputer;
+pub use memo_hmac_signer::MemoHmacSigner;
+pub use ring_signer::{InputSecret, OneTimeKeyDeriveData, RingSigner, SignableInputRing};
 
 /// A trait which captures all the functionality of a set of account keys needed
 /// to check ones balance and sign transactions.
@@ -23,8 +24,8 @@ pub use ring_signer::{RingSigner, SignableInputRing, OneTimeKeyDeriveData, Input
 /// This is meant to be generic over, a locally stored set of account keys, and
 /// a connection to a hardware wallet.
 ///
-/// Note: We still should add more functionality around computation of public address,
-/// computation of view private key, subaddress view private key. At the moment,
-/// this is difficult because public address is part of mc-account-keys and this
-/// creates a circular dependency.
+/// Note: We still should add more functionality around computation of public
+/// address, computation of view private key, subaddress view private key. At
+/// the moment, this is difficult because public address is part of
+/// mc-account-keys and this creates a circular dependency.
 pub trait AbstractAccountKeys: KeyImageComputer + MemoHmacSigner + RingSigner {}
--- a/abstract-account-keys/src/lib.rs
+++ b/abstract-account-keys/src/lib.rs
@@ -1,0 +1,30 @@
+#![no_std]
+#![deny(missing_docs)]
+
+//! This crate provides traits that either AccountKeys objects or Hardware Wallets
+//! objects can implement to support signing a Tx, checking balance, etc.
+
+// FIXME
+extern crate alloc;
+
+mod error;
+mod key_image_computer;
+mod memo_hmac_signer;
+mod ring_signer;
+
+pub use error::Error;
+pub use key_image_computer::{KeyImageComputer};
+pub use memo_hmac_signer::{MemoHmacSigner};
+pub use ring_signer::{RingSigner, SignableInputRing, OneTimeKeyDeriveData, InputSecret};
+
+/// A trait which captures all the functionality of a set of account keys needed
+/// to check ones balance and sign transactions.
+///
+/// This is meant to be generic over, a locally stored set of account keys, and
+/// a connection to a hardware wallet.
+///
+/// Note: We still should add more functionality around computation of public address,
+/// computation of view private key, subaddress view private key. At the moment,
+/// this is difficult because public address is part of mc-account-keys and this
+/// creates a circular dependency.
+pub trait AbstractAccountKeys: KeyImageComputer + MemoHmacSigner + RingSigner {}
--- a/abstract-account-keys/src/memo_hmac_signer.rs
+++ b/abstract-account-keys/src/memo_hmac_signer.rs
@@ -1,0 +1,12 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
+use crate::Error;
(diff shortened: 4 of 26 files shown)


account-keys/src/impls.rs
```diff
@@ -0,0 +1,89 @@
+//! Implement mc-abstract-account-keys traits for concrete AccountKeys
+
+use crate::{AccountKeys};
+use curve25519_dalek::scalar::Scalar;
+use mc_abstract_account_keys::{RingSigner, MemoHmacSigner, KeyImageComputer, Error};
+use mc_crypto_ring_signature::{KeyImage,SignableInputRing, onetime_keys::recover_onetime_private_key};
+
+// Note: We should delete `LocalRingSigner` and just use `AccountKeys` where we were using that.
+impl RingSinger for AccountKeys {
```
nick-mobilecoin, line 9: :upside_down_face: 
I'm curious what kind of Singer a RingSinger is, perhaps it's like [dio](https://youtu.be/PrBUjXaRSUQ)
```suggestion
impl RingSigner for AccountKey {
```

account-keys/src/impls.rs
```diff
@@ -0,0 +1,89 @@
+//! Implement mc-abstract-account-keys traits for concrete AccountKeys
+
+use crate::{AccountKeys};
```
nick-mobilecoin, line 3: :upside_down_face: Since this module isn't actually included in the crate not sure it's full intent, but should this be `AccountKey` or is this anticipating some new type?
```suggestion
use crate::{AccountKey};
```

account-keys/src/impls.rs
```diff
@@ -0,0 +1,89 @@
...
+        )?)
+    }
+}
+
+impl MemoHmacSigner for AccountKeys {
+    fn compute_memo_hmac_sig(&self, receiving_subaddress_view_public: &CompressedRistrettoPublic, tx_out_public_key: &CompressedRistrettoPublic, memo_type: &[u8; 2], memo_data_sans_hmac: &[u8; 48]) -> Result<[u8; 16], Error> {
+
+        let shared_secret = self
+            .subaddress_spend_private_key(DEFAULT_SUBADDRESS_INDEX)
+            .key_exchange(receiving_subaddress_view_public_key);
```
nick-mobilecoin, line 65: :upside_down_face: 
```suggestion
            .key_exchange(receiving_subaddress_view_public);
```

account-keys/src/impls.rs
```diff
@@ -0,0 +1,89 @@
...
+
+        let hmac_value = compute_category1_hmac(
+            shared_secret.as_ref(),
+            tx_out_public_key,
+            Self::MEMO_TYPE_BYTES,
```
nick-mobilecoin, line 70: :wrench: why pass in the `memo_type` if this is going to override it?
My guess is there shouldn't be a need to pass it in, but not sure on the future plans
```suggestion
            memo_type,
```

account-keys/src/impls.rs
```diff
@@ -0,0 +1,89 @@
...
+}
+
+impl MemoHmacSigner for AccountKeys {
+    fn compute_memo_hmac_sig(&self, receiving_subaddress_view_public: &CompressedRistrettoPublic, tx_out_public_key: &CompressedRistrettoPublic, memo_type: &[u8; 2], memo_data_sans_hmac: &[u8; 48]) -> Result<[u8; 16], Error> {
+
+        let shared_secret = self
+            .subaddress_spend_private_key(DEFAULT_SUBADDRESS_INDEX)
+            .key_exchange(receiving_subaddress_view_public_key);
+
+        let hmac_value = compute_category1_hmac(
```
nick-mobilecoin, line 67: :wrench: from https://github.com/mobilecoinfoundation/mobilecoin/blob/master/CONTRIBUTING.md, https://doc.rust-lang.org/1.0.0/style/style/imports.html#prefer-fully-importing-types/traits-while-module-qualifying-functions.
```suggestion
        let hmac_value = mc_crypto_memo_mac::compute_category1_hmac(
```

account-keys/src/impls.rs
```diff
@@ -0,0 +1,89 @@
...
+            }
+        };
+
+        // Check if this is the correct one-time private key
+        if RistrettoPublic::from(&onetime_private_key) != target_key {
+            return Err(Error::TrueInputNotOwned);
+        }
+
+        // Note: Some implementations might be able to cache this generator
+        let generator = generators(*ring.input_secret.amount.token_id);
```
nick-mobilecoin, line 43: :wrench: it was like this before so might be better to leave alone for this PR
from https://github.com/mobilecoinfoundation/mobilecoin/blob/master/CONTRIBUTING.md, https://doc.rust-lang.org/1.0.0/style/style/imports.html#prefer-fully-importing-types/traits-while-module-qualifying-functions.
```suggestion
        let generator = mc-crypto_ring_signature::generators(*ring.input_secret.amount.token_id);
```

crypto/memo-mac/README.md
```diff
@@ -0,0 +1,7 @@
+mc-crypto-memo-mac
+==================
+
+This crate contains details for computing the RTH memo hmac value. This is in it's
```
nick-mobilecoin, line 4: :upside_down_face: "it's" == "it is"
```suggestion
This crate contains details for computing the RTH memo hmac value. This is in its
```

transaction/std/src/memo/authenticated_common.rs
```diff
@@ -64,7 +31,7 @@ pub fn validate_authenticated_sender(
         shared_secret.as_ref(),
         tx_out_public_key,
         memo_type_bytes,
-        memo_data,
+        memo_data[..48].try_into().unwrap(),
     );
     let found_hmac: [u8; 16] = memo_data[(64 - 16)..].try_into().unwrap();
```
nick-mobilecoin, line 36: :wrench: mixing `48` and `64-16` 
```suggestion
        memo_data[..48].try_into().unwrap(),
    );
    let found_hmac: [u8; 16] = memo_data[48..].try_into().unwrap();
```

nick-mobilecoin (changes requested):

nick-mobilecoin:
Sorry forgot to write info on the review.
The main blocker is the `impls.rs` module. it doesn't compile and isn't actually included in any crate.

garbageslam:
Thanks!

abstract-account-keys/src/memo_hmac_signer.rs
```diff
@@ -0,0 +1,12 @@
...
+
+use crate::Error;
+use mc_crypto_keys::{CompressedRistrettoPublic};
+
+/// An object that can compute the appropriate hmac of a MCIP #4-style Authenticated Sender memo,
+/// for spend keys which have been abstracted.
+/// This generally requires the subaddress spend private key.
+pub trait MemoHmacSigner {
+    /// Compute the hmac which "signs" a given MCIP #4 Authenticated sender memo
+    fn compute_memo_hmac_sig(&self, receiving_subaddress_view_public: &CompressedRistrettoPublic, tx_out_public_key: &CompressedRistrettoPublic, memo_type: &[u8; 2], memo_data_sans_hmac: &[u8; 48]) -> Result<[u8; 16], Error>;
```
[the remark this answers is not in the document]
garbageslam: Changing the memo size is really hard after we go to block version 2 in mainnet, i dont expect we're actually going to do that. But yea, named constants are good

abstract-account-keys/src/key_image_computer.rs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
+use crate::Error;
+use mc_crypto_keys::{CompressedRistrettoPublic};
+use mc_crypto_ring_signature::KeyImage;
+
+/// An object that can compute the key image of a TxOut, given a subset of its data and the subaddress index.
+/// This generally requires the subaddress spend private key.
+pub trait KeyImageComputer {
+    /// Compute the key image of a TxOut given its public key, and the subaddress index on which this account owns it.
```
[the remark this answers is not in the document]
garbageslam: English is hard at 2 am lol

garbageslam:
I'm going to have to restructure this a bit, I've got a circular depenency once the `impl` part is involved:

```
error: cyclic package dependency: package `mc-account-keys v2.0.0 (/tmp/mobilenode/account-keys)` depends on itself. Cycle:
package `mc-account-keys v2.0.0 (/tmp/mobilenode/account-keys)`
    ... which satisfies path dependency `mc-account-keys` (locked to 2.0.0) of package `mc-crypto-ring-signature v2.0.0 (/tmp/mobilenode/crypto/ring-signature)`
    ... which satisfies path dependency `mc-crypto-ring-signature` (locked to 2.0.0) of package `mc-abstract-account-keys v2.0.0 (/tmp/mobilenode/abstract-account-keys)`
    ... which satisfies path dependency `mc-abstract-account-keys` (locked to 2.0.0) of package `mc-account-keys v2.0.0 (/tmp/mobilenode/account-keys)`
    ... which satisfies path dependency `mc-account-keys` (locked to 2.0.0) of package `mc-api v2.0.0 (/tmp/mobilenode/api)`
    ... which satisfies path dependency `mc-api` (locked to 2.0.0) of package `mc-fog-report-api v2.0.0 (/tmp/mobilenode/fog/report/api)`
    ... which satisfies path dependency `mc-fog-report-api` (locked to 2.0.0) of package `go-grpc-gateway-testing v2.0.0 (/tmp/mobilenode/go-grpc-gateway/testing)`
```

abstract-account-keys/src/memo_hmac_signer.rs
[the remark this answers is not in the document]
garbageslam: >     Is there a reason not to expose Hmac<Sha512> as the return value? Perhaps a generic 512bit Hmac. The main reason is to leverage rusts type system and checks, avoiding u8 slices being passed around.

Yeah, this won't have the correct size. We could make a special named type, but I see little advantage at this moment.

account-keys/src/impls.rs
garbageslam: Eran typically comments that we should use imports at the top of the file, rather than insert module names in the middle of the file

garbageslam:
Unfortunately, I can't prioritize fixing this right now, we need to focus on things targetted for v3 release. After we branch for v3 we can figure out what should happen here

garbageslam:
Gonna set this back to draft status.

abstract-account-keys/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+displaydoc = { version = "0.2", default-features = false }
+rand_core = { version = "0.6", default-features = false }
+serde = { version = "1", default-features = false }
+subtle = { version = "2", default-features = false }
+zeroize = { version = "1", default-features = false }
+
+# MobileCoin dependencies
+mc-crypto-keys = { path = "../crypto/keys", default-features = false }
+mc-crypto-ring-signature = { path = "../crypto/ring-signature", default-features = false }
+mc-transaction-types = { path = "../transaction/types", default-features = false }
```
remoun, line 19: Optional/Equivalent: Keep the MC deps first
```suggestion
# MobileCoin dependencies
mc-crypto-keys = { path = "../crypto/keys", default-features = false }
mc-crypto-ring-signature = { path = "../crypto/ring-signature", default-features = false }
mc-transaction-types = { path = "../transaction/types", default-features = false }

# External dependencies
displaydoc = { version = "0.2", default-features = false }
rand_core = { version = "0.6", default-features = false }
serde = { version = "1", default-features = false }
subtle = { version = "2", default-features = false }
zeroize = { version = "1", default-features = false }
```

abstract-account-keys/src/lib.rs
```diff
@@ -0,0 +1,31 @@
+#![no_std]
+#![deny(missing_docs)]
+
+//! This crate provides traits that either AccountKeys objects or Hardware
+//! Wallets objects can implement to support signing a Tx, checking balance,
+//! etc.
+
+// FIXME
```
remoun, line 8: What's this FIXME for?

crypto/memo-mac/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+[features]
+test-only = []
+
+[dependencies]
+# External dependencies
+hmac = "0.12"
+sha2 = { version = "0.10", default-features = false }
+
+# MobileCoin dependencies
+mc-crypto-keys = { path = "../../crypto/keys", default-features = false }
```
remoun, line 16: Ditto (MC first)
```suggestion
# MobileCoin dependencies
mc-crypto-keys = { path = "../../crypto/keys", default-features = false }

# External dependencies
hmac = "0.12"
sha2 = { version = "0.10", default-features = false }
```

abstract-account-keys/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "mc-abstract-account-keys"
```
remoun, line 2: Optional/Nit: `abstract` seems out of place in Rust. How about `mc-account-key-traits`?

abstract-account-keys/src/lib.rs
remoun:
```suggestion
// FIXME: stop depending on alloc. This crate needs to be core-only to work on hardware wallets
```

ryankurte: i think the abstract parts of this could end up in `mc_core` now?

jcape:
I'm going to close this since it looks like the relevant bits are in other (fresher) PRs. Feel free to re-open if that's not the case.

cbeck88:
Thank you

PR #2591, opened 2022-09-22
Add "fill_fractional_output_at"

garbageslam:
This adds a helper function "fill_fractional_output_at" which can help someone using an SCI to figure out exactly how much to fill it to get the desired amount of some output.

We probably need to create more helper functions as well for dealing with collections of SCIs, but in the interest of trying to keep this PR small, we're just adding one function with tests here.


garbageslam:
I think we should close this, I think this API isn't actually going to be very useful.
Need to start again from scratch

PR #2595, opened 2022-09-22
Create db cleaner executable

samdealy:
### Motivation

This PR creates the executable that cleans up the egress keys. Right now, it only implements the dry run functionality, which prints out the keys to the screen. 

### Future Work
* The next PR will handle the non-dry run scenario, which will make changes to the ingest_invocation table.

### DIFF
--- a/fog/sql_recovery_db/cleanup/src/config.rs
+++ b/fog/sql_recovery_db/cleanup/src/config.rs
@@ -1,0 +1,20 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
+//! Specifies configuration for the Fog SQL Recovery DB cleanup binary.
+
+use clap::Parser;
+use serde::Serialize;
+
+/// Configuration parameters for the Fog SQL recovery DB cleanup task.
+#[derive(Clone, Parser, Serialize)]
+#[clap(version)]
+pub struct SqlRecoveryDbCleanupConfig {
+    /// If set to true, performs cleanup task for unused egress keys.
+    #[clap(long)]
+    pub egress_keys: bool,
+
+    /// If set to true, prints out any DB entries that would be cleared by the
+    /// command and doesn't execute the deletion.
+    #[clap(long)]
+    pub dry_run: bool,
+}
--- a/fog/sql_recovery_db/cleanup/src/main.rs
+++ b/fog/sql_recovery_db/cleanup/src/main.rs
@@ -1,0 +1,30 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
+//! Runs a cleanup utility for the Fog Sql Recovery DB.
+
+use crate::{config::SqlRecoveryDbCleanupConfig, db_cleaner::DbCleaner};
+use chrono::Duration;
+use clap::Parser;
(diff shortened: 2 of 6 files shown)


fog/sql_recovery_db/cleanup/src/config.rs
```diff
@@ -0,0 +1,20 @@
...
+use clap::Parser;
+use serde::Serialize;
+
+/// Configuration parameters for the Fog SQL recovery DB cleanup task.
+#[derive(Clone, Parser, Serialize)]
+#[clap(version)]
+pub struct SqlRecoveryDbCleanupConfig {
+    /// If set to true, performs cleanup task for unused egress keys.
+    #[clap(long)]
+    pub egress_keys: bool,
```
garbageslam, line 14: maybe this should be the other way, so, cleanup of these should happen unless you opt out explicitly? I think usually you would want to do this

garbageslam (approved):

fog/sql_recovery_db/cleanup/src/config.rs
samdealy: I would normally agree, but James mentioned that this tool might be used for other types of DB cleanups. In that scenario, there might be multiple types of cleanups, and I think it might be better to have the Ops engineer specify exactly what type of cleanup they want. That said, happy to change it this perspective isn't convincing.

fog/sql_recovery_db/cleanup/src/config.rs
```diff
@@ -0,0 +1,20 @@
...
+
+use clap::Parser;
+use serde::Serialize;
+
+/// Configuration parameters for the Fog SQL recovery DB cleanup task.
+#[derive(Clone, Parser, Serialize)]
+#[clap(version)]
+pub struct SqlRecoveryDbCleanupConfig {
+    /// If set to true, performs cleanup task for unused egress keys.
+    #[clap(long)]
```
nick-mobilecoin, line 13: :facepalm: I saw `long` and thought this was storing a long int.... not that it was a long argument `--egress-keys`

fog/sql_recovery_db/cleanup/src/main.rs
```diff
@@ -0,0 +1,30 @@
...
+use chrono::Duration;
+use clap::Parser;
+use mc_common::logger::create_app_logger;
+use mc_fog_sql_recovery_db::SqlRecoveryDb;
+use std::env;
+
+mod config;
+mod db_cleaner;
+
+static EXPIRATION_DAYS: i64 = 2;
```
nick-mobilecoin, line 15: :thought_balloon: I think this could be a future PR if desired, should the expiration time and database_url be args as well.

nick-mobilecoin (approved):

fog/sql_recovery_db/cleanup/src/config.rs
samdealy: yeah, this is standard clap usage.

fog/sql_recovery_db/cleanup/src/main.rs
samdealy: Sounds good.

PR #2600, opened 2022-09-22
the rand folks accepted my patch, so we can use upstream trait now

garbageslam:


crypto/ring-signature/src/ring_signature/mlsag.rs
```diff
@@ -20,17 +20,6 @@ use crate::{
...
-/// A trait which implies RNGCore and CryptoRng
-///
-/// This is needed because &mut (dyn RngCore + CryptoRng) is not valid in rust
-/// right now and we need this to make the ring signer trait work as desired
-//
-// Note: This trait can go away if it is upstreamed to rand-core:
-// https://github.com/rust-random/rand/pull/1230
-pub trait CryptoRngCore: RngCore + CryptoRng {}
-
-impl<T> CryptoRngCore for T where T: RngCore + CryptoRng {}
```
garbageslam, line 32: removing this is the point of this PR

nick-mobilecoin (approved):

BrianCorbin (approved):

PR #2606, opened 2022-09-23
Update fog readme for new mob prompt behavior

nick-mobilecoin:
Also improved test utility postgres connection error reporting

### Future Work

[Soundtrack of this PR]()

### DIFF
--- a/fog/README.md
+++ b/fog/README.md
@@ -58,7 +58,10 @@ To run the unittest tests locally, you need to start a local postgres instance, 
 ./mob prompt
 
 # Start postgres
-service postgresql start
+supo service postgresql start
+
+# create a postgres user
+sudo -u postgres createuser --superuser $USER
 
 # Run the tests
 cargo test
--- a/fog/README.md
+++ b/fog/README.md
@@ -58,7 +58,7 @@ To run the unittest tests locally, you need to start a local postgres instance, 
 ./mob prompt
 
 # Start postgres
-supo service postgresql start
+sudo service postgresql start
 
 # create a postgres user
 sudo -u postgres createuser --superuser $USER
--- a/fog/sql_recovery_db/src/test_utils.rs
+++ b/fog/sql_recovery_db/src/test_utils.rs
@@ -100,7 +100,7 @@ impl SqlRecoveryDbTestContext {
                 .take(TOTAL_RETRY_COUNT),
             || PgConnection::establish(url),
         )
-        .unwrap_or_else(|_| panic!("Cannot connect to PG database '{}'", url))
+        .unwrap_or_else(|err| panic!("Cannot connect to PG database '{}: {}'", url, err))
     }
 }
 


nick-mobilecoin:
* master
  * **PR #2606** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2606" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2606?utm_source=stack-comment).

fog/README.md
```diff
@@ -58,7 +58,10 @@ To run the unittest tests locally, you need to start a local postgres instance,
 ./mob prompt
 
 # Start postgres
-service postgresql start
+supo service postgresql start
```
wjuan-mob, line 61: Should this be sudo?

nick-mobilecoin: doh, thanks!

fog/README.md
```diff
@@ -58,7 +58,10 @@ To run the unittest tests locally, you need to start a local postgres instance,
 ./mob prompt
 
 # Start postgres
-service postgresql start
+sudo service postgresql start
+
+# create a postgres user
+sudo -u postgres createuser --superuser $USER
```
garbageslam, line 64: why is this not needed for circle ci i wonder?

nick-mobilecoin: My understanding is because `mob prompt` isn't used in CI.
The recent changes to `mob prompt` set some env variables, https://github.com/mobilecoinfoundation/mobilecoin/blob/master/mob#L211 
These get picked up by https://github.com/mobilecoinofficial/docker-rust-sgx-base/blob/main/entrypoint-builder-install.sh#L20

This means only if those env values are set will the user info propagate into the container, otherwise the containers run as before, which is as sudo

garbageslam: can we add this line to the entrypoint.sh instead?

nick-mobilecoin: @jgreat do you have thoughts on expanding the entrypoint.sh to create a postgres user, since it's looking at the user env already.

jgreat: We could put those pg bootstrapping steps in the entrypoint, but longer term we should look at extracting pg from the mob prompt container and utilize the GHA `services:` options for pg usage.  The pg container does all the boot strapping for you. 

We could extend mob prompt to wrap docker compose for external services required for local development.

jgreat: Actually we already do the user creation in the image build: https://github.com/mobilecoinofficial/docker-rust-sgx-base/blob/06741c8b14c3cd734ecc371874d9aafc272192b9/Dockerfile#L93-L96

Could add `service postgresql start` to the entrypoint.sh

jgreat (approved):
LGTM, but would be unnecessary if you add `service postgresql start` to the entrypoint.sh script in https://github.com/mobilecoinofficial/docker-rust-sgx-base.

If you do add the start to the entrypoint script, you may want to add instructions on testing out if the db is working here.

fog/README.md
nick-mobilecoin: that creates the `root` user, we need to create one for the localuser

wjuan-mob (approved):

fog/README.md
nick-mobilecoin: With #2607 and the comment of 
> onger term we should look at extracting pg from the mob prompt container and utilize the GHA services: options for pg usage. The pg container does all the boot strapping for you.

> We could extend mob prompt to wrap docker compose for external services required for local development.

I'm going to resolve and merge this.  Once that longer term solution happens it will require removing both postgres steps from the readme

PR #2623, opened 2022-09-28
Feature/circuit oram

wjuan-mob:
### Future Work

[Soundtrack of this PR]()


fog/ingest/enclave/impl/src/rng_store.rs
```diff
@@ -24,18 +24,18 @@ type KeySize = U32;
...
+type ObliviousRAMAlgo<OSC> = CircuitORAM4096Z2Creator<McRng, OSC>;
 
 // These are the requirements on the storage, this is imposed by the choice of
 // oram algorithm
 
 /// The storage data size which OSC must be able to support
 pub type StorageDataSize = U4096;
 /// The storage meta size which OSC must be able to support
-pub type StorageMetaSize = U64;
+pub type StorageMetaSize = U32;
```
garbageslam, line 38: this is going down because the bucket size is two now?

wjuan-mob: Yes. Because we're going from 16x4 to 16x2.

PR #2617, opened 2022-09-28
Modify MVQ to use NonceMessage

samdealy:
### Motivation
The MultiViewStoreQuery uses `attest::Message` queries but needs to use `attest::NonceMessage` queries now that FVR is using Nonce based sessions.  This PR makes this change but stops short of adding a `query` enclave method that takes a `NonceMessage`. 

NOTE: CI is failing for this PR for reasons that don't seem to be associated with the new functionality. I'm going to investigate this CI failure in parallel, but it shouldn't block this PR. 

### Future Work
- Create a `query` enclave method that takes a `NonceMessage`.

PR #2549, opened 2022-09-16
Add inbound frontend connection support crypto-ake-enclave

jcape:
- Add `AkeEnclaveState::frontends` to keep track of nonce-session-based sessions
- Plumb new `frontend_accept` and `frontend_close` methods to add/remove elements from it.

### Motivation

Earlier PRs developed a specific session type for explicit nonce sessions, this allows their use by `AkeEnclaveState` to allow backend enclaves to track sessions from frontends.

PR #2625, opened 2022-09-28
Replace client with nonce paradigm for backends

samdealy:
### Motivation
@jcape created new `Nonce`-based structs that allow Fog View Router (FVR) and Fog VIew Stores to communicate with messages that contain the nonce used in the noise protocol that attests the communication. #2549 added support for inbound frontend messages, and this PR adds the support for outbound "backend" messages. 

Note that this PR stops short of modifying the `MultiviewStoreQueryRequest` and `MultiviewStoreQueryResponse`, and instead supplies a blank `attest::Message` for the conversion that the `router_request_handler` uses. The following PR (#2617) implements this change.

RFC: In the `backend_decrypt` method, I use the `peek_nonce` method rather than the `get_nonce` method to decrypt the message from the backend. See comment for more info but could use confirmation that this is correct.

Note: I included changes to some Cargo.toml files that needed their deps to be sorted.

### Future Work
* #2617: Modify `MultiViewStoreQueryRequest` + `Response` to use `attest::NonceMessage`
* Create a `query_for_nonce` enclave method that fulfills query requests when an `attest::NonceMessage` is used.
* Finish frontend work by using the auth methods in FVR.

### DIFF
--- a/crypto/ake/enclave/src/lib.rs
+++ b/crypto/ake/enclave/src/lib.rs
@@ -560,7 +560,7 @@ impl<EI: EnclaveIdentity> AkeEnclaveState<EI> {
         // Ensure lock gets released as soon as we're done decrypting.
         let mut backends = self.backends.lock()?;
         backends
-            .get_mut(&responder_id)
+            .get_mut(responder_id)
             .ok_or(Error::NotFound)
             .and_then(|session| {
                 Ok(session.decrypt_with_nonce(&msg.aad, &msg.data, msg.channel_id.peek_nonce())?)
--- a/crypto/ake/enclave/src/lib.rs
+++ b/crypto/ake/enclave/src/lib.rs
@@ -554,8 +554,8 @@ impl<EI: EnclaveIdentity> AkeEnclaveState<EI> {
 
     pub fn backend_decrypt(
         &self,
-        responder_id: ResponderId,
-        msg: EnclaveMessage<NonceSession>,
+        responder_id: &ResponderId,
+        msg: &EnclaveMessage<NonceSession>,
     ) -> Result<Vec<u8>> {
         // Ensure lock gets released as soon as we're done decrypting.
         let mut backends = self.backends.lock()?;
--- a/crypto/ake/enclave/src/lib.rs
+++ b/crypto/ake/enclave/src/lib.rs
@@ -239,12 +239,12 @@ impl<EI: EnclaveIdentity> AkeEnclaveState<EI> {
         Ok(())
     }
 
-    /// Constructs a ClientAuthRequest to be sent to an enclave backend.
+    /// Constructs a NonceAuthRequest to be sent to an enclave backend.
     ///
     /// Differs from peer_init in that this enclave does not establish a peer
     /// connection to the enclave described by `backend_id`. Rather, this
     /// enclave serves as a client to this other backend enclave.
-    pub fn backend_init(&self, backend_id: ResponderId) -> Result<ClientAuthRequest> {
+    pub fn backend_init(&self, backend_id: ResponderId) -> Result<NonceAuthRequest> {
         let mut csprng = McRng::default();
 
         let initiator = Start::new(backend_id.to_string());
@@ -264,7 +264,7 @@ impl<EI: EnclaveIdentity> AkeEnclaveState<EI> {
     pub fn backend_connect(
         &self,
         backend_id: ResponderId,
-        backend_auth_response: ClientAuthResponse,
+        backend_auth_response: NonceAuthResponse,
     ) -> Result<()> {
         let initiator = self
             .backend_auth_pending
@@ -530,7 +530,7 @@ impl<EI: EnclaveIdentity> AkeEnclaveState<EI> {
     pub fn reencrypt_sealed_message_for_backends(
         &self,
         sealed_client_message: &SealedClientMessage,
-    ) -> Result<Vec<EnclaveMessage<ClientSession>>> {
+    ) -> Result<Vec<EnclaveMessage<NonceSession>>> {
         let (client_query_bytes, _) = sealed_client_message.data.unseal_raw()?;
 
         let mut backends = self.backends.lock()?;
@@ -538,8 +538,9 @@ impl<EI: EnclaveIdentity> AkeEnclaveState<EI> {
             .iter_mut()
             .map(|(_, encryptor)| {
                 let aad = sealed_client_message.aad.clone();
-                let data = encryptor.encrypt(&aad, &client_query_bytes)?;
-                let channel_id = sealed_client_message.channel_id.clone();
+                let (data, nonce) = encryptor.encrypt_with_nonce(&aad, &client_query_bytes)?;
+                let channel_id =
+                    NonceSession::new(sealed_client_message.channel_id.clone().into(), nonce);
                 Ok(EnclaveMessage {
                     aad,
                     channel_id,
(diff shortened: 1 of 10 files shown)


samdealy:
* feature/fog-view-router
  * **PR #2625** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2625" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2625?utm_source=stack-comment).

crypto/ake/enclave/src/lib.rs
```diff
@@ -554,14 +555,16 @@ impl<EI: EnclaveIdentity> AkeEnclaveState<EI> {
...
+        msg: EnclaveMessage<NonceSession>,
     ) -> Result<Vec<u8>> {
         // Ensure lock gets released as soon as we're done decrypting.
         let mut backends = self.backends.lock()?;
         backends
             .get_mut(&responder_id)
             .ok_or(Error::NotFound)
-            .and_then(|session| Ok(session.decrypt(&msg.aad, &msg.data)?))
+            .and_then(|session| {
+                Ok(session.decrypt_with_nonce(&msg.aad, &msg.data, msg.channel_id.peek_nonce())?)
```
samdealy, line 566: Should this be `get_nonce` or `peek_nonce`? I figured it should be peek nonce because we're just using it to decrypt and I think the nonce should only increment when it encrypts a message back to the sender?

fog/api/src/conversions.rs
```diff
@@ -5,26 +5,27 @@
...
 
-impl From<Vec<attest::Message>> for MultiViewStoreQueryRequest {
-    fn from(attested_query_messages: Vec<attest::Message>) -> MultiViewStoreQueryRequest {
+impl From<Vec<attest::NonceMessage>> for MultiViewStoreQueryRequest {
+    fn from(_attested_query_messages: Vec<attest::NonceMessage>) -> MultiViewStoreQueryRequest {
         let mut multi_view_store_query_request = MultiViewStoreQueryRequest::new();
-        multi_view_store_query_request.set_queries(attested_query_messages.into());
+        // TODO: Once MultiViewStoreQueryRequest.queries is modified to be a
+        // Vec<attest::NonceMessage> change this to use the
+        // _attested_query_messages_field.
```
nick-mobilecoin, line 27: :+1: nice description in the todo

fog/view/enclave/impl/src/lib.rs
```diff
@@ -263,15 +264,20 @@ where
...
+        // shard query responses.
+        let shard_query_response = shard_query_responses
+            .clone()
+            .into_iter()
             .next()
-            .expect("Shard query responses must have at least one response.")
-            .clone();
-        let shard_query_response_plaintext =
-            self.ake.client_decrypt(encrypted_shard_query_response)?;
+            .expect("Shard query responses must have at least one response.");
```
nick-mobilecoin, line 277: :wrench: Is the `clone()` necessary here?
(sorry github suggestion failure here)
```rust
        let shard_query_response = shard_query_responses
            .into_iter()
            .next()
            .expect("Shard query responses must have at least one response.");
```

fog/view/server/src/router_request_handler.rs
```diff
@@ -233,7 +233,7 @@ async fn authenticate_view_store<E: ViewEnclaveProxy>(
     logger: Logger,
 ) -> Result<(), RouterServerError> {
     let view_store_id = view_store_url.responder_id()?;
-    let client_auth_request = enclave.view_store_init(view_store_id.clone())?;
+    let nonce_auth_request = enclave.view_store_init(view_store_id.clone())?;
```
nick-mobilecoin, line 236: :question: Whats the reason for the `nonce_` prefix.  Would this be more stable to be `auth_request`.  I'm coming in fresh, so may not fully understand the Nonce types and usage, but it feels like `nonce_` is the data type and ideally this code shouldn't need a change if the underlying data types change.

Similar thought on the `nonce_auth_response` name

crypto/ake/enclave/src/lib.rs
nick-mobilecoin: (I'm coming in fresh so probably am lacking some understanding)
I think peeking makes sense.  If somewhere else in the code has a need to decrypt `msg` then a get here will prevent future decryptions.

One thing I'm not clear on is the message back.  Most of the steps look to be post incrementing the nonce.  This means unless there is an explicit pre-increment the first reply might re-use the same nonce.

fog/view/server/src/router_request_handler.rs
awygle: I agree with Nick here.

crypto/ake/enclave/src/lib.rs
awygle: My belief is that either `peek_nonce` or `get_nonce` would work here, as we're taking ownership of the `EnclaveMessage<NoiseSession>` and not returning it, and `NoiseSession` doesn't hold the nonce value by reference or anything like that.

As for Nick's comment, `decrypt_with_nonce` calls `set_nonce` and then `decrypt`, which will post-increment the nonce. So the only case in which we'd be possibly in trouble would be a sequence that looks like this:

`Node1 sends message A with nonce 0 -> Node2 decrypts message A with nonce 0 -> Node2 sends message B with nonce 1 -> Node2 decrypts message A with nonce 0 (for a second time) -> Node2 sends message C with nonce 1 (key is now compromised)`.

Is `EnclaveMessage<NonceSession>` `Clone` (or `Copy` I guess)? Because if so, the above can happen (although it would have to be us that screwed it up rather than client code, so at least that's nice).

jcape: Updating from offline: `peek_nonce()` is fine in this case, because `channel_id` is ephemeral, and as a general rule, `const > mut`.

It's probably advisable to refactor `peek_nonce()` into an immutable `nonce()` method, and delete `get_nonce()` altogether. I wrote those accessors before I decided apps didn't need to shoulder the responsibility for maintaining nonces apart from `Ready<>`.

samdealy: SGTM.

fog/view/server/src/router_request_handler.rs
samdealy: The `nonce` helps the reader understand that it corresponds to an actual type `NonceAuthRequest` rather than a nebulous `auth_request` that might be passed to a grpc call. This shouldn't get changed again so I don't see the staleness being an issue.

I'd prefer to keep it because I often have a hard time understanding what the types are in the handler because we're doing so many conversions between bytes and structs. That said, if you both feel strongly, I'll change it.

nick-mobilecoin: I'm fine with leaving it and see how it plays out.

awygle: Yeah my feelings are not that strong.

awygle (approved):

fog/view/enclave/impl/src/lib.rs
jcape: This ends up cloning the entire list in order to use the first element, but presumably `backend_decrypt()` is taking a reference? If so, what you want to do is something like this:

```
let plaintext = {
    let response = shard_query_responses
        .iter()
        .next()
        .expect("Shard query responses must have at least one response");
    self.ake.backend_decrypt(response.0, response.1)?
};
```

fog/view/server/src/router_request_handler.rs
jcape: FWIW, I don't think there's any value in calling this `nonce_auth_response`. There's only one type of auth that's used in this entire function. ;-)

fog/view/enclave/impl/src/lib.rs

PR #2627, opened 2022-09-29
Create query_backend enclave methods

samdealy:
### Motivation
PR #2617 updated the MVQ request and responses to use `attest::NonceMessage`, and modified the Fog View Store API to accept this new MVQ but didn't implement the logic to actually fulfill this nonce-based query. This PR implements this by providing new enclave methods.

### DIFF
--- a/crypto/ake/enclave/src/lib.rs
+++ b/crypto/ake/enclave/src/lib.rs
@@ -216,7 +216,6 @@ impl<EI: EnclaveIdentity> AkeEnclaveState<EI> {
     }
 
     pub fn frontend_decrypt(&self, msg: EnclaveMessage<NonceSession>) -> Result<Vec<u8>> {
-        // Ensure lock gets released as soon as we're done decrypting.
         let mut frontends = self.frontends.lock()?;
         frontends
             .get_mut(&msg.channel_id)
--- a/crypto/ake/enclave/src/lib.rs
+++ b/crypto/ake/enclave/src/lib.rs
@@ -196,6 +196,36 @@ impl<EI: EnclaveIdentity> AkeEnclaveState<EI> {
         }
     }
 
+    pub fn frontend_encrypt(
+        &self,
+        session_id: &NonceSession,
+        aad: &[u8],
+        data: &[u8],
+    ) -> Result<EnclaveMessage<NonceSession>> {
+        let mut frontends = self.frontends.lock()?;
+        let session = frontends.get_mut(session_id).ok_or(Error::NotFound)?;
+        let (data, nonce) = session.encrypt_with_nonce(aad, data)?;
+        let channel_id = NonceSession::new(session.binding().to_owned(), nonce);
+
+        // Return message
+        Ok(EnclaveMessage {
+            aad: aad.to_vec(),
+            channel_id,
+            data,
+        })
+    }
+
+    pub fn frontend_decrypt(&self, msg: EnclaveMessage<NonceSession>) -> Result<Vec<u8>> {
+        // Ensure lock gets released as soon as we're done decrypting.
+        let mut frontends = self.frontends.lock()?;
+        frontends
+            .get_mut(&msg.channel_id)
+            .ok_or(Error::NotFound)
+            .and_then(|session| {
+                Ok(session.decrypt_with_nonce(&msg.aad, &msg.data, msg.channel_id.nonce())?)
+            })
+    }
+
     /// Accept an explicit-nonce session from a frontend service (router) to
     /// ourselves (acting as a store).
     pub fn frontend_accept(
--- a/fog/view/enclave/api/src/lib.rs
+++ b/fog/view/enclave/api/src/lib.rs
@@ -83,6 +83,9 @@ pub enum ViewEnclaveRequest {
     /// An encrypted fog_types::view::QueryRequest
     /// Respond with fog_types::view::QueryResponse
     Query(EnclaveMessage<ClientSession>, UntrustedQueryResponse),
+    /// An encrypted fog_types::view::QueryRequest
+    /// Respond with fog_types::view::QueryResponse.
+    QueryBackend(EnclaveMessage<NonceSession>, UntrustedQueryResponse),
     /// Request from untrusted to add encrypted tx out records to ORAM
     AddRecords(Vec<ETxOutRecord>),
     /// Takes a client query message and returns a SealedClientMessage
@@ -161,6 +164,13 @@ pub trait ViewEnclaveApi: ReportableEnclave {
         payload: EnclaveMessage<ClientSession>,
         untrusted_query_response: UntrustedQueryResponse,
     ) -> Result<Vec<u8>>;
+
+    /// Service a query response from a frontend.
+    fn query_backend(
+        &self,
+        payload: EnclaveMessage<NonceSession>,
+        untrusted_query_response: UntrustedQueryResponse,
+    ) -> Result<(Vec<u8>, u64)>;
 
     /// SERVER-FACING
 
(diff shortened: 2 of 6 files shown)


samdealy:
* feature/fog-view-router
  * **PR #2625** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2625" target="_blank">Graphite</a> 
      * **PR #2627** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2627" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2627?utm_source=stack-comment).

crypto/ake/enclave/src/lib.rs
```diff
@@ -196,6 +196,36 @@ impl<EI: EnclaveIdentity> AkeEnclaveState<EI> {
...
+        // Return message
+        Ok(EnclaveMessage {
+            aad: aad.to_vec(),
+            channel_id,
+            data,
+        })
+    }
+
+    pub fn frontend_decrypt(&self, msg: EnclaveMessage<NonceSession>) -> Result<Vec<u8>> {
+        // Ensure lock gets released as soon as we're done decrypting.
```
nick-mobilecoin, line 219: :wrench: I'm not sure what this comment is trying to communicate.  It feels like it might have come from a scoped block that no longer exists
```suggestion
```

fog/view/enclave/api/src/lib.rs
```diff
@@ -162,6 +165,13 @@ pub trait ViewEnclaveApi: ReportableEnclave {
         untrusted_query_response: UntrustedQueryResponse,
     ) -> Result<Vec<u8>>;
 
+    /// Service a query response from a frontend.
+    fn query_backend(
```
nick-mobilecoin, line 169: :wrench: should the comment say `backend`?
```suggestion
    /// Service a query response from a backend.
    fn query_backend(
```

fog/view/enclave/impl/src/lib.rs
```diff
@@ -61,6 +61,63 @@ where
...
     }
+
+    fn query_impl(
+        &self,
+        plaintext_request: &[u8],
+        untrusted_query_response: UntrustedQueryResponse,
+    ) -> Result<Vec<u8>> {
+        let req: QueryRequest = mc_util_serial::decode(plaintext_request).map_err(|e| {
+            log::error!(self.logger, "Could not decode user request: {}", e);
+            Error::ProstDecode
```
nick-mobilecoin, line 72: :wrench: It looks like this was `ProstDecode` previously, but looking at `mc_util_serial::decode()` it looks like this is a serde decode
```suggestion
            Error::SerdeDecode
```

fog/view/enclave/api/src/lib.rs
```diff
@@ -162,6 +165,13 @@ pub trait ViewEnclaveApi: ReportableEnclave {
...
+        &self,
+        payload: EnclaveMessage<NonceSession>,
+        untrusted_query_response: UntrustedQueryResponse,
+    ) -> Result<(Vec<u8>, u64)>;
```
nick-mobilecoin, line 173: :question: perhaps it was due to `fog/view/enclave/src/lib.rs` dealing with the raw values, but was the idea of returning an `EnclaveMessage<>` considered?

fog/view/enclave/impl/src/lib.rs
```diff
@@ -127,64 +184,31 @@ where
...
         Ok(response.data)
     }
 
+    fn query_backend(
+        &self,
+        msg: EnclaveMessage<NonceSession>,
+        untrusted_query_response: UntrustedQueryResponse,
+    ) -> Result<(Vec<u8>, u64)> {
+        let channel_id = msg.channel_id.clone();
+        let user_plaintext = self.ake.frontend_decrypt(msg)?;
```
nick-mobilecoin, line 202: :question: I'm a little confused here was there a mistake in the naming with `backend` and `frontend`?  Or is this a case where it's relative names i.e. this is a middle player and it's backend talks to another's frontend?

If the latter is the case I think there should be some clarifying comments, even if this is documented in some kind of architecture diagram, if someone lands here it's going to cause some head scratching

fog/view/server/src/fog_view_service.rs
```diff
@@ -193,15 +193,14 @@ where
...
         tracer.in_span("query_nonce_impl", |_cx| {
-            // TODO: Create query_nonce enclave method that does what query currently does
-            // but for NonceMessage. It should produce data and a nonce that is
-            // then set on the nonce_message.
-            let _untrusted_query_response =
+            let untrusted_query_response =
                 self.create_untrusted_query_response(request.get_aad(), &tracer)?;
-            let data = vec![0; 0];
-            let nonce = 0;
+            let (data, nonce) = tracer.in_span("enclave_query", |_cx| {
```
nick-mobilecoin, line 199: :wrench: 
```suggestion
            let (data, nonce) = tracer.in_span("enclave_query_backend", |_cx| {
```

fog/view/enclave/impl/src/lib.rs
jcape: We shouldn't have any serde in encrypted messages.

nick-mobilecoin: nvm, my bad
```rust
pub use prost::{DecodeError, EncodeError, Message};
...
pub fn decode<T: Message + Default>(buf: &[u8]) -> Result<T, DecodeError> {
    T::decode(buf)
}
```
That looks like prost.
I was looking at the `pub mod decode`

fog/view/enclave/api/src/lib.rs
samdealy: No, the response is coming from a frontend.

crypto/ake/enclave/src/lib.rs
samdealy: removed.

fog/view/enclave/api/src/lib.rs
samdealy: Yes, and now that I'm thinking about it, James and I discussed modifying this PR to implement that. 

Originally I chose to implement it as a tuple because most Fog enclave methods return bytes rather than the EnclaveMessage. However, it's clearer in this case to return the message. Done.

fog/view/enclave/impl/src/lib.rs
samdealy: I can see the confusion. I changed the name to `query_store` so that it's clear that it's the query method for Fog View Store's (rather than client-facing Fog View instances).

fog/view/server/src/fog_view_service.rs
samdealy: Done (but used `store` instead)

fog/view/enclave/api/src/lib.rs
```diff
@@ -83,6 +83,9 @@ pub enum ViewEnclaveRequest {
     /// An encrypted fog_types::view::QueryRequest
     /// Respond with fog_types::view::QueryResponse
     Query(EnclaveMessage<ClientSession>, UntrustedQueryResponse),
+    /// An encrypted fog_types::view::QueryRequest
+    /// Respond with fog_types::view::QueryResponse.
+    QueryBackend(EnclaveMessage<NonceSession>, UntrustedQueryResponse),
```
jcape, line 88: I think this may be the source of Nick's comment later, this should probably be called `QueryStore` to match the `query_store` method name...?

jcape (approved):
LGTM pending `QueryBackend` rename.

fog/view/enclave/api/src/lib.rs
samdealy: Done.

fog/view/enclave/impl/src/lib.rs
samdealy: Done.

PR #2629, opened 2022-09-29
Make the test client exercise the partial fill swaps

garbageslam:
This is analogous to this PR:

https://github.com/mobilecoinfoundation/mobilecoin/pull/1979

except now we exercise the partial fill swap functionality.

This gives test coverage over the new functionality in a deployed network.


fog/test-client/src/test_client.rs
```diff
@@ -669,33 +684,50 @@ impl TestClient {
...
 
     /// Conduct an atomic swap transfer between two clients
-    /// Returns the transaction and the block count of the node it was submitted
-    /// to.
     ///
-    /// The target client builds a signed contingent input, then the source
-    /// client incorporates this into a Tx and submits it.
+    /// The source client builds an SCI, and the target client builds the swap
+    /// transaction. This only builds and submits the transactions, it does not
+    /// confirm it.
```
garbageslam, line 690: i decided its more logical for the source client to be the originator and build the SCI, and the target client should be the counterparty that fills it

fog/test-client/src/test_client.rs
```diff
@@ -669,33 +684,50 @@ impl TestClient {
...
-    ) -> Result<TransferData, TestClientError> {
+        is_partial_fill: bool,
+    ) -> Result<SwapTransferData, TestClientError> {
         self.tx_info.clear();
         let target_address = target_client.get_account_key().default_subaddress();
+
+        let mut rng = McRng::default();
+
+        let tok1_val = 1 + thread_rng().gen_range(0, self.policy.transfer_amount);
+        let tok2_val = 1 + thread_rng().gen_range(0, self.policy.transfer_amount);
```
garbageslam, line 717: whether or not we are doing partial fills, we choose random numbers for tok1 and tok2 value, to get better test coverage on rounding, boundary conditions, etc.

fog/test-client/src/test_client.rs
```diff
@@ -730,30 +761,61 @@ impl TestClient {
         let fee = Amount::new(fee_value, token_id1);
 
         // Build swap proposal
-        // Note: We are adding fee-value here to avoid "SCI Unprofitable" errors,
-        // when transfer_amount is very small
-        let signed_input = target_client
+        let signed_input = source_client
             .build_swap_proposal(
-                Amount::new(self.policy.transfer_amount + fee_value, token_id1),
```
garbageslam, line 737: this hack of adding fee_value is no longer necessary

garbageslam:
this passed CD

fog/sample-paykit/src/client.rs
```diff
@@ -541,6 +582,51 @@ impl Client {
                 .or_default() += req_amount.value;
         }
 
+        let sci_token_id = TokenId::from(sci.pseudo_output_amount.token_id);
+        let sci_value = sci.pseudo_output_amount.value;
+
+        // Now we have to case out on the partial-fill vs. non-partial fill flow
+        if let Some(fill_amount) = fill_amount {
+            // Compute the parameter that we need to pass to the sci builder
+            // FIXME: Don't unwrap here
```
garbageslam, line 591: This is actually okay strictly speaking because validated is called earlier

fog/test-client/src/test_client.rs
wjuan-mob: Maybe we should use that language then, if that's how we're wording it? If we call them sci originator, and sci counterparty would that add clarity?

eranrund (approved):

fog/test-client/Cargo.toml
```diff
@@ -40,6 +40,7 @@ lazy_static = "1.4"
 maplit = "1.0"
 more-asserts = "0.3"
 once_cell = "1.15"
+rand = "0.7"
```
remoun, line 43: Nit: `rand` is currently at 0.8.5

```suggestion
rand = "0.8"
```

fog/test-client/src/test_client.rs
remoun, line 717: Why not use the `rng` declared right before?
```suggestion
        let tok1_val = 1 + rng.gen_range(0, self.policy.transfer_amount);
        let tok2_val = 1 + rng.gen_range(0, self.policy.transfer_amount);
```

remoun (approved):
LGTM

fog/test-client/src/test_client.rs
garbageslam: sure, but at a higher level in the test, the language source-client and target-client is used. but in this function i guess we could do that

fog/test-client/src/test_client.rs
garbageslam: it doesn't compile, this function requires the `Rng` trait, but `McRng` only implement `RngCore`. This is ultimately because we don't pull `rand` into the enclave, we pull `rand-core`.

fog/test-client/src/test_client.rs
garbageslam: Let's make an issue to track this in a subsequent PR, I plan to make some follow-on changes for the sci interface anyways.

garbageslam: I changed the code comment, but not the variable names

PR #2637, opened 2022-09-30
fixup cargo toml

garbageslam:
https://github.com/mobilecoinfoundation/mobilecoin/commit/bf4216551790b05c70a29321e248a8a58b1739b4

In referenced commit, we started using features from rand-core 0.6.4, after dependabot had bumped us to that version.

However, this can break downstream, because we only had 0.6 in the Cargo.toml, so cargo could potentially select 0.6.3 still.

Thanks to Davey for reporting the build breakage.

### DIFF
--- a/crypto/ring-signature/Cargo.toml
+++ b/crypto/ring-signature/Cargo.toml
@@ -30,7 +30,7 @@ mc-util-serial = { path = "../../util/serial" }
 # Enable all default features not known to break code coverage builds
 proptest = { version = "1.0", default-features = false, features = ["default-code-coverage"], optional = true }
 prost = { version = "0.11", default-features = false, features = ["prost-derive"] }
-rand_core = { version = "0.6", default-features = false }
+rand_core = { version = "0.6.4", default-features = false }
 serde = { version = "1.0", default-features = false, features = ["derive"] }
 subtle = { version = "2.4.1", default-features = false, features = ["i128"] }
 zeroize = { version = "1", default-features = false }


crypto/ring-signature/Cargo.toml
```diff
@@ -30,7 +30,7 @@ mc-util-serial = { path = "../../util/serial" }
 # Enable all default features not known to break code coverage builds
 proptest = { version = "1.0", default-features = false, features = ["default-code-coverage"], optional = true }
 prost = { version = "0.11", default-features = false, features = ["prost-derive"] }
-rand_core = { version = "0.6", default-features = false }
+rand_core = { version = "0.6.4", default-features = false }
```
nick-mobilecoin, line 33: :question: Should we be updating all of our deps to use "0.6.4"?
Even if we want to update all eventually, I see no problem only doing this one, especially if it's quick turnaround to get someone going.
```shell
➜  mobilecoin git:(fixup-cargo-toml) rg rand_core -ttoml -uu
transaction/core/Cargo.toml
19:rand_core = { version = "0.6", default-features = false }

transaction/std/Cargo.toml
17:rand_core = { version = "0.6", default-features = false }

util/from-random/Cargo.toml
10:rand_core = "0.6"

util/keyfile/Cargo.toml
41:rand_core = "0.6.4"

mobilecoind/Cargo.toml
83:rand_core = "0.6"
...
```

jcape (approved):

PR #2638, opened 2022-09-30
Remove `get_nonce` from NonceSession, rename `peek_nonce`->`nonce`

awygle:
Also sort Cargo.toml files due to precommit hook.

### Motivation

There's no reason to expose `get_nonce`, and it could cause some issues, so it's being removed. Without `get_nonce` to contrast against, there's no reason not to call `peek_nonce` just `nonce`.

### DIFF
--- a/attest/api/src/conversions.rs
+++ b/attest/api/src/conversions.rs
@@ -126,7 +126,7 @@ impl From<EnclaveMessage<NonceSession>> for NonceMessage {
         retval.set_aad(src.aad);
         // it doesn't matter if we don't bump the nonce when retrieving it,
         // src.channel_id will be discarded anyways.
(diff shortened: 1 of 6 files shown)


attest/api/src/conversions.rs
```diff
@@ -126,7 +126,7 @@ impl From<EnclaveMessage<NonceSession>> for NonceMessage {
         retval.set_aad(src.aad);
         // it doesn't matter if we don't bump the nonce when retrieving it,
         // src.channel_id will be discarded anyways.
```
nick-mobilecoin, line 128: :wrench: Seems like this comment is no longer pertinent
```suggestion
```

nick-mobilecoin (approved):

jcape (approved):

PR #2652, opened 2022-10-04
Updating local network README files

dolanbernard:
Updated README in local-network to include optional block-version parameter
Updated README in fog-local-network to include postgresql setup
Updated default block version in local_network.py from 2 to 3

### Motivation

These changes should be made in order to have more complete, up-to-date instructions for running and testing on a local network.


nick-mobilecoin (approved):
:+1:

eranrund (approved):

PR #2654, opened 2022-10-04
fixup to partial fills -- input rules check for zero partial fill change

garbageslam:
this was suggested in review of the mcip, and brings the implementation in line with the spec. Previously this was only checked in the SCI validate routine.

It doesn't create a bug in the enclave to omit
this check, but it also simplifies review if we just have one spec for everywhere and enforce it everywhere.

### DIFF
--- a/transaction/core/src/input_rules.rs
+++ b/transaction/core/src/input_rules.rs
@@ -100,9 +100,18 @@ impl InputRules {
             // There is a partial fill change output. Let's try to unblind its amount.
             let (partial_fill_change_amount, _) = partial_fill_change.reveal_amount()?;
             // If the min fill value exceeds the fractional change, then the SCI is
-            // ill-formed
+            // ill-formed.
+            // This check ensures that the expression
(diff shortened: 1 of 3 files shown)


transaction/core/src/input_rules.rs
```diff
@@ -100,11 +100,20 @@ impl InputRules {
...
+            // `partial_fill_change_amount.value - self.min_partial_fill_value`
+            // does not panic.
             if partial_fill_change_amount.value < self.min_partial_fill_value {
                 return Err(InputRuleError::MinPartialFillValueExceedsPartialFillChange);
             }
 
+            // Per spec, check if the partial fill change amount is zero.
+            // This is always invalid, and can lead to division by zero on clients.
+            if partial_fill_change_amount.value == 0 {
+                return Err(InputRuleError::ZeroPartialFillChange);
```
remoun, line 114: Consider adding a test case that exercises this path

remoun (approved):

eranrund (approved):

nick-mobilecoin (approved):

PR #2656, opened 2022-10-04
master: [Fixes] CI/CD - Run CD on PR to master, release/ branches.

jgreat:
### Motivation

Make sure that CD is run when we have a PR to `master` or `release/` branches.  Skip CI/CD when changes are only to `.md` files.

When to require CI/CD runs:
- PR to `master`
- PR to `release/`

Opt into CD:
- create a branch with `feature/`

How to opt out of CI/CD runs:
- add the GHA built in `[skip ci]` to your commit message.

### Changes:

- `on.pull_request` CD workflow
- fix some of the skip filters to line up with the jobs in CD
- update cd docs

### Future Work
- add more file exclusions
- Add logic to skip just CD or just CI.


.internal-ci/README.md
```diff
@@ -36,58 +37,61 @@ v0.0.0-my-awesome-feature.21.sha-abcd1234
...
+
+
 ## CI triggers
 
 This workflow is set up to trigger of certain branch patterns.
 
-### Feature Branches - `feature/*`
+### Branches - `feature/*, release/*, master`
 
 Feature branches will trigger a build that will create a dynamic development environment and run integration tests against the environment.
```
nick-mobilecoin, line 79: :wrench: Should the text be genericised as well?
```suggestion
### Branches - `feature/*, release/*, master`

Branches will trigger a build that will create a dynamic development environment and run integration tests against the environment.
```

.internal-ci/README.md
```diff
@@ -36,58 +37,61 @@ v0.0.0-my-awesome-feature.21.sha-abcd1234
...
 
 | Tags | SGX_MODE | IAS_MODE | Signer | Description |
 | --- | --- | --- | --- | --- |
 | `v0.0.0-my-awesome-feature.21.sha-abcd1234` | `HW` | `DEV` | CI Signed Development | For use in development environments. |
 
-### Release Branches - `release/*`
+### Semver tags - `v2.0.0`
 
 Release branches will trigger a build that will create a set of release artifacts.
```
nick-mobilecoin, line 87: :wrench: similar to above should the text be updated for _tags_
```suggestion
### Semver tags - `v2.0.0`

Tags will trigger a build that will create a set of release artifacts.
```

nick-mobilecoin (approved):

.internal-ci/README.md
jgreat: Thanks!

.internal-ci/README.md
jgreat: Thanks!

garbageslam (changes requested):
I meant to leave this comment on this one: https://github.com/mobilecoinfoundation/mobilecoin/pull/2762#pullrequestreview-1152146656

I think this is a bad change to merge right now, and it's going to blow up the dev cluster.

If we do this, then every dependabot PR (which opens against master), will create it's own network, in order to do CD on that PR.

This will consume all of the cluster resources and prevent us from testing our actual work.

Without a fix for this issue I don't think we can merge this change.

garbageslam:
i think currently dependabot opens as many as 12 PRs at once typically and the configured limit is like 25. But it seems that once we have 5 or 6 concurrent networks, they all start failing?

If we also reduce dependabot limits to like 2 or 3, so they happen in the middle of the night and are then deleted, it might be okay? Can we do that dependabot change in this PR?

garbageslam:
Eran also suggested that we dont want to run CD against dependabot at all. I think that is correct.

joekottke (approved):

garbageslam (approved):

eranrund (approved):

.github/workflows/mobilecoin-dev-cd.yaml
```diff
@@ -19,15 +26,20 @@ on:
     - release/*
     tags:
       - v[0-9]+*
+    paths-ignore:
+    - '**.md'
 
 # don't run more than one at a time for a branch/tag
-concurrency: mobilecoin-dev-cd-${{ github.ref }}
+concurrency:
+  group: mobilecoin-dev-cd-${{ github.head_ref || github.ref }}
```
remoun, line 34: `head_ref` doesn't seem right here -- IIUC this would result in two PRs to `master` clashing

```suggestion
  group: mobilecoin-dev-cd-${{ github.ref }}
```

remoun: NVM, I mixed up `head_ref` and `base_ref`

PR #2662, opened 2022-10-06
make `tools/local-network/local-network.py` not do ip checks

garbageslam:
This was the simplest way I could see to turn off default-features only on mobilecoind (thanks Nick!)

### DIFF
--- a/tools/local-network/local_network.py
+++ b/tools/local-network/local_network.py
@@ -469,6 +469,7 @@ class Network:
 
         # Note: bypass-ip-check feature of mobilecoind is used here,
         # because ip-check is sometimes problematic for devs or CI machines.
+        # If the service rate limits you then mobilecoind fails to start.
         subprocess.run(
             f'cd {PROJECT_DIR} && CONSENSUS_ENCLAVE_PRIVKEY="{enclave_pem}" cargo build -p mc-consensus-service -p mc-ledger-distribution -p mc-admin-http-gateway -p mc-util-grpc-admin-tool -p mc-mobilecoind -p mc-crypto-x509-test-vectors -p mc-consensus-mint-client -p mc-util-seeded-ed25519-key-gen --feature mc-mobilecoind/bypass-ip-check {CARGO_FLAGS}',
             shell=True,
--- a/tools/local-network/local_network.py
+++ b/tools/local-network/local_network.py
@@ -470,8 +470,9 @@ class Network:
         # Note: no-default-features is set here to avoid the ip_check feature
         # of mc-mobilecoind, which is sometimes problematic for devs or CI machines.
         # If service rate limits you then you can't start mobilecoind.
+        # None of the other crates here have a default feature.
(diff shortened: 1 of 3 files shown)


garbageslam:
An alternative could be:

* Add a feature called `no-ip-check` to `mc-mobilecoind`
* It is off by default
* If enabled, then this overrides `ip-check` being on.
* Then we can simply do `cargo build ... -p mc-mobilecoind/no-ip-check ...`

tools/local-network/local_network.py
```diff
@@ -467,8 +467,11 @@ def build_binaries(self):
                 check=True,
             )
 
+        # Note: no-default-features is set here to avoid the ip_check feature
+        # of mc-mobilecoind, which is sometimes problematic for devs or CI machines.
+        # If service rate limits you then you can't start mobilecoind.
         subprocess.run(
-            f'cd {PROJECT_DIR} && CONSENSUS_ENCLAVE_PRIVKEY="{enclave_pem}" cargo build -p mc-consensus-service -p mc-ledger-distribution -p mc-admin-http-gateway -p mc-util-grpc-admin-tool -p mc-mobilecoind -p mc-crypto-x509-test-vectors -p mc-consensus-mint-client -p mc-util-seeded-ed25519-key-gen {CARGO_FLAGS}',
+            f'cd {PROJECT_DIR} && CONSENSUS_ENCLAVE_PRIVKEY="{enclave_pem}" cargo build --no-default-features -p mc-consensus-service/default -p mc-ledger-distribution/default -p mc-admin-http-gateway/default -p mc-util-grpc-admin-tool/default -p mc-mobilecoind -p mc-crypto-x509-test-vectors/default -p mc-consensus-mint-client/default -p mc-util-seeded-ed25519-key-gen/default {CARGO_FLAGS}',
```
nick-mobilecoin, line 474: Sorry I was incorrect on my original online suggestion, looks like none of the crates other than mobilecoind currnetly have default so it can be, the downside is if a default comes it will be turned off here :(
```suggestion
            f'cd {PROJECT_DIR} && CONSENSUS_ENCLAVE_PRIVKEY="{enclave_pem}" cargo build --no-default-features -p mc-consensus-service -p mc-ledger-distribution -p mc-admin-http-gateway -p mc-util-grpc-admin-tool -p mc-mobilecoind -p mc-crypto-x509-test-vectors -p mc-consensus-mint-client -p mc-util-seeded-ed25519-key-gen {CARGO_FLAGS}',
```

nick-mobilecoin (approved):

remoun (commented):
How about we make a `bypass-ip-check` feature? Cargo doesn't support removing features AFAIK, but we could add that feature that skips the check, and it would only be invokable via build flags.

garbageslam:
Okay, I will take that approach instead

david1alvarez (approved):
lgtm, verified this works

remoun (approved):

PR #2682, opened 2022-10-08
improve fog ingest logging when unique constraint is violated (release/v2)

garbageslam:
this doesn't happen very often, but when it did happen in testnet, we couldn't figure out why very easily and went database spelunking. we still don't know why, and in the future we want these details logged.

(see also https://github.com/mobilecoinfoundation/mobilecoin/pull/2681 which is this PR on master)

### DIFF
--- a/fog/sql_recovery_db/src/lib.rs
+++ b/fog/sql_recovery_db/src/lib.rs
@@ -588,10 +588,13 @@ impl SqlRecoveryDb {
             // errors.
             Err(Error::Orm(diesel::result::Error::DatabaseError(
                 diesel::result::DatabaseErrorKind::UniqueViolation,
-                _,
-            ))) => Ok(AddBlockDataStatus {
-                block_already_scanned_with_this_key: true,
-            }),
+                details,
+            ))) => {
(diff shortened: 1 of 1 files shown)


fog/sql_recovery_db/src/lib.rs
```diff
@@ -588,10 +588,13 @@ impl SqlRecoveryDb {
             // errors.
             Err(Error::Orm(diesel::result::Error::DatabaseError(
                 diesel::result::DatabaseErrorKind::UniqueViolation,
-                _,
-            ))) => Ok(AddBlockDataStatus {
-                block_already_scanned_with_this_key: true,
-            }),
+                details,
```
nick-mobilecoin, line 591: :question: There are should be no sensitive info in the error (`details`), correct?

nick-mobilecoin (approved):

fog/sql_recovery_db/src/lib.rs
garbageslam: Yup, nothing in this DB is sensitive

garbageslam: https://docs.rs/diesel/latest/diesel/result/trait.DatabaseErrorInformation.html

eranrund (approved):

PR #2683, opened 2022-10-08
Add TxSummary object and incorporate it into MLSAG signing digest

garbageslam:
The motivation here is that, hardware wallets like Ledger have an expectation that the device can verify what is the balance change that will result from signing a given transaction and who the outbound money will be paid to.

However, currently we have scoped things so that hardware wallets just are responsible for signing the Ring MLSAG, and they don't see large parts of the Tx, including the TxPrefix.

The only way that the set of outputs is connected to the RingMLSAG is that the TxPrefix is hashed into the "message", which is hashed with more things to produce the "extended_message_digest", which is the 32 byte digest that the RingMLSAG actually signs.

For the hardware wallet to really know what it is signing then, it would have to reproduce the entire extended message digest. The problem with this is that the TxPrefix can be > 100kb, which is much larger than the few kb that the hardware device currently has to handle, and is much too big to fit in the device's memory.

Instead, we introduce a `TxSummary` object and add an extra step where it is hashed in just at the very end. The computer can then prove that the hash the MLSAG is signing is connected to a given TxSummary, and provide enough information from there to understand the balance delta and where the money is going. This manages to avoid sending all the merkle proofs for all the mixins and so it can be done sending only a few kb.

The hope is to do this for v3, otherwise this issue may become the long pole for shipping hardware wallet support.

For other clients and servers, this doesn't meaningfully affect the time to sign or verify a Tx.

### DIFF
--- a/transaction/core/src/input_rules.rs
+++ b/transaction/core/src/input_rules.rs
@@ -13,7 +13,7 @@ use crate::{
     tx::{Tx, TxOut},
     BlockVersion, RevealedTxOut, RevealedTxOutError,
 };
-use alloc::{collections::BTreeSet, vec::Vec};
+use alloc::{collections::BTreeMap, vec::Vec};
 use displaydoc::Display;
 use mc_crypto_digestible::Digestible;
 use mc_crypto_keys::CompressedRistrettoPublic;
@@ -70,17 +70,17 @@ pub struct InputRules {
 }
 
 impl InputRules {
-    /// Get all tx target keys associated to outputs appearing in these rules
-    pub fn associated_tx_target_keys(&self) -> BTreeSet<CompressedRistrettoPublic> {
-        let mut result = BTreeSet::default();
+    /// Get all TxOut's appearing in these rules, keyed by public key
+    pub fn associated_tx_outs(&self) -> BTreeMap<CompressedRistrettoPublic, TxOut> {
+        let mut result = BTreeMap::default();
         for output in self.required_outputs.iter() {
-            result.insert(output.target_key);
+            result.insert(output.public_key, output.clone());
         }
         for output in self.partial_fill_outputs.iter() {
-            result.insert(output.tx_out.target_key);
+            result.insert(output.tx_out.public_key, output.tx_out.clone());
         }
         for output in self.partial_fill_change.iter() {
-            result.insert(output.tx_out.target_key);
+            result.insert(output.tx_out.public_key, output.tx_out.clone());
         }
         result
     }
--- a/transaction/core/src/input_rules.rs
+++ b/transaction/core/src/input_rules.rs
@@ -13,9 +13,10 @@ use crate::{
     tx::{Tx, TxOut},
     BlockVersion, RevealedTxOut, RevealedTxOutError,
 };
-use alloc::vec::Vec;
+use alloc::{collections::BTreeSet, vec::Vec};
 use displaydoc::Display;
 use mc_crypto_digestible::Digestible;
+use mc_crypto_keys::CompressedRistrettoPublic;
 use prost::Message;
 use serde::{Deserialize, Serialize};
 use zeroize::Zeroize;
@@ -69,6 +70,21 @@ pub struct InputRules {
 }
 
 impl InputRules {
+    /// Get all tx target keys associated to outputs appearing in these rules
+    pub fn associated_tx_target_keys(&self) -> BTreeSet<CompressedRistrettoPublic> {
+        let mut result = BTreeSet::default();
+        for output in self.required_outputs.iter() {
+            result.insert(output.target_key);
+        }
+        for output in self.partial_fill_outputs.iter() {
+            result.insert(output.tx_out.target_key);
+        }
+        for output in self.partial_fill_change.iter() {
+            result.insert(output.tx_out.target_key);
+        }
+        result
+    }
+
     /// Verify that a Tx conforms to the rules.
     pub fn verify(&self, block_version: BlockVersion, tx: &Tx) -> Result<(), InputRuleError> {
         // NOTE: If this function gets too busy, we should split it up
--- a/transaction/core/src/input_rules.rs
+++ b/transaction/core/src/input_rules.rs
@@ -1,4 +1,4 @@
-// Copyright (c) 2022 The MobileCoin Foundation
+// Copyright (c) 2018-2022 The MobileCoin Foundation
 
 //! Input rules, described in MCIP #31, specify any additional criteria that the
 //! Tx must satisfy to be valid.
@@ -18,13 +18,14 @@ use displaydoc::Display;
 use mc_crypto_digestible::Digestible;
 use prost::Message;
 use serde::{Deserialize, Serialize};
+use zeroize::Zeroize;
 
 /// A representation of rules on a transaction, imposed by the signer of some
 /// input in the transaction.
 ///
 /// Any rule could conceivably be added here if it can be evaluated against a
 /// `Tx`.
-#[derive(Clone, Digestible, PartialEq, Eq, Message, Serialize, Deserialize)]
+#[derive(Clone, Deserialize, Digestible, Eq, Hash, Message, PartialEq, Serialize, Zeroize)]
 pub struct InputRules {
     /// Outputs that are required to appear in the Tx prefix for the transaction
     /// to be valid
--- a/transaction/core/src/revealed_tx_out.rs
+++ b/transaction/core/src/revealed_tx_out.rs
@@ -7,10 +7,11 @@ use mc_crypto_digestible::Digestible;
 use mc_crypto_ring_signature::Scalar;
 use prost::Message;
 use serde::{Deserialize, Serialize};
+use zeroize::Zeroize;
 
 /// A TxOut together with its amount shared secret, which can be used to reveal
 /// the amount and token id and check them against the commitment data
-#[derive(Clone, Digestible, PartialEq, Eq, Message, Serialize, Deserialize)]
+#[derive(Clone, Deserialize, Digestible, Eq, Hash, Message, PartialEq, Serialize, Zeroize)]
 pub struct RevealedTxOut {
     /// The TxOut which is being revealed
     #[prost(message, required, tag = "1")]
--- a/transaction/core/src/ring_ct/rct_bulletproofs.rs
+++ b/transaction/core/src/ring_ct/rct_bulletproofs.rs
@@ -1045,7 +1045,7 @@ mod rct_bulletproofs_tests {
         ring_signature::{
             generators, Error as RingSignatureError, KeyImage, PedersenGens, ReducedTxOut,
         },
-        tx::TxPrefix,
+        tx::{TxIn, TxPrefix},
         CompressedCommitment, TokenId,
     };
     use alloc::vec::Vec;
@@ -1171,6 +1171,11 @@ mod rct_bulletproofs_tests {
                         blinding,
                     },
                 });
+
+                // This is janky, but needed so that the compute_mlsag_signing_digest
+                // code can find one `TxIn` for each input ring, and determine that
+                // there are no InputRules associated to any of them.
+                tx_prefix.inputs.push(TxIn::default());
             }
 
             // Create one output with the same value as each input.
--- a/transaction/core/src/ring_ct/rct_bulletproofs.rs
+++ b/transaction/core/src/ring_ct/rct_bulletproofs.rs
@@ -22,7 +22,7 @@ use mc_crypto_ring_signature::{
 };
 use mc_crypto_ring_signature_signer::{RingSigner, SignableInputRing};
 use mc_util_serial::prost::Message;
-use mc_util_zip_exact::zip_exact;
+use mc_util_zip_exact::{zip_exact, ZipExactError};
 use rand_core::{CryptoRng, RngCore};
 use serde::{Deserialize, Serialize};
 use zeroize::Zeroize;
@@ -33,7 +33,7 @@ use crate::{
     range_proofs::{check_range_proofs, generate_range_proofs},
     ring_ct::{Error, GeneratorCache},
     tx::TxPrefix,
-    Amount, BlockVersion, TxSummary,
+    Amount, BlockVersion, TxInSummary, TxSummary,
 };
 
 /// A presigned RingMLSAG and ancillary data needed to incorporate it into a
@@ -162,6 +162,13 @@ impl SigningData {
     /// * `check_value_is_preserved` - If true, check that the value of inputs
     ///   equals value of outputs.
     /// * `rng` - randomness
+    ///
+    /// Returns:
+    /// * The SigningData
+    /// * The TxSummary
+    /// * The extended message digest
+    ///
+    /// (The latter two are MCIP #52 and only relevant on small HW wallets)
     pub fn new<CSPRNG: RngCore + CryptoRng>(
         block_version: BlockVersion,
         tx_prefix: &TxPrefix,
@@ -170,7 +177,7 @@ impl SigningData {
         fee: Amount,
         check_value_is_preserved: bool,
         rng: &mut CSPRNG,
-    ) -> Result<Self, Error> {
+    ) -> Result<(Self, TxSummary, Vec<u8>), Error> {
         if !block_version.masked_token_id_feature_is_supported() && fee.token_id != 0 {
             return Err(Error::TokenIdNotAllowed);
         }
@@ -243,11 +250,10 @@ impl SigningData {
         let pseudo_output_blindings = compute_pseudo_output_blindings(rings, output_secrets, rng)?;
 
         // Create Range proofs for outputs and pseudo-outputs.
-        let pseudo_output_values_and_blindings: Vec<(u64, Scalar)> = rings
-            .iter()
-            .zip(pseudo_output_blindings.iter())
-            .map(|(ring, blinding)| (ring.amount().value, *blinding))
-            .collect();
+        let pseudo_output_values_and_blindings: Vec<(u64, Scalar)> =
+            zip_exact(rings.iter(), pseudo_output_blindings.iter())?
+                .map(|(ring, blinding)| (ring.amount().value, *blinding))
+                .collect();
 
         // Create a pedersen generator cache
         let mut generator_cache = GeneratorCache::default();
@@ -369,13 +375,13 @@ impl SigningData {
 
         // Compute the mlsag_signing_digest in whatever way is appropriate for this
         // block version.
-        let mlsag_signing_digest = compute_mlsag_signing_digest(
+        let (mlsag_signing_digest, tx_summary, extended_message) = compute_mlsag_signing_digest(
             block_version,
             tx_prefix,
             &pseudo_output_commitments,
             &range_proof,
             &range_proofs,
-        );
+        )?;
 
         let mut pseudo_output_token_ids: Vec<u64> =
             rings.iter().map(|ring| *ring.amount().token_id).collect();
@@ -394,15 +400,19 @@ impl SigningData {
             assert!(!range_proofs.is_empty());
         }
 
-        Ok(Self {
-            mlsag_signing_digest,
-            pseudo_output_blindings,
-            pseudo_output_commitments,
-            range_proof_bytes: range_proof,
-            range_proofs,
-            pseudo_output_token_ids,
-            output_token_ids,
-        })
+        Ok((
+            Self {
+                mlsag_signing_digest,
+                pseudo_output_blindings,
+                pseudo_output_commitments,
+                range_proof_bytes: range_proof,
+                range_proofs,
+                pseudo_output_token_ids,
+                output_token_ids,
+            },
+            tx_summary,
+            extended_message,
+        ))
     }
 }
 
@@ -716,13 +726,14 @@ impl SignatureRctBulletproofs {
         }
 
         // Extend the message with the range proof and pseudo_output_commitments.
-        let mlsag_signing_digest = compute_mlsag_signing_digest(
-            block_version,
-            tx_prefix,
-            &self.pseudo_output_commitments,
-            &self.range_proof_bytes,
-            &self.range_proofs,
-        );
+        let (mlsag_signing_digest, _tx_summary, _extended_message_digest) =
+            compute_mlsag_signing_digest(
+                block_version,
+                tx_prefix,
+                &self.pseudo_output_commitments,
+                &self.range_proof_bytes,
+                &self.range_proofs,
+            )?;
 
         // Each MLSAG must be valid.
         for (i, ring) in rings.iter().enumerate() {
@@ -775,16 +786,20 @@ fn sign_with_balance_check<CSPRNG: RngCore + CryptoRng, S: RingSigner + ?Sized>(
     signer: &S,
     rng: &mut CSPRNG,
 ) -> Result<SignatureRctBulletproofs, Error> {
-    let SigningData {
-        mlsag_signing_digest,
-        pseudo_output_blindings,
-        pseudo_output_commitments,
-        range_proofs,
-        range_proof_bytes,
-        pseudo_output_token_ids,
-        output_token_ids,
-        ..
-    } = SigningData::new(
+    let (
+        SigningData {
+            mlsag_signing_digest,
+            pseudo_output_blindings,
+            pseudo_output_commitments,
+            range_proofs,
+            range_proof_bytes,
+            pseudo_output_token_ids,
+            output_token_ids,
+            ..
+        },
+        _tx_summary,
+        _extended_message,
+    ) = SigningData::new(
         block_version,
         tx_prefix,
         rings,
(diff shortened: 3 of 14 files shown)


garbageslam:
There is now an MCIP for this, we should bring this code into alignment with the MCIP spec: https://github.com/mobilecoinfoundation/mcips/pull/52

transaction/core/src/input_rules.rs
```diff
@@ -18,13 +18,14 @@ use displaydoc::Display;
...
 use serde::{Deserialize, Serialize};
+use zeroize::Zeroize;
 
 /// A representation of rules on a transaction, imposed by the signer of some
 /// input in the transaction.
 ///
 /// Any rule could conceivably be added here if it can be evaluated against a
 /// `Tx`.
-#[derive(Clone, Digestible, PartialEq, Eq, Message, Serialize, Deserialize)]
+#[derive(Clone, Deserialize, Digestible, Eq, Hash, Message, PartialEq, Serialize, Zeroize)]
```
remoun, line 28: Consider adding `ZeroizeOnDrop` along with `Zeroize` (throughout this PR)
```suggestion
#[derive(Clone, Deserialize, Digestible, Eq, Hash, Message, PartialEq, Serialize, Zeroize, ZeroizeOnDrop)]
```
(requires `use`)

transaction/core/src/ring_ct/rct_bulletproofs.rs
```diff
@@ -148,15 +162,22 @@ impl SigningData {
...
     ///   equals value of outputs.
     /// * `rng` - randomness
+    ///
+    /// Returns:
+    /// * The SigningData
+    /// * The TxSummary
+    /// * The extended message digest
+    ///
+    /// (The latter two are MCIP #52 and only relevant on small HW wallets)
     pub fn new<CSPRNG: RngCore + CryptoRng>(
```
remoun, line 172: Optional/Equivalent: Rather than changing the return type of `new` and all its callers, add a `new_with_summary` that returns all 3, and update `new` to `new_with_summary(...).0`

transaction/types/src/block_version.rs
```diff
@@ -13,9 +13,22 @@ use serde::{Deserialize, Serialize};
...
-/// block version numbers.
+/// If you need to manipulate block versions that come from the network, you
+/// should use u32 to represent that.
+///
+/// Then, if you need to e.g. sign a transaction,
+/// you should try to convert them to BlockVersion.
+/// If that conversion fails, it means that this set of rules
+/// is not understood by your version of `mc-transaction-core`.
+/// This means that your build has reached end-of-life, and the user needs to
+/// update.
```
remoun, line 24: Nit: Mixing "you" and "the user"
```suggestion
/// This means that your build has reached end-of-life, and needs an update.
```

transaction/core/src/tx_summary.rs
```diff
@@ -0,0 +1,187 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
+//! The TxSummary is meant to reduce the complexity of implementing a hardware
+//! wallet. (See MCIP 52 also.)
```
remoun, line 4: Nit: "also" seems premature here
```suggestion
//! wallet. (See MCIP 52)
```

transaction/core/src/tx_summary.rs
```diff
@@ -0,0 +1,187 @@
...
+//! are interested is part of the root hash.
+//!
+//! So, we:
+//! * Take the existing extended message digest (32 bytes)
+//! * Use that to start a new Merlin transcript
+//! * Martial the TxSummary into this new Merlin transcript
+//! * A 32 byte digest resulting from that is new "extended message with tx
+//!   summary digest", and this is what the MLSAG actually sign.
+//!
+//! We propose that the TxSummary contains:
```
remoun, line 40: Nit: "We propose" seems out of place in a code comment
```suggestion
//! The TxSummary contains:
```

transaction/core/src/tx_summary.rs
```diff
@@ -0,0 +1,187 @@
...
+//! * For each output in TxPrefix.outputs, the public key, target key, and
+//!   masked amount.
+//! * The list of pseudo output commitments.
+//!
+//! These data can be produced easily during Tx validation without major
+//! changes. They don't contain any secrets that the enclave isn't supposed to
```
remoun, line 46: What enclave?

transaction/core/src/tx_summary.rs
```diff
@@ -0,0 +1,187 @@
...
+//! have.
+//!
+//! For the hardware wallet, what we can do is:
+//! * When we want to start signing MLSAGs, we send the 32 byte extended message
+//!   digest, and the TxSummary. The device can compute the "extended message
+//!   with tx summary digest" and sign MLSAGs over that.
+//! * For each output in the summary, the device expects the computer to
```
remoun, line 53: What computer?
```suggestion
//! * For each output in the summary, the device expects the host computer to
```

transaction/core/src/tx_summary.rs
```diff
@@ -0,0 +1,187 @@
...
+//!   actually sign, the TxSummaryUnblindingData can include the amount and
+//!   blinding factor. The computer actually has this for all of the pseudo
+//!   output commitments anyways.
+//!
+//! At this point, the device would then know the amount (value and token id) of
+//! every input in the Tx, and the amount and recipient of every output in the
+//! Tx. So it could completely account for the transfer of value caused by the
+//! Tx, it could identify change outputs, and it could display b58 encodings of
+//! the public addresses for outbound transfers. This should only require
+//! sending a few kb in the worst case.
```
remoun, line 77:
```suggestion
//! sending a few KB in the worst case.
```

transaction/core/src/tx_summary.rs
```diff
@@ -0,0 +1,187 @@
...
+    CompressedCommitment, InputRules, MaskedAmount,
+};
+use alloc::vec::Vec;
+use mc_crypto_digestible::Digestible;
+use mc_crypto_keys::CompressedRistrettoPublic;
+use prost::Message;
+use serde::{Deserialize, Serialize};
+use zeroize::Zeroize;
+
+/// A subset of the data in a Tx which enables efficient verification (by e.g. a
```
remoun, line 104:
```suggestion
/// A subset of the data in a Tx which enables efficient verification (e.g. by a
```

transaction/core/src/tx_summary.rs
```diff
@@ -0,0 +1,187 @@
...
+
+/// A subset of the data of a TxOut.
+///
+/// Fog hint and memo are omitted to reduce size and complexity on HW device,
+/// which can't really do much with those and isn't very interested in them
+/// anyways.
+#[derive(Clone, Deserialize, Digestible, Eq, Hash, Message, PartialEq, Serialize, Zeroize)]
+pub struct TxOutSummary {
+    /// The amount being sent.
+    #[prost(oneof = "MaskedAmount", tags = "1, 6")]
```
remoun, line 151: Why not use `1, 2`?

transaction/core/src/tx_summary.rs
```diff
@@ -0,0 +1,187 @@
...
+    #[digestible(name = "amount")]
```
remoun, line 152: What does this gain us? Why not `masked_amount`?

garbageslam: 👍

transaction/core/src/tx_summary.rs
garbageslam: 👍

garbageslam:
Still to do:
- [ ] Update spec for the InputRules stuff per discussion
- [ ] Implement the unblinding data
- [ ] Implement the streaming support for TxSummary + TxSummaryUnblindingData verification

transaction/core/src/input_rules.rs
garbageslam: Let's make a ticket to discuss this more, and do all of the `transaction-core` stuff using `zeroize(drop)`, if we want this. In this PR I would rather just follow existing patterns.

There are also tradeoffs here, because `Drop` types cannot be `Copy`.

I think there's may also be a world where objects like `Tx` are zeroize on drop, but lower level objects like `TxOut` are just zeroize. Not sure, needs more investigation.

None of this was possible until recently because `RistrettoPublic` did not implement Zeroize.

garbageslam: This is now https://github.com/mobilecoinfoundation/mobilecoin/issues/2717

transaction/core/src/tx_summary.rs
```diff
@@ -0,0 +1,210 @@
...
+
+/// A subset of the data of a TxOut.
+///
+/// Fog hint and memo are omitted to reduce size and complexity on HW device,
+/// which can't really do much with those and isn't very interested in them
+/// anyways.
+#[derive(Clone, Deserialize, Digestible, Eq, Hash, Message, PartialEq, Serialize, Zeroize)]
+pub struct TxOutSummary {
+    /// The amount being sent.
+    #[prost(oneof = "MaskedAmount", tags = "1, 2")]
```
garbageslam, line 181: this actually has to be 1,6 because i'm using the Enum from `transaction-core` and these tags are also duplicated in that enum.

transaction/core/src/tx_summary.rs
```diff
@@ -0,0 +1,210 @@
...
+    #[prost(message, required, tag = "3")]
+    pub target_key: CompressedRistrettoPublic,
+
+    /// The per output tx public key
+    #[prost(message, required, tag = "4")]
+    pub public_key: CompressedRistrettoPublic,
+
+    /// Whether or not this output is associated to an input with rules
+    #[prost(bool, tag = "5")]
+    pub associated_to_input_rules: bool,
```
nick-mobilecoin, line 194: :question: Why is the input rules flag here?  If the client wants to ensure input rules shouldn't it know the rules themselves?

Maybe hw wallets won't used with SCI, but if the customer wants to swap 100mob and get back 100eusd, the application could change this to swap 100mob for 10eusd and nothing would say differently here.

transaction/core/src/tx_summary.rs
```diff
@@ -0,0 +1,210 @@
...
+/// omitting the Ring and the proofs of membership.
+#[derive(Clone, Deserialize, Digestible, Eq, Hash, Message, PartialEq, Serialize, Zeroize)]
+pub struct TxInSummary {
+    /// Commitment of value equal to the real input.
+    #[prost(message, required, tag = "1")]
+    pub pseudo_output_commitment: CompressedCommitment,
+
+    /// Whether there are input rules associated to this input
+    #[prost(bool, tag = "2")]
+    pub has_input_rules: bool,
```
nick-mobilecoin, line 209: :question: Similar to above what value does this flag provide in the summary?

nick-mobilecoin (approved):

transaction/core/src/ring_ct/rct_bulletproofs.rs
garbageslam: I think it is consistent with a lot of our other implementations, e.g. `ConsensusServiceEnclave`. but it's a pretty minor change, i'm okay to do this if you want.

to some extent I want to smoke out the implementations and make sure they do something with these extra things, at least in debug mode. so that's a good reason to keep it this way for the moment.

transaction/core/src/tx_summary.rs
garbageslam: can't easily explain right now, but the goal is basically to reduce the bandwidth to the device and not send all of the input rules, which may be quite big. The device doesn't actually need to see the rules to meet its requirements, the only reason it needs this flag is so that it can know that it shouldn't expect to see a public address for this output.

transaction/core/src/tx_summary.rs
garbageslam: it tells the device that this input does not come from us (the signer), but rather from some counterparty, so it should not be reflected as deducting from our balance, but rather someone else's.

transaction/core/src/ring_ct/rct_bulletproofs.rs
remoun: Ah OK, I thought the other impls were intentionally not using those bits. It sounds like you're planning on utilizing the returned values elsewhere, so feel free to ignore the suggestion

transaction/core/src/ring_ct/rct_bulletproofs.rs
```diff
@@ -877,7 +906,62 @@ fn compute_pseudo_output_blindings<CSPRNG: RngCore + CryptoRng>(
...
+/// TODO: When support for block version < 2 is deprecated,
+/// we can make this return `[u8; 32]` instead of Vec<u8>, which is nicer for
+/// the hardware wallet implementation.
+fn compute_mlsag_signing_digest(
+    block_version: BlockVersion,
+    tx_prefix: &TxPrefix,
+    pseudo_output_commitments: &[CompressedCommitment],
+    range_proof_bytes: &[u8],
+    range_proofs: &[Vec<u8>],
+) -> Result<(Vec<u8>, TxSummary, Vec<u8>), ZipExactError> {
```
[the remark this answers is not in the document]
garbageslam: I am planning to change these to `[u8; 32]`... But point taken, we should probably introduce types for these

transaction/core/src/tx_summary.rs
nick-mobilecoin: I'm sure I'm missing some of the bigger picture, and I'm not fully aware of the threat model (let alone lacking experience to point out weaknesses reliably)
Let us say the application builds up a Tx:
1. 10mob this client that is willing to trade (no input rules)
2. 10 eusd for the trade (input rules from partial fill originator)
3. 200mob (input rules) but this is coming from this client's wallet  

Could 3 happen?  The application is attempting to transfer 200mob from the client to some address, with some arbitrary input rule that can be met.  Would the HW wallet hide this 200mob transaction from the client.

It seems like input rules are designed to evolve over time.  If this is the case, keying off `has_input_rules` may quickly turn into a misnomer.  What if someone decides to add a date/time expirey via input rules.  Then the input rules could be for this client and not a counterparty.

transaction/core/src/tx_summary.rs
nick-mobilecoin: See comment below, regarding what is the threat model.
I'm assuming that the HW wallet is the trusted device and that the application it's talking to isn't trustworthy.
If the application is fully trustworthy then it's probably ok, but if not, I don't have enough understanding to say one way or another, but it feels like this is leaving out sufficient information for the HW wallet.

transaction/core/src/tx_summary.rs
garbageslam: these are great questions -- lets try to move the discussion to MCIP 52, i still haven't actually written out what I'm trying to do and what my assumptions are. it's possible I've messed up, but let me take some time to write why I think this rev is okay

transaction/core/src/ring_ct/rct_bulletproofs.rs
garbageslam: I think you guys are right and new_with_summary is better (having now actually developed the next PR).
Gonna follow that route now for this change

transaction/core/src/input_rules.rs
```diff
@@ -68,6 +70,21 @@ pub struct InputRules {
 }
 
 impl InputRules {
+    /// Get all tx target keys associated to outputs appearing in these rules
+    pub fn associated_tx_target_keys(&self) -> BTreeSet<CompressedRistrettoPublic> {
```
garbageslam, line 74: This should probably just get the set of all outputs appearing in the rules instead

transaction/core/src/ring_ct/rct_bulletproofs.rs
```diff
@@ -753,26 +778,30 @@ impl SignatureRctBulletproofs {
 /// * `rng` - randomness
 fn sign_with_balance_check<CSPRNG: RngCore + CryptoRng, S: RingSigner + ?Sized>(
     block_version: BlockVersion,
-    message: &[u8; 32],
+    tx_prefix: &TxPrefix,
```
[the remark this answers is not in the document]
garbageslam: this function is actually being deleted in #2725 which is staged after this, so i'm going to leave it in this PR

(it's actually getting moved somewhere else, and I rewrote the docs there)

garbageslam:
Hmm, i think I had a bad rebase, will fix

transaction/core/src/tx_summary.rs
```diff
@@ -0,0 +1,220 @@
...
+//!
+//! It's reasonable to ask, what if the host computer lies to the device in the
+//! TxSummary. What if, the TxPrefix actually says one thing, and the TxSummary
+//! it gives to the device says another. The device cannot detect this if it
+//! doesn't have insight into the extended message digest. However, in this
+//! case, the transaction will simply be invalid when it is submitted, because
+//! the consensus enclave will derive a different value for the TxSummary when
+//! it attempts to validate the transaction, and that value will actually be
+//! based on the TxPrefix. So the host computer can gain no advantage by lying
+//! to the device in this way.
```
isis-mc, line 87: So this possibly constitutes a signing oracle in a way that could rely a bit more than would be ideal on the collision resistance of the transcript? I think it would depend on the implementation details of the hardware wallet, e.g. if you can just "plug it in and boop the 'OK' button" to sign, or if there is some PIN, etc., that is required. To be clear, this isn't broken or necessarily bad as-in, it's mostly just that I'd like to keep on the backburner to brainstorm potential ways to do it better.

transaction/core/src/tx_summary.rs
isis-mc: Did this get updated?

isis-mc (approved):
LGTM, small nits detailed above.

transaction/core/src/ring_ct/signing_digest.rs
```diff
@@ -0,0 +1,152 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
```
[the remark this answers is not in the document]
garbageslam: We have a commit hook that checks for copyright statements and it has 2018-2022 as the target. At some point James and I asked Shoshana if its important to keep tge original date on every file and she didn't seem to think so, we can update it later and this is only a matter of 3 years on like a 75 year copyright, so the commit hook just puts 2018-2022 on everything

transaction/core/src/ring_ct/rct_bulletproofs.rs
```diff
@@ -16,7 +16,7 @@ use curve25519_dalek::{
     traits::Identity,
 };
 use mc_common::HashSet;
-use mc_crypto_digestible::{DigestTranscript, Digestible, MerlinTranscript};
+use mc_crypto_digestible::Digestible;
```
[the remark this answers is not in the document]
garbageslam: I just tried to move the merlin stuff to a separate module since it's getting larger and more complicated

transaction/core/src/revealed_tx_out.rs
```diff
@@ -7,10 +7,11 @@ use mc_crypto_digestible::Digestible;
 use mc_crypto_ring_signature::Scalar;
 use prost::Message;
 use serde::{Deserialize, Serialize};
+use zeroize::Zeroize;
 
 /// A TxOut together with its amount shared secret, which can be used to reveal
 /// the amount and token id and check them against the commitment data
-#[derive(Clone, Digestible, PartialEq, Eq, Message, Serialize, Deserialize)]
+#[derive(Clone, Deserialize, Digestible, Eq, Hash, Message, PartialEq, Serialize, Zeroize)]
```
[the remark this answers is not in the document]
garbageslam: It was always Digestible, but it wasn't Hash or Zeroize before. I am not particularly sure why it wasn't zeroize, and it wasn't Hash because it wasn't needed before.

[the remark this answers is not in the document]
garbageslam: I think it's because I added `Zeroize` to `InputRules`

transaction/core/src/ring_ct/signing_digest.rs
```diff
@@ -0,0 +1,152 @@
...
+/// version.
+///
+/// Returns:
+/// * The MLSAG Signing Digest
+/// * The TxSummary
+/// * The extended_message_digest.
+///
+/// TODO: When support for block version < 2 is deprecated,
+/// we can make this return `[u8; 32]` instead of Vec<u8>, which is nicer for
+/// the hardware wallet implementation.
```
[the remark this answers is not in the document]
garbageslam: We already have block version 2 in prod -- we can deprecate building block version < 2 at any time, and we talked about this earlier in a release planning meeting and people seemed on board.

transaction/core/src/ring_ct/signing_digest.rs
```diff
@@ -0,0 +1,152 @@
...
+pub fn compute_mlsag_signing_digest(
+    block_version: BlockVersion,
+    tx_prefix: &TxPrefix,
+    pseudo_output_commitments: &[CompressedCommitment],
+    range_proof_bytes: &[u8],
+    range_proofs: &[Vec<u8>],
```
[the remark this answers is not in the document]
garbageslam: There's some docu in the other file, probably should document this one also though.

The issue is like:

In block version 2, there is only one range proof that covers all of the pseudo-output and output commitments. This works because in block version 2, there are no mixed transactions, so they all have the same token id and the same Pedersen generators. (in this version, range_proofs is empty)

In block version 3, mixed transactions are allowed, but the bulletproofs implementation doesn't support mixing of Pedersen generators. So we have to have a range proof per token id in the transaction. (in this version, range_proof_bytes is empty)

This function works with either block version, but after block version 3 is in production, range_proof_bytes can go away.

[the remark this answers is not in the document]
garbageslam: If the upstream bulletproofs supported mixing of Pedersen generators then we wouldn't have to do this and could just have one range proof always.

transaction/core/src/ring_ct/signing_digest.rs
```diff
@@ -0,0 +1,152 @@
...
+        digest_extended_message(
+            message,
+            pseudo_output_commitments,
+            range_proof_bytes,
+            range_proofs,
+        )
+        .to_vec()
+    } else {
+        // Old-style extended message
+        extend_message(message, pseudo_output_commitments, range_proof_bytes)
```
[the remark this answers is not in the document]
garbageslam: yup

transaction/core/src/ring_ct/signing_digest.rs
```diff
@@ -0,0 +1,152 @@
...
+    })
+}
+
+/// Computes a merlin digest of message, pseudo_output_commitments, range proof
```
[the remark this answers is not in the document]
garbageslam: i think its just an oversight

transaction/core/src/ring_ct/rct_bulletproofs.rs
```diff
@@ -104,8 +104,19 @@ pub struct SignedInputRing {
...
-    /// Transaction extended message.
-    pub extended_message_digest: Vec<u8>,
+    /// The bytes actually signed by MLSAG signatures.
+    /// This is different depending on what block version we are in.
+    /// * In the oldest block versions, this is a large number of bytes called
+    ///   the "extended message".
+    /// * In block version 2, this is instead 32 bytes called the "extended
+    ///   message digest".
+    /// * In block version 3, this is instead 32 bytes called the
+    ///   "extended-message-and-tx-summary digest".
```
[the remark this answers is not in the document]
garbageslam: That's not what i intended i guess, I'm intending that `mlsag_signing_digest` refers to "the thing that MLSAGs are supposed to sign, no matter what block version you are in", and "extended-message-and-tx-summary digest" is a block version 3 specific digest.

transaction/core/src/ring_ct/signing_digest.rs
```diff
@@ -0,0 +1,152 @@
...
+    message: &[u8],
+    pseudo_output_commitments: &[CompressedCommitment],
+    range_proof_bytes: &[u8],
+    range_proofs: &[Vec<u8>],
+) -> [u8; 32] {
+    let mut transcript = MerlinTranscript::new(EXTENDED_MESSAGE_DOMAIN_TAG.as_bytes());
+    message.append_to_transcript(b"message", &mut transcript);
+    pseudo_output_commitments.append_to_transcript(b"pseudo_output_commitments", &mut transcript);
+    range_proof_bytes.append_to_transcript_allow_omit(b"range_proof_bytes", &mut transcript);
+    range_proofs.append_to_transcript_allow_omit(b"range_proofs", &mut transcript);
```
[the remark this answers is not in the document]
garbageslam: The way it works is, one of them is empty whenever the other one is not empty. The first one is present in block version 2, and the second one is present in block version 3

transaction/core/src/tx_summary.rs
```diff
@@ -0,0 +1,220 @@
...
+//!   ADDITONALLY supply this subset of the arguments to TxOut::new:
+//!   * block_version,
+//!   * amount,
+//!   * recipient,
+//!   * tx_private_key,
+//! * These are not part of the TxSummary (because the consensus enclave can't
+//!   know them) but can be part of TxSummaryUnblindingData or some such thing.
+//! * For a given public key, target key, and masked amount, it is intractable
+//!   to find a different amount, recipient, and tx private key that leads to
+//!   the same data (discrete log hard). So the device can be convinced that
```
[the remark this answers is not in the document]
garbageslam: This part isn't about the `TxSummary`, it's basically about how hard is it to find a collision in `TxOut::new`. I think it's discrete log, it reduces to solving equations nontrivially when multiplying recipient public keys by the tx private key.

transaction/core/src/tx_summary.rs
garbageslam: Yeah, I tried to think about this a bit i the MCIP, for me the issue is like:

* If the goal is that we don't send the whole Tx to the hardware wallet (to reduce bandwidth), then we have to accept that we might be signing something that isn't well-formed
* We could imagine restructuring the MLSAG signing digest so that it's like, first we digest the range proofs, merkle proofs, etc., then we digest the TxPrefix (which is reorganized to look more like the TxSummary), and that's the whole digest and there's no third step. But I'm not sure if this gets you away from the issue of, if there is a collision in merlin then you can be tricked about the first part of what you are signing (which was abbreviated to a hash)

Maybe I'm oversimplifying, but that's my justification anyways for why this state is okay. But it probably simplifies the situation if we move towards replacing the TxPrefix with the TxSummary in the Tx

transaction/core/src/tx_summary.rs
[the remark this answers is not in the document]
garbageslam: Happy to discuss offline

transaction/core/src/ring_ct/signing_digest.rs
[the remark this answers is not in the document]
garbageslam: added some docu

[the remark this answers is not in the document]
garbageslam: the `_allow_omit` part means that if the sequence is empty, it can just be skipped from the transcript

PR #2695, opened 2022-10-11
Add `--check` option to rust lint.sh script

nick-mobilecoin:
Previously the lint.sh script would always check. Now the script will
attempt to fix and will only check if the `--check` flag is provided.

### Future Work

[Soundtrack of this PR]()

### DIFF
--- a/tools/lint.sh
+++ b/tools/lint.sh
@@ -4,9 +4,32 @@
 
 set -e  # exit on error
 
-if [[ ! -z "$1" ]]; then
-    cd "$1"
-fi
+CHECK=""
+
+while [ "$1" != "" ]; do
+	case "$1" in
+		--check)
+		  CHECK="--check";;
+		-h | --help)
+			echo "Lints rust files in the repo"
+                        echo "Linters that can fix will fix the files, by default."
+                        echo "Fixing linters include 'cargo sort' and 'cargo fmt'."
+			echo ""
+			echo "Usage: $(basename $0) [--check]"
+			echo ""
+			echo "Args:"
+			echo "        --check  Limit linters that can fix to only check"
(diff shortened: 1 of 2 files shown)


nick-mobilecoin:
* master
  * **PR #2695** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2695" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2695?utm_source=stack-comment).

tools/lint.sh
```diff
@@ -4,9 +4,32 @@
 
 set -e  # exit on error
 
-if [[ ! -z "$1" ]]; then
-    cd "$1"
-fi
```
nick-mobilecoin, line 9: I didn't see any place that the changing of directory was being used so omitted it from the command line parsing.

tools/lint.sh
```diff
@@ -4,9 +4,32 @@
...
+CHECK=""
+
+while [ "$1" != "" ]; do
+	case "$1" in
+		--check)
+		  CHECK="--check";;
```
garbageslam, line 12: Maybe this is meant to be indented 8 spaces? Not sure

garbageslam (approved):
LGTM thank you!

tools/lint.sh
nick-mobilecoin: Looks like a copy pasta fail.  Those are tab characters that got pulled in. 
I normalized the indent to 2 spaces as that's what was there previously with the for loop at the bottom.

jcape (approved):

PR #2703, opened 2022-10-11
Use slices in fog view enclave impl methods

samdealy:
### Future Work

[Soundtrack of this PR]()

PR #2694, opened 2022-10-11
Implement debug for Fog View Router client error

samdealy:
### Motivation
The FVR integration tests unwraps a result with the `mc_fog_view_connection::Error`, and this requires that it implement `Debug`.

PR #2692, opened 2022-10-10
Close gRPC channel when client is dropped

samdealy:
### Motivation
As described in the grpcio [docs](https://docs.rs/grpcio/latest/grpcio/type.ClientDuplexSender.html), a `ClientDuplexSender` needs to be closed before it's dropped. In our case this means we need to close the `request_sender` field before we drop the `FogViewRouterGrpcClient`.

We weren't doing this before and it outputted and error log to the message that said `ERROR: Remote stopped...`. After implementing this call to close upon dropping, the error message is no longer outputted.

PR #2672, opened 2022-10-07
Fix reencrypt

samdealy:
### Motivation
Previously, the `reencrypt_sealed_message_for_backends` method used the sealed client message's channel id in the `EnclaveMessages` sent to the stores from the router.  This is incorrect- the channel id for the specific encryptor should be used because it corresponds to the nonce session between the router and the store.

PR #2711, opened 2022-10-12
Update block tracker to use sharding strategy

samdealy:
### Motivation
This PR makes the `BlockTracker` aware of the store's `ShardingStrategy`. This change is cosmetic in that it makes sharding logic cleaner- the sharding strategy now exists in the `BlockTracker`, which is a struct responsible for tracking which blocks have been processed.

### DIFF
--- a/fog/view/server/src/db_fetcher.rs
+++ b/fog/view/server/src/db_fetcher.rs
@@ -179,10 +179,9 @@ where
     db: DB,
     stop_requested: Arc<AtomicBool>,
     shared_state: Arc<Mutex<DbFetcherSharedState>>,
-    block_tracker: BlockTracker,
+    block_tracker: BlockTracker<SS>,
     num_queued_records_limiter: Arc<(Mutex<usize>, Condvar)>,
     readiness_indicator: ReadinessIndicator,
-    sharding_strategy: SS,
     logger: Logger,
 }
 
@@ -206,10 +205,9 @@ where
             db,
             stop_requested,
             shared_state,
-            block_tracker: BlockTracker::new(logger.clone()),
+            block_tracker: BlockTracker::new(logger.clone(), sharding_strategy),
             num_queued_records_limiter,
             readiness_indicator,
-            sharding_strategy,
             logger,
         };
         thread.run();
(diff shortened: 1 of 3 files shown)


samdealy:
* master
      * **PR #2711** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2711" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2711?utm_source=stack-comment).

fog/view/server/src/db_fetcher.rs
```diff
@@ -200,13 +202,14 @@ where
...
         readiness_indicator: ReadinessIndicator,
         sharding_strategy: SS,
+        block_tracker: Arc<Mutex<BlockTracker>>,
         logger: Logger,
     ) {
         let thread = Self {
             db,
             stop_requested,
             shared_state,
-            block_tracker: BlockTracker::new(logger.clone()),
```
garbageslam, line 209: I am skeptical of the idea to fuse these two block trackers, there are two block trackers because they are tracking different things, and merging them seems likely to lead to bugs.

One block tracker is tracking which blocks we have fetched from the database, and which blocks we still need to fetch.

The other block tracker is tracking which blocks we have actually loaded into the enclave.

If we lose this distinction, bad things happen.
If a block is fetched from DB and not yet loaded into enclave, and there is only one block tracker, then either this block tracker thinks this block is processed, or it isn't.

* If the block tracker doesn't think it is processed, then the DB fetcher will decide that we need to fetch it again, so now we can fetch blocks over and over again. This is bad.
* If the block tracker thinks it is processed, even though it's not in the enclave yet, then when we have to report the "highest_processed_block_count" to the user, the number we report will be wrong, and the user can computer an incorrect balance, because we will have broken the fog view protocol.

So I don't see right now there is any way that merging the block trackers can work. But, I'm willing to be wrong.

nick-mobilecoin (approved):

fog/view/server/src/db_fetcher.rs
nick-mobilecoin: Just to double check this is comment is relating to an earlier version that fully removed the `block_tracker` from `DbFetcherThread`?  Looking at the current change it seems that there are still 2 trackers.

samdealy: Yes, this is an outdated comment. Graphite chose to link the new changes to this outdated PR.

jcape (approved):

PR #2722, opened 2022-10-17
Implement new highest processed block count logic

samdealy:
### Motivation
Now that the stores are returning their highest processed block counts in addition the block ranges their responsible for, the router will calculate the highest processed block count for the client by finding the lowest store highest processed block count at which a store has not finished processing its block range. If all block ranges have been processed, it returns the max highest processed block count.

### DIFF
--- a/fog/view/enclave/impl/src/lib.rs
+++ b/fog/view/enclave/impl/src/lib.rs
@@ -364,7 +364,7 @@ where
     fn get_minimum_highest_processed_block_count(
         mut responses: Vec<DecryptedMultiViewStoreQueryResponse>,
     ) -> u64 {
-        responses.sort_by(|a, b| a.block_range.start_block.cmp(&b.block_range.start_block));
+        responses.sort_unstable_by_key(|response| response.block_range.start_block);
 
         // Find the first time in which a highest processed block count does not equate
         // to the final block that the shard is responsible for.
--- a/fog/view/enclave/impl/src/lib.rs
+++ b/fog/view/enclave/impl/src/lib.rs
@@ -376,12 +376,12 @@ where
                 max_highest_processed_block_count = response_highest_processed_block_count;
             }
 
-            // In this case, the shard hasn't processed all the blocks it's responsible for.
-            // And, as such, those blocks might not be processed. Let's return
-            // this number. TODO: Consider implementing logic that accounts for
-            // overapping block ranges. If ranges overlap, then the next server
-            // might have processed those blocks that this shard did not process
-            // (but is responsible for).
+            // In this case, the shard hasn't processed all the blocks it's responsible for,
+            // and, as such, those blocks might not be processed so we should return this
+            // number.
+            // TODO: Consider implementing logic that accounts for overlapping block ranges.
+            //   If ranges overlap, then the next server might have processed those blocks
+            //   that this shard did not process (but is responsible for).
             if response_highest_processed_block_count < response.block_range.end_block {
                 return max_highest_processed_block_count;
             }
--- a/fog/view/enclave/impl/src/lib.rs
+++ b/fog/view/enclave/impl/src/lib.rs
@@ -36,6 +36,16 @@ use mc_oblivious_traits::ORAMStorageCreator;
 use mc_sgx_compat::sync::Mutex;
 use mc_sgx_report_cache_api::{ReportableEnclave, Result as ReportableEnclaveResult};
 
+/// Helper struct that contains the decrypted `QueryResponse` and the
+/// `BlockRange` the shard is responsible for.
+#[derive(Clone)]
+struct DecryptedMultiViewStoreQueryResponse {
+    /// Decrypted `QueryResponse`
+    query_response: QueryResponse,
+    /// The `BlockRange` that the shard is meant to process.
+    block_range: BlockRange,
+}
+
 pub struct ViewEnclave<OSC>
 where
     OSC: ORAMStorageCreator<StorageDataSize, StorageMetaSize>,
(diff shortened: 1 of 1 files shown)


fog/view/enclave/impl/src/lib.rs
```diff
@@ -352,10 +360,31 @@ where
         )
     }
 
+    // Takes each MultiViewStoreResponseQuery
```
nick-mobilecoin, line 363: :thought_balloon: I would suggest creating a unit test for this. This is a nice isolated piece of logic.  

The `query_response` might be a bit hefty to create for each test...

fog/view/enclave/impl/src/lib.rs
```diff
@@ -352,10 +360,31 @@ where
...
     fn get_minimum_highest_processed_block_count(
-        &self,
-        _shard_query_responses: Vec<(QueryResponse, BlockRange)>,
+        responses: &[DecryptedMultiViewStoreQueryResponse],
     ) -> u64 {
-        todo!()
+        let responses_with_gaps = responses
```
nick-mobilecoin, line 367: :+1: Really digging the isolated logic of this function. It's easy to reason about in isolation without worrying about the other logic in creating the response

fog/view/enclave/impl/src/lib.rs
```diff
@@ -352,10 +360,31 @@ where
...
+            .iter()
+            .cloned()
+            .filter(|response| {
+                response.query_response.highest_processed_block_count
+                    != response.block_range.end_block
```
nick-mobilecoin, line 372: :wrench: Since `end_block` is exclusive should this be something like
```suggestion
                response.query_response.highest_processed_block_count + 1
                    != response.block_range.end_block
```

Looking around some tests, I don't think so, but I think there may need to be some better disambiguation in the descriptions (follow on PR/issue if agreed)

fog/view/enclave/impl/src/lib.rs
```diff
@@ -352,10 +360,31 @@ where
...
+                .min_by_key(|response| response.query_response.highest_processed_block_count)
+        } else {
+            responses
+                .iter()
+                .max_by_key(|response| response.query_response.highest_processed_block_count)
+        };
+
+        highest_processed_block_count.map_or(u64::MIN, |response| {
+            response.query_response.highest_processed_block_count
+        })
```
nick-mobilecoin, line 388: :wrench: The cloning and double iteration stood out as a code smell.  I will admit even authoring this suggestion it feels like it might be harder to grok
```suggestion
        let min_block_from_gaps = responses
            .iter()
            .filter(|response| {
                response.query_response.highest_processed_block_count
                    != response.block_range.end_block
            })
            .min_by_key(|response| response.query_response.highest_processed_block_count);

        let highest_processed_block_count = match min_block_from_gaps {
            Some(response) => Some(response),
            _ => responses
                .iter()
                .max_by_key(|response| response.query_response.highest_processed_block_count),
        };

        highest_processed_block_count.map_or(u64::MIN, |response| {
            response.query_response.highest_processed_block_count
        })
```

nick-mobilecoin (approved):

fog/view/enclave/impl/src/lib.rs
```diff
@@ -36,6 +36,16 @@ use mc_oblivious_traits::ORAMStorageCreator;
...
 use mc_sgx_report_cache_api::{ReportableEnclave, Result as ReportableEnclaveResult};
 
+/// Helper struct that contains the decrypted `QueryResponse` and the
+/// `BlockRange` the shard is responsible for.
+#[derive(Clone)]
+struct DecryptedMultiViewStoreQueryResponse {
+    /// Decrypted `QueryResponse`
+    pub query_response: QueryResponse,
+    /// The `BlockRange` that the shard is meant to process.
+    pub block_range: BlockRange,
```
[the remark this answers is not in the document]
samdealy: Done.

fog/view/enclave/impl/src/lib.rs
samdealy: I changed the logic that calculates this, but the underlying idea is the same. If the block_count is less than the end range block index (which is non inclusive), the store hasn't finished processing its blocks.

fog/view/enclave/impl/src/lib.rs
samdealy: I changed the logic here. So I'll resolve this.

fog/view/enclave/impl/src/lib.rs
samdealy: Do we have unit tests in enclaves?

nick-mobilecoin: blarg, I thought there is a way if it doesn't depend on the SGX SDK directly, but maybe it's only for support libs the enclaves rely on.
I can understand punting on it for this PR due to the possible complexity there :(.

nick-mobilecoin (approved):

fog/view/enclave/impl/src/lib.rs
samdealy: I just double checked with james, and you're right- since it doesn't depend on SGX SDK we can unit test it. I actually change this method slightly in a follow up PR so I'll add it there.

fog/view/enclave/impl/src/lib.rs
```diff
@@ -350,10 +360,33 @@ where
...
     }
 
+    // Takes each MultiViewStoreResponseQuery
     fn get_minimum_highest_processed_block_count(
-        &self,
-        _shard_query_responses: Vec<(QueryResponse, BlockRange)>,
+        mut responses: Vec<DecryptedMultiViewStoreQueryResponse>,
     ) -> u64 {
-        todo!()
+        responses.sort_by(|a, b| a.block_range.start_block.cmp(&b.block_range.start_block));
```
jcape, line 367: Maximum effort version:
```suggestion
        responses.sort_unstable_by_key(|resp| resp.block_range.start_block);
```

jcape (approved):
LGTM (comment is mostly just noting fancy sort options in rust)...

fog/view/enclave/impl/src/lib.rs

PR #2720, opened 2022-10-17
Add block range to MVQ response

samdealy:
### Motivation
The highest processed block range was being calculated incorrectly previously. The router's enclave was using the minimum highest processed block from all the stores as the highest processed block that it returned to the client. The stores calculated their highest processed block based on its enclave block tracker that only incremented its processed block count when it added a block to it's enclave. 

This resulted in unexpected behavior because each store returned the highest block that it added to its enclave RATHER than the highest block that it observed (the expected behavior). The new paradigm is as follows: each store will continue returning this number and will now additionally add it's block range. The router will then look at each block range, and find the first time a store hasn't processed all of it's block range. This will be returned to the client as the highest processed block range. 

This PR makes the small change of adding a new `block_range` field to the MVQ proto. The next PRs will implement the rest of the logic.

PR #2660, opened 2022-10-05
Fix fog conformance OOM failure by reducing `frontends` allocation

samdealy:
### Motivation
Previously, we allocated space for 10,000 `frontends` in `AkeEnclaveState`. This caused an OOM failure when the fog conformance tests ran. After tinkering around with the numbers, I found that an allocation of `1000` frontends works fine, and then I cut that number in half because we should never need more than 300 frontends given our FVR needs.

PR #2653, opened 2022-10-04
Implement nonce based auth for Fog View Store

samdealy:
### Motivation
This PR updates Fog View stores to implement nonce based auth that serves as the foundation for FVR <-> Fog View Store communication.

PR #2634, opened 2022-09-30
Feature/fog view router

samdealy:
### Future Work

[Soundtrack of this PR]()

PR #2726, opened 2022-10-18
Set block-related data more accurately

samdealy:
### Motivation
Previously we were only setting `highest_processed_block_count` by finding the first response that wasn't fully processed. This PR also sets `highest_processed_block_signature_timestamp` and `last_known_block_count` using this response. It also adds much needed unit test for the `get_block_data` method.

### DIFF
--- a/fog/view/enclave/impl/src/lib.rs
+++ b/fog/view/enclave/impl/src/lib.rs
@@ -2,7 +2,7 @@
 
 //! View Enclave Implementation
 
-#![no_std]
+#![cfg_attr(not(test), no_std)]
 
 extern crate alloc;
 
@@ -44,6 +44,40 @@ struct DecryptedMultiViewStoreQueryResponse {
     query_response: QueryResponse,
     /// The `BlockRange` that the shard is meant to process.
     block_range: BlockRange,
+}
+
+/// Helper struct that contains block data for the client `QueryResponse`
+#[derive(Clone)]
+struct BlockData {
+    /// The highest processed block count that will be returned to the client.
+    highest_processed_block_count: u64,
+    /// The timestamp for the highest processed block count
+    highest_processed_block_signature_timestamp: u64,
+    /// The last known block count that will be returned to the client.
+    last_known_block_count: u64,
+}
+
+impl BlockData {
+    fn new(
+        highest_processed_block_count: u64,
+        highest_processed_block_signature_timestamp: u64,
+        last_known_block_count: u64,
+    ) -> Self {
+        Self {
+            highest_processed_block_count,
+            highest_processed_block_signature_timestamp,
+            last_known_block_count,
+        }
+    }
+}
+impl Default for BlockData {
+    fn default() -> Self {
+        Self {
+            highest_processed_block_count: u64::MIN,
+            last_known_block_count: u64::MIN,
+            highest_processed_block_signature_timestamp: u64::MIN,
+        }
+    }
 }
 
 pub struct ViewEnclave<OSC>
@@ -339,8 +373,12 @@ where
 
         shard_query_response.tx_out_search_results =
             Self::get_collated_tx_out_search_results(client_query_request, &shard_query_responses)?;
+        let block_data = get_block_data(shard_query_responses);
         shard_query_response.highest_processed_block_count =
-            Self::get_minimum_highest_processed_block_count(shard_query_responses);
+            block_data.highest_processed_block_count;
+        shard_query_response.highest_processed_block_signature_timestamp =
+            block_data.highest_processed_block_signature_timestamp;
+        shard_query_response.last_known_block_count = block_data.last_known_block_count;
 
         Ok(shard_query_response)
     }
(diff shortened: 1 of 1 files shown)


samdealy:
* master
    * **PR #2720** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2720" target="_blank">Graphite</a> 
      * **PR #2711** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2711" target="_blank">Graphite</a> 
        * **PR #2721** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2721" target="_blank">Graphite</a> 
          * **PR #2722** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2722" target="_blank">Graphite</a> 
            * **PR #2726** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2726" target="_blank">Graphite</a>  👈
                * **PR #2712** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2712" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2726?utm_source=stack-comment).

fog/view/enclave/impl/src/lib.rs
```diff
@@ -46,6 +46,40 @@ struct DecryptedMultiViewStoreQueryResponse {
...
 
+/// Helper struct that contains block data for the client `QueryResponse`
+#[derive(Clone)]
+struct BlockData {
+    /// The highest processed block count that will be returned to the client.
+    highest_processed_block_count: u64,
+    /// The timestamp for the highest processed block count
+    highest_processed_block_signature_timestamp: u64,
+    /// The last known block count that will be returned to the client.
+    last_known_block_count: u64,
```
nick-mobilecoin, line 57: :question: I'm guessing this is coming from `UntrustedQueryResponse` which could clarify some
1. Is the timestamp meant to be generic or does it have units like milliseconds?
2. I'm guessing _last known_ means _highest_ that we know exists but we haven't been able to decode or whatever versus highest processed which is the _highest_ block that we've been able to muck with and can provide meaningful input on

fog/view/enclave/impl/src/lib.rs
```diff
@@ -359,34 +397,155 @@ where
...
+            first_query_response.highest_processed_block_count
+        );
+        assert_eq!(
+            result.highest_processed_block_signature_timestamp,
+            first_query_response.highest_processed_block_signature_timestamp
+        );
+        assert_eq!(
+            result.last_known_block_count,
+            first_query_response.last_known_block_count
+        );
```
nick-mobilecoin, line 549: :wrench: Might consider something like the suggestion below
It requires adding `yare` to the Cargo.toml as well as deriving `Debug` and `PartialEq` on the `BlockData`
When doing a loop with an early return one should test; 0, the first, a middle element, the last element, exiting the loop, the current tests are only looking at going all the way through the loop and a middle element.  Adding test cases for the other 2 will probably blow this file up.
I'm not fond of the tuple for the input parameters to the test function, but putting a type there would have put more noise in the array/slice.
Since the time and and last known block are just passthroughs, I used an offset to make sure they copied correctly.

:question:It does bring up a question about `last_known` should we find the highest of that one?  I'm guessing the `last_known` should be pretty close for most shards so probably won't matter.

```suggestion
    #[test]
    fn no_responses() {
        let decrypted_query_responses = vec![];
        assert_eq!(get_block_data(decrypted_query_responses), BlockData::new(0, 0, 0));
    }

    // The tuple is (block_start, block_end, highest_processed)
    #[parameterized(
    all = { &[(0, 4, 4), (5, 8, 8), (9, 11, 11)], 2 },
    first_partial = { &[(0, 4, 3), (5, 8, 8), (9, 11, 11)], 0 },
    middle_partial = { &[(0, 4, 4), (5, 8, 7), (9, 11, 11)], 1 },
    last_partial = { &[(0, 4, 4), (5, 8, 8), (9, 11, 10)], 2 },
    )]
    fn block_data(block_data: &[(u64, u64, u64)], expected_index: usize) {
        let time_shift = 3;
        let known_shift = 2;
        let mut decrypted_query_responses = Vec::with_capacity(block_data.len());
        decrypted_query_responses.extend(block_data.into_iter().map(|b| {
            let block_range = BlockRange::new(b.0, b.1);
            let query_response = create_query_response(b.2, b.2 + time_shift, b.2 + known_shift);
            DecryptedMultiViewStoreQueryResponse {
                query_response,
                block_range,
            }
        }));

        let data = block_data[expected_index];
        let expected_block = BlockData::new(data.2, data.2 + time_shift, data.2 + known_shift);
        assert_eq!(get_block_data(decrypted_query_responses), expected_block);
```

nick-mobilecoin (approved):

fog/view/enclave/impl/src/lib.rs
samdealy: 1. Yea it's coming from UntrustedQueryResponse. 
2. Yes.

fog/view/enclave/impl/src/lib.rs
```diff
@@ -2,7 +2,7 @@
 
 //! View Enclave Implementation
 
-#![no_std]
+#![cfg_attr(not(test), no_std)]
```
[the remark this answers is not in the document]
samdealy: Done.

fog/view/enclave/impl/src/lib.rs
samdealy: While the parametrized version is more concise, I find it unreadable. The the tuple arguments to the `#[parametrized]]` block and devices such as `time_shift` and `known_shift` are hard to follow. I tend to prefer keeping test set up declarative so that you don't have to do much mental gymnastics to see what's going on. 

I am ok with adding those additional test cases, but I think that every case is covered (early return and non-early return). Not sure we have this 0, first ,middle, and last requirement for unit tests elsewhere, but if you feel strongly I will add them.

samdealy: And I think it makes sense to return the highest `last_known_block_index` so I  changed that.

nick-mobilecoin: - parameterized vs not. I can understand the concerns especially in this instance. I didn't like the array of tuples but couldn't think of a clearer approach. I would encourage keeping the tool availability in ones mind as I feel there are is a large amount of workflow tests in the codebase. I think having people see parameterized tests in the actual codebase will make others consider it over some of the workflow style tests.

. "every case covered"  This is one of the gotcha's of code coverage.  Code coverage is  negative metric, meaning if you don't have it, you know you didn't meet some metric so you know you for sure didn't meet the higher metric of "sufficent" tests.  Having code coverage means that you're somewhere between the code coverage bar and "sufficient" tests bar, but you don't know where.  I quoted sufficient as it is very subjective and there are strong debates about code coverage vs not.
Considering testing coverage, for searching an iterable, one should have:
   - empty iterable, nothing found, this ensures the default value is correct. What if the logic maxed out and had some kind of state of initialized with item from iterable.
   - only one item, this is the boundary for 0 items.  It also ensures someone doesn't have a stray next() call
   - first item, this ensures one doesn't return prior to copying the item value.  Could probably be omitted with a one item test.
   - middle item, this ensures subsequent items don't overwrite it
   - last item, this ensures it's copied correctly and prevents bailing if there is some kind of current <=> next comparison happening
   
I'm coming from C so it's possible there are some language constructs that prevent these so they can be omitted...
Also this is me coming from an opaque box perspective and other times I will argue it's OK to simplify tests by knowledge of the code implementation, thus is the hypocrisy of me :)

nick-mobilecoin: Here's an article that might explain it better than me, https://www.guru99.com/equivalence-partitioning-boundary-value-analysis.html
It's not actually "searching an iterable" so much as boundary checking the iterable with the addition of empty
```


        ┌────────┬────────┬────────┬────────┐
        │        │        │        │        │
        └────────┴────────┴────────┴────────┘
      ▲     ▲                 ▲         ▲     ▲
      │     │                 │         │     │
```
The boundary is "before" the iterable, the first item, the last item, after the iterable, and then you want one from the middle for paritioning equivalence

samdealy: I think it's fine to check for these cases. We may consider implementing this as a standard elsewhere in the codebase as well. Done.

fog/view/enclave/impl/src/lib.rs
```diff
@@ -359,34 +397,451 @@ where
...
+    }
+
+    #[test]
+    fn second_response_zero_processed_blocks() {
+        const STORE_COUNT: usize = 4;
+        let mut decrypted_query_responses = Vec::with_capacity(STORE_COUNT);
+
+        // Make the first response fully processed.
+        let fully_processed_block_count = 3;
+        let fully_processed_timestamp = 3;
```
jcape, line 544: Should these be 10, 10?

jcape (approved):

awygle (approved):

PR #2746, opened 2022-10-21
add scheduled times and timezones for dependabot PRs

garbageslam:
Currently, the dev-cluster can only build and run tests for a handful of PRs at a time. If dependabot makes 7 PRs at once, this will swamp the cluster, and all engineers will be blocked from iterating on CI for several hours (even if CD is not run).

This change makes dependabot only create PRs for the root workspace at midnight in the pacific timezone.

The other workspaces are assigned to 1 AM in the pacific timezone.

Given that most engineers are in one of the four US continental timezones, this will always be between midnight and 4 AM for most employees, so people won't have to kill the dependabot CI runs manually in order to be productive.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,8 @@ updates:
   directory: "/"
   schedule:
     interval: daily
+    time: "00:00"
+    timezone: "America/Los_Angeles"
   open-pull-requests-limit: 25
   labels:
     - "blocks-v3.x"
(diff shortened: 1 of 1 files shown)


eranrund (approved):

.github/dependabot.yml
```diff
@@ -49,4 +59,6 @@ updates:
   directory: "/"
   schedule:
     interval: daily
+    time: "01:00"
```
remoun, line 62: Maybe have the GHA runs later?
```suggestion
    time: "02:00"
```

remoun (approved):

.github/dependabot.yml
garbageslam: sure

PR #2772, opened 2022-10-25
master - ci/cd - Reduce dependabot pr limit, set ci/cd concurrency logic

jgreat:
### Motivation

Set concurrency logic for ci and cd workflows:

- Limit to one concurrent process on master and release/*.
- Limit other events, PRs and other branches, by actor (user or dependabot).
- Cancel in progress if master or release/*, but keep queue for PRs and other actors.

Reduce the dependabot open-pull-requests-limit to 10 for / and 5 for github actions (we only have a couple external actions at this point)

### Future Work

- Wish GHA logic was cleaner or regex was available.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,7 +6,7 @@ updates:
     interval: daily
     time: "00:00"
     timezone: "America/Los_Angeles"
-  open-pull-requests-limit: 25
+  open-pull-requests-limit: 10
   labels:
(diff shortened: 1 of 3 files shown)


jgreat:
Tested a second "no-op" push,  got a second job queued up as expected.

jgreat:
A third job canceled the second still pending job. That's not the desired behavior.

PR #2785, opened 2022-10-27
propagate mc-crypto-dalek backends, serde + prost feature flags

ryankurte:
- add backend helpers for `mc-crypto-dalek`
  - replaces per-crate backend flags (also re-exports `dalek` components, could replace `*-dalek` deps)
  - allows target/backend detection when `default` feature is enabled (required for tests to run per-crate / w/out a top level backend definition)
- propagate `prost` and `serde` features through ring-signer related types

blocked on #2525 (now merged!)

cc. @jcape @garbageslam

### Motivation

towards streaming ring signatures / in an effort to isolate the procedural and cryptographic changes.

PR #2786, opened 2022-10-27
support streaming ring signature signing and verification

ryankurte:
- split ring signing into `mlsag_sign` and `mlsag_verify` components supporting streaming logic, exposed via `internals` feature flag
- add `Ring` trait, generic over compressed / decompressed rings for storage / performance optimisation
- rework existing `RingMLSAG` to call back to these components (so there's only one implementation)
  - now gated behind `alloc` feature
- move `ReducedTxOut`, `fn challenge`, etc. to support this

blocked on: 
- #2525
- #2785 

### Motivation

hardware wallets don't have the memory to hold full ring computations, so there needs to be a way to compute them in a streaming manner...


crypto/ring-signature/src/ring_signature/mlsag_sign.rs
```diff
@@ -0,0 +1,343 @@
...
+
+        // Responses `r_{0,0}, r_{0,1}, ... , r_{ring_size-1,0}, r_{ring_size-1,1}`.
+        r.iter_mut()
+            .for_each(|v| *v = CurveScalar::from(Scalar::zero()));
+
+        for i in 0..params.ring_size {
+            if i == params.real_index {
+                continue;
+            }
+            r[2 * i].scalar = Scalar::random(&mut rng);
```
ryankurte, line 167: i found an issue with a hw RNG because all but the real response were being set to zero... would it be reasonable to add a check / retry / fail to make sure this can never happen in reality?

garbageslam: I think such check should be in the implementation of the object that implements CryptoRng

garbageslam: Compare with: https://github.com/mobilecoinfoundation/mobilecoin/blob/master/crypto/rand/src/rdrandrng/retry.rs

ryankurte: hmm yeah that's probably fair. as a corollary, are the properties of the ring maintained _if_ one or more responses are zero?

crypto/ring-signature/src/ring_signature/mlsag.rs
```diff
@@ -112,6 +105,7 @@ impl RingMLSAG {
     // * `check_value_is_preserved` - If true, check that the value of inputs equals
     //   value of outputs.
     // * `rng` - Randomness.
+    #[allow(unreachable_code, unused_variables)]
```
briancorbin, line 108: Is the purpose of this to allow for streaming?

crypto/ring-signature/src/ring_signature/mlsag.rs
```diff
@@ -124,114 +118,46 @@ impl RingMLSAG {
...
-        }
-
-        let G = B_BLINDING;
-        debug_assert!(
-            generator.B_blinding == G,
-            "basepoint for blindings mismatch"
+        // Setup buffers
+        let (mut responses, mut decompressed_ring) = (
+            alloc::vec![CurveScalar::from(Scalar::zero()); 2 * ring_size],
+            alloc::vec![(RistrettoPublic::default(), Commitment::default()); ring_size],
```
briancorbin, line 128: Is the idea that the buffers would get set up off of the ledger, and then the signing of each ring happens individually on the ledger to fill the buffers?

crypto/ring-signature/src/ring_signature/mlsag_verify.rs
```diff
@@ -0,0 +1,116 @@
...
+use crate::{
+    ring_signature::{
+        challenge, hash_to_point, CurveScalar, Error, KeyImage, Ring, Scalar, B_BLINDING,
+    },
+    Commitment, CompressedCommitment,
+};
+
+/// MLSAG Verification object
+pub struct MlsagVerify<'a, R: Ring> {
+    /// Key image too be verified
```
briancorbin, line 16:
```suggestion
    /// Key image to be verified
```

crypto/ring-signature/src/ring_signature/mlsag.rs
```diff
@@ -124,114 +118,46 @@ impl RingMLSAG {
         check_value_is_preserved: bool,
         // Note: this `mut rng` can just be `rng` if this is merged upstream:
         // https://github.com/dalek-cryptography/curve25519-dalek/pull/394
```
briancorbin, line 120:
```suggestion
```

briancorbin: Remove comment now that `mut` has been removed

briancorbin (changes requested):

crypto/ring-signature/src/ring_signature/mlsag.rs
ryankurte: not sure what you mean by buffers? the streaming is to mitigate packet size constraints and avoid holding these (challenge/response and ring entry) buffers on the ledger when constructing each ring signature.

the process from the devices perspective (ie. APDUs over USB) is documented [here](https://mobilecoinofficial.github.io/ledger-mob/ledger_mob_core/#executing-a-transaction) (which matches a [state machine](https://mobilecoinofficial.github.io/ledger-mob/src/ledger_mob_core/engine/ring.rs.html#118) on the device), but from a host perspective it's just an implementation of [`RingSigner`](https://mobilecoinofficial.github.io/ledger-mob/ledger_mob/tx/struct.TransactionHandle.html#impl-RingSigner-for-TransactionHandle%3CT%3E) that manages this [internally](https://mobilecoinofficial.github.io/ledger-mob/src/ledger_mob/tx/ring.rs.html#61).

crypto/ring-signature/src/ring_signature/mlsag_sign.rs
ryankurte: the conclusion was, no this is definitely a problem. added an RNG check to the ledger app to mitigate.

crypto/ring-signature/src/ring_signature/mlsag.rs
ryankurte: oh hmm, the allow unused was to avoid a compiler error when `alloc` feature is disabled but i changed the approach so may be able to remove it now.

briancorbin (approved):

crypto/ring-signature/src/ring_signature/mlsag.rs
isis-mc: Did we determine if we could remove this?

crypto/ring-signature/src/ring_signature/mlsag.rs
```diff
@@ -122,116 +116,46 @@ impl RingMLSAG {
         output_blinding: &Scalar,
         generator: &PedersenGens,
         check_value_is_preserved: bool,
-        // Note: this `mut rng` can just be `rng` if this is merged upstream:
-        // https://github.com/dalek-cryptography/curve25519-dalek/pull/394
-        mut rng: &mut dyn CryptoRngCore,
+        rng: &mut dyn CryptoRngCore,
```
isis-mc, line 119: Oops, this was on me, sorry about that.

crypto/ring-signature/src/ring_signature/mlsag_sign.rs
```diff
@@ -0,0 +1,343 @@
...
+        // "Close the loop" by computing responses for the real index.
+        let (key_image, c_zero) = sign_ctx.finalise(self)?;
+
+        Ok((key_image, c_zero))
+    }
+}
+
+/// MLSAG signing context, supports incremental computation for ring signing.
+///
+/// WARNING: MISUSE OF THIS API MAY GENERATE INVALID RINGS,
```
isis-mc, line 94: nit: maybe `# Warning` for rustdoc formatting

crypto/ring-signature/src/ring_signature/mlsag_sign.rs
```diff
@@ -0,0 +1,343 @@
...
+/// SEE [`MlsagSign`] FOR A PREFERRED HIGHER LEVEL ABSTRACTION
+#[derive(Debug, Zeroize)]
+pub struct MlsagSignCtx<R: AsMut<[CurveScalar]>> {
+    alpha_0: Scalar,
+    alpha_1: Scalar,
+    G: RistrettoPoint,
+    I: RistrettoPoint,
+
+    /// Cache real input for balance checking
+    #[zeroize(skip)]
```
isis-mc, line 104: Why do we skip zeroize on the real input?

crypto/ring-signature/src/ring_signature/mlsag_sign.rs
```diff
@@ -0,0 +1,343 @@
...
+    key_image: KeyImage,
+
+    #[zeroize(skip)]
+    output_commitment: Commitment,
+
+    complete: bool,
+}
+
+#[allow(dead_code)]
```
isis-mc, line 125: What are/were the dead code warnings here? Can we remove this flag? Or maybe move the "dead code" to a more internal function that's `#[inline(always)]` or something like that?

crypto/ring-signature/src/ring_signature/mlsag_sign.rs
```diff
@@ -0,0 +1,343 @@
...
+    pub fn init<'a>(
+        params: &'a MlsagSignParams<'a>,
+        mut rng: impl CryptoRngCore,
+        mut responses: R,
+    ) -> Result<Self, Error> {
+        let G = B_BLINDING;
+        debug_assert!(
+            params.generator.B_blinding == G,
+            "basepoint for blindings mismatch"
+        );
```
isis-mc, line 138: Is this because of the multi asset types now?

crypto/ring-signature/src/ring_signature/mlsag_sign.rs
```diff
@@ -0,0 +1,343 @@
...
+            zeroth_challenge: None,
+            responses,
+            complete: false,
+        })
+    }
+
+    /// Update signing context with provided tx_out.
+    ///
+    /// NOTE THIS _MUST_ FOLLOW A CALL TO INIT AND TRAVERSE THE RING STARTING
+    /// WITH THE REAL TXOUT
```
isis-mc, line 191: It's annoying and complicated to change/maintain but maybe we could consider leaning on the type system a bit more here? Such that this Cannot Be Called without calling the proper init

crypto/ring-signature/src/ring_signature/mod.rs
```diff
@@ -4,27 +4,45 @@
...
-    error::Error,
-    key_image::KeyImage,
-    mlsag::{ReducedTxOut, RingMLSAG},
+#[cfg(feature = "alloc")]
+pub use self::mlsag::RingMLSAG;
+
+pub use self::{curve_scalar::CurveScalar, error::Error, key_image::KeyImage};
+
+use crate::{
+    domain_separators::{HASH_TO_POINT_DOMAIN_TAG, RING_MLSAG_CHALLENGE_DOMAIN_TAG},
```
isis-mc, line 34: Yay, tyvm for adding the domain separator.

crypto/ring-signature/src/ring_signature/mod.rs
```diff
@@ -96,6 +120,117 @@ pub fn hash_to_point(ristretto_public: &RistrettoPublic) -> RistrettoPoint {
...
+    fn size(&self) -> usize;
+
+    /// Access a decompressed ring element by index
+    ///
+    /// (a-la `core::ops::Index`, defined here to avoid orphan rules)
+    fn index(&self, index: usize) -> Result<(RistrettoPublic, Commitment), Error>;
+
+    /// Ensure ring decompresses (no-op for pre-decompressed rings)
+    fn check(&self) -> Result<(), Error>;
+}
```
isis-mc, line 181: :tada:

isis-mc (approved):
Some very minor nits but overall looks great! Thanks!

crypto/ring-signature/src/ring_signature/mlsag.rs
```diff
@@ -1,52 +1,45 @@
...
 use mc_crypto_digestible::Digestible;
-use mc_crypto_hashes::{Blake2b512, Digest};
-use mc_crypto_keys::{CompressedRistrettoPublic, RistrettoPrivate, RistrettoPublic};
+
+#[cfg(feature = "prost")]
 use prost::Message;
-use rand_core::CryptoRngCore;
+
+#[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
```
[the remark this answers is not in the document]
ryankurte: there didn't used to be feature-gating for serde/prost, so these are all (where all is spread over however many piecewise PRs) new afaik

crypto/ring-signature/src/ring_signature/mlsag_sign.rs
ryankurte: at the time of writing there wasn't a zeroize impl for that type, i could rebase and remove if there now is

crypto/ring-signature/src/ring_signature/mlsag_sign.rs
ryankurte: ahh i'll have to go back and check, but i think the inline trick would solve it

crypto/ring-signature/src/ring_signature/mlsag_sign.rs
ryankurte: err, i just moved that from mlsag.rs:135?

crypto/ring-signature/src/ring_signature/mlsag_sign.rs
ryankurte: i though about type-stating the MlsagSignCtx to mitigate that, but because of the need to stream we can't associate any state with the referenced buffers (responses etc.) so there's still a sharp edge there (and you just end up wrapping it in an enum so you can hold it anyway).

could refactor to that, or add an `initialised` flag and runtime check that should get optimised out?

PR #2808, opened 2022-11-02
Retry ledger "index out of bounds" errors

garbageslam:
Test client sometimes fails in CD with errors like this:

> 2022-11-01 23:40:23.511798629 UTC CRIT thread main on test_client panicked! panicked at 'Unexpected error SubmitTx(TxRejected(Ledger, "Ledger error: `Index out of bounds: 111964`."))', fog/test-client/src/bin/main.rs:111:23    0:     0x564d793cb0c9 - backtrace::backtrace::trace::he6de28a0cae2390a

https://github.com/mobilecoinfoundation/mobilecoin/actions/runs/3373323306/jobs/5598519127

(After this commit we can actually see the error message: https://github.com/mobilecoinfoundation/mobilecoin/commit/68c6b1afee693b9d1548a266bb247b18278306e9)

I believe that this may be caused by a race condition:
* Fog is reading from one of the consensus nodes, which is ahead of another
* The test client decides that some number of tx outs are in the ledger based on what fog says
* The test client samples ring inputs (or selects tx outs) that don't appear in the ledger of the consensus node that it actually submits to
* The submission fails with this index out of bounds error during merkle proof validation

If we retry, then the slower consensus node will have a chance to catch up and perhaps the submission will succeed, so this seems like a good error to retry on.

---

This branch makes a new grpc consensus api status code for this error condition, so that it can be cleanly identified and retried on. It also makes the thick client do that. Some clients don't use the `ThickClient` object, like SDKs, so if they want to follow suit they will need downstream changes.

### DIFF
--- a/connection/src/error.rs
+++ b/connection/src/error.rs
@@ -42,6 +42,7 @@ impl Error {
         match self {
             Error::Grpc(_) => true,
             Error::Attestation(err) => err.should_retry(),
+            Error::TransactionValidation(ProposeTxResult::LedgerTxOutIndexOutOfBounds, _) => true,
             _ => false,
         }
     }
--- a/consensus/api/proto/consensus_common.proto
+++ b/consensus/api/proto/consensus_common.proto
@@ -97,6 +97,7 @@ enum ProposeTxResult {
     InputRuleInvalidAmountSharedSecret = 51;
     InputRuleTxOutConversion = 52;
     InputRuleAmount = 53;
+    LedgerTxOutIndexOutOfBounds = 54;
 }
 
 /// Response from TxPropose RPC call.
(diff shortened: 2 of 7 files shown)


consensus/api/proto/consensus_common.proto
```diff
@@ -97,6 +97,7 @@ enum ProposeTxResult {
     InputRuleInvalidAmountSharedSecret = 51;
     InputRuleTxOutConversion = 52;
     InputRuleAmount = 53;
+    LedgerTxOutIndexOutOfBounds = 54;
```
garbageslam, line 100: This is the new status code that indicates the condition that we want clients to be able to retry on (without retrying all Ledger errors, many of which may not be recoverable)

connection/src/error.rs
```diff
@@ -42,6 +42,7 @@ impl Error {
         match self {
             Error::Grpc(_) => true,
             Error::Attestation(err) => err.should_retry(),
+            Error::TransactionValidation(ProposeTxResult::LedgerTxOutIndexOutOfBounds, _) => true,
```
garbageslam, line 45: this makes the thick client attempt to retry if it encounters this new error code

nick-mobilecoin (approved):

eranrund (approved):

PR #2809, opened 2022-11-02
Make the TxSummary include a digest of input rules, per discussion

garbageslam:
There is little downside to this since the TxInSummary is already very small compared to the TxOutSummary, and it allows that in the future we could also send the InputRules to the hardware device and verify them, if we decide that that is helpful.

This was a review comment in MCIP 52 from Ryan

### DIFF
--- a/transaction/core/src/input_rules.rs
+++ b/transaction/core/src/input_rules.rs
@@ -15,7 +15,7 @@ use crate::{
 };
 use alloc::{collections::BTreeMap, vec::Vec};
 use displaydoc::Display;
-use mc_crypto_digestible::Digestible;
+use mc_crypto_digestible::{Digestible, MerlinTranscript};
 use mc_crypto_keys::CompressedRistrettoPublic;
 use prost::Message;
 use serde::{Deserialize, Serialize};
@@ -83,6 +83,11 @@ impl InputRules {
             result.insert(output.tx_out.public_key, output.tx_out.clone());
         }
         result
+    }
+
+    /// Compute MCIP-52 canonical digest of input rules
+    pub fn canonical_digest(&self) -> [u8; 32] {
+        self.digest32::<MerlinTranscript>(b"mc-input-rules")
     }
 
     /// Verify that a Tx conforms to the rules.
--- a/transaction/core/src/tx_summary.rs
+++ b/transaction/core/src/tx_summary.rs
@@ -143,7 +143,7 @@ impl TxSummary {
                         ..Default::default()
                     };
                     if let Some(rules) = &input.input_rules {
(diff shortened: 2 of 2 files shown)


transaction/core/src/input_rules.rs
```diff
@@ -85,6 +85,11 @@ impl InputRules {
         result
     }
 
+    /// Compute MCIP-52 canonical digest of input rules
+    pub fn canonical_digest(&self) -> [u8; 32] {
```
nick-mobilecoin, line 89: :question: Is there a cryptographic domain usage for *canonical*?
Seeing it's usage implies that there is another *non_canonical* digest in use somewhere?

transaction/core/src/tx_summary.rs
```diff
@@ -214,7 +214,8 @@ pub struct TxInSummary {
     #[prost(message, required, tag = "1")]
     pub pseudo_output_commitment: CompressedCommitment,
 
-    /// Whether there are input rules associated to this input
-    #[prost(bool, tag = "2")]
-    pub has_input_rules: bool,
+    /// If there are input rules associated to this input, the canonical digest
+    /// of these (per MCIP 52). If not, then this field is empty.
+    #[prost(bytes, tag = "2")]
+    pub input_rules_digest: Vec<u8>,
```
nick-mobilecoin, line 220: :wrench: It looks like the current [MCIP-52](https://github.com/garbageslam/mcips/blob/tx_summary_digest/text/0052-tx-summary-digest.md) is still saying "flag".  While one might argue the presence or absence of this is a flag, I wonder if it would be better to be more specific in the MCIP
It could also be the iteration race of documenting and implementing and thus hasn't been updated yet.

nick-mobilecoin (approved):

transaction/core/src/tx_summary.rs
garbageslam: Yea im in the process of updating

transaction/core/src/input_rules.rs
garbageslam: Yea, its usually important for bytes / digests to be canonical when they are being signed or verified in some way.

One of the things mc-crypto-digestible achieves is that it produces canonical digests of objects regardless of any issues around serialization. For comparison, protobuf serializations are not canonical -- there are many valid serializations of an object and protobuf encoders can change these bytes without it being considered a breaking change, as long as its forwards and backwards compatible. So digesting protobuf bytes is not a canonical digest. This one is, and that's what the word canonical is meant to signal. Changing the bytes of this digest at all for a given object is considered an (unacceptable) breaking change.

remoun (approved):

PR #2815, opened 2022-11-03
fix unnecessary type parameter in build_unsigned

garbageslam:
The `build_unsigned` function takes an RNG type parameter but it is not bound to any actual inputs, or used in the function body. So all the callers have to supply a pointless type parameter.

This is a random defect that I saw when developing #2725. Pulling this out to ease review of that.


nick-mobilecoin (approved):

remoun (approved):
Thank you!

PR #2816, opened 2022-11-03
Move signing function to SigningData in rct_bulletproofs

garbageslam:
This is a change that I wanted to do for a while, that I ended up doing as part of #2725 because it was necessary to do the hw-wallet signing flow appropriately.

Pulling out of #2725 to ease review, this is just code movement and should not be a functional change.

### DIFF
--- a/transaction/core/src/ring_ct/rct_bulletproofs.rs
+++ b/transaction/core/src/ring_ct/rct_bulletproofs.rs
@@ -561,7 +561,7 @@ impl SignatureRctBulletproofs {
     ///   amount commitment.
     /// * `fee` - Value of the implicit fee output.
     /// * `signer` - The ring signer entity (with spend private key)
-    /// * `rng` -
+    /// * `rng` - randomness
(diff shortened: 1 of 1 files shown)


transaction/core/src/ring_ct/rct_bulletproofs.rs
```diff
@@ -511,7 +560,8 @@ impl SignatureRctBulletproofs {
     /// * `output_values_and_blindings` - Value and blinding for each output
     ///   amount commitment.
     /// * `fee` - Value of the implicit fee output.
-    /// * `token id` - This determines the pedersen generator for commitments
+    /// * `signer` - The ring signer entity (with spend private key)
+    /// * `rng` -
```
nick-mobilecoin, line 564: :upside_down_face: 
```suggestion
    /// * `rng` - randomness
```

nick-mobilecoin (approved):

eranrund (approved):

PR #2844, opened 2022-11-08
master - CD - restore 1.1.3 ledger and db from backup

jgreat:
### Motivation

Master branch version of #2802

Bootstrap environment from S3/PG backups. Reduce the amount of steps required to deploy and test.

### Work done but not captured in this PR

Build out `v1.1.3-dev` environment, initialize ledger, run block_version 0 testing and capture block chain s3 buckets and fog recovery database. 

Backups are located in the development ledger bucket: `mobilecoin.eu.development.chain`

- `/prebuilt/${VERSION}/chain/node[1-3]/`
- `/prebuilt/${VERSION}/sql/fog_recovery.sql`

### Changes

- Break setup steps into their own reusable workflow (more DRY)
- Use github `environments` feature for name spacing required secrets/configuration values.  (Step towards reusing workflows for alpha/beta/test/main artifact builds and deployments)
- Add bootstrap reusable workflow that will restore from a specified backup in s3.
- Remove all v1.1.3 deployment and testing steps. The results of these tests are captured in the backups.
- Clean up dev env after runs on a pull_request, a tag, and release/* branches.  Keep master and feature/* until the branches are deleted.

### Major changes in dev env workings.

- Reduce deployment/testing complexity by only using pre-built tokens.json, token governor keys, token signer keys and  wallet/ledger seeds.
- Create static tokens.json with IDs 0, 1 and 8192
- Remove `client_auth` option.  Just run the test environments like we run MC prod envs.
- Don't launch dev envs with persistent disk.  Faster deploys and more density on SGX nodes
- Move to 3 consensus and 2 view nodes.
- Reduce "sgx" requests and limits for more density on SGX nodes.

### Future Work
- Restore from v3.0.0-dev network state.


.github/workflows/mobilecoin-workflow-dev-setup-environment.yaml
```diff
@@ -0,0 +1,235 @@
...
+        echo -n "${{ secrets.MINTING_8192_GOVERNOR_1_PRIVATE }}" > "${MINTING_BASE_PATH}/token_8192_governor_1.private.pem"
+        echo -n "${{ secrets.MINTING_8192_GOVERNOR_1_PUBLIC }}" > "${MINTING_BASE_PATH}/token_8192_governor_1.public.pem"
+        echo -n "${{ secrets.MINTING_8192_SIGNER_1_PRIVATE }}" > "${MINTING_BASE_PATH}/token_8192_signer_1.private.pem"
+        echo -n "${{ secrets.MINTING_8192_SIGNER_1_PUBLIC }}" > "${MINTING_BASE_PATH}/token_8192_signer_1.public.pem"
+
+        # values for v3.0.0-dev release
+        echo -n "${{ secrets.MINTING_8192_GOVERNOR_1_PRIVATE }}" > "${MINTING_BASE_PATH}/minter8192_governor.private.pem"
+        echo -n "${{ secrets.MINTING_8192_GOVERNOR_1_PUBLIC }}" > "${MINTING_BASE_PATH}/minter8192_governor.public.pem"
+        echo -n "${{ secrets.MINTING_8192_SIGNER_1_PRIVATE }}" > "${MINTING_BASE_PATH}/token8192_signer.private.pem"
+        echo -n "${{ secrets.MINTING_8192_SIGNER_1_PUBLIC }}" > "${MINTING_BASE_PATH}/token8192_signer.public.pem"
```
nick-mobilecoin, line 171: :question: These look to be the same values as 5 linus up why write to another file instead of re-using?

.internal-ci/test/minting-config-tx-test.sh
```diff
@@ -66,13 +57,4 @@ mc-consensus-mint-client generate-and-submit-mint-config-tx \
...
-new_block_count=0
-echo "-- Waiting for mint config tx to commit to the block chain"
-
-while [[ $block_count -ge $new_block_count ]]
-do
-    sleep 15
-    new_block_count=$(get_block_count)
-    echo "  Current block count: $new_block_count"
-done
+sleep 15
```
nick-mobilecoin, line 60: :question: Did the logic to wait for the blocks mean to get removed?

nick-mobilecoin (approved):

.internal-ci/test/minting-config-tx-test.sh
jgreat: Good catch

.github/workflows/mobilecoin-workflow-dev-setup-environment.yaml
jgreat: This is for compatibility with the previous release. 

The testing scripts are bundled in the release its self. That makes them internally consistent for a specific stage of the release, but it does mean that we need to make sure we keep compatible configs for as long as we are using the older release as part of the CD upgrade process.

.github/workflows/mobilecoin-dev-cd.yaml
```diff
@@ -74,7 +76,8 @@ jobs:
 # Build binaries
 #########################################
   build-rust-hardware-projects:
-    if: github.actor != 'dependabot[bot]'
+    needs:
+    - generate-metadata
     runs-on: [self-hosted, Linux, large]
```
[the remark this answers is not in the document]
jgreat: yes, not sure how I missed this.

aweis89 (approved):
LGTM

joekottke (approved):
Approving after walkthrough with @jgreat

PR #2839, opened 2022-11-08
Add unary apis to FogViewRouter

samdealy:
### Future Work

[Soundtrack of this PR]()

PR #2766, opened 2022-10-24
Calculate last known block txo count and add tests

samdealy:
### Motivation
A previous PR added `last_known_block_count` collation, and this one adds `last_known_block_cumulative_txo_count` collation. I also took the opportunity to refactor this into a separate \
method and add unit tests.

PR #2845, opened 2022-11-08
Make query batch logic compatible with sharding strategy

samdealy:
### Motivation
I merged `master` into `feature/fog-view-router` and didn't accurately resolve merge conflicts for changes introduced by PR #2708. This PR fixes that merge issue and also updates the `sql_recovery_db` method to use a `BlockRange` (this is not strictly necessary for this PR but thought it was a good and easy cleanup)

### DIFF
--- a/fog/sql_recovery_db/src/lib.rs
+++ b/fog/sql_recovery_db/src/lib.rs
@@ -979,8 +979,7 @@ impl SqlRecoveryDb {
     ///
     /// Arguments:
     /// * ingress_key: The ingress key we need ETxOutRecords from
-    /// * block_index: The first block we need ETxOutRecords from
-    /// * block_count: How many consecutive blocks to also request data for.
+    /// * block_range: The range of blocks to get ETxOutRecords from.
     ///
     /// Returns:
     /// * The sequence of ETxOutRecord's, from consecutive blocks starting from
@@ -988,8 +987,7 @@ impl SqlRecoveryDb {
     fn get_tx_outs_by_block_range_and_key_retriable(
         &self,
         ingress_key: CompressedRistrettoPublic,
-        block_index: u64,
-        block_count: usize,
+        block_range: &BlockRange,
     ) -> Result<Vec<Vec<ETxOutRecord>>, Error> {
         let conn = self.pool.get()?;
 
@@ -1005,8 +1003,8 @@ impl SqlRecoveryDb {
             use schema::ingested_blocks::dsl;
             dsl::ingested_blocks
                 .filter(dsl::ingress_public_key.eq(key_bytes))
-                .filter(dsl::block_number.ge(block_index as i64))
-                .limit(block_count as i64)
+                .filter(dsl::block_number.ge(block_range.start_block as i64))
+                .limit(block_range.len() as i64)
                 .select((dsl::block_number, dsl::proto_ingested_block_data))
                 .order(dsl::block_number.asc())
         };
@@ -1014,12 +1012,12 @@ impl SqlRecoveryDb {
         // We will get one row for each hit in the table we found
         let rows: Vec<(i64, Vec<u8>)> = query.load(&conn)?;
 
-        if rows.len() > block_count {
+        if rows.len() > (block_range.len() as usize) {
             log::warn!(
                 self.logger,
                 "When querying, more responses than expected: {} > {}",
                 rows.len(),
-                block_count
+                block_range.len(),
             );
         }
 
@@ -1033,11 +1031,11 @@ impl SqlRecoveryDb {
 
         let mut result = Vec::new();
         for (idx, (block_number, proto)) in rows.into_iter().enumerate() {
-            if block_index + idx as u64 == block_number as u64 {
+            if block_range.start_block + idx as u64 == block_number as u64 {
                 let proto = ProtoIngestedBlockData::decode(&*proto)?;
                 result.push(proto.e_tx_out_records);
             } else {
-                log::warn!(self.logger, "When querying for block index {} and up to {} blocks on, the {}'th response has block_number {} which is not expected. Gaps in the data?", block_index, block_count, idx, block_number);
+                log::warn!(self.logger, "When querying for block index {} and up to {} blocks on, the {}'th response has block_number {} which is not expected. Gaps in the data?", block_range.start_block, block_range.len(), idx, block_number);
                 break;
             }
         }
@@ -1473,8 +1471,7 @@ impl RecoveryDb for SqlRecoveryDb {
     ///
     /// Arguments:
     /// * ingress_key: The ingress key we need ETxOutRecords from
-    /// * block_index: The first block we need ETxOutRecords from
-    /// * block_count: How many consecutive blocks to also request data for.
+    /// * block_range: The range of blocks to get ETxOutRecords from.
     ///
     /// Returns:
     /// * The sequence of ETxOutRecord's, from consecutive blocks starting from
@@ -1482,11 +1479,10 @@ impl RecoveryDb for SqlRecoveryDb {
     fn get_tx_outs_by_block_range_and_key(
         &self,
         ingress_key: CompressedRistrettoPublic,
-        block_index: u64,
-        block_count: usize,
+        block_range: &BlockRange,
     ) -> Result<Vec<Vec<ETxOutRecord>>, Self::Error> {
         our_retry(self.get_retries(), || {
-            self.get_tx_outs_by_block_range_and_key_retriable(ingress_key, block_index, block_count)
+            self.get_tx_outs_by_block_range_and_key_retriable(ingress_key, block_range)
         })
     }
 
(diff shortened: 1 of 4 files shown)


samdealy:
* master
    * **PR #2845** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2845" target="_blank">Graphite</a>  👈
    * **PR #2803** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2803" target="_blank">Graphite</a> 
      * **PR #2804** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2804" target="_blank">Graphite</a> 
        * **PR #2839** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2839" target="_blank">Graphite</a> 
          * **PR #2840** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2840" target="_blank">Graphite</a> 
    * **PR #2783** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2783" target="_blank">Graphite</a> 
    * **PR #2781** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2781" target="_blank">Graphite</a> 
    * **PR #2766** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2766" target="_blank">Graphite</a> 
    * **PR #2726** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2726" target="_blank">Graphite</a> 
    * **PR #2721** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2721" target="_blank">Graphite</a> 
    * **PR #2722** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2722" target="_blank">Graphite</a> 
    * **PR #2720** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2720" target="_blank">Graphite</a> 
    * **PR #2711** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2711" target="_blank">Graphite</a> 
      * **PR #2730** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2730" target="_blank">Graphite</a> 
      * **PR #2712** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2712" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2845?utm_source=stack-comment).

nick-mobilecoin (changes requested):
The main issue is that I'm not sure this should be skipping ranges when the start isn't in the shard as it feels like the end could still be in the shard.

fog/view/server/src/block_tracker.rs
```diff
@@ -90,6 +91,38 @@ where
...
+        &self,
+        block_index: u64,
+        block_count: usize,
+    ) -> Result<BlockRange, String> {
+        if !self
+            .sharding_strategy
+            .get_block_range()
+            .contains(block_index)
+        {
+            return Err(format!(
```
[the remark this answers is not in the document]
samdealy: Yeah, I thought about this when designing it. The thing to keep in mind is that the block index of 200 will eventually be supplied to this function because the caller is going sequentially through each block. I figured it would be cleaner to only process blocks if the provided index is in range. I think this because the query batch size is really a greedy attempt to load more blocks, and I think it would be confusing if we load more blocks for a block index that isn't in range.

fog/sql_recovery_db/src/lib.rs
```diff
@@ -1005,21 +1003,21 @@ impl SqlRecoveryDb {
...
+                .limit(block_range.len() as i64)
                 .select((dsl::block_number, dsl::proto_ingested_block_data))
                 .order(dsl::block_number.asc())
         };
 
         // We will get one row for each hit in the table we found
         let rows: Vec<(i64, Vec<u8>)> = query.load(&conn)?;
 
-        if rows.len() > block_count {
+        if rows.len() > (block_range.len() as usize) {
```
nick-mobilecoin, line 1015: :upside_down_face: it's already a usize
```suggestion
        if rows.len() > block_range.len() {
```

fog/sql_recovery_db/src/lib.rs
```diff
@@ -1005,21 +1003,21 @@ impl SqlRecoveryDb {
             use schema::ingested_blocks::dsl;
             dsl::ingested_blocks
                 .filter(dsl::ingress_public_key.eq(key_bytes))
-                .filter(dsl::block_number.ge(block_index as i64))
-                .limit(block_count as i64)
+                .filter(dsl::block_number.ge(block_range.start_block as i64))
+                .limit(block_range.len() as i64)
```
nick-mobilecoin, line 1007: :wrench: might consider storing `block_range.len()` in a local variable and re-using the local variable.

nick-mobilecoin (approved):
:+1: Like the simplification in this change interation

fog/types/src/common.rs
```diff
@@ -39,6 +47,16 @@ impl BlockRange {
     pub fn overlaps(&self, other: &BlockRange) -> bool {
         self.start_block < other.end_block && other.start_block < self.end_block
     }
+
+    /// Returns the length of the BlockRange, i.e. the number of blocks.
+    pub fn len(&self) -> usize {
+        (self.end_block - self.start_block) as usize
```
remoun, line 53: It seems safer to keep this as a `u64` -- `usize` could be 32 bits
```suggestion
    pub fn len(&self) -> u64 {
        self.end_block - self.start_block
```

fog/api/proto/view.proto
```diff
@@ -273,6 +273,13 @@ message TxOutSearchResult {
     /// It is be zero-padding in the other cases.
     /// FIXME: MC-1491 ensure this happens either in enclave or db, or wait for ORAM
     bytes ciphertext = 3;
+
+    /// The length of the payload that is encrypted in the ciphertext. Ciphertexts will always be 255 in length, but
+    /// the payload may be less than that, so any additional bytes should not be interpreted by the client. This value
+    /// tells the client which bytes to interpret.
+    /// TODO: Remove this field and replace with an optional `ciphertext_padding` field. This new field will contain
```
remoun, line 280: Adding a field with a TODO to remove it seems odd. How much work would it take to replace this with `ciphertext_padding`?

fog/types/src/common.rs
nick-mobilecoin: This got changed in response to a comment I made, https://github.com/mobilecoinfoundation/mobilecoin/pull/2845#discussion_r1017221789
> :question: Doing some quick spot checks it seems most len() functions return a usize? i'm guessing this internal start and end are u64 due to prost, but it seems plausible the length here could be usize.
>
> Not really sure on this. I was mainly looking at through std for len() so it could just be std

I'm fine if we think u64 is the more robust option. My argument was mainly for consistency with other APIs.
For usize to be an issue on 32 bit systems the block range would have to exceed 4billion blocks.  Which could be probable as the chain grows. Not sure the likely hood or if this code would actually ever be used on 32 bit systems.  My understanding is the enclave code is more or less limited to 64 bit sgx platforms, but this isn't guaranteed to be enclave only.

fog/view/enclave/impl/src/oblivious_utils.rs
```diff
@@ -56,23 +57,14 @@ fn maybe_overwrite_tx_out_search_result(
...
+        .payload_length
+        .conditional_assign(
+            &(shard_ciphertext_length as u32),
+            should_overwrite_tx_out_search_result,
+        );
     for idx in 0..shard_ciphertext_length {
-        // Offset the client ciphertext by 1 because the first byte is reserved for the
-        // length delta.
-        client_tx_out_search_result.ciphertext[idx + 1].conditional_assign(
+        client_tx_out_search_result.ciphertext[idx].conditional_assign(
```
jcape: Is there a conditional assign for slices that we could use instead of doing this byte-at-a-time here?

fog/view/enclave/impl/src/oblivious_utils.rs
```diff
@@ -589,18 +583,25 @@ mod tests {
 
         assert!(result.is_err());
     }
+
     #[test]
     fn collate_shard_query_responses_different_ciphertext_lengths_returns_correct_client_ciphertexts(
     ) {
         let client_search_keys: Vec<Vec<u8>> = (0..3).map(|num| vec![num; 10]).collect();
-        let ciphertext_values = [28u8, 5u8, 128u8];
+        let ciphertext_values = [5u8, 28u8, 128u8];
```
jcape: Why the re-order?

fog/view/enclave/impl/src/oblivious_utils.rs
```diff
@@ -569,14 +562,15 @@ mod tests {
...
         let shard_tx_out_search_results: Vec<TxOutSearchResult> = client_search_keys
             .iter()
             .map(|search_key| TxOutSearchResult {
                 search_key: search_key.clone(),
                 result_code: TxOutSearchResultCode::NotFound as u32,
-                ciphertext: vec![0u8; CLIENT_CIPHERTEXT_LENGTH],
+                ciphertext: vec![0u8; CLIENT_CIPHERTEXT_LENGTH + 1],
+                payload_length: (CLIENT_CIPHERTEXT_LENGTH + 1) as u32,
             })
             .collect();
```
jcape: Cleanup for a future PR: https://doc.rust-lang.org/book/ch11-01-writing-tests.html#checking-for-panics-with-should_panic

fog/types/src/common.rs
remoun: A `BlockRange` greater than 4 billion blocks seems more plausible than an in-memory container with that many entries. So it's safer to use `u64`.

With that said, I don't feel strongly about this, and y'all have more context about this code, so I won't block approval on this.

samdealy: Yeah, I think `u64` is more robust because it ensures we never have to convert `u64` to `usize`. Done.

fog/view/server/test-utils/src/lib.rs
```diff
@@ -231,16 +231,16 @@ pub async fn assert_e_tx_out_records(
...
+        .map(|record| TxOutSearchResult {
             search_key: record.search_key.clone(),
             result_code: TxOutSearchResultCode::Found as u32,
             ciphertext: record.payload.clone(),
-        });
-    }
+            payload_length: record.payload.len() as u32,
+        })
+        .collect::<Vec<_>>();
+    expected_results.sort_by_key(|result| result.ciphertext.clone());
```
jcape: This is a unit test so it's not super critical, but you probably don't want `sort_by_key` in this situation, since `Vec::clone()` is going to be very expensive.

jcape (approved):
LGTM, but it it seems like the `index, count` -> BlockRange` mechanical change could be split out if/when repeated...

fog/view/enclave/impl/src/oblivious_utils.rs
samdealy: No, I don't think there is. That's why we decided to do it this way.

remoun (approved):

fog/view/enclave/impl/src/oblivious_utils.rs
samdealy: That change was from another PR that got merged to this branch prematurely, but it makes it easier to test the results.

PR #2853, opened 2022-11-10
`mob prompt`: limit number of cores based on memory

nick-mobilecoin:
Building the entire workspace requires ~2GB per core.
Some machines have many cores but a small amount of ram resulting in out
of memory issues. When using `mob prompt` the number of jobs will be
limited based on the physical memory size. Developers can override this
by explicitly setting the `CARGO_BUILD_JOBS` environment variable to the
desired number of jobs `CARGO_BUILD_JOBS=4 ./mob prompt`.

### Future Work

[Soundtrack of this PR]()

### DIFF
--- a/mob
+++ b/mob
@@ -56,7 +56,6 @@ import grp
 import os
 import pathlib
 import platform
-import psutil
 import subprocess
 import sys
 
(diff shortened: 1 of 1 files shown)


nick-mobilecoin:
* master
  * **PR #2853** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2853" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2853?utm_source=stack-comment).

mob
```diff
@@ -56,6 +56,7 @@ import grp
 import os
 import pathlib
 import platform
+import psutil
```
nick-mobilecoin, line 59: not sure if we want to depend on this.  I had it installed on my machine. I don't recall intentionally adding it

jcape (approved):
LGTM

mob
```diff
@@ -249,6 +250,22 @@ if git_commit:
...
+# This logic will limit the build jobs to prevent out of memory issues.
+# https://github.com/rust-lang/cargo/issues/9157 talks about trying to fix this
+# in cargo.
+jobs = os.environ.get("CARGO_BUILD_JOBS")
+if jobs is None:
+    cpus = psutil.cpu_count()
+    ram = psutil.virtual_memory().total
+    # Remove 5GB for local PC overhead
+    max_jobs = (ram - 5_000_000_000) // 2_000_000_000
+    jobs = min(max_jobs, cpus)
```
garbageslam, line 265: What if this number is zero or negative?

garbageslam (changes requested):
Set a floor of 1 on the number of jobs, and perhaps emit a warning that the build may fail?

mob
nick-mobilecoin: good catch, added logic for that

garbageslam (approved):

PR #2856, opened 2022-11-10
Allow Custom S3 Regions with Ledger Distribution

briancorbin:
This enables us to use localstack (a mock AWS instance) with an insecure connection (HTTP)

This is mainly for local network setups where secure connections are not required. A small change was made to the rusoto library to allow HTTP connections and currently lives in a fork. This will be reverted once the change makes it into the base library.

### DIFF
--- a/ledger/distribution/Cargo.toml
+++ b/ledger/distribution/Cargo.toml
@@ -25,8 +25,8 @@ displaydoc = "0.2"
 protobuf = "2.27.1"
 retry = "2.0"
 # TODO: Replace with https://github.com/awslabs/aws-sdk-rust when it is ready.
-rusoto_core = { git = "https://github.com/legokichi/rusoto.git", branch = "enable-http1", features = ["rustls"], default_features = false }
-rusoto_s3 = { git = "https://github.com/legokichi/rusoto.git", branch = "enable-http1", features = ["rustls"], default_features = false }
+rusoto_core = { git = "https://github.com/mobilecoinfoundation/rusoto.git", features = ["rustls"], default_features = false }
+rusoto_s3 = { git = "https://github.com/mobilecoinfoundation/rusoto.git", features = ["rustls"], default_features = false }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 tokio = { version = "1", features = ["rt-multi-thread"] }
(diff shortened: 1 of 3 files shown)


ledger/distribution/Cargo.toml
```diff
@@ -22,8 +22,8 @@ displaydoc = "0.2"
 protobuf = "2.27.1"
 retry = "2.0"
 # TODO: Replace with https://github.com/awslabs/aws-sdk-rust when it is ready.
-rusoto_core = { version = "0.48.0", features = ["rustls"], default_features = false }
-rusoto_s3 = { version = "0.48.0", features = ["rustls"], default_features = false }
+rusoto_core = { git = "https://github.com/legokichi/rusoto.git", branch = "enable-http1", features = ["rustls"], default_features = false }
```
jgreat, line 25: That's a pretty old branch off somebodies personal fork, I'm not sure using this is a good from a sec and update practice.  Maybe if we fork the upstream and make adjustments?

I'm looking at https://rusoto.github.io/rusoto/rusoto_core/region/enum.Region.html

There seems to be an option for custom region and their example uses a http endpoint, does this work?

jgreat (dismissed):

ledger/distribution/Cargo.toml
briancorbin: That's the same version as 0.48.0

briancorbin: it's only 1 commit ahead on branch "enable-http1"

briancorbin: Happy to perform the fork on our end and use that instead if that's preferable. Their example does not work, as they made changes in 0.48.0 that disabled the ability to use http connections. I believe it was unintentional but the dev that's in charge of rusoto has been unresponsive.

jgreat: Oh well the upstream hasn't been updated in a while either :)

jgreat: Well that's a shame on the broken upstream.   I'll leave the using this fork vs making our own to @jcape.

briancorbin: Already forked it and updated, so it should be good :)

remoun (approved):

jgreat (approved):

NotGyro (approved):
Looks good to me.

PR #2857, opened 2022-11-11
Account for epochs in block tracker highest processed block count logic

samdealy:
### Motivation
In #2852, I modified the `smoke_tests` to use the Fog View Router's `FogViewAPI` implementation (i.e. the new `unary` API). This caused the `test_middle_missing_range_with_decommission` to fail at line 828, which checks that the highest processed block count is 15. Instead, the test was producing 5. 

This occurs because the Fog View Server `enclave_block_tracker` doesn't track blocks for the first ingress key because its epoch doesn't account for any of that ingress key's blocks. As such, the `process_block_per_ingress_key` didn't have any entries for the first ingress key. This caused the check on line 159 to always fail, triggering the `else` block and causing the block count to never increment. 

The solution is to just skip over any keys that don't overlap with the shard's epoch.This allows the highest processed block count calculation to only incorporate keys / blocks that the shard is responsible for.

### DIFF
--- a/fog/recovery_db_iface/src/types.rs
+++ b/fog/recovery_db_iface/src/types.rs
@@ -72,32 +72,45 @@ impl IngressPublicKeyRecord {
     ///
     /// The function computes whether the view server needs to try to load a
     /// block connected to this ingress key.
+    pub fn covers_block_index(&self, block_index: u64) -> bool {
+        self.get_block_range().contains(block_index)
+    }
+
+    /// Computes the block range for which this record is responsible for.
     ///
-    /// Behavior:
     /// Intuitively an IngressPublicKeyRecord corresponds to a range of
     /// consecutive block indices, that the view server needs to try to load
     /// for the users to make progress correctly.
     ///
-    /// If the key is not retired or lost, the range is
-    /// [ start_block , infinity )
+    /// If the key is lost, the range is
+    /// [ start_block, last_scanned_block ]
+    ///
     /// If the key is retired and not lost, the range is
     /// [ start_block , pubkey_expiry )
     ///
-    /// If the key is lost, the range is
-    /// [ start_block, last_scanned_block ]
+    /// If the key is neither retired nor lost, the range is
+    /// [ start_block , infinity )
+    ///
     /// and if the key is lost and there is no last scanned block, the range is
     /// empty.
-    pub fn covers_block_index(&self, block_index: u64) -> bool {
+    pub fn get_block_range(&self) -> BlockRange {
         if self.status.lost {
-            self.status.start_block <= block_index
-                && self
-                    .last_scanned_block
-                    .map(|idx| block_index <= idx)
-                    .unwrap_or(false)
-        } else {
-            self.status.start_block <= block_index
-                && (!self.status.retired || block_index < self.status.pubkey_expiry)
+            return self.get_lost_block_range();
         }
+        if self.status.retired {
+            return BlockRange::new(self.status.start_block, self.status.pubkey_expiry);
+        }
+
+        BlockRange::new(self.status.start_block, u64::MAX)
+    }
+
+    fn get_lost_block_range(&self) -> BlockRange {
+        if let Some(last_scanned_block) = self.last_scanned_block {
+            // Add 1 because the last_scanned_block is inclusive.
+            return BlockRange::new(self.status.start_block, last_scanned_block + 1);
+        }
+
+        BlockRange::new(0, 0)
     }
 
     /// The next block index that needs to be scanned with this key.
--- a/fog/recovery_db_iface/src/types.rs
+++ b/fog/recovery_db_iface/src/types.rs
@@ -110,6 +110,8 @@ impl IngressPublicKeyRecord {
             return BlockRange::new(self.status.start_block, last_scanned_block + 1);
         }
 
+        // If we lost this key before it scanned anything, then return a
+        // default "empty" block range.
         BlockRange::new(0, 0)
     }
 
--- a/fog/view/server/src/block_tracker.rs
+++ b/fog/view/server/src/block_tracker.rs
@@ -148,39 +148,21 @@ where
             // Go over all known ingress keys and check if
             // any of them need to provide this block and have not provided it
             for rec in ingress_keys {
-                // If this ingress key isn't responsible to provide this block index, we can
-                // move on
-                if !rec.covers_block_index(next_block_index) {
-                    continue;
-                }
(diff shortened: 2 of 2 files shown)


samdealy:
* feature/fog-view-router
  * **PR #2845** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2845" target="_blank">Graphite</a> 
    * **PR #2803** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2803" target="_blank">Graphite</a> 
      * **PR #2804** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2804" target="_blank">Graphite</a> 
        * **PR #2839** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2839" target="_blank">Graphite</a> 
          * **PR #2852** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2852" target="_blank">Graphite</a> 
          * **PR #2840** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2840" target="_blank">Graphite</a> 
          * **PR #2857** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2857" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2857?utm_source=stack-comment).

fog/view/server/src/block_tracker.rs
```diff
@@ -154,6 +154,25 @@ where
                     continue;
                 }
 
+                // Skip the the below checks when the next block index is less than the shard's
+                // start block. This prevents prior missed blocks from from impeding the shard's
```
nick-mobilecoin, line 158: :upside_down_face: 
```suggestion
                // start block. This prevents prior missed blocks from impeding the shard's
```

fog/view/server/src/block_tracker.rs
```diff
@@ -154,6 +154,25 @@ where
...
+                // this check, the highest processed block count will be stuck on 2 (the
+                // `last_processed_block`). Even if the shard processes blocks 5 through 8, this
+                // highest processed block count would stay at 2 forever.
+                //
+                // Note: we don't skip the below checks for the shard's end block because the shard
+                // may not have seen the higher blocks. Therefore we could potentially be incrementing
+                // the highest processed block count for blocks that don't exist yet. If there
+                // are missed blocks after the shard's end block, then that's fine because the logic
+                // below ensures that each block will be incremented by one until the first processed
+                // block after the missed blocks.
```
nick-mobilecoin, line 171: :question: I'm not quite following this statement
> If there are missed blocks after the shard's end block, then that's fine because the logic below ensures that each block will be incremented by one until the first processed block after the missed blocks.

It seems like the `next_block_index > *last_processed_block` would be hit, which feels like it would be the start of the first "missed" block after the shard's end block

nick-mobilecoin (approved):

fog/view/server/src/block_tracker.rs
samdealy: Removed by later commits. Marking as resolved.

fog/view/server/src/block_tracker.rs
samdealy: Removed in later commits. Marking as resolved.

fog/view/server/src/block_tracker.rs
```diff
@@ -148,20 +148,21 @@ where
             // Go over all known ingress keys and check if
             // any of them need to provide this block and have not provided it
             for rec in ingress_keys {
-                // If this ingress key isn't responsible to provide this block index, we can
-                // move on
-                if !rec.covers_block_index(next_block_index) {
+                let epoch = self.sharding_strategy.get_block_range();
+                let is_key_responsible = rec.get_block_range().overlaps(&epoch)
```
nick-mobilecoin, line 152: :question: the previous attempt intentionally excluded the "end" in the check, this will now check and skip over blocks after `end`.  Just want to be sure that is desired.

nick-mobilecoin (approved):

fog/view/server/src/block_tracker.rs
```diff
@@ -148,20 +148,20 @@ where
...
+                let epoch = self.sharding_strategy.get_block_range();
+                let is_key_responsible = rec.get_block_range().overlaps(&epoch) && rec.covers_block_index(next_block_index);
+                if !is_key_responsible {
                     continue;
                 }
 
                 // Check if the last block we actually loaded with this key is less than
                 // next_block_index, if so then this is what we are stuck on
-                if let Some(last_processed_block) =
+                if let Some(last_processed_block_index) =
```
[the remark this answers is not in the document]
samdealy: Yes. I think we should try to update any variables that don't disambiguate whenever possible.

fog/view/server/src/block_tracker.rs
samdealy: Yes, this is desired. My logic earlier was flawed.

garbageslam:
@samdealy can we run this through CD? do you think that's helpful? that would help to convince me that it's working

fog/recovery_db_iface/src/types.rs
```diff
@@ -72,32 +72,45 @@ impl IngressPublicKeyRecord {
...
+        BlockRange::new(self.status.start_block, u64::MAX)
+    }
+
+    fn get_lost_block_range(&self) -> BlockRange {
+        if let Some(last_scanned_block) = self.last_scanned_block {
+            // Add 1 because the last_scanned_block is inclusive.
+            return BlockRange::new(self.status.start_block, last_scanned_block + 1);
+        }
+
+        BlockRange::new(0, 0)
```
garbageslam, line 113:
```suggestion
        /// If we lost this key before it scanned anything, then return a default "empty" block range.
        BlockRange::new(0, 0)
```

garbageslam (approved):
this makes sense to me

fog/recovery_db_iface/src/types.rs
samdealy: Done.

samdealy:
> @samdealy can we run this through CD? do you think that's helpful? that would help to convince me that it's working

We will be able to run this through CD once we make the necessary changes to our deployment. At that time, if it doesn't work, then we can revisit.

jcape (approved):

PR #2876, opened 2022-11-16
feat(mob): mob tool cleanup and support other images

jgreat:
### Motivation

`mob` tool enhancements.

- add support for `signing-tool` image - tools for singing code.
- bump `builder-install` to version with fixes for MacOS duplicate group problem.
- "Clean up" and reorder how options get added into the `docker run` command. 
- only show container pull, and other image startup messages when using `--verbose`
- add arbitrary additional environmental vars with `--env SOME_ENV_VAR=value`.
- remove `--expose` option - expose is really only useful you want to share services between docker containers with a `--link` command, but better to use docker-compose at that point.

### Important Change: "run command" behavior

Before this change you could specify `mob prompt --cmd "ls -al"`.  This functionality didn't  seem to quite work as I expected.  I couldn't get multi argument commands to work right. 

I've changed this functionality to use either the action of `prompt`  or a command like `ls`, or a multi argument command if you quote the command and arguments `"ls -al"`

mob prompt usage - default to builder-install image and hide startup messages.
```bash
./mob prompt
jgreat@b91a8cd51d94:/tmp/mobilenode$ 
```

mob prompt with verbose
```bash
➜ ./mob --verbose prompt 
$ command -v docker > /dev/null 2>&1
Mapping SSH_AUTH_SOCKET into container
$ cd /home/jgreat/git/mcf/mobilecoin
$ git describe --always --dirty=-modified
$ docker image pull mobilecoin/builder-install:v0.0.19
v0.0.19: Pulling from mobilecoin/builder-install
Digest: sha256:6c9496f8e80cac0b3b27670c9eb9f968687aa6fe7c6e4b6b48d29820d4323a7a
Status: Image is up to date for mobilecoin/builder-install:v0.0.19
docker.io/mobilecoin/builder-install:v0.0.19
$ docker run --rm --env=ENTRYPOINT_VERBOSE=1 --volume=/home/jgreat/git/mcf/mobilecoin:/tmp/mobilenode --workdir=/tmp/mobilenode --env=EXTERNAL_UID=1000 --env=EXTERNAL_GID=1000 --env=EXTERNAL_USER=jgreat --env=EXTERNAL_GROUP=jgreat --env=CARGO_TARGET_DIR=/tmp/mobilenode/target/docker --env=MC_CHAIN_ID=local --env=TEST_DATABASE_URL=postgres://localhost --env=RUST_BACKTRACE=1 --env=CARGO_BUILD_JOBS=16 --env=SGX_MODE=SW --env=IAS_MODE=DEV --env=GIT_COMMIT=45b677f9 --cap-add=SYS_PTRACE -it --publish 8080:8080 --publish 8081:8081 --publish 8443:8443 --publish 3223:3223 --publish 3225:3225 --publish 3226:3226 --publish 3228:3228 --publish 4444:4444 --env=SSH_AUTH_SOCK --volume=/run/user/1000/keyring/ssh:/run/user/1000/keyring/ssh mobilecoin/builder-install:v0.0.19 /bin/bash
Executing /usr/local/bin/entrypoint.sh with command /bin/bash
-- Found User ID 1000, setting up and switching to that user.
-- Create group: 1000 jgreat
-- Create user: 1000 jgreat 
-- Set up .mob directory for cargo and build caching
-- Setup user .bashrc
-- Setup no password sudo access.
-- Set permissions for build tools.
-- Using /tmp/mobilenode as base directory
-- Change to external user
jgreat@69077b5999a1:/tmp/mobilenode$
```

run arbitrary command in container:
```bash
➜ ./mob "ls -al"
total 536
drwxrwxr-x 43 jgreat jgreat   4096 Nov 16 20:14 .
drwxrwxrwt  1 root   root     4096 Nov 16 22:58 ..
drwxrwxr-x  2 jgreat jgreat   4096 Oct 26 14:42 .cargo
drwxrwxr-x  2 jgreat jgreat   4096 Nov  8 17:01 .circleci
drwxrwxr-x  2 jgreat jgreat   4096 Nov  8 17:01 .config
...
```

run cargo commands:
```bash
➜ ./mob "cargo build"
  Downloaded indexmap v1.9.1
  Downloaded time v0.3.15
  Downloaded 2 crates (147.7 KB) in 0.40s
   Compiling indexmap v1.9.1
   Compiling ring v0.16.20
   Compiling time v0.3.15
...
```

use single quotes to reference a variable that's inside the container.
```bash
➜ echo $CARGO_TARGET_DIR
# < no set so no return>

# CARGO_TARGET_DIR is set in the container
➜ ./mob 'echo $CARGO_TARGET_DIR'
/tmp/mobilenode/target/docker
```

### Future Work

[Track playing at the time of this PR](https://open.spotify.com/track/2VtySl4QKhfIm7mJ6vfxkF)


joekottke (approved):
When I run it on macOS, I still see some of the warnings about `ssh-agent` settings (because you can't map the macOS socket into the Linux docker environment).  I'm not sure if this is important, or if it needs changing.

The command works and doesn't complain about IDs and such, it does spit out what appears to be the `/etc/group` line that will be used when I run it:

```
$ ./mob prompt
SSH Agent authentication disabled. You will need to run the following commands to build: 
eval `ssh-agent`
ssh-add /root/.ssh/<your-ssh-private-key>
staff:x:50:
otto@0b53ea44eb4d:/tmp/mobilenode$
```

jgreat:
> When I run it on macOS, I still see some of the warnings about `ssh-agent` settings (because you can't map the macOS socket into the Linux docker environment). I'm not sure if this is important, or if it needs changing.
> 
> The command works and doesn't complain about IDs and such, it does spit out what appears to be the `/etc/group` line that will be used when I run it:
> 
> ```
> $ ./mob prompt
> SSH Agent authentication disabled. You will need to run the following commands to build: 
> eval `ssh-agent`
> ssh-add /root/.ssh/<your-ssh-private-key>
> staff:x:50:
> otto@0b53ea44eb4d:/tmp/mobilenode$
> ```

Let me add an OS check around that part of the setup.  We'll just do the agent if we're on linux.

.mobconf
```diff
@@ -1,3 +1,9 @@
 [image]
 url = mobilecoin/builder-install
-tag = v0.0.18
+tag = v0.0.19
```
remoun, line 3: IIUC this `image` section isn't used

```suggestion
```

mob
```diff
@@ -209,157 +226,201 @@ while not os.path.exists(os.path.join(top_level, ".mobconf")):
...
+##
+# Set up docker environment and base options
+##
+ports = []
+container_ssh_dir = ""
+docker_run = ["docker", "run", "--rm"]
 
 ##
-# docker run flags
+# Turn off entrypoint messages with not in verbose mode
```
remoun, line 237:
```suggestion
# Turn off entrypoint messages when not in verbose mode
```

mob
```diff
@@ -209,157 +226,201 @@ while not os.path.exists(os.path.join(top_level, ".mobconf")):
...
+    else:
+        docker_run.append('--env=SGX_MODE=SW')
 
-for pair in build_env:
-    docker_run.extend(["--env", pair])
+    # set IAS_MODE options
+    if args.ias_prod:
+        docker_run.append('--env=IAS_MODE=PROD')
+    else:
+        docker_run.append('--env=IAS_MODE=DEV')
```
remoun, line 300: Optional/DRY: Merge append calls
```suggestion
    docker_run.append('--env=IAS_MODE={}'.format('PROD' if args.ias_prod else 'DEV'))
```

mob
```diff
@@ -209,157 +226,201 @@ while not os.path.exists(os.path.join(top_level, ".mobconf")):
...
-    docker_run.extend(["--volume", os.path.expanduser("~/.ssh") + ":" + "/root/.ssh"])
-elif args.ssh_agent:
-    docker_run.extend(["--env", "SSH_AUTH_SOCK=/tmp/ssh_auth_sock"])
-    docker_run.extend(["--volume", os.environ["SSH_AUTH_SOCK"] + ":" + "/tmp/ssh_auth_sock"])
+##
+# Add extra environment variables.
+##
+if args.env:
+    for e in args.env:
+        docker_run.append('--env={}'.format(e))
```
remoun, line 361: IIUC this `if` is unnecessary
```suggestion
for e in args.env:
    docker_run.append('--env={}'.format(e))
```

remoun (approved):

mob
jgreat: I think we should prefer more verbose syntax over condensed syntax for readability and long term maintenance.  

This is also why I also have moved to bash/shell statements on separate lines and avoid "one-liners" when possible.  I find coming back to "one-liners" after a couple of months requires a lot of decoding to figure out what was intended. 

I think it also helps those who are a bit more causal in the code base and bounce around between languages a lot.

mob
jgreat: We need an if here because arg.env isn't iterable unless you specify an option.

```
Traceback (most recent call last):
  File "/home/jgreat/git/mcf/mobilecoin/./mob", line 359, in <module>
    for e in args.env:
TypeError: 'NoneType' object is not iterable
```

.mobconf
jgreat: Left here for external scripts like in the release-scripts repo.  Will clean up in a later pr after we update or retire those scripts.

PR #2889, opened 2022-11-22
Pull in protobuf changes from milliec/ledger-router-dev

NotGyro:
- Defines necessary protobuf messages for key image lookups in a Fog ledger router/store system.
- Moves AddShardRequest to fog_common.proto (see #2888)

### Motivation

As part of the ongoing project to improve Fog's scalability and stability, we're working on implementing a system that divides Fog ledger servers into a router (which routes requests from the client to a number of ledger store backends) and many stores (which actually hold relevant ledger data). Currently, only key image look-ups are supported.

### Future Work

* Pulling in the rest of the fog-ledger-router work from [milliec/ledger-router-dev into this branch](https://github.com/mobilecoinfoundation/mobilecoin/tree/milliec/ledger-router-dev) - so enclave API changes, implementation of the new enclave API methods, grpc servers, tests, etc.

PR #2891, opened 2022-11-22
Protobuf conversion functions for Fog Ledger Router

awygle:
Pull in protobuf object conversion functions for Fog Ledger Router

### Motivation

As part of the Fog Ledger Router project, we need functions to convert from protobuf objects to various Rust objects. This pulls in the changes made to conversions.rs for upstreaming.

### Future Work

The next PR in this series will be the stubs of the APIs added to the Fog Ledger enclave to support the Fog Ledger Router.

### DIFF
--- a/fog/api/src/conversions.rs
+++ b/fog/api/src/conversions.rs
@@ -2,7 +2,9 @@
 //
 // Contains helper methods that enable conversions for Fog Api types.
 
-use crate::{fog_common, ingest_common, view::MultiViewStoreQueryRequest};
+use crate::{
+    fog_common, ingest_common, ledger::MultiKeyImageStoreRequest, view::MultiViewStoreQueryRequest,
+};
 use mc_api::ConversionError;
 use mc_attest_api::attest;
 use mc_attest_enclave_api::{EnclaveMessage, NonceSession};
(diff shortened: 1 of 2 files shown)


awygle:
* feature/fog-ledger-router
  * **PR #2889** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2889" target="_blank">Graphite</a> 
    * **PR #2891** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2891" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2891?utm_source=stack-comment).

fog/api/src/conversions.rs
```diff
@@ -30,6 +32,25 @@ impl From<Vec<attest::NonceMessage>> for MultiViewStoreQueryRequest {
...
+    fn from(enclave_messages: Vec<EnclaveMessage<NonceSession>>) -> MultiKeyImageStoreRequest {
+        enclave_messages
+            .into_iter()
+            .map(|enclave_message| enclave_message.into())
+            .collect::<Vec<attest::NonceMessage>>()
+            .into()
+    }
+}
+
+impl From<Vec<attest::NonceMessage>> for MultiKeyImageStoreRequest {
```
nick-mobilecoin, line 45: :question: I'm not opposed to this implementation as I'm generally pro functionalizing things.
Just curious if there is another use case for `Vec<attest::NonceMessage>` -> `MultiKeyImageStoreRequest` outside of the `Vec<EnclaveMessage<NonceSession>>` -> `MultiKeyImageStoreRequest`

nick-mobilecoin (approved):

fog/api/src/conversions.rs
NotGyro: A lot of Rust code (I can't be any less vague than this, I just see it very often) is genericized such that you can pass any type which implements `From<T>` to certain functions, to provide some extra convenience. `From` is reflexive, so you can pass `T` or anything that can be converted to `T` to a function which takes `From<T>` like that. 

This is just ergonomics, of course, and having a function that does this explicitly might be clearer / more readable than having the type system do it for you implicitly. I wouldn't really mind either way.

NotGyro (approved):
Looks good to me - although I wrote the first draft of this code, if I'm remembering correctly, so take my review with a grain of salt.

fog/api/src/conversions.rs
awygle: The conversions between AKE messages and domain-specific messages are pretty hairy in general, IMO. I believe that I copied an existing pattern for this.

We still `cargo check` without the `From<Vec<attest::NonceMessage>>` impl, so lifting the actual `MultiKeyImageStoreRequest` stuff into the other `From` block would work fine. But I think it's fine to leave this as-is.

nick-mobilecoin: sounds good I don't really have a strong preference one way or another

PR #2894, opened 2022-11-22
master (CI/CD) feat: CD don't use Github Environments

jgreat:
### Motivation

Undo using GitHub Environments feature.  It's nice for organizing the secrets in a clean way, but it just makes a mess of the PR conversation history because of its integration with Deployments. Every action that references an environment creates its own "Deployment" in the pr log.

- fix: `actionlint` errors and warnings from `shellcheck`
  - `actionlint` is checking shell code with `shellcheck` in GHA workflows.  This is good, but required some cleanup.
- fix: remove references to `dev` Environment
- fix: check and cleanup `secrets` names
  - Just update and remap some var names for better organization

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -177,13 +177,17 @@ jobs:
             grep -v -e mc-fog -e mc-consensus | \
             awk "{ print \"-p \" \$1 }" | \
             sort > /tmp/test-packages
-          split -n l/$RUNNER_INDEX/$NUM_RUNNERS /tmp/test-packages | \
+
+          split -n "l/$RUNNER_INDEX/$NUM_RUNNERS" /tmp/test-packages | \
             tee /tmp/mc-test-packages
+
           # Hack: mc-util-sample-ledger needs mc-util-keyfile bins.
           # TODO: Replace with artifact deps when that does not require
           # additional cargo flags.
-          grep -q generate-sample-ledger /tmp/mc-test-packages && \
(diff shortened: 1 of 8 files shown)


nick-mobilecoin (approved):

eranrund (approved):

.github/workflows/ci.yml
```diff
@@ -623,27 +630,27 @@ jobs:
...
           # Mint 1 million token1's to the first 4 accounts
           echo "Minting"
           for ACCOUNT_NUM in $(seq 0 4); do
               "$BIN_DIR/mc-consensus-mint-client" \
-                  "generate-and-submit-mint-tx" \
+                  generate-and-submit-mint-tx \
                   --node insecure-mc://localhost:3200/ \
                   --signing-key "$BIN_DIR/mc-local-network/minting-keys/minter1" \
-                  --recipient $(cat keys/account_keys_${ACCOUNT_NUM}.b58pub) \
+                  --recipient "$(cat "keys/account_keys_${ACCOUNT_NUM}.b58pub")" \
```
joekottke, line 643: This looks like it might have issues with quoting and escaping?

joekottke (approved):
One nit, as I think it might not work due to quoting and escaping.  But I could be wrong.

.github/workflows/ci.yml
jgreat: This does work, its weird, but because of the `()` the internally quoted considered separately.

Issue #2577, opened 2022-09-20
Consensus Node - Prefix existing metrics with "consensus_service"


jgreat:
There are a bunch of metrics that the consensus service presents, that have very generic names.  We should prefix all the metrics presented by a service with the service name. 

example from consensus 2.0

```
# HELP duration Duration for a request, in units of time
# TYPE duration histogram
duration_bucket{method="admin.AdminAPI.GetInfo",le="0.005"} 1
duration_bucket{method="admin.AdminAPI.GetInfo",le="0.01"} 1
duration_bucket{method="admin.AdminAPI.GetInfo",le="0.025"} 1
duration_bucket{method="admin.AdminAPI.GetInfo",le="0.05"} 1
duration_bucket{method="admin.AdminAPI.GetInfo",le="0.1"} 1
duration_bucket{method="admin.AdminAPI.GetInfo",le="0.25"} 1
duration_bucket{method="admin.AdminAPI.GetInfo",le="0.5"} 1
duration_bucket{method="admin.AdminAPI.GetInfo",le="1"} 1
duration_bucket{method="admin.AdminAPI.GetInfo",le="2.5"} 1
duration_bucket{method="admin.AdminAPI.GetInfo",le="5"} 1
duration_bucket{method="admin.AdminAPI.GetInfo",le="10"} 1
duration_bucket{method="admin.AdminAPI.GetInfo",le="+Inf"} 1
duration_sum{method="admin.AdminAPI.GetInfo"} 0.000241801
duration_count{method="admin.AdminAPI.GetInfo"} 1
duration_bucket{method="admin.AdminAPI.GetPrometheusMetrics",le="0.005"} 9371
duration_bucket{method="admin.AdminAPI.GetPrometheusMetrics",le="0.01"} 9371
duration_bucket{method="admin.AdminAPI.GetPrometheusMetrics",le="0.025"} 9371
duration_bucket{method="admin.AdminAPI.GetPrometheusMetrics",le="0.05"} 9371
duration_bucket{method="admin.AdminAPI.GetPrometheusMetrics",le="0.1"} 9371
duration_bucket{method="admin.AdminAPI.GetPrometheusMetrics",le="0.25"} 9371
duration_bucket{method="admin.AdminAPI.GetPrometheusMetrics",le="0.5"} 9371
duration_bucket{method="admin.AdminAPI.GetPrometheusMetrics",le="1"} 9371
[...]

PR #2908, opened 2022-11-29
Metrics with service names

varsha888:
### Motivation

Prepend service names to metrics for better tracking info
https://github.com/mobilecoinfoundation/mobilecoin/issues/2577 

Screen Shot 2022-11-30 at 1 51 54 PM


fog/ledger/server/src/lib.rs
```diff
@@ -9,9 +9,15 @@ mod merkle_proof_service;
...
+
 pub use block_service::BlockService;
 pub use config::LedgerServerConfig;
 pub use key_image_service::KeyImageService;
 pub use merkle_proof_service::MerkleProofService;
 pub use server::LedgerServer;
 pub use untrusted_tx_out_service::UntrustedTxOutService;
+
+lazy_static::lazy_static! {
+    pub static ref SVC_COUNTERS: ServiceMetrics = ServiceMetrics::new_and_registered("ledger");
```
jgreat, line 22: can we make this `fog_ledger` ?

fog/ingest/server/src/lib.rs
```diff
@@ -31,6 +31,12 @@ mod worker;
 
 use core::fmt::Display;
 use itertools::Itertools;
+use mc_util_metrics::ServiceMetrics;
+
+lazy_static::lazy_static! {
+    /// Generates service metrics for tracking
+    pub static ref SVC_COUNTERS: ServiceMetrics = ServiceMetrics::new_and_registered("ingest");
```
jgreat, line 38: Can we make this `fog_ingest`?

fog/report/server/src/lib.rs
```diff
@@ -12,3 +12,10 @@ pub use crate::{
     config::{Config, Error, Materials},
     server::Server,
 };
+
+use mc_util_metrics::ServiceMetrics;
+
+lazy_static::lazy_static! {
+    /// Generates service metrics for tracking
+    pub static ref SVC_COUNTERS: ServiceMetrics = ServiceMetrics::new_and_registered("report");
```
jgreat, line 20: Can we make this `fog_report`?

util/grpc/src/lib.rs
```diff
@@ -308,6 +308,8 @@ pub fn rpc_logger(ctx: &RpcContext, logger: &Logger) -> Logger {
 }
 
 lazy_static::lazy_static! {
+    /// Generates service metrics for tracking
+    pub static ref SVC_COUNTERS: ServiceMetrics = ServiceMetrics::new_and_registered("grpc");
```
jgreat, line 312: I see the screen shot shows metrics from the AdminAPI, is this used across projects for grpc services/routes? 

Ideally this would be prefixed with the service name and not just `grpc`.  Something like `consensus_grpc` or `fog_ledger_grpc`.  Is there a way to pass the service name in as a parameter when these metrics get registered?

In any case this is better that just  `duration_bucket`  :+1:

fog/ledger/server/src/lib.rs
varsha888: will do 👍

util/grpc/src/lib.rs
varsha888: I'll check if I can pass the service name as a param as Its a bigger change if it has to be `consensus_grpc` etc., as its used across the project.

fog/report/server/src/service.rs
```diff
@@ -16,9 +16,9 @@ use mc_fog_sig_report::Signer as ReportSigner;
 use mc_util_grpc::{
     check_request_chain_id, rpc_database_err, rpc_internal_error, rpc_logger, send_result,
 };
-use mc_util_metrics::SVC_COUNTERS;
 use prost::DecodeError;
 use signature::{Error as SignatureError, Signature};
+use crate::SVC_COUNTERS;
```
iamalwaysuncomfortable, line 21: Nit: crate imports should be placed before other imports

util/metrics/src/service_metrics.rs
```diff
@@ -64,31 +65,45 @@ pub struct ServiceMetrics {
...
 }
+impl Default for ServiceMetrics {
+    fn default() -> Self {
+        let args = std::env::args().next().unwrap_or_default();
+        let mut arg = Path::new(&args).file_stem().unwrap_or_default().to_str().unwrap_or_default().replace('-', "_");
+        if !arg.is_empty() {
+            arg.push('_');
+        }
+        arg.push_str("grpc");
+        let svc = ServiceMetrics::new(arg);
```
awygle, line 76: It looks like this call could be to `new_and_registered` instead of replicating the logic here.

util/grpc/src/lib.rs
```diff
@@ -308,6 +308,9 @@ pub fn rpc_logger(ctx: &RpcContext, logger: &Logger) -> Logger {
 }
 
 lazy_static::lazy_static! {
+    /// Generates service metrics for tracking
+    pub static ref SVC_COUNTERS: ServiceMetrics = ServiceMetrics::default();
+    // call default insteadof new and registered
```
awygle, line 313: Nit: I'd include why you're calling a different method here in the comment, if it's important. I'd also move the comment above the definition and add a newline between the definition and the start of the `RPC_LOGGER_CLIENT_ID_SEED` comment.

awygle (approved):
This looks good implementation-wise, I just had a couple of minor suggestions.

jgreat (approved):

jcape:
Unfortunately this isn't passing CI, and it looks like it's hanging in `consensus-tests 2, 2` because `Cancelled after 360m` (i.e. all the tests hang for 6 hours), my suspicion here is that something about that test is preventing us from having two `SVC_COUNTERS` variables active at a given time (i.e. the test assumes there's only one).

PR #2910, opened 2022-11-30
Update to newest `rust-mbedtls`

the-real-adammork:
### Motivation

I made a patch to our `rust-mbedtls` crate that [fixes big-num multiplication](https://github.com/mobilecoinfoundation/rust-mbedtls/pull/66) in static libraries compiled on w/ macOS. This brings that update to the `mobilecoin` repo. I had to upgrade our version of `heapless` to 0.8 to resolve a dependency issue with `spin`.

removed `cmake` revision patch, this isn't needed in the main `mobilecoin` repo, it will be specified in the `libmobilecoin` repo instead.

[Los York's - Cielo = Sunny](https://www.youtube.com/watch?v=078j7mERXVY)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1839,14 +1839,12 @@ dependencies = [
 
 [[package]]
 name = "heapless"
-version = "0.7.16"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


the-real-adammork:
might need to make one more change, closing for now

PR #2896, opened 2022-11-23
Ledger enclave support for router and store

NotGyro:
- Adds `ledger_store_init()`, `frontend_accept()`, and `ledger_store_connect()` to the Fog ledger enclave API for general use with a fog ledger router/store system
- Adds `check_key_image_store()`, `decrypt_and_seal_query()`, `create_multi_key_image_store_query_data()`, and `collate_shard_query_responses()` to the Fog ledger enclave API, which are currently specific to key-image store queries
- Implements all of the above new API methods, including stub methods for testing. 
- New EnclaveCall variants for the above methods. 

Depends on #2889

### Motivation

As part of the ongoing project to improve Fog's scalability and stability, we're working on implementing a system that divides Fog ledger servers into a router (which routes requests from the client to a number of ledger store backends) and many stores (which actually hold relevant ledger data). Currently, only key image look-ups are supported.

### Future Work

* Generalize such that the ledger router/store system can be used for other ledger queries, not just key image queries (for example, merkele proofs).

PR #2913, opened 2022-12-02
Oblivious collation of key image store responses

awygle:
Obliviously collates responses from Key Image Stores, to maintain security properties under the new router/store model.

### Motivation

We want to maintain our security properties even though we are splitting out the Fog Ledger service into a router/store model. In order to do this, we obliviously collate the responses from Key Image Stores in encrypted SGX memory so that an attacker cannot determine which shard a given response came from.

### DIFF
--- a/fog/ledger/enclave/impl/src/lib.rs
+++ b/fog/ledger/enclave/impl/src/lib.rs
@@ -38,6 +38,8 @@ use mc_oblivious_traits::ORAMStorageCreator;
 use mc_sgx_compat::sync::Mutex;
 use mc_sgx_report_cache_api::{ReportableEnclave, Result as ReportableEnclaveResult};
 
+mod oblivious_utils;
+
 /// In-enclave state associated to the ledger enclaves
 pub struct SgxLedgerEnclave<OSC>
 where
@@ -237,7 +239,7 @@ where
         }
         let channel_id = sealed_query.channel_id.clone();
         let client_query_plaintext = self.ake.unseal(&sealed_query)?;
-        let _client_query_request: CheckKeyImagesRequest =
+        let client_query_request: CheckKeyImagesRequest =
             mc_util_serial::decode(&client_query_plaintext).map_err(|e| {
                 log::error!(self.logger, "Could not decode client query request: {}", e);
                 Error::ProstDecode
@@ -256,42 +258,44 @@ where
 
         // NOTES:
         // num_blocks = min(responses.num_blocks)
-        // global_txo_count = min(global_txo_count) TODO CONFIRM
+        // global_txo_count = min(global_txo_count)
         // results = cat(responses.results)
         // latest_block_version = max(responses.latest_block_version)
         // max_block_version = max(latest_block_version,
-        // mc_transaction_core::MAX_BLOCK_VERSION
+        // mc_transaction_core::MAX_BLOCK_VERSION)
 
-        // TODO no unwraps
         let num_blocks = shard_query_responses
             .iter()
             .map(|query_response| query_response.num_blocks)
             .min()
-            .unwrap();
+            .expect("this is only None when the iterator is empty but we early-exit in that case");
         let global_txo_count = shard_query_responses
             .iter()
             .map(|query_response| query_response.global_txo_count)
             .min()
-            .unwrap();
+            .expect("this is only None when the iterator is empty but we early-exit in that case");
         let latest_block_version = shard_query_responses
             .iter()
             .map(|query_response| query_response.latest_block_version)
             .max()
-            .unwrap();
-        // TODO I believe this needs to be implemented in an oblivious way to meet the
-        // security requirements. I'm not 100% sure what an oblivious approach
-        // to this looks like, though. In general this kind of thing needs to be
-        // talked about.
-        let results = shard_query_responses
+            .expect("this is only None when the iterator is empty but we early-exit in that case");
+
+        let plaintext_results = shard_query_responses
             .into_iter()
             .flat_map(|query_response| query_response.results)
             .collect();
+
+        let oblivious_results = oblivious_utils::collate_shard_key_image_search_results(
+            client_query_request.queries,
+            plaintext_results,
+        );
+
         let max_block_version = max(latest_block_version, *MAX_BLOCK_VERSION);
 
         let client_query_response = CheckKeyImagesResponse {
             num_blocks,
             global_txo_count,
-            results,
+            results: oblivious_results,
             latest_block_version,
             max_block_version,
         };
--- a/fog/ledger/enclave/impl/src/lib.rs
+++ b/fog/ledger/enclave/impl/src/lib.rs
@@ -29,7 +29,8 @@ use mc_common::{
 use mc_crypto_ake_enclave::{AkeEnclaveState, NullIdentity};
 use mc_crypto_keys::X25519Public;
 use mc_fog_ledger_enclave_api::{
-    Error, KeyImageData, LedgerEnclave, OutputContext, Result, UntrustedKeyImageQueryResponse,
+    Error, KeyImageData, KeyImageResult, LedgerEnclave, OutputContext, Result,
+    UntrustedKeyImageQueryResponse,
 };
 use mc_fog_types::ledger::{
     CheckKeyImagesRequest, CheckKeyImagesResponse, GetOutputsRequest, GetOutputsResponse,
@@ -280,14 +281,14 @@ where
             .max()
             .expect("this is only None when the iterator is empty but we early-exit in that case");
 
-        let plaintext_results = shard_query_responses
+        let plaintext_results: Vec<KeyImageResult> = shard_query_responses
             .into_iter()
             .flat_map(|query_response| query_response.results)
             .collect();
 
         let oblivious_results = oblivious_utils::collate_shard_key_image_search_results(
             client_query_request.queries,
-            plaintext_results,
+            &plaintext_results,
         );
 
         let max_block_version = max(latest_block_version, *MAX_BLOCK_VERSION);
--- a/fog/ledger/enclave/impl/src/oblivious_utils.rs
+++ b/fog/ledger/enclave/impl/src/oblivious_utils.rs
@@ -1,0 +1,454 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
+//! Contains methods that allow a Fog View Router enclave to combine all of the
+//! Fog View Shard's query responses into one query response that'll be returned
+//! for the client.
+
+use aligned_cmov::{
+    subtle::{Choice, ConstantTimeEq},
+    CMov,
+};
+use alloc::vec::Vec;
+use mc_fog_types::ledger::{KeyImageQuery, KeyImageResult, KeyImageResultCode};
+use mc_watcher_api::TimestampResultCode;
+
+/// The default KeyImageResultCode used when collating the shard responses.
+const DEFAULT_KEY_IMAGE_SEARCH_RESULT_CODE: KeyImageResultCode = KeyImageResultCode::NotSpent;
+
+pub fn collate_shard_key_image_search_results(
+    client_queries: Vec<KeyImageQuery>,
+    shard_key_image_search_results: Vec<KeyImageResult>,
+) -> Vec<KeyImageResult> {
+    let mut client_key_image_search_results: Vec<KeyImageResult> = client_queries
+        .iter()
+        .map(|client_query| KeyImageResult {
+            key_image: client_query.key_image,
+            spent_at: 1, // not 0 because it's defined to be >0 in the .proto file
+            timestamp: u64::MAX,
+            timestamp_result_code: TimestampResultCode::TimestampFound as u32,
+            key_image_result_code: DEFAULT_KEY_IMAGE_SEARCH_RESULT_CODE as u32,
+        })
+        .collect();
+
+    for shard_key_image_search_result in shard_key_image_search_results.iter() {
+        for client_key_image_search_result in client_key_image_search_results.iter_mut() {
+            maybe_overwrite_key_image_search_result(
+                client_key_image_search_result,
+                shard_key_image_search_result,
+            );
+        }
+    }
+
+    client_key_image_search_results
+}
+
+fn maybe_overwrite_key_image_search_result(
+    client_key_image_search_result: &mut KeyImageResult,
+    shard_key_image_search_result: &KeyImageResult,
+) {
+    let should_overwrite_key_image_search_result = should_overwrite_key_image_search_result(
+        client_key_image_search_result,
+        shard_key_image_search_result,
+    );
+
+    client_key_image_search_result.key_image_result_code.cmov(
+        should_overwrite_key_image_search_result,
+        &shard_key_image_search_result.key_image_result_code,
+    );
+
+    client_key_image_search_result.spent_at.cmov(
+        should_overwrite_key_image_search_result,
+        &shard_key_image_search_result.spent_at,
+    );
+
+    client_key_image_search_result.timestamp.cmov(
+        should_overwrite_key_image_search_result,
+        &shard_key_image_search_result.timestamp,
+    );
+
+    client_key_image_search_result.timestamp_result_code.cmov(
+        should_overwrite_key_image_search_result,
+        &shard_key_image_search_result.timestamp_result_code,
+    );
+}
+
+fn should_overwrite_key_image_search_result(
+    client_key_image_search_result: &KeyImageResult,
+    shard_key_image_search_result: &KeyImageResult,
+) -> Choice {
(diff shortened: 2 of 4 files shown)


awygle:
* feature/fog-ledger-router
  * **PR #2889** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2889" target="_blank">Graphite</a> 
    * **PR #2896** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2896" target="_blank">Graphite</a> 
      * **PR #2897** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2897" target="_blank">Graphite</a> 
        * **PR #2891** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2891" target="_blank">Graphite</a> 
          * **PR #2898** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2898" target="_blank">Graphite</a> 
            * **PR #2913** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2913" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2913?utm_source=stack-comment).

awygle:
* feature/fog-ledger-router
  * **PR #2889** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2889" target="_blank">Graphite</a> 
    * **PR #2896** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2896" target="_blank">Graphite</a> 
      * **PR #2897** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2897" target="_blank">Graphite</a> 
        * **PR #2891** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2891" target="_blank">Graphite</a> 
          * **PR #2898** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2898" target="_blank">Graphite</a> 
            * **PR #2913** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2913" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2913?utm_source=stack-comment).

fog/ledger/enclave/impl/src/oblivious_utils.rs
```diff
@@ -0,0 +1,454 @@
...
+                spent_at: shard_spent_at,
+                timestamp: shard_timestamp,
+                timestamp_result_code: shard_timestamp_result_code as u32,
+                key_image_result_code: shard_key_image_result_code as u32,
+            };
+            let result: bool =
+                should_overwrite_key_image_search_result(&client_result, &shard_result).into();
+            assert_eq!(result, expected_result, "{}", panic_message);
+        }
+    }
```
nick-mobilecoin, line 258: :wrench: A couple of flags on this test that makes it hard to parse:
- The tuples with 12 items, generally functions should be kept to 5 or less arguments. I would argue tuples should be kept to 3 or less.
- The for looping over a vector for test cases, that often indicates parameterized test cases, but not always
For this suggestion:
- I created a function `default_client_key_image()` and I updated the `Vec` in `collate_shard_key_image_search_results()` to use it.
- For boolean expected values it's often clearer to have separate test functions instead of passing the expected as one of the parameters
- With dedicated parameterized tests names i think the failure messages can be omitted
- it's a bit easier to see that all 9 permutations  + keys not matching are accounted for (3 permutations were missing from previous)
```suggestion
    use yare::parameterized;

    #[test]
    fn differing_key_images_do_not_update() {
        let client_result = default_client_key_image(1.into());
        let shard_result = default_client_key_image(2.into());

        let result: bool =
            should_overwrite_key_image_search_result(&client_result, &shard_result).into();
        assert_eq!(result, false);
    }

    #[parameterized(
    client_default_shard_is_spent = { KeyImageResultCode::NotSpent, KeyImageResultCode::Spent},
    client_default_shard_is_error = { KeyImageResultCode::NotSpent, KeyImageResultCode::KeyImageError},
    client_error_shard_is_spent = { KeyImageResultCode::KeyImageError, KeyImageResultCode::Spent},
    client_spent_shard_is_spent = { KeyImageResultCode::Spent, KeyImageResultCode::Spent},
    )]
    fn should_update(client_code: KeyImageResultCode, shard_code: KeyImageResultCode) {
        let mut client_result = default_client_key_image(1.into());
        let mut shard_result = client_result.clone();
        client_result.key_image_result_code = client_code as u32;
        shard_result.key_image_result_code = shard_code as u32;

        let result: bool =
            should_overwrite_key_image_search_result(&client_result, &shard_result).into();
        assert_eq!(result, true);
    }

    #[parameterized(
    client_default_shard_is_not_spent = { KeyImageResultCode::NotSpent, KeyImageResultCode::NotSpent},
    client_error_shard_is_not_spent = { KeyImageResultCode::KeyImageError, KeyImageResultCode::NotSpent},
    client_error_shard_is_error = { KeyImageResultCode::KeyImageError, KeyImageResultCode::KeyImageError},
    client_spent_shard_is_not_spent = { KeyImageResultCode::Spent, KeyImageResultCode::NotSpent},
    client_spent_shard_is_error = { KeyImageResultCode::Spent, KeyImageResultCode::KeyImageError},
    )]
    fn should_not_update(client_code: KeyImageResultCode, shard_code: KeyImageResultCode) {
        let mut client_result = default_client_key_image(1.into());
        let mut shard_result = client_result.clone();
        client_result.key_image_result_code = client_code as u32;
        shard_result.key_image_result_code = shard_code as u32;

        let result: bool =
            should_overwrite_key_image_search_result(&client_result, &shard_result).into();
        assert_eq!(result, false);
    }
```

fog/ledger/enclave/impl/src/oblivious_utils.rs
```diff
@@ -0,0 +1,454 @@
...
+use alloc::vec::Vec;
+use mc_fog_types::ledger::{KeyImageQuery, KeyImageResult, KeyImageResultCode};
+use mc_watcher_api::TimestampResultCode;
+
+/// The default KeyImageResultCode used when collating the shard responses.
+const DEFAULT_KEY_IMAGE_SEARCH_RESULT_CODE: KeyImageResultCode = KeyImageResultCode::NotSpent;
+
+pub fn collate_shard_key_image_search_results(
+    client_queries: Vec<KeyImageQuery>,
+    shard_key_image_search_results: Vec<KeyImageResult>,
```
nick-mobilecoin, line 20: :wrench: Does this need to take ownership?
```suggestion
    shard_key_image_search_results: &Vec<KeyImageResult>,
```

fog/ledger/enclave/impl/src/oblivious_utils.rs
```diff
@@ -0,0 +1,454 @@
...
+
+        let all_images_found = result.iter().all(|key_image_result| {
+            key_image_result.key_image_result_code == KeyImageResultCode::Spent as u32
+        });
+
+        assert!(
+            all_images_found,
+            "{}",
+            "all images are present but some were not collated"
+        );
```
nick-mobilecoin, line 304: :wrench: A couple things on these _tests_:
- It looks like after each assert it's a new test case, suggest making each one it's own function
- Be more explicit on the expected, it looks like only the counts and the result code is compared, but the entire image should be copied over.
- There is no test for mixed results codes, i.e. image 1 has spent, 2 has error

Suggest something like the following:
- it makes it easier to have more than one item per Vec
- it makes it more explicit what the expected is
- The duplicate entries are more explicit.
- Chose a `TimestampResultCode` that differs from the default (only needs to be done in one test case
```suggestion
    #[test]
    fn all_available() {
        let range = 1..=3;
        let client_queries = range
            .clone()
            .map(|key_image| KeyImageQuery {
                key_image: key_image.into(),
                start_block: 0,
            })
            .collect::<Vec<_>>();
        let mut shard_results = range
            .map(|key_image| KeyImageResult {
                key_image: key_image.into(),
                spent_at: key_image + 1,
                timestamp: key_image + 10,
                timestamp_result_code: TimestampResultCode::WatcherBehind as u32,
                key_image_result_code: KeyImageResultCode::Spent as u32,
            })
            .collect::<Vec<_>>();

        let mut results = collate_shard_key_image_search_results(client_queries, &shard_results);
        results.sort_by_key(|r| r.key_image);
        shard_results.sort_by_key(|r| r.key_image);
        assert_eq!(results, shard_results);
    }

    #[test]
    fn duplicate_shard_results_returns_one_result() {
        let client_queries = vec![KeyImageQuery {
            key_image: 1.into(),
            start_block: 0,
        }];
        let key_image_result = KeyImageResult {
            key_image: 1.into(),
            spent_at: 2,
            timestamp: 3,
            timestamp_result_code: TimestampResultCode::WatcherBehind as u32,
            key_image_result_code: KeyImageResultCode::Spent as u32,
        };
        let shard_results = vec![key_image_result.clone(), key_image_result.clone()];
        let mut results = collate_shard_key_image_search_results(client_queries, &shard_results);
        results.sort_by_key(|r| r.key_image);
        assert_eq!(results, vec![key_image_result]);
    }
```

fog/ledger/enclave/impl/src/oblivious_utils.rs
```diff
@@ -0,0 +1,454 @@
...
+
+#[cfg(test)]
+mod tests {
+    extern crate std;
+
+    use super::*;
+    use std::vec;
+
+    #[test]
+    fn should_overwrite_tests() {
```
nick-mobilecoin, line 116: :+1: I like how this behavior was separated out into it's own test

fog/ledger/enclave/impl/src/oblivious_utils.rs
```diff
@@ -0,0 +1,454 @@
...
+
+        let all_images_found = result.iter().all(|key_image_result| {
+            key_image_result.key_image_result_code == KeyImageResultCode::Spent as u32
+        });
+
+        assert!(
+            !all_images_found,
+            "{}",
+            "all images show results but no result provided"
+        );
```
nick-mobilecoin, line 394: :wrench: If this stays in it's current form it should be _positively_ checking, i.e. it should probably be looking for `NotSpent`, but I would lean towards the more explicit version below
```suggestion
        assert_eq!(result, vec![default_client_key_image(1.into())]);
```

fog/ledger/enclave/impl/src/oblivious_utils.rs
```diff
@@ -0,0 +1,454 @@
...
+
+        let all_images_found = result.iter().all(|key_image_result| {
+            key_image_result.key_image_result_code == KeyImageResultCode::Spent as u32
+        });
+
+        assert!(
+            !all_images_found,
+            "{}",
+            "all images show results but some had no result provided"
+        );
```
nick-mobilecoin, line 452: :wrench: Similar to the not found case above this should be asserting that 1 == spent and 2 == not spent or the default

nick-mobilecoin (approved):
Mostly suggestions on testing

fog/ledger/enclave/impl/src/oblivious_utils.rs
```diff
@@ -0,0 +1,454 @@
...
+    //   At a high level, we want to prioritize "spent" responses.
+    // Error responses are "retriable" errors that the client will retry
+    // after a backoff. The "not spent" response is the default response and
+    // gets overwritten by any other response.
+    // spent > error > not spent
+    do_key_images_match
+           // Always write a Found code
+        & (shard_code_is_spent
+            // Write an error code IFF the client code is NotFound.
+            | ((shard_code_is_error) & client_code_is_not_spent))
```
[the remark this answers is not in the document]
samdealy: Re: "client_code_is_not_spent generally suggest trying to avoid negative names" I tend to agree but I think in this case it's justified because the thing itself is called `NotSpent` and if we call the variable `shard_code_is_spent` then that implies that the thing is `KeyImageResultCode::Spent`, which it might not be. This variable in this case means that it has a result code of `NotSpent`, and it doesn't mean that it has a result code of `Spent.`

fog/ledger/enclave/impl/src/oblivious_utils.rs
```diff
@@ -0,0 +1,454 @@
...
+                TimestampResultCode::TimestampFound,
+                KeyImageResultCode::NotSpent,
+                654321,
+                1,
+                123456,
+                TimestampResultCode::TimestampFound,
+                KeyImageResultCode::Spent,
+                false,
+                "key images don't match, but overwritten anyway!",
+            ),
```
samdealy, line 132: Nit: Make a struct that holds these values. It's difficult to understand what these variables mean without looking at the for loop below. This would also simplify the for loop and allow you to access fields rather than writing variable names in the tuple.

samdealy (approved):

fog/ledger/enclave/impl/src/lib.rs
```diff
@@ -256,42 +258,44 @@ where
...
-        // mc_transaction_core::MAX_BLOCK_VERSION
+        // mc_transaction_core::MAX_BLOCK_VERSION)
 
-        // TODO no unwraps
         let num_blocks = shard_query_responses
             .iter()
             .map(|query_response| query_response.num_blocks)
             .min()
-            .unwrap();
+            .expect("this is only None when the iterator is empty but we early-exit in that case");
```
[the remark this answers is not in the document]
jcape: FWIW, I think this is fine to leave as `expect()`. I'd rather have the enclave panic when something like this happens vs. continuing with `num_blocks = 0`.

fog/ledger/enclave/impl/src/oblivious_utils.rs
jcape: It doesn't, based on line 33, but as a general rule you want `&[Foo]` instead of a `&Vec<Foo>`
```suggestion
    shard_key_image_search_results: &[KeyImageResult],
```

fog/ledger/enclave/impl/src/oblivious_utils.rs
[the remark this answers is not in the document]
jcape: No need to guess about what the compiler will do (also recommend doing the more verbose version w/ better names):

https://godbolt.org/z/zKzGjGh96

fog/ledger/enclave/impl/src/oblivious_utils.rs
jcape: 💯

jcape (approved):
Approved assuming comments are addressed.

fog/ledger/enclave/impl/src/oblivious_utils.rs
```diff
@@ -0,0 +1,454 @@
...
+) -> Choice {
+    let client_key_image: &[u8] = client_key_image_search_result.key_image.as_ref();
+    let shard_key_image: &[u8] = shard_key_image_search_result.key_image.as_ref();
+    let do_key_images_match = client_key_image.ct_eq(shard_key_image);
+
+    let client_key_image_search_result_code = client_key_image_search_result.key_image_result_code;
+    let shard_key_image_search_result_code = shard_key_image_search_result.key_image_result_code;
+
+    let client_code_is_not_spent: Choice =
+        client_key_image_search_result_code.ct_eq(&(KeyImageResultCode::NotSpent as u32));
```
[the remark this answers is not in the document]
awygle: Hm, I could definitely argue this either way but I don't have strong feelings either way. I'll make this change.

fog/ledger/enclave/impl/src/oblivious_utils.rs
```diff
@@ -0,0 +1,276 @@
...
+        let results = collate_shard_key_image_search_results(client_queries, &shard_results);
+
+        assert_eq!(
+            results[0].key_image_result_code,
+            KeyImageResultCode::Spent as u32
+        );
+        assert_eq!(
+            results[1].key_image_result_code,
+            DEFAULT_KEY_IMAGE_SEARCH_RESULT_CODE as u32
+        );
```
nick-mobilecoin, line 274: :wrench: Consider the following to more explicitly ensure only 2 entries in the results and that all the fields were populated.  For the all the fields correctly, I don't think it's necessary since other tests more or less checked that, but it keeps this test consistent with the others.
```suggestion
        let shard_results = vec![key_image_result.clone()];

        let mut results = collate_shard_key_image_search_results(client_queries, &shard_results);
        results.sort_by_key(|r| r.key_image);
        assert_eq!(
            results,
            vec![key_image_result, default_client_key_image(2.into())]
        );
```

nick-mobilecoin (approved):
LGTM thanks!

Issue #2717, opened 2022-10-17
Using zeroize-on-drop in mc-transaction-core


garbageslam:
Historically, most objects in transaction-core did not implement `Zeroize`.
This is because `curve25519-dalek::RistrettoPoint` did not implement `Zeroize` and so it was impossible.

Now, it does implement `Zeroize`, but not `drop(Zeroize)`. This means, we know how to zeroize it, but it is not automatically zeroized on drop.

Generally, we want "secrets" to be zeroized on drop. However, in a typical transaction almost everything is a secret in our model. Not only are your private keys a secret, but also who you are sending money to is a secret.
* `tx_private_key` should be zeroized on drop
* `tx_out_shared_secret` should be zeroized on drop
* Your friend's public address should be zeroized on drop
* Which inputs you selected to spend, and which ring input is the real input, should be zeroized on drop.
* Pseudo-output blinding factors should be zeroized on drop.
* Possibly the whole Tx should be zeroized, but this might be overkill.

Zeroizing on drop has additional consequences.
* `Drop` types cannot be `Copy`. Some types like `CompressedRistrettoPublic` are `Copy`.
* I am not sure how it interacts with `Prost`.

This issue tracks discussion and implementation around:
- [ ] What types if any should be Zeroized on Drop?

PR #2925, opened 2022-12-05
Make the `Tx` object zeroize itself on drop

cbeck88:
This is an experimental change to improve hygeine of memory in the consensus enclave (but also in the clients)

This makes a bunch of objects derive `Zeroize` in transaction-core, and makes the `Tx` object zeroize itself on drop.

For examples of why this is a good change, consider that, in the `mc-connection` ThickClient crate, after we serialize a Tx to protobuf and encrypt it for the enclave, we zeroize the plaintext. However, before this change, we do not zeroize the `Tx` object (since it would not have been possible), so another copy of the data that we zeroized continues to exist in memory. After this change, the `Tx` will zeroize itself automatically as well.

The main reason not to do this would be if it hurts the performance of the consensus nodes, but I consider this unlikely, because, the untrusted side never actually sees `Tx` objects so SCP should not be meaningfully impacted by this, only the transaction validation handles the decrypted `Tx`. And the elliptic curve operations done to validate a Tx should be many orders of magnitude more expensive than a zeroizing operation. So I expect no measurable performance difference as a result of this change.

See github issue #2717

### Future Work

[Soundtrack of this PR]()

### DIFF
--- a/ledger/db/src/test_utils/mod.rs
+++ b/ledger/db/src/test_utils/mod.rs
@@ -310,7 +310,7 @@ pub fn initialize_ledger(
                 );
 
                 let key_images = tx.key_images();
-                (tx.prefix.outputs, key_images)
+                (tx.prefix.outputs.clone(), key_images)
             }
             None => {
                 // Create an origin block.
--- a/transaction/core/src/tx.rs
+++ b/transaction/core/src/tx.rs
@@ -103,7 +103,8 @@ impl fmt::Debug for TxHash {
 }
 
 /// A CryptoNote-style transaction.
-#[derive(Clone, Eq, PartialEq, Serialize, Deserialize, Message, Digestible)]
+#[derive(Clone, Eq, PartialEq, Serialize, Deserialize, Message, Digestible, Zeroize)]
+#[zeroize(drop)]
 pub struct Tx {
     /// The transaction contents.
     #[prost(message, required, tag = "1")]
@@ -160,7 +161,7 @@ impl Tx {
 ///
 /// Note: If you add something here, consider if it should be added to the
 /// TxSummary also for hardware wallet visibility.
-#[derive(Clone, Deserialize, Eq, PartialEq, Serialize, Message, Digestible)]
+#[derive(Clone, Deserialize, Digestible, Eq, Message, PartialEq, Serialize, Zeroize)]
 pub struct TxPrefix {
     /// List of inputs to the transaction.
     #[prost(message, repeated, tag = "1")]
@@ -244,7 +245,7 @@ impl TxPrefix {
 }
 
 /// An "input" to a transaction.
-#[derive(Clone, Deserialize, Eq, PartialEq, Serialize, Message, Digestible)]
+#[derive(Clone, Deserialize, Digestible, Eq, Message, PartialEq, Serialize, Zeroize)]
 pub struct TxIn {
     /// A "ring" of outputs containing the single output that is being spent.
     /// It would be nice to use [TxOut; RING_SIZE] here, but Prost only works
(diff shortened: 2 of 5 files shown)


ledger/db/src/test_utils/mod.rs
```diff
@@ -310,7 +310,7 @@ pub fn initialize_ledger(
                 );
 
                 let key_images = tx.key_images();
-                (tx.prefix.outputs, key_images)
+                (tx.prefix.outputs.clone(), key_images)
```
cbeck88, line 313: this was needed because rustc says we cannot move out of `Tx` anymore since it is a `Drop` type now

transaction/core/src/tx.rs
```diff
@@ -103,7 +103,8 @@ impl fmt::Debug for TxHash {
 }
 
 /// A CryptoNote-style transaction.
-#[derive(Clone, Eq, PartialEq, Serialize, Deserialize, Message, Digestible)]
+#[derive(Clone, Eq, PartialEq, Serialize, Deserialize, Message, Digestible, Zeroize)]
+#[zeroize(drop)]
```
cbeck88, line 107: this is the main change

nick-mobilecoin, line 107: :question: These [docs](https://docs.rs/zeroize/latest/zeroize/) are hard to parse.
> Zeroing memory is hard
> ...
> This crate has custom derive support for the Zeroize trait, gated under the zeroize crate’s zeroize_derive Cargo feature, which automatically calls zeroize() on all members of a struct or tuple struct.

The _automatically calls_ gave me the impression this was during drop.  In their defense they have examples after with comments saying `ZeorizeOnDrop` happens on drop.

Anyway should the `ZeroizeOnDrop` be added to more types in addition to the plain `Zeroize`?

nick-mobilecoin (approved):
looks good sans the other instance that needs to clone
```shell
error[E0509]: cannot move out of type `Tx`, which implements the `Drop` trait
   --> consensus/enclave/mock/src/lib.rs:287:28
    |
287 |             outputs.extend(tx.prefix.outputs.into_iter());
    |                            ^^^^^^^^^^^^^^^^^
    |                            cannot move out of here
    |                            move occurs because `tx.prefix.outputs` has type `Vec<TxOut>`, which does not implement the `Copy` trait

For more information about this error, try `rustc --explain E0509`.
```

transaction/core/src/tx.rs
remoun: Yes, looks like we want both. That's what the Zeroize tests do: https://github.com/RustCrypto/utils/blob/master/zeroize/tests/zeroize_derive.rs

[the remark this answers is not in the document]
cbeck88: thanks, didn't realize that

cbeck88: > Anyway should the ZeroizeOnDrop be added to more types in addition to the plain Zeroize?

I am not sure right now
I tried to make a list of things that should maybe be zeroized in #2717 

I figured starting with `Tx` was a reasonable place to start, and I wrote a bunch of rationale in the code comments.

It is causing wierd rust compilation issues to make the type Drop when it wasn't previously, so I'd rather not do this to a lot of types in this PR. And the other thing is, zeroizing operations may have nontrivial overhead, so we either have to measure the impact carefully or have a very good reason why it shouldn't impact network performance (ideally both).

iamalwaysuncomfortable (commented):
Is there any to stand this hypothesis up in one of the test networks?

cbeck88:
In this draft PR, I developed a benchmark which clones and drops a Tx with the `ZeroizeOnDrop` feature.

https://github.com/mobilecoinfoundation/mobilecoin/pull/2932

It looks that it takes about .24 milliseconds, which is a lot slower than I would have hoped for.

cbeck88:
See also this issue: https://github.com/RustCrypto/utils/issues/743

jcape:
@cbeck88 Is this still live, or can we close it?

cbeck88:
I plan to test this properly once we have slam working properly

cbeck88:
Idk, i mean, if slam looks the same before and after this change then i feel like it should be okay to go in, but feel free to tell me otherwise

jcape:
@cbeck88 Ping on this now that we've got slam going...

cbeck88:
Yea lets try to slam it

I also opened a PR in zeroize crate, if they take that should also give us a good way to optimize this

cbeck88:
https://github.com/RustCrypto/utils/pull/841

PR #2926, opened 2022-12-06
add transaction signing API and types, offline signer

ryankurte:
adds `mc-transaction-signer` crate, defining types for communication with view-only accounts (offline, ledger, etc.), as well as helpers for implementing these consistently and the offline signer implementation.

### Motivation

this API doesn't really overlap with the frontend JSON or backend GRPC/proto, but needs to be specified well to support different backends. providing this as a module with types and _super prescriptive) helpers both simplifies other implementations and helps keep these in sync.

blocked on:
- [x] https://github.com/mobilecoinfoundation/mobilecoin/pull/2901 for `ViewAccount` and `TxOutPublicKey` types

replaces: 
- https://github.com/mobilecoinfoundation/mobilecoin/pull/2902

### DIFF
--- a/transaction/signer/Cargo.toml
+++ b/transaction/signer/Cargo.toml
@@ -1,0 +1,39 @@
+[package]
+name = "mc-transaction-signer"
+version = "2.1.0-pre0"
+authors = ["MobileCoin"]
+edition = "2021"
+readme = "README.md"
+
+[features]
+
+
+[dependencies]
+# External dependencies
+anyhow = "1.0.61"
+displaydoc = { version = "0.2", default-features = false }
+serde = { version = "1.0",  default-features = false, features = ["derive"] }
+serde_json = "*"
+subtle = { version = "2.4.1", default-features = false, features = ["i128"] }
+zeroize = { version = "1", default-features = false }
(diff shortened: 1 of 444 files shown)


briancorbin (approved):
This looks great! Thanks so much for migrating this and refactoring it from Full Service :)

transaction/signer/src/lib.rs
```diff
@@ -0,0 +1,319 @@
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
+//! Transaction signer types, used for communication with external signer
+//! implementations
```
cbeck88, line 4: This description is rather sparse, can we try to add some description of what an external signer does, or maybe talk about an example?

transaction/signer/Cargo.toml
```diff
@@ -0,0 +1,39 @@
+[package]
+name = "mc-transaction-signer"
+version = "4.0.1"
+authors = ["MobileCoin"]
+edition = "2021"
+readme = "README.md"
```
cbeck88, line 6: it looks that there is no README.md, can we add one?

cbeck88 (approved):
looks good, my only ask is for some more high level docu for this crate

PR #2940, opened 2022-12-08
master - fix: update sgx runtime packages 2.18

jgreat:
### Motivation

Bump runtime container packages for SGX 2.18


cbeck88 (changes requested):
Can we please not do cherry-picks between master and release branch?

Can we please start merging release/4.0 into master periodically?

They cherry-pick workflow doesn't work very well and doesn't scale, and leads to us failing to merge fixes from release branches into master. We agreed to stop doing this starting in the 4.0 release cycle.

Let's redevelop this as a merge of release/4.0 into master please.

PR #2943, opened 2022-12-08
release/v4.0 - fix:  CD workflow issues #2941 and #2942

### LINKED ISSUE #2941: CI/CD - Possible race condition for runtime-base image.
Right now each build updates runtime-base:latest.  

We could get in a situation where a build with and earlier version of the runtime container gets used in a place where we want the latest. For example: when we update sgx packages. 

There are a couple of possibilities: 

- I think we can make images downstream of runtime-base use a specific version.
- We make a separate workflow for runtime-base that just watched that dockerfile for changes and also does periodic runs to apply OS package updates..

### LINKED ISSUE #2942: CI/CD - run on any/all PRs
We were not using graphite to do the chained PRs when we set up the CD workflow.  We need to run CD on all the PRs/branches in the chain, so I guess we need to turn off the filter there.

jgreat:
### Motivation

- fixes #2941 : Use docker sha tag instead of latest on FROM image for runtime containers.
- fixes #2942 : Run CD on all PRs to make sure graphite branch/merge chains get tested.

### DIFF
--- a/.github/workflows/mobilecoin-dev-cd.yaml
+++ b/.github/workflows/mobilecoin-dev-cd.yaml
@@ -275,7 +275,7 @@ jobs:
 
 #########################################
 # Build/Publish public artifacts
-######################################### 1a94367
+#########################################
   docker:
     runs-on: [self-hosted, Linux, small]
     needs:
--- a/.github/workflows/mobilecoin-dev-cd.yaml
+++ b/.github/workflows/mobilecoin-dev-cd.yaml
@@ -11,10 +11,6 @@ env:
 
 on:
   pull_request:
-    branches:
-    - master
-    - main
-    - release/*
     paths-ignore:
     - '**.md'
   push:
@@ -279,7 +275,7 @@ jobs:
 
 #########################################
 # Build/Publish public artifacts
-#########################################
+######################################### 1a94367
   docker:
     runs-on: [self-hosted, Linux, small]
     needs:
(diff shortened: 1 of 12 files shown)


transaction/extra/src/tx_summary_unblinding/verifier.rs
```diff
@@ -313,7 +313,14 @@ impl TxSummaryStreamingVerifier {
...
         fee: Amount,
         tombstone_block: u64,
-    ) -> ([u8; 32], TxSummaryUnblindingReport) {
+    ) -> Result<([u8; 32], TxSummaryUnblindingReport), Error> {
+        if self.output_count != self.expected_num_outputs {
+            return Err(Error::StillExpectingMoreOutputs);
+        }
+        if self.input_count != self.expected_num_inputs {
+            return Err(Error::StillExpectingMoreInputs);
+        }
```
cbeck88, line 322: It's kind of weird that this change is in the diff

jgreat: Wait what?

jgreat: I guess I needed a rebase.  Its gone now.

joekottke (approved):

.github/workflows/mobilecoin-dev-cd.yaml
```diff
@@ -279,7 +275,7 @@ jobs:
 
 #########################################
 # Build/Publish public artifacts
-#########################################
+######################################### 1a94367
```
nick-mobilecoin, line 278: :upside_down_face: guessing this was some  left over debug code
```suggestion
#########################################
```

nick-mobilecoin (approved):

jcape (approved):

PR #2780, opened 2022-10-26
Bump all versions to 3.0.0

jcape:
### Motivation

We need to tag and build 3.0.0.`

### In this PR
* Bump all crate versions to 3.0.0.

PR #2944, opened 2022-12-08
proposal: BRANCHING.md [skip-ci]

cbeck88:
Add a proposal for handling release branches by merging them

[Rendered Proposal](https://github.com/mobilecoinfoundation/mobilecoin/blob/branching-strategy/BRANCHING.md)

### DIFF
--- a/BRANCHING.md
+++ b/BRANCHING.md
@@ -42,6 +42,14 @@ considered by a future merge and can never create a conflict in a future merge.
 (This is a major difference with squash merging and rebase merging. This is very
 important when merging long-lived branches and not small PRs.)
 
+If there are no conflicts, we can simply open a PR from e.g. `release/v4.0` to
+`master` in github. The release branches are protected, so merging this PR will
+not delete the branch.
+
+If there are conflicts, this won't exactly work, because we will have to resolve
+the conflicts, but we don't want to push more commits to `release/v4.0` to accomplish
+this.
+
 The local workflow for this may look something like this:
 
 ```
@@ -59,10 +67,6 @@ $ git push origin merge-release-v4.0
 
 Here, we create a topic branch (off of master), merge the release branch into it,
 resolve any conflicts, and then propose to merge this to master.
-
-If there are no conflicts, then this can be simplified, we can simply open a PR
-from `release/v4.0` to `master`. If there are conflicts, then we wouldn't be able
-to merge that PR without making changes to `release/v4.0`, which is undesirable.
 
 When to merge the release branch
 --------------------------------
--- a/BRANCHING.md
+++ b/BRANCHING.md
@@ -6,7 +6,8 @@ Branching and Releasing
 Changes are created via PRs, reviewed, and squash-merged into `master`.
 
 Sometimes long running feature branches are created. These are, still, eventually
-squash merged into `master`.
+squash merged into `master`. (Alternatively, this could be a normal git merge
+and not a squash merge, it doesn't much matter.)
 
 Release branches
 ----------------
--- a/BRANCHING.md
+++ b/BRANCHING.md
@@ -17,6 +17,8 @@ of master.
 Minor releases are created by creating a release branch e.g. `release/v4.1` off of
 the predecessor `release/v4.0`.
 
+Any bugs that are found are fixed in the release branch first.
+
 Release candidates are tagged on release branches.
 
 Propagating changes
@@ -56,8 +58,8 @@ Currently, we want this merge to happen via a PR and go through (usually perfunc
 review. Reviewers should check that this is not a squash merge or a rebase merge.
 
 Right now we think it's desirable that whenever the release branch moves forwards,
-a PR could be opened automatically that proposes merges the release branch to master,
-exposing any git conflicts, which can then be reviewed and resolved.
+a PR could be opened automatically that proposes to merge the release branch to master,
+exposing any git conflicts, which can then be reviewed and resolved in this PR.
 
 Making changes
 --------------
--- a/BRANCHING.md
+++ b/BRANCHING.md
@@ -38,10 +38,6 @@ considered by a future merge and can never create a conflict in a future merge.
 (This is a major difference with squash merging and rebase merging. This is very
 important when merging long-lived branches and not small PRs.)
 
-This leads to a diagram like the following:
-
-[![](https://mermaid.ink/img/pako:eNqtU8FuwjAM_ZXIZ7aBaJbR86addto1F5OYNippq5CiIcS_L5QxlS2FViKnOO8lfn6O96AqTZACCysz_t1hncuStUtV1hofj5YOS5UzR2vCDT1tk0saMzqVgAlfSIghy2fx8g_xmAVomzxOO1BOqqgazyyaMi7FksuoV8mg6JzkejlqNu0pRwsxv6F5oMqbSn5Nmt0n4Yge81jpxPk8bspKiGRUI7uXvxYdq_-Uwvvd6kLO7R5qo4rIg4M-wxUzBym46Bcf-alPEUwgZA4sHUZ0f8Qk-JwsSUjDVqMrjq8eAg8bX33uSgWpdw1NoKk1eno1mDm0kK5wvQmnpI2v3Mdp5tvRPzPfWuSHePgGwSE-5w?type=png)](https://mermaid.live/edit#pako:eNqtU8FuwjAM_ZXIZ7aBaJbR86addto1F5OYNippq5CiIcS_L5QxlS2FViKnOO8lfn6O96AqTZACCysz_t1hncuStUtV1hofj5YOS5UzR2vCDT1tk0saMzqVgAlfSIghy2fx8g_xmAVomzxOO1BOqqgazyyaMi7FksuoV8mg6JzkejlqNu0pRwsxv6F5oMqbSn5Nmt0n4Yge81jpxPk8bspKiGRUI7uXvxYdq_-Uwvvd6kLO7R5qo4rIg4M-wxUzBym46Bcf-alPEUwgZA4sHUZ0f8Qk-JwsSUjDVqMrjq8eAg8bX33uSgWpdw1NoKk1eno1mDm0kK5wvQmnpI2v3Mdp5tvRPzPfWuSHePgGwSE-5w)
-
 When to merge the release branch
 --------------------------------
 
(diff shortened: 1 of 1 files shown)


BRANCHING.md
```diff
@@ -0,0 +1,167 @@
...
+  they are instead very similar commits touching the same code.
+* You will have to resolve this conflict, which usually won't be too hard.
+* To avoid this situation, make sure you consider whether a change should go
+  to master or the release branch. If in doubt, ask in the channel.
+
+This leads to a git history like the following:
+
+[![](https://mermaid.ink/img/pako:eNqtU8FuwjAM_ZXIZ7aBaJbR86addto1F5OYNippq5CiIcS_L5QxlS2FViKnOO8lfn6O96AqTZACCysz_t1hncuStUtV1hofj5YOS5UzR2vCDT1tk0saMzqVgAlfSIghy2fx8g_xmAVomzxOO1BOqqgazyyaMi7FksuoV8mg6JzkejlqNu0pRwsxv6F5oMqbSn5Nmt0n4Yge81jpxPk8bspKiGRUI7uXvxYdq_-Uwvvd6kLO7R5qo4rIg4M-wxUzBym46Bcf-alPEUwgZA4sHUZ0f8Qk-JwsSUjDVqMrjq8eAg8bX33uSgWpdw1NoKk1eno1mDm0kK5wvQmnpI2v3Mdp5tvRPzPfWuSHePgGwSE-5w?type=png)](https://mermaid.live/edit#pako:eNqtU8FuwjAM_ZXIZ7aBaJbR86addto1F5OYNippq5CiIcS_L5QxlS2FViKnOO8lfn6O96AqTZACCysz_t1hncuStUtV1hofj5YOS5UzR2vCDT1tk0saMzqVgAlfSIghy2fx8g_xmAVomzxOO1BOqqgazyyaMi7FksuoV8mg6JzkejlqNu0pRwsxv6F5oMqbSn5Nmt0n4Yge81jpxPk8bspKiGRUI7uXvxYdq_-Uwvvd6kLO7R5qo4rIg4M-wxUzBym46Bcf-alPEUwgZA4sHUZ0f8Qk-JwsSUjDVqMrjq8eAg8bX33uSgWpdw1NoKk1eno1mDm0kK5wvQmnpI2v3Mdp5tvRPzPfWuSHePgGwSE-5w)
```
nick-mobilecoin, line 97: :question: curious why an image link versus the raw mermaid text?  I imagine if mermaid.live goes down mermaid rendering may be gone as well so probably no harm linking to the image.

BRANCHING.md
```diff
@@ -0,0 +1,167 @@
...
+
+Any bugs that are found are fixed in the release branch first.
+
+Release candidates are tagged on release branches.
+
+Propagating changes
+-------------------
+
+Changes which are made in release branches are propagated forward to master,
+by *merging* the release branch into master.
```
nick-mobilecoin, line 30: :question: It seems like the changelog is a manual at release time creation. Will this have any negative affects generating it?  
I wouldn't think so, as long as the history for release N is looked at for the changelog for release N, it will make it to master in the merge and then the master side of the history can be looked at to get the "additional" changes.

BRANCHING.md
```diff
@@ -0,0 +1,169 @@
...
+It is acceptable to merge the release branch to master at any of the following times:
+
+* After making a change in the release branch
+* On a periodic basis
+* After finishing the release
+
+Currently, we want this merge to happen via a PR and go through (usually perfunctory)
+review. Reviewers should check that this is not a squash merge or a rebase merge.
+
+Right now we think it's desirable that whenever the release branch moves forwards,
```
nick-mobilecoin, line 62: :question: Do we know how hard it will be to resolve or ignore version bumps of the crates, thinking things like #2780.  
In this case master should have already been on 4.0.0-pre0.  We could go to 3.0.0 and then back to 4.0.0-pre0 in a subsequent commit, but it seems like we'd want to resolve during merge.

BRANCHING.md
cbeck88: IDK, I wanted to do inline mermaid, but it seems like the github markdown doesn't support the gitGraph mermaid type. The mermaid.live editor gave me this blob to paste right into markdown so that's what I went with.

BRANCHING.md
cbeck88: I think we should be able to make this work either using a `git merge -Xours` or `git merge -Xtheirs` (https://git-scm.com/docs/git-merge), or whipping up a shell-script to help with these conflicts.

BRANCHING.md
cbeck88: I think we were trying to get into a pattern where if you merge a notable change, you also make a changelog entry. This would ensure that these changes are merged forward into master and not abandoned in the release branch

nick-mobilecoin (approved):

jgreat (approved):
I think I understand the proposed workflow. 

Right now, we only have the "Squash Commits" option for PRs turned on.  I believe if we want to follow this proposed workflow we need to turn on "Allow Merge Commits".   

Looking at the Github docs,  I believe this summary matches their recommendation on how to handle long term head branches.  

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges

The "Allow Merge Commits" option is going to do a `--no-ff` merge, which after refreshing  some git knowledge makes sense.

eranrund (commented):
I think overall this makes sense, and more specifically I think the potential big win from implementing this strategy is combating the last drawback mentioned at the bottom - forgetting to bring back to `master` any important changes that happened in a release branch. In theory this should combat that, especially if there is automation involved and we are eager to merge back any commits.
However, it is not yet clear to me how this behaves if we ever want to make a change that we don't want to bring back. It's rare, but I seem to recall this did happen a few times (but unfortunately I can't remember the specifics). Maybe the workaround for that is to then followup with a commit to master to undo whatever got brought back that we didn't like?
I'm also not yet sure how this behaves if a change in the release branch causes merge conflicts into `master`, how and when are these conflicts addressed?

A thing I am concerned about is that people working on PRs need to be mindful of where they should be merged into, which essentially means they need to have some visibility into the release process to know whether the thing they are PRing is required in the release or can wait for the next one. My understanding is that the moment a PR is merged into master instead of a release branch, we're back in cherrypick territory. But I guess this is like that today too so maybe it's not a problem, and with some cultural shift we can get people to be better at asking the question of "on which branch should I develop this" before they start working on new PRs.

I'd like to hear @jcape's thoughts on this since he is one of the people who are most involved in the release process, and have definitely felt the pains of managing the release branches.

Thank you for putting all of this in writing!

BRANCHING.md
```diff
@@ -0,0 +1,169 @@
...
+a PR could be opened automatically that proposes to merge the release branch to master,
+exposing any git conflicts, which can then be reviewed and resolved in this PR.
```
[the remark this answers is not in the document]
cbeck88: I think that's the process, yeah. There's a topic branch for the merge that is created off the release branch, then you try to merge it to master, and if there are conflicts, you try to resolve them, and put up that PR

BRANCHING.md
```diff
@@ -0,0 +1,169 @@
...
+made it back.
+
+It is acceptable to merge the release branch to master at any of the following times:
+
+* After making a change in the release branch
+* On a periodic basis
+* After finishing the release
+
+Currently, we want this merge to happen via a PR and go through (usually perfunctory)
+review. Reviewers should check that this is not a squash merge or a rebase merge.
```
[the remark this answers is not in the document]
cbeck88: Maybe you're right, maybe this is only visible to the person who actually clicks merge

cbeck88:
> However, it is not yet clear to me how this behaves if we ever want to make a change that we don't want to bring back. It's rare, but I seem to recall this did happen a few times (but unfortunately I can't remember the specifics). Maybe the workaround for that is to then followup with a commit to master to undo whatever got brought back that we didn't like?
I'm also not yet sure how this behaves if a change in the release branch causes merge conflicts into master, how and when are these conflicts addressed?

I guess I'm struggling to think of a case where this is needed.

* Crate versions? For example, maybe in making release 4.1 crate version bumps, we end up with a conflict with master branch which is on 5.0 or something. I think (1) it's probably okay for master branch crate versions to match the latest release branch crate versions, and then getting this change is not bad (2) if you disagree with (1), I think you still end up choosing not to bring back the change you don't want at the time of resolving the git conflict during the merge.
* Some SGX thing? For example, you need to bump SGX SDK version in the release branch for some fix, but master branch is already on a later SGX SDK version. In this case I think, again, you end up choosing not to bring back the change during merge conflict resolution. In most cases though, if the release branch needs to bump SGX SDK version, then master branch does too, so I think this scenario is unlikely.

If you have a change that you only want in the release branch, and it doesn't conflict when it merges to master, you can simply revert that commit in master afterwards.

In projects I've worked on that used this strategy before, it was pretty rare that there was a change in the release branch that shouldn't land on master.  Even things around project configuration or packaging files for release process are usually helpful to merge forwards. I think changelog is another example of something that will be helped by this.

cbeck88:
I'll try to add more docu about what it looks like to merge the release branch to master

cbeck88:
Also, here's a thread where people are trying to do the same thing we are proposing here, and discussion about the git settings. It seems they opted to make it so that only admins can do a normal merge commit.

https://github.com/dotnet/runtime/issues/260

BRANCHING.md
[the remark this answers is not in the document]
david1alvarez: This can be restricted at the repo level I believe as part of its permissions. I am not sure if you can set the restrictions on a branch-by-branch basis, but you can set a global one.

BRANCHING.md
```diff
@@ -0,0 +1,214 @@
...
+  This leads to regressions in the next release. Because we never merge the release
+  branch back to master, and all the cherry-picks have different hashes from their
+  originals, it's very hard to ever be sure that we got all the changes.
+
+This merge-based policy hopes to rectify these issues.
+
+Why squash merge at all
+-----------------------
+
+The best argument for why topic branches should be squash merged is that,
```
david1alvarez, line 196: It also has the major benefits of ensuring the history of any base branch (`master`/`main`/`release/*`) is readable, and discourages their histories from being filled with unhelpful commit messages like "debugging" and "fixes". It also dramatically improves the usefulness of tools that rely on git-blame (e.g. GitLense in VSCode), as the commits they point to offer an insight into the the reason for the change from their title.

david1alvarez (approved):

BRANCHING.md
cbeck88: Having worked in projects for years that don't do any squash merging, I guess I don't think this is a benefit of squash merging.

In projects that don't do any squash merging, you don't allow people to merge a PR with commit messages like "debugging" and "fixes". They have to either use `git commit --amend` so the fixups are part of the previous commit, or `git rebase` on their branch and squash and edit things to make their commit history useful before merging.

Anyways, you can still navigate to the merge commit and read the same commit message you would have gotten in a squash merge. It's only like another click or two.

jcape (approved):

PR #2952, opened 2022-12-12
feat: HAProxy Ingress Blocklists

jgreat:
### Motivation

Update helm chart ingress objects to support enabling IP blocklists.

### Future Work

We need to merge these changes up to the master branch and back port them to the release/v3.0 branch.

### DIFF
--- a/.internal-ci/helm/consensus-node-config/values.yaml
+++ b/.internal-ci/helm/consensus-node-config/values.yaml
@@ -60,8 +60,8 @@ global:
     # tokensConfig:
     #   tokensSignedJson: |-
 
-  ### Enable haproxy IP blacklist for ingress
-  # pattern is the object in the configmap shared between infra-haproxy-blacklist and haproxy kubernetes-ingress
+  ### Enable haproxy IP blocklist for ingress
+  # pattern is the object in the configmap shared between infra-haproxy-blocklist and haproxy kubernetes-ingress
   blocklist:
     enabled: false
     pattern: patterns/blocked-countries
--- a/.internal-ci/helm/consensus-node-config/values.yaml
+++ b/.internal-ci/helm/consensus-node-config/values.yaml
@@ -59,3 +59,9 @@ global:
     # Add signed tokens.json with --set-file=global.node.tokensConfig.tokensSignedJson=tokens.signed.json
     # tokensConfig:
     #   tokensSignedJson: |-
+
+  ### Enable haproxy IP blacklist for ingress
+  # pattern is the object in the configmap shared between infra-haproxy-blacklist and haproxy kubernetes-ingress
+  blocklist:
+    enabled: false
+    pattern: patterns/blocked-countries
--- a/.internal-ci/helm/fog-services/templates/fog-report-grpc-ingress.yaml
+++ b/.internal-ci/helm/fog-services/templates/fog-report-grpc-ingress.yaml
@@ -13,6 +13,9 @@ metadata:
     {{- if $.Values.global.certManagerClusterIssuer }}
     cert-manager.io/cluster-issuer: {{ $.Values.global.certManagerClusterIssuer }}
     {{- end }}
+    {{- if (include "fogServices.blocklist.enabled" $) }}
+    haproxy.org/blacklist: {{ include "fogServices.blocklist.pattern" $ }}
+    {{- end }}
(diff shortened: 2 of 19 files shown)


.internal-ci/helm/fog-services/templates/fog-report-grpc-ingress.yaml
```diff
@@ -13,6 +13,9 @@ metadata:
     {{- if $.Values.global.certManagerClusterIssuer }}
     cert-manager.io/cluster-issuer: {{ $.Values.global.certManagerClusterIssuer }}
     {{- end }}
+    {{- if (include "fogServices.blacklist.enabled" $) }}
+    haproxy.org/blacklist: {{ include "fogServices.blacklist.pattern" $ }}
```
nick-mobilecoin, line 17: :question: Does the `$` change behaviour versus the other changes that use `.`?
The other locations use `include "blah" .`
Looking at https://helm.sh/docs/howto/charts_tips_and_tricks/#using-the-include-function I'm not seeing information on `$` vs `.`

.internal-ci/helm/fog-services/templates/fog-report-http-ingress.yaml
```diff
@@ -13,6 +13,9 @@ metadata:
     {{- if $.Values.global.certManagerClusterIssuer }}
     cert-manager.io/cluster-issuer: {{ $.Values.global.certManagerClusterIssuer }}
     {{- end }}
+    {{- if (include "fogServices.blacklist.enabled" $) }}
+    haproxy.org/blacklist: {{ include "fogServices.blacklist.pattern" $ }}
```
nick-mobilecoin, line 17: :question: may need `.` similar to above if the `$` is an issue.

.internal-ci/helm/consensus-node-config/values.yaml
```diff
@@ -59,3 +59,9 @@ global:
     # Add signed tokens.json with --set-file=global.node.tokensConfig.tokensSignedJson=tokens.signed.json
     # tokensConfig:
     #   tokensSignedJson: |-
+
+  ### Enable haproxy IP blacklist for ingress
+  # pattern is the object in the configmap shared between infra-haproxy-blacklist and haproxy kubernetes-ingress
```
nick-mobilecoin, line 64: :question: Where is this pattern? I did some quick grepping of `infra-haproxy-blacklist` the mobilecoin repo and the github orgs but came up empty

nick-mobilecoin (approved):

.internal-ci/helm/fog-services/templates/fog-report-grpc-ingress.yaml
jgreat: Because this is a loop, you need to use `$` to access the top level scope. `.` would only access the inner loop scope.

.internal-ci/helm/consensus-node-config/values.yaml
jgreat: The configmap name is defined on when you launch the haproxy, that's an internal repo that has all of those definitions.  The `data` object will be called `blocked-countries`.  This object will be a newline separated list of CIDRs to be blocked. The list is generated by on the scripts/deployment in `infra-haproxy-blacklist`.

.internal-ci/helm/fog-services/templates/fog-report-grpc-ingress.yaml
```diff
@@ -13,6 +13,9 @@ metadata:
     {{- if $.Values.global.certManagerClusterIssuer }}
     cert-manager.io/cluster-issuer: {{ $.Values.global.certManagerClusterIssuer }}
     {{- end }}
+    {{- if (include "fogServices.blocklist.enabled" $) }}
+    haproxy.org/blacklist: {{ include "fogServices.blocklist.pattern" $ }}
```
remoun, line 17: I notice most other templates pass `.`, while these `fog-report*` templates use `$`. What's the difference?

remoun (approved):
Thank you!

.internal-ci/helm/fog-services/templates/fog-report-grpc-ingress.yaml
jgreat: See nick's question above :) TL;DR this template is in a loop so to access the outer context you need to use `$`

PR #2955, opened 2022-12-14
Feature/v4.0 dev faucet

aweis89:
- Adds dev faucet traffic generator service to the mobilecoind container
- Modifies mobilcoind helm chart and image

### Motivation
Adds dev faucet traffic generator for performance testing

### Future Work

- Consider seperating out into it's own deployment
- Upgrade posgresql bitnami chart to version ^12.0.0

### DIFF
--- a/.internal-ci/docker/Dockerfile.bootstrap-tools
+++ b/.internal-ci/docker/Dockerfile.bootstrap-tools
@@ -43,7 +43,6 @@ COPY ${RUST_BIN_PATH}/mc-util-seeded-ed25519-key-gen /usr/local/bin/
 COPY ${RUST_BIN_PATH}/fog-report-cli /usr/local/bin/
 COPY ${RUST_BIN_PATH}/read-pubfile /usr/local/bin/
 COPY ${RUST_BIN_PATH}/mc-util-grpc-token-generator /usr/local/bin/
-COPY ${RUST_BIN_PATH}/mc-mobilecoind-dev-faucet /usr/local/bin/
 
 # Test wrappers and util scripts
 COPY .internal-ci/test/ /test/
--- a/.internal-ci/docker/Dockerfile.bootstrap-tools
+++ b/.internal-ci/docker/Dockerfile.bootstrap-tools
@@ -43,6 +43,7 @@ COPY ${RUST_BIN_PATH}/mc-util-seeded-ed25519-key-gen /usr/local/bin/
 COPY ${RUST_BIN_PATH}/fog-report-cli /usr/local/bin/
 COPY ${RUST_BIN_PATH}/read-pubfile /usr/local/bin/
 COPY ${RUST_BIN_PATH}/mc-util-grpc-token-generator /usr/local/bin/
+COPY ${RUST_BIN_PATH}/mc-mobilecoind-dev-faucet /usr/local/bin/
 
 # Test wrappers and util scripts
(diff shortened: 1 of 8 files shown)


.internal-ci/docker/Dockerfile.bootstrap-tools
```diff
@@ -43,6 +43,7 @@ COPY ${RUST_BIN_PATH}/mc-util-seeded-ed25519-key-gen /usr/local/bin/
 COPY ${RUST_BIN_PATH}/fog-report-cli /usr/local/bin/
 COPY ${RUST_BIN_PATH}/read-pubfile /usr/local/bin/
 COPY ${RUST_BIN_PATH}/mc-util-grpc-token-generator /usr/local/bin/
+COPY ${RUST_BIN_PATH}/mc-mobilecoind-dev-faucet /usr/local/bin/
```
aweis89, line 46: Not sure this is really useful here?

.internal-ci/helm/mobilecoind/templates/mobilecoind-json-service.yaml
```diff
@@ -14,3 +14,6 @@ spec:
   - name: json
     port: 9090
     targetPort: json
+  - name: json-http
+    port: 80
```
remoun, line 18: Why add this port?

remoun (approved):
LGTM

.internal-ci/helm/mobilecoind/templates/mobilecoind-json-service.yaml
aweis89: Nice catch. The idea was to use port 80 on the service so that calling services don't need to specify a port. Generally I think we should prefer to use port `80` on services to abstract that away from the client. I think that preferable than trying to match the internal port of the pod and exposing that implementation detail to clients

.internal-ci/docker/entrypoints/mobilecoind.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
jgreat, line 1: Needs copyright

.internal-ci/helm/mobilecoind/templates/mobilecoind-dev-faucet-configmap.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: v1
```
jgreat, line 1: Needs copyright

.internal-ci/helm/mobilecoind/templates/mobilecoind-json-service.yaml
jgreat: We can map ports in 2 ways:
1. use a single hostname and define multiple ports.
2. use a hostname per service and use a single standard port for the service (80, 443, 5432...) 

I think this is the right approach, using the second where we just point at a sane abstracted service name and use the standard port.

jgreat: This does get a little bit fuzzy where a pod might be listening on multiple "standard" ports.  Like metrics on "9090".  

I think in general where an app is publishing an "HTTP" (or GRPC) port, mapping it to 80 on the k8s "Service" load-balancer makes sense and can ease the configuration burden

jgreat (approved):
Small nits around adding copyrights.

Issue #2966, opened 2022-12-16
[FVR] Correlate View Shards with their expected block ranges


samdealy:
This will allow the View Router to ensure that each shard provides the expected block range response.

PR #2979, opened 2023-01-03
Implement optelemetry tracing for FVR

samdealy:
### Motivation
Implements opemtelemtry tracing for Fog View Router. Makes use of `async` opentelemtry tracing primitives to accommodate the router's async code.

PR #2981, opened 2023-01-04
Ensure shards provide correct blocks

samdealy:
### Motivation

As described in #2966, the Fog View Router (FVR) has no way of knowing if a shard provides the "correct" block range - i.e. the block range for which it was configured to process. This PR makes FVR aware of the shards' block ranges and throws an error if a shard provides an incorrect block range.

### DIFF
--- a/fog/view/server/src/bin/router.rs
+++ b/fog/view/server/src/bin/router.rs
@@ -8,9 +8,9 @@ use mc_common::{logger::log, time::SystemTimeProvider};
 use mc_fog_api::view_grpc::FogViewStoreApiClient;
 use mc_fog_view_enclave::{SgxViewEnclave, ENCLAVE_FILE};
 use mc_fog_view_server::{
-    config::{FogViewRouterConfig, ShardingStrategy::Epoch},
+    config::FogViewRouterConfig,
     fog_view_router_server::{FogViewRouterServer, Shard},
-    sharding_strategy::ShardingStrategy,
+    sharding_strategy::{EpochShardingStrategy, ShardingStrategy},
 };
 use mc_util_cli::ParserWithBuildInfo;
 use mc_util_grpc::ConnectionUriGrpcioChannel;
@@ -47,23 +47,17 @@ fn main() {
             .name_prefix("Main-RPC".to_string())
             .build(),
     );
-    for (i, shard_uri) in config.shard_uris.clone().into_iter().enumerate() {
+    for shard_uri in config.shard_uris.clone() {
         let fog_view_store_grpc_client = FogViewStoreApiClient::new(
             ChannelBuilder::default_channel_builder(grpc_env.clone())
                 .connect_to_uri(&shard_uri, &logger),
         );
 
-        let sharding_strategy = config
-            .sharding_strategies
-            .get(i)
-            .unwrap_or_else(|| panic!("Couldn't find shard at index {}", i));
-        let Epoch(epoch_sharding_strategy) = sharding_strategy;
+        // TODO: update this logic once we introduce other types of sharding strategies.
+        let epoch_sharding_strategy = EpochShardingStrategy::try_from(shard_uri.clone())
+            .expect("Could not get sharding strategy");
         let block_range = epoch_sharding_strategy.get_block_range();
-        let shard = Shard::new(
-            shard_uri.clone(),
-            Arc::new(fog_view_store_grpc_client),
-            block_range,
-        );
+        let shard = Shard::new(shard_uri, Arc::new(fog_view_store_grpc_client), block_range);
         shards.push(shard);
     }
     let shards = Arc::new(RwLock::new(shards));
--- a/fog/view/server/src/bin/router.rs
+++ b/fog/view/server/src/bin/router.rs
@@ -8,12 +8,13 @@ use mc_common::{logger::log, time::SystemTimeProvider};
 use mc_fog_api::view_grpc::FogViewStoreApiClient;
 use mc_fog_view_enclave::{SgxViewEnclave, ENCLAVE_FILE};
 use mc_fog_view_server::{
-    config::FogViewRouterConfig, fog_view_router_server::FogViewRouterServer,
+    config::{FogViewRouterConfig, ShardingStrategy::Epoch},
+    fog_view_router_server::{FogViewRouterServer, Shard},
+    sharding_strategy::ShardingStrategy,
 };
 use mc_util_cli::ParserWithBuildInfo;
 use mc_util_grpc::ConnectionUriGrpcioChannel;
 use std::{
-    collections::HashMap,
     env,
     sync::{Arc, RwLock},
 };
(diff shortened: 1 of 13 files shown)


samdealy:
* feature/fog-view-router
  * **PR #2976** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2976" target="_blank">Graphite</a> 
    * **PR #2979** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2979" target="_blank">Graphite</a> 
      * **PR #2981** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2981" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/2981?utm_source=stack-comment).

fog/view/server/src/bin/router.rs
```diff
@@ -40,28 +41,39 @@ fn main() {
...
-        fog_view_store_grpc_clients.insert(shard_uri, Arc::new(fog_view_store_grpc_client));
+
+        let sharding_strategy = config
+            .sharding_strategies
+            .get(i)
+            .unwrap_or_else(|| panic!("Couldn't find shard at index {}", i));
+        let Epoch(epoch_sharding_strategy) = sharding_strategy;
+        let block_range = epoch_sharding_strategy.get_block_range();
+        let shard = Shard::new(
+            shard_uri.clone(),
```
jcape, line 63: I'm pretty sure this clone is unnecessary, since you're cloning the collection at line 50.

fog/view/server/src/config.rs
```diff
@@ -121,8 +121,14 @@ pub struct FogViewRouterConfig {
...
-    #[clap(long, env = "MC_CLIENT_LISTEN_URI")]
+    /// Sharding Strategies for each Shard. Should be indexed the same as the
+    /// `shard_uris` field.
+    #[clap(long, env = "MC_VIEW_SHARDING_STRATEGIES")]
+    pub sharding_strategies: Vec<ShardingStrategy>,
+
+    /// gRPC listening URI for Fog View Stores. Should be indexed the same as
+    /// the `sharding_stratgies` field.
+    #[clap(long, env = "MC_VIEW_SHARD_URIS")]
     pub shard_uris: Vec<FogViewStoreUri>,
```
jcape, line 132: I'm a little uncomfortable with this, it feels like this maybe should be a URI parameter on the StoreURI so we're not doing this kind of indexing.

e.g. the way we do `?responder-id=some.other.name` could be used to do a `?strategy=epoch`.

fog/view/server/src/bin/router.rs
```diff
@@ -40,28 +41,39 @@ fn main() {
...
         let fog_view_store_grpc_client = FogViewStoreApiClient::new(
             ChannelBuilder::default_channel_builder(grpc_env.clone())
                 .connect_to_uri(&shard_uri, &logger),
         );
-        fog_view_store_grpc_clients.insert(shard_uri, Arc::new(fog_view_store_grpc_client));
+
+        let sharding_strategy = config
+            .sharding_strategies
+            .get(i)
+            .unwrap_or_else(|| panic!("Couldn't find shard at index {}", i));
```
jcape, line 59: This will do the same thing:
```suggestion
        let sharding_strategy = config.sharding_strategies[i];
```

fog/view/server/src/fog_view_router_server.rs
```diff
@@ -43,6 +41,33 @@ where
...
+#[derive(Clone)]
+pub struct Shard {
+    /// The uri that this shard listens on.
+    pub uri: FogViewStoreUri,
+
+    /// The gRPC client that is used to communicate with the shard.
+    pub grpc_client: Arc<view_grpc::FogViewStoreApiClient>,
+
+    /// The `BlockRange` that this shard is responsible for providing.
+    pub block_range: BlockRange,
```
jcape, line 54: Do these values need to be `pub`, or can they be protected behind, e.g. `block_range(&self) -> &BlockRange`-style accessors?

My default opinion on public struct members is a very defensive "only when absolutely necessary, e.g. for serialization/DMA."

fog/view/server/src/fog_view_router_service.rs
```diff
@@ -69,11 +65,11 @@ where
             let logger = logger.clone();
             // TODO: Confirm that we don't need to perform the authenticator logic. I think
             // we don't  because of streaming...
```
jcape, line 67: Total aside: Could you clarify this TODO?

jcape (approved):
Some minor nits, but I propose putting the sharding strategy on the ViewStoreURI as a parameter.

fog/view/server/src/config.rs
awygle: I agree this is a bit sketchy vis a vis trying to keep these collections in sync. Not sure how clap would handle `Vec<(FogViewStoreUri, ShardingStrategy)>` either. James' suggestion seems reasonable but I haven't really poked around in how `ResponderId` is implemented as a URI parameter so I'm not sure if it'd have any other consequences.

fog/view/server/src/router_admin_service.rs
```diff
@@ -44,8 +43,13 @@ impl FogViewRouterAdminService {
...
             )
         })?;
-        let mut shard_clients = self.shard_clients.write().expect("RwLock Poisoned");
-        if shard_clients.keys().contains(&view_store_uri) {
+        let mut shards = self.shards.write().expect("RwLock Poisoned");
+        if shards
+            .iter()
+            .map(|shard| shard.uri.clone())
+            .collect::<HashSet<FogViewStoreUri>>()
+            .contains(&view_store_uri)
```
awygle, line 51: This seems like it should be expressable as `shards.iter().any(|shard| shard.uri.clone() == view_store_uri)` or something similar, without having to collect up a whole `HashSet`.

fog/view/server/src/router_admin_service.rs
```diff
@@ -62,7 +66,11 @@ impl FogViewRouterAdminService {
             ChannelBuilder::default_channel_builder(grpc_env)
                 .connect_to_uri(&view_store_uri, logger),
         );
-        shard_clients.insert(view_store_uri, Arc::new(view_store_client));
+        // TODO: Add block range or sharding strategy to this...
```
awygle, line 69: I think this PR would resolve this TODO?

fog/view/server/src/router_admin_service.rs
```diff
@@ -62,7 +66,11 @@ impl FogViewRouterAdminService {
...
+        // Check to make sure this block range isn't already covered...
```
awygle, line 70: I actually think overlapping block ranges is fine, desirable even.

fog/view/server/src/router_request_handler.rs
```diff
@@ -27,7 +28,7 @@ const RETRY_COUNT: usize = 3;
 /// Handles a series of requests sent by the Fog Router client.
 pub async fn handle_requests<E>(
     method_name: GrpcMethodName,
-    shard_clients: Vec<Arc<FogViewStoreApiClient>>,
+    shards: Vec<Shard>,
```
awygle, line 31: This is doing two additional `Clone`s compared to the previous, cloning a URI and a block range. I believe this should be cheap, so I'm not too worried, but the cost of cloning a URI should at least be looked at.

fog/view/server/src/router_request_handler.rs
```diff
@@ -215,8 +202,8 @@ where
             query_responses.push(multi_view_store_query_response);
         }
 
-        shard_clients = processed_shard_response_data.shard_clients_for_retry;
-        if shard_clients.is_empty() {
+        shards = processed_shard_response_data.shards_for_retry;
+        if shards.is_empty() {
             break;
```
awygle, line 207: Eventually we are going to want to change this termination logic to allow for the case where multiple shards are capable of serving the same range (at least, I believe that's functionality @jcape has expressed desire for). We should probably have an issue for that, as the "next step" of this PR chain.

awygle (changes requested):
I agree with James about the config - we need to either address that or affirmatively convince ourselves it's not important.

fog/view/server/src/bin/router.rs
samdealy: Done.

fog/view/server/src/config.rs
samdealy: I added `FogViewUrl` parsing logic for a query param named `sharding_strategy.` Users can pass `?sharding_strategy=0-100` etc. now.

fog/view/server/src/bin/router.rs
samdealy: This code is no longer present given the new url parsing logic.

fog/view/server/src/fog_view_router_service.rs
samdealy: I'm pretty sure that the authenticator enforces unary authentication logic, so I don't think it applies for streaming. I can look more into this though, but will make changes in a follow-up PR if necessary.

fog/view/server/src/router_admin_service.rs
samdealy: Done.

fog/view/server/src/router_admin_service.rs
samdealy: Done.

fog/view/server/src/router_admin_service.rs
samdealy: Yes that's true. This comment means that there shouldn't be any duplicate block ranges (which is taken care of by the above logic, so I removed this comment).

fog/view/server/src/router_request_handler.rs
samdealy: Ack.

fog/view/server/src/router_request_handler.rs
samdealy: Ack.

fog/view/server/src/fog_view_router_server.rs
samdealy: Done.

awygle (approved):

fog/view/server/src/fog_view_router_server.rs
jcape: Yep, follow-up is fine, ty

fog/view/server/src/fog_view_router_service.rs
jcape: Yeah, I guess what I'm saying is: what logic would go in here?

Issue #2965, opened 2022-12-16
[FVR] Use `const_cmp` feature in `e_tx_out_store`


samdealy:
Depends on #2964 

#2880 updated  `e_tx_out_store` `find_records` method to use the `FixedTxOutSearchResult`, and the logic, as it stands, is not constant time because it's copying a variable number of bytes. To fix this, we need to use a `const` version of `core::cmp::min` to figure out the constant byte length to copy. The `const` version of this function is only available under the `const_cmp` feature flag, which itself is only available in later versions of rust.

PR #2999, opened 2023-01-10
Update rust to 2023-01-04

samdealy:
### Motivation
The repo is currently using  of the `nightly-2022-04-29` version of rust. This is problematic because it prevents us from using some of the latest features (e.g. #2965). This PR updates Rust and fixes a few clippy errors such as:
- [Inline text formatting](https://rust-lang.github.io/rust-clippy/master/index.html#uninlined_format_args)
- [Result error variant too large](https://rust-lang.github.io/rust-clippy/master/index.html#result_large_err): note, we disabled this lint because we don't want to introduce an `alloc` and the change is pretty substantial (you have to `Box` a lot of errors and are therefore unable to destructure inner error fields because `Box` is an opaque type, thus decreasing readability)
- Removal of unnecssary `clone` or references that clippy didn't catch previously.

If you see any instances of changes outside of those outlined above, feel free to leave a comment and I'll double check that it should be included.

### Future Work
* Update repos that use mobilecoin.git as a submodule.


cbeck88 (approved):
LGTM

attest/ake/src/lib.rs
```diff
@@ -7,6 +7,7 @@
 //! (client) and responders (server).
 
 #![allow(clippy::type_complexity)]
+#![allow(clippy::result_large_err)]
```
nick-mobilecoin, line 10: :thought_balloon: TL;DR this is fine for this PR of updating rust, just curious if we want to follow up later.

It looks like without the suppression clippy outputs:
```console
error: the `Err`-variant returned from this function is very large
  --> attest/ake/src/responder.rs:39:10
   |
39 |     ) -> Result<(Ready<Cipher>, AuthResponseOutput), Error>
   |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
  ::: attest/ake/src/error.rs:28:5
   |
28 |     ReportVerification(VerifierError),
   |     --------------------------------- the largest variant contains at least 696 bytes
```

we may not be too worried that we're always returning a 700byte structure, but I wonder if we should consider creating a follow on issue to consider addressing the error size.

There is also a configuration value if we agree that the default size is too restrictive
> Configuration
>
> This lint has the following configuration variables:
>
>    large-error-threshold: u64(defaults to 128): The maximum size of the Err-variant in a Result returned from a function

nick-mobilecoin (approved):

attest/ake/src/lib.rs
samdealy: @jcape and I discussed this, and we think that this is not worth pursuing. Boxing these errors comes at a significant code readability cost- a lot of the error consumption code destructures these errors. If you box them, the inner field (i.e. what used to be just the error) is now an opaque (box) type, which prevents destructuring and makes you do some ugly box dereferences. 

There's also no indication that the size of these errors has any impact on the performance of our code. So given these two things, we decided to not implement it right now. If we find that it becomes a problem in the future, then we can reconsider.

PR #3020, opened 2023-01-11
Cleanup in preparation for 2023-01-04 rust 

samdealy:
### Motivation
The repo is currently using  of the `nightly-2022-04-29` version of rust. This is problematic because it prevents us from using some of the latest features (e.g. #2965). This PR makes some updates  and fixes a few clippy errors such as:
- [Inline text formatting](https://rust-lang.github.io/rust-clippy/master/index.html#uninlined_format_args)

If you see any instances of changes outside of those outlined above, feel free to leave a comment and I'll double check that it should be included.

### Future Work
* Actually update rust in a separate PR and ignore the`result_large_err` lint
* Update repos that use mobilecoin.git as a submodule.


consensus/scp/src/slot.rs
```diff
@@ -2626,7 +2626,7 @@ mod ballot_protocol_tests {
         }
 
         // Force ballot timeout timer to fire.
-        slot.next_ballot_at = Some(Instant::now() - Duration::from_secs(1));
+        slot.next_ballot_at = Some(Instant::now().checked_sub(Duration::from_secs(1)).unwrap());
```
awygle, line 2629: For those like me who are surprised to see this change, see https://github.com/rust-lang/rust-clippy/issues/9371

mobilecoind-dev-faucet/src/bin/main.rs
```diff
@@ -4,6 +4,7 @@
 
 #![deny(missing_docs)]
 #![feature(proc_macro_hygiene, decl_macro)]
+#![allow(clippy::let_unit_value)]
```
awygle, line 7: What was triggering this lint? Is there a reason we're allowing it instead of changing the code to no longer trigger it?

util/metrics/src/service_metrics.rs
```diff
@@ -95,7 +95,9 @@ impl ServiceMetrics {
             .unwrap(),
         }
     }
+}
 
+impl ServiceMetrics {
```
awygle, line 100: Why split this impl block? I don't see a Clippy lint for this or anything (actually there's a negative Clippy lint but it's restricted)

awygle: And actually shouldn't the upper impl block (line 68) be `impl Default for ServiceMetrics`?

awygle (approved):
LGTM, couple of minor questions.

mobilecoind-dev-faucet/src/bin/main.rs
samdealy: I couldn't figure out what caused this lint. The lint is normally triggered when you do something like `let value = ()`, but that's not going on here. I figured it had something to do with the internals of `rocket` and decided to just turn it off.

consensus/scp/src/slot.rs
nick-mobilecoin: Nice, that was one of my gripes with _safe_ rust. debug would always panic on underflow while release would wrap around
> Performs subtraction that returns None instead of wrapping around on underflow.
unintentional underflow and overflow are sources of many bugs in industry, very nice to see this warning on by default

util/metrics/src/service_metrics.rs
samdealy: Yes, you're right. Done.

common/src/logger/mod.rs
```diff
@@ -119,7 +119,7 @@ cfg_if::cfg_if! {
             fn basic_trace_time() {
                 let logger = create_test_logger("basic_trace_time".to_string());
 
-                slog_scope::scope(&logger.clone(), || {
+                slog_scope::scope(&logger, || {
```
awygle, line 122: As requested by James, I reviewed what `slog_scope::scope` is doing here. I see why it looked sketchy, but I think it's actually OK. Because the `ScopeGuard` created in `slog_scope::scope` isn't returned, there's no way to keep the logger's pointer in `TL_SCOPES` past the execution of the closure, and you can't `drop(logger)` in the closure because the closure is called in the execution of `scope` while the borrowed `logger` argument is still in play, so you can't move the logger. So I don't think there's a way to make this Do Bad Things without doing something `unsafe` yourself (and at that point bets are mostly off anyway).

cbeck88 (approved):

jcape (approved):

PR #3029, opened 2023-01-13
Remove cd triggers for PRs, main and master branches

jgreat:
### Motivation

We're finding there's a lot of pain in the dynamic environments at the moment.  We don't want to loose the production like deployment testing, but we're going to roll back the CD requirement for new work going towards master branch.

- Limit trigger to PRs with 'release/*' as the base ref
- remove push trigger on main, master, and release/* branch
- change clean up tasks CD to match new triggers.
- Add mobilecoin-cd-complete dummy job that is just successful if the rest of the workflow is successful.  This way we have a "standard" job with a static name we can use for branch protection checks.

### DIFF
--- a/.github/workflows/mobilecoin-dev-cd.yaml
+++ b/.github/workflows/mobilecoin-dev-cd.yaml
@@ -10,19 +10,11 @@ env:
   RELEASE_3X_TAG: v3.0.0-dev
 
 on:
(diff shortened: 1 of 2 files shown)


jcape (approved):

joekottke (approved):
LGTM.

PR #3047, opened 2023-01-26
release/v4.1 - feat(cd/testing): add ability to manually deploy a test/main network to dev

jgreat:
### Motivation

Its currently a fiddly process to deploy a main/test signed artifacts to the dev cluster for smoke testing. This improves the process from 1h+ manual effort to 10m of automated "push-button" deployment.

- scripts to detect and set correct signed tokens.json based on version and network tier.
- add option to bootstrap a dispatch deploy from backup.
- update dispatch deploy defaults for v4 release.
- remove need to specify a tokens version.
- add secrets (really configs because the values are not secrets) for testnet/mainnet tokens.json
- add secrets for test/main ias 

This will be promoted to release/v4.1 -> master.

### DIFF
--- a/.internal-ci/util/.shared_functions
+++ b/.internal-ci/util/.shared_functions
@@ -1,0 +1,44 @@
+#!/bin/bash
+# Copyright (c) 2018-2023 The MobileCoin Foundation
+
+shopt -s extglob
+
+# 1: semver tag
+# returns: network tier dev|test|main
+get_network_tier()
+{
+    case "${1}" in
+        v+([0-9])\.+([0-9])\.+([0-9])-test )
+            echo "Found Test version ${1}" >&2
+            echo "test"
+        ;;
+        v+([0-9])\.+([0-9])\.+([0-9]) )
+            echo "Found Main version ${1}" >&2
+            echo "main"
+        ;;
+        * )
+            echo "Found Development version ${1}" >&2
+            echo "dev"
+        ;;
+    esac
+}
+
+# 1: semver tag
+# returns: major version
+get_major_version()
+{
+    # Trim off version from - "dev" metadata
+    version=${1%%-*}
+    echo "DEBUG: version: $version" >&2
+    # Trim off the "patch"
+    v_major_minor=${version%.*}
+    echo "DEBUG: v_major_minor $v_major_minor" >&2
+    # Trim off the minor
+    v_major=${v_major_minor%.*}
+    echo "DEBUG: v_major $v_major" >&2
+    # Trim off the v
+    major=${v_major#v}
+
+    echo "${major}"
+}
+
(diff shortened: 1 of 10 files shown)


jcape:
This should go into 4.1.

aweis89 (approved):

.internal-ci/util/.shared_functions
```diff
@@ -0,0 +1,44 @@
...
+{
+    # Trim off version from - "dev" metadata
+    version=${1%%-*}
+    echo "DEBUG: version: $version" >&2
+    # Trim off the "patch"
+    v_major_minor=${version%.*}
+    echo "DEBUG: v_major_minor $v_major_minor" >&2
+    # Trim off the minor
+    v_major=${v_major_minor%.*}
+    echo "DEBUG: v_major $v_major" >&2
```
nick-mobilecoin, line 38: :upside_down_face: is there a reason not to just stop at the first `.` the intermediate trimmings seem unecessary
```suggestion
    # Trim off version from - "dev" metadata
    v_major=${1%%.*}
```

.internal-ci/util/set_tokens_config_version.sh
```diff
@@ -0,0 +1,83 @@
...
+            exit 1
+        fi
+    ;;
+    *)
+        echo "ERROR: Unknown network ${network}" >&2
+        exit 1;
+    ;;
+esac
+
+echo "${config}"
```
nick-mobilecoin, line 83: :upside_down_face: the suggestion might be too indirect of a solution
`get_network_tier` returns `dev` for the default so the unknown network case won't be hit
```suggestion
if [[ ${major} -eq 0 ]]
then
    version="V2"
elif [[ ${major} -ge 1 ]] && [[ ${major} -le 3 ]]
then
    version="V1"
elif [[ ${major} -ge 4 ]]
then
    version="V2"
else
    echo "Major version is invalid? ${1} ${major}" >&2
    exit 1
fi

token_json="${network^^}_TOKENS_CONFIG_${version}_JSON"
echo "Using ${token_json}" >&2
echo "${!token_json}"
```

nick-mobilecoin (approved):
just some minor alternative suggestions

.internal-ci/util/set_tokens_config_version.sh
jgreat: I was debating if it would be more long term supportable (understandable by someone else in 6 months)  to use the longer case/if or construct the variable name.  I think I'll take your suggestions here and tweak them a bit and add some comments for the parameter expansion bits.  I can never remember what they do with out googling.

.internal-ci/util/.shared_functions
jgreat: And now I remember why I was breaking it down from the back up.  This is valid semver and used in some places like the dev builds to add metadata but reduce char count.  `v0-v4-0-smoketest-release.1753.sha-33ea8c0f`

Splitting on the first `.` gives you `v0-v4-0-smoketest-release` which isn't the version.

nick-mobilecoin: oh good to know maybe a comment in that function for future maintainers

PR #3062, opened 2023-01-27
minor improvements to faucet worker

cbeck88:


mobilecoind-dev-faucet/src/worker.rs
```diff
@@ -529,6 +533,12 @@ impl WorkerTokenState {
             }
         });
 
+        // Update the status of in-flight txs.
+        // Doing this before checking for new UTXOs avoids some races, where we find out
+        // that our TX settled but still think that one of the UTXOs that it used is
+        // unspent.
+        self.check_on_in_flight_txs(client, logger);
```
cbeck88, line 540: Moving this line here before the GetUnspentTxOutListRequest, instead of after, is the only functional change of this PR

mobilecoind-dev-faucet/src/worker.rs
```diff
@@ -406,14 +409,15 @@ struct WorkerTokenState {
...
-    // the same UTXO concurrently.
-    known_utxos: HashMap<KeyImage, UtxoTracker>,
+    // Trackers for the UTXOs that we have queued.
+    // A tracker is created for a UTXO at the time that we get it from mobilecoind,
+    // and added here. The tracker contains a one-shot channel that the other side
+    // uses to let us know what happened with this UTXO.
+    // As long as a UTXO is in this list, we won't requeue it, until its tracker
+    // has resolved either in success or an error. We must avoid requeuing it
+    // to avoid handing it out to two different consumers who then conflict with
+    // eachother.
```
wjuan-mob, line 419:
```suggestion
    // each other.
```

mobilecoind-dev-faucet/src/worker.rs
wjuan-mob: I think that the associated comment on check_on_in_flight_txs should also be updated to reflect this move. I'm sorry to have to ask for additional clarification but I'm a little confused about how this reduces the race condition. If the tx settles after getunspenttxoutlist is called, doesn't that still cause problems?

cbeck88: The race condition that I think I see in testing is:

0. Worker has a tx is inflight, spending a TXO, lets call it A.
1. worker asks mobilecoind for its utxos, and this list  includes A.
2. worker asks mobilecoind what txs are in flight, and finds that the Tx is now settled and A is not unspent anymore. But, A is still in the list it got before. This causes it to be selected for future Tx's later.

If we do it the other way, it's possible that we think a Tx is still in flight, but one of the UTXOs it spent isn't on the list anymore. However, this doesn't cause the same problems, because we don't end up selecting a UTXO that isn't valid anymore. I hope that makes sense. I'm not sure what more comments I can make that would be appropriate without being TMI.

wjuan-mob: That does make sense. I think that the example helped a lot. Thank you.
My other comment was just that the function check_on_in_flight_txs says that it's called before step_3/4, which is no longer accurate.

wjuan-mob (approved):

mobilecoind-dev-faucet/src/worker.rs
cbeck88: okay thank you!

mobilecoind-dev-faucet/src/worker.rs
```diff
@@ -406,20 +409,23 @@ struct WorkerTokenState {
...
+    // each other.
+    queued_utxo_trackers: HashMap<KeyImage, UtxoTracker>,
     // The queue of UTXOs with the target value
     sender: async_channel::Sender<UtxoRecord>,
     // If we submit a rebalancing transaction, the response we can use to track it
     // Only one of these will be used at a time, and split txs cannot be submitted
     // while this is in-flight.
     in_flight_rebalancing_tx_state: Option<SubmitTxResponse>,
+    // Track the key images associated to utxos in the rebalancing tx
+    in_flight_rebalancing_key_images: HashSet<KeyImage>,
```
cbeck88, line 428: I also added this as a minor improvement (I'm not sure it's really needed but it can't hurt)

mobilecoind-dev-faucet/src/worker.rs
```diff
@@ -1012,6 +1039,7 @@ impl WorkerTokenState {
     // Check if a key image is part of an in-flight transaction
     fn key_image_is_in_flight(&self, key_image: &KeyImage) -> bool {
         self.in_flight_split_tx_states.contains_key(key_image)
+            || self.in_flight_rebalancing_key_images.contains(key_image)
```
cbeck88, line 1042: i added this also to try to avoid (and added the self.in_flight_rebalancing_key_images variable) to try to reduce the "key image already spent" errors. I'm not sure it will really fix it but worth trying

jgreat:
The CD failure here seems to happen a whole lot more after we integrated the faucet.   The cd-integration-tests stalls out on one of the python scripts that sets up monitors waiting for the blocks to sync. Mobilecoind doesn't show any errors when this happens. I'm guessing some kind of weirdness/race when multiple clients are trying to watch monitors?

cbeck88:
I would kind of like to lean into the faucet and not use "drain accounts"? What do you think about that?

cbeck88:
Maybe it will be better with the "activate" option? So the faucet wont do anything while drain accounts is trying to work? https://github.com/mobilecoinfoundation/mobilecoin/pull/3063

nick-mobilecoin (approved):

mobilecoind-dev-faucet/src/lib.rs
```diff
@@ -64,13 +64,19 @@ pub struct Config {
     /// Target Queue Depth. When the queue for a token id is less than this in
     /// depth, the worker attempts to make a split Tx to produce more TxOuts
     /// for the queue.
-    #[clap(long, default_value = "10000", env = "MC_TARGET_QUEUE_DEPTH")]
+    #[clap(long, default_value = "100000", env = "MC_TARGET_QUEUE_DEPTH")]
```
wjuan-mob, line 67: Why did we up this default?

cbeck88: because in the graphs we generated, we didn't really sustain the peak in a way that makes it clear what the peak is

historically slam used 100,000 txs -- I went down to lower numbers for ease of development and initial testing, and i was kind of hopefuly that 10 or 20 thousand would be enough to see the peak, but it looks like in a deployed network it's not enough

mobilecoind-dev-faucet/src/lib.rs
```diff
@@ -247,6 +257,9 @@ impl State {
         &self,
         req: &JsonFaucetRequest,
     ) -> Result<api::SubmitTxResponse, String> {
+        if !self.worker.activate() {
```
wjuan-mob, line 260: Would it make sense to centralize this logic somewhere? Maybe I'm thinking about this in the wrong way, but it feels like if we want it to be the case that if we activate the worker on any type of post request, we should be doing that in the rocket code to call activate on post requests instead of having it activate on these specific functions.

cbeck88: we could make a "self.activate" function, sure

wjuan-mob (approved):

PR #3063, opened 2023-01-28
Add an "activate" option for the mobilecoind-dev-faucet

cbeck88:
Jason requested that, the mobilecoind-dev-faucet should not immediately start sending lots of transactions after it is deployed, because that makes some other deployment-related things harder. By default it does that because it wants to have its queue full.

The change we made is:

* The faucet worker now has an "is_active" flag, which if false causes it to skip its loop, and this defaults to false.
* The flag is set to true if `--activate` is passed.
* The flag is set true whenever an HTTP POST occurs (post to get payment from faucet, or to start a slam), whether or not the POST results in a successful operation.
* Docu is updated so that for dev testing it demonstrates `--activate` and the existing documented flows should still work.
* CI is updated to use `--activate`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -505,7 +505,7 @@ jobs:
           # Run mobilecoind-dev-faucet
           MC_LOG="info" \
           "$BIN_DIR"/mobilecoind-dev-faucet \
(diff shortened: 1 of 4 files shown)


jgreat (approved):
Thanks!

PR #3078, opened 2023-02-01
Make mobilecoind, dev-faucet, fog-test-client use MRENCLAVE and support many css

cbeck88:
This makes all the targets that create attestation verifiers in this repo use MRENCLAVE for verification, and accept multiple CSS files (rather than just one) as a CLI arg or env variable so that they can easily be made to trust several versions of the enclave.

For things like dev-faucet and fog-test-client, this may simplify things slightly operationally, since you won't have to rebuild against an old releases' CSS files, you can simply install the old release CSS with the deployed binary.

For mobilecoind, this improves the trust model, since we're really using SGX now to ensure that the correct code is running in the network. It also ensures we're checking fog ingest using the MRENCLAVE validation.


mobilecoind/src/config.rs
```diff
@@ -316,6 +328,31 @@ impl PeersConfig {
             .collect()
     }
 
+    /// Create a consensus enclave verifier, which allows the baked-in
+    /// sigstruct, and any sigstructs produced at run-time via the
+    /// MC_CONSENSUS_ENCLAVE_CSS list. This uses MRENCLAVE verification.
+    pub fn create_consensus_verifier(&self, logger: &Logger) -> Verifier {
```
cbeck88, line 334: this is probably the most sensitive part of the change -- it changes how mobilecoind is constructing its consensus enclave verifier

mobilecoind/src/bin/main.rs
```diff
@@ -31,18 +30,12 @@ fn main() {
...
 
-    let mut mr_signer_verifier =
-        MrSignerVerifier::from(mc_consensus_enclave_measurement::sigstruct());
-    mr_signer_verifier
-        .allow_hardening_advisories(mc_consensus_enclave_measurement::HARDENING_ADVISORIES);
-
-    let mut verifier = Verifier::default();
-    verifier.mr_signer(mr_signer_verifier).debug(DEBUG_ENCLAVE);
-
-    log::debug!(logger, "Verifier: {:?}", verifier);
```
cbeck88, line 42: this is where mobilecoind previously constructed the consensus enclave verifier

mobilecoind/src/config.rs
```diff
@@ -87,8 +90,8 @@ pub struct Config {
 
     /// Fog ingest enclave CSS file (needed in order to enable sending
     /// transactions to fog recipients).
-    #[clap(long, value_parser = load_css_file, env = "MC_FOG_INGEST_ENCLAVE_CSS")]
-    pub fog_ingest_enclave_css: Option<Signature>,
+    #[clap(long, value_parser = load_css_file, env = "MC_FOG_INGEST_ENCLAVE_CSS", use_value_delimiter = true)]
+    pub fog_ingest_enclave_css: Vec<Signature>,
```
cbeck88, line 94: I believe that using `use_value_delimiter=true` is correct here, and will allow that a list like `foo.css,bar.css` can be the value of `MC_FOG_INGEST_ENCLAVE_CSS` environment variable, and then this will be split by clap, and then the arguments passed one by one to `load_css_file`. I did not actually test this, I probably should.

fog/sample-paykit/src/client_builder.rs
```diff
@@ -255,85 +234,97 @@ impl ClientBuilder {
...
-            mc_fog_ingest_enclave_measurement::get_mr_signer_verifier(None)
-        };
+            verifier.mr_enclave(mr_enclave_verifier);
+        }
+
+        log::debug!(self.logger, "Fog Ingest Verifier: {:?}", verifier);
 
-        let mut verifier = Verifier::default();
-        verifier.debug(DEBUG_ENCLAVE).mr_signer(mr_signer_verifier);
         verifier
```
cbeck88, line 281: this logic is somewhat repetitive, it might be nice to make a helper function somewhere, but I didn't want to go too deep on refactoring right now.

mobilecoind/src/config.rs
cbeck88: okay, i added some more tests of this to build confidence

mobilecoind/src/config.rs
```diff
@@ -166,26 +169,27 @@ impl Config {
...
-                let mut mr_signer_verifier = MrSignerVerifier::new(
-                    signature.mrsigner().into(),
-                    signature.product_id(),
-                    signature.version(),
-                );
-                mr_signer_verifier.allow_hardening_advisories(&[
-                    "INTEL-SA-00334",
-                    "INTEL-SA-00615",
-                    "INTEL-SA-00657",
-                ]);
```
cbeck88, line 181: I think this earlier code was somewhat wrong anyways, there is a better conversion like `MrSignerVerifier::from(..._sigstruct())` that was being used elsewhere, and the hardening advisories are now in a measurement crate list

fog/test-client/src/config.rs
```diff
@@ -106,21 +106,21 @@ pub struct TestClientConfig {
...
-    #[clap(long, env = "MC_CONSENSUS_ENCLAVE_CSS")]
-    pub consensus_enclave_css: Option<String>,
+    /// Consensus enclave CSS file(s) (additional to the build-time CSS)
+    #[clap(long, env = "MC_CONSENSUS_ENCLAVE_CSS", use_value_delimiter = true)]
+    pub consensus_enclave_css: Vec<String>,
 
-    /// Fog ingest enclave CSS file (overriding the build-time CSS)
-    #[clap(long, env = "MC_INGEST_ENCLAVE_CSS")]
-    pub ingest_enclave_css: Option<String>,
+    /// Fog ingest enclave CSS file(s) (additional the build-time CSS)
```
wjuan-mob, line 113: Nit: Here and below
```suggestion
    /// Fog ingest enclave CSS file(s) (additional to the build-time CSS)
```

fog/test-client/src/config.rs
```diff
@@ -106,21 +106,21 @@ pub struct TestClientConfig {
...
+    #[clap(long, env = "MC_INGEST_ENCLAVE_CSS", use_value_delimiter = true)]
+    pub ingest_enclave_css: Vec<String>,
 
-    /// Fog ledger enclave CSS file (overriding the build-time CSS)
-    #[clap(long, env = "MC_LEDGER_ENCLAVE_CSS")]
-    pub ledger_enclave_css: Option<String>,
+    /// Fog ledger enclave CSS file(s) (additional the build-time CSS)
```
wjuan-mob, line 117:
```suggestion
    /// Fog ledger enclave CSS file(s) (additional to the build-time CSS)
```

fog/test-client/src/config.rs
```diff
@@ -106,21 +106,21 @@ pub struct TestClientConfig {
...
+    #[clap(long, env = "MC_LEDGER_ENCLAVE_CSS", use_value_delimiter = true)]
+    pub ledger_enclave_css: Vec<String>,
 
-    /// Fog view enclave CSS file (overriding the build-time CSS)
-    #[clap(long, env = "MC_VIEW_ENCLAVE_CSS")]
-    pub view_enclave_css: Option<String>,
+    /// Fog view enclave CSS file(s) (additional the build-time CSS)
```
wjuan-mob, line 121:
```suggestion
    /// Fog view enclave CSS file(s) (additional to the build-time CSS)
```

fog/sample-paykit/src/client_builder.rs
```diff
@@ -42,11 +42,11 @@ pub struct ClientBuilder {
     // Address book, for memos
     address_book: Vec<PublicAddress>,
 
-    // Optional sigstructs for attested services
-    consensus_sigstruct: Option<Signature>,
-    fog_ingest_sigstruct: Option<Signature>,
-    fog_ledger_sigstruct: Option<Signature>,
-    fog_view_sigstruct: Option<Signature>,
+    // Optional additional sigstructs for attested services
+    consensus_sigstruct: Vec<Signature>,
```
wjuan-mob, line 46: Should these names be pluralized to match with the change?

mobilecoind/src/config.rs
```diff
@@ -203,8 +209,6 @@ impl Config {
 
         let conn = GrpcFogReportConnection::new(self.peers_config.chain_id.to_owned(), env, logger);
 
-        let verifier = self.get_fog_ingest_verifier();
```
wjuan-mob, line 206: Sorry for my ignorance, but was there a reason to move this line?

cbeck88: It resolved a borrow checker issue, we could also have cloned the logger

fog/sample-paykit/src/client_builder.rs
cbeck88: Yea okay

nick-mobilecoin (approved):

fog/sample-paykit/src/client_builder.rs
```diff
@@ -255,85 +234,97 @@ impl ClientBuilder {
...
-                signature.version(),
-            );
-            mr_signer_verifier
+        let mut verifier = Verifier::default();
+        verifier.debug(DEBUG_ENCLAVE);
+
+        let mut mr_enclave_verifier =
+            MrEnclaveVerifier::from(mc_consensus_enclave_measurement::sigstruct());
+        mr_enclave_verifier
+            .allow_hardening_advisories(mc_consensus_enclave_measurement::HARDENING_ADVISORIES);
```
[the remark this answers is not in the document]
nick-mobilecoin: Thinking more, a `From()` trait wouldn't work unless it was a `ConsensusSignature` since one needs to know which advisories to use.

mobilecoind/src/config.rs
```diff
@@ -358,3 +395,115 @@ impl PeersConfig {
...
+            "mobilecoind",
+            "--chain-id",
+            "foo",
+            "--peer",
+            "mc://test",
+            "--tx-source-url",
+            "example.com",
+            "--consensus-enclave-css",
+            valid_css_path,
+        ]);
```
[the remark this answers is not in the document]
cbeck88: okay sounds good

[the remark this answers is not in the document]
cbeck88: maybe `.unwrap()` is better than `assert!(_.is_ok()))`

fog/test-client/src/test_client.rs
```diff
@@ -166,40 +166,44 @@ impl TestClient {
...
     }
 
-    /// Set the consensus sigstruct used by the clients
+    /// Set the consensus sigstruct(s) trusted by the clients, besides
+    /// build-time CSS
     #[must_use]
-    pub fn consensus_sigstruct(mut self, sig: Option<Signature>) -> Self {
-        self.consensus_sig = sig;
+    pub fn consensus_sigstructs(mut self, sigs: Vec<Signature>) -> Self {
+        self.consensus_sigs = sigs;
```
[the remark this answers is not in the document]
cbeck88: yeah the test-client thing is not actually a "builder pattern" like the client builder is. maybe it should be.

happy to change this to make the semantics align, i went back and forth a couple times

cbeck88:
this needs to be rebased, and I want to wait for the latest roll up from 4.1 to land: https://github.com/mobilecoinfoundation/mobilecoin/pull/3078

mobilecoind-dev-faucet/src/lib.rs
```diff
@@ -81,6 +83,16 @@ pub struct Config {
     ///     env MC_PEER=mc://foo:123,mc://bar:456
     #[clap(long = "peer", env = "MC_PEER", use_value_delimiter = true)]
     pub peers: Option<Vec<ConsensusClientUri>>,
+
+    /// Consensus enclave CSS file(s) to trust. Any sigstruct provided here will
+    /// be trusted in attestation with consensus. This can be used to allow
+    /// attestation with a previous version of consensus as well as the
+    /// current version. Note that we also bake in a CSS file at build time
+    /// so it may be okay to omit this parameter in a lot of testing.
```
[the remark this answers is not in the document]
cbeck88: > The reason being the person may want to slam a specific network and limiting to the CLI provided values ensures they aren't hitting a network they didn't intend.

I'm hoping that chain-id is another failsafe to prevent that. It seems very convenient to keep the build-time one, because if we are willing to trust the build-time one, why would we suddenly not trust that if CSS are provided at runtime? Shouldn't the set of trusted things only grow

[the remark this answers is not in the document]
jcape: All previously released versions should _not_ be supported: an enclave that's vulnerable to LVI should not be trusted, for example.

jcape (changes requested):
This PR is necessary, but insufficient. Each MRENCLAVE value must be accompanied by an individualized list of acceptable advisories, so you're going to need more than just Vec<Signature>. I'd offer that what you really want to come out of this config is a `Vec<StatusVerifier>`, with an appropriate function that will parse a single string into that, e.g.:

```
file:///path/to/enclave.css?advisories=INTEL-SA-00334,INTEL-SA-00615,INTEL-SA-00657
0xMRENCLAVE?advisories=FOO
0xMRSIGNER?product=1&svn=5&advisories=FOOBAR
```

cbeck88:
Okay, I think I understand, let me just repeat back to you to make sure:

Suppose that, in the next release, we learn that there is a new advisory INTEL-SA-00666 about nasal demons in the microcode or some such thing.

Intel will soon (but not yet) begin flagging our enclaves in the wild with this advisory and attestations will fail at that point, since it isn't in any of the existing softwares' advisories list. All the clients will have to be updated.

For the new clients to work, they have to work with the servers that are deployed today (that are potentially affected by nasal demons, but still in the grace period), and the next released (mitigated) enclave. If the clients aren't compatible with new and old then we won't be able to complete the migration.

So you'd want the new client to have Intel-SA-00666 listed as mitigated in the new release (and subsequent releases), but not treat that advisory as mitigated when checking older enclaves? So after the grace period is up, and intel is actually issuing Intel-SA-00666 from IAS, attestation stops working with the older clients, and even with the newer clients.

So this is why you want it to be possible to have different advisory lists for different sigstructs.

cbeck88:
1. Did I get that right?
2. What if, we make some kind of file format, like, json or protobuf, to list pairs of sigstruct bytes and advisory lists, instead of putting the association in a query string? I feel like creating data structures in CLI parameters is fraught
3. Is it necessary to block this change on this request? It seems to me that the issue you are raising is orthogonal to the issue I am trying to fix right now.

The thing I am trying to fix right now is, let's use MRENCLAVE instead of MRSIGNER, we seem to agree that there's no good reason to be using MRSIGNER, and it's not even really making packaging or releasing simpler.

The issue you are now raising is, you don't want the new set of advisories to the apply to the old enclave. But that is actually the status quo. We currently can only apply one list, and we use MRSIGNER with the older version number, so the "mitigated advisories" for the next release apply also to the enclave from the previous release. So this is not a new issue in my PR, it's pre-existing.

Can we not improve the situation incrementally rather than trying to fix everything at once in one giant PR? Using MRENCLAVE  rather than MRSIGNER is a significant improvement over the status quo imo.

cbeck88:
I see, so maybe it's like, let's only change the command-line interface to these programs once, rather than changing it twice and making ops and wallet teams change things around twice?

cbeck88:
okay, let's chat some more later, i may close this and make another PR. gonna think about it somemore

cbeck88:
gonna close this and continue in #3085

PR #3080, opened 2023-02-01
write additional helpful data in the json slam report

cbeck88:
this stores the units of the "prepare_time" and "submit_time" in the names of these json elements.

it also computes the average throughput of the slam operation -- total transactions divided by total time. this is a rough approximation to the peak tx/s that should be reasonably accurate for a sustained slam.

this number is obtained programmatically from the tool conducting the slam, rather than by humans looking at graphs, and so doesn't have the fuzziness of people judging lines, which in my mind gives it some merit to be worth reporting.
(not that it should replace the number we normally use)


NotGyro (approved):
Looks good to me.

wjuan-mob (approved):

PR #3081, opened 2023-02-01
Make Fog View Store return externally-addressable URI to Router

samdealy:
### Motivation
When deploying Fog View Router to a local network, we discovered that the router was trying to contact the store via a url with host == `0.0.0.0`. This occurred because the store used its `client_listen_url` to construct the URI that it gave to the router during the auth process. The router was unable to contact the store because it operated on a different host, not the same host as the router.

To fix this, I added a `responder_id` query param to the `FogViewStoreUri`, and the value is the externally addressable responder_id in "{HOST}:{PORT}" format. This enables the router to contact the store running in the same local network on a different machine.

### DIFF
--- a/fog/types/src/view.rs
+++ b/fog/types/src/view.rs
@@ -697,7 +697,7 @@ mod tests {
     fn tx_out_search_result_conversion(payload_length: usize) {
         let fixed_tx_out_search_results = (0..10)
             .map(|i| {
-                let payload = vec![i; payload_length as usize];
+                let payload = vec![i; payload_length];
                 FixedTxOutSearchResult::new(vec![i], &payload, TxOutSearchResultCode::Found)
             })
             .collect::<Vec<_>>();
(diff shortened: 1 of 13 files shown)


samdealy:
* feature/fog-view-router
  * **PR #3081** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3081" target="_blank">Graphite</a>  👈
    * **PR #3082** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3082" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3081?utm_source=stack-comment).

awygle (approved):
This looks like it works. One question though - what's the difference between the `responder-id` query param on the `FogViewStoreUri` and the `client_responder_id` config parameter?

samdealy:
> This looks like it works. One question though - what's the difference between the `responder-id` query param on the `FogViewStoreUri` and the `client_responder_id` config parameter?

@jcape and I thought it would be cleaner to put all info that identifies the store in the URL. We are probably going to remove the `CLIENT_RESPONDER_ID` field.

jcape (approved):

PR #3083, opened 2023-02-01
add memo-mac crate, backport changes for ledger desktop wallet support

ryankurte:
backporting a minimum changeset from `ledger-wip` to support building the desktop wallet against `release/4.1`.

- adds `crypto/memo-mac` from #2585
- ~adds `memo` functionality to `core` (this could equally be part of `memo-mac`?~
- moves `ShortAddressHash` from `account-keys` to `core`
- minor changes to tx summary interfaces for embedded use (substantial refactoring of `transaction` is planned in further PRs)

cc. @briancorbin


transaction/extra/src/tx_summary_unblinding/verifier.rs
```diff
@@ -143,7 +155,7 @@ impl TxSummaryStreamingVerifier {
     /// Returns:
     /// * A properly initialized TxSummaryStreamingVerifier
     pub fn new(
-        extended_message_digest: &[u8; 32],
+        extended_message_digest: &[u8],
```
cbeck88, line 158: Is there a good reason for this change?

I think this may change the meaning of the hash at line 154/166 in this diff. When you digest a fixed size array, I don't remember if it prepends the length. Even if it works, what advantage is there to this?

cbeck88: It seems much better to fail fast int he extended message digest is not 32 bytes -- this will always be broken and give a garbage output if it is not 32 bytes.

transaction/extra/Cargo.toml
```diff
@@ -9,6 +9,7 @@ edition = "2021"
 cfg-if = "1.0"
 displaydoc = { version = "0.2", default-features = false }
 hmac = "0.12"
+mc-account-keys = { path = "../../account-keys" }
```
cbeck88, line 12: why is this being moved? this breaks our convention

crypto/ring-signature/signer/src/traits.rs
```diff
@@ -73,6 +73,10 @@ impl From<RistrettoPrivate> for OneTimeKeyDeriveData {
 ///
 /// A transaction builder can be built around this.
 pub trait RingSigner {
+    /// Implementation errors must be convertable to standard [SignerError] type
+    /// for common handling
+    type Error: Into<SignerError>;
```
cbeck88, line 78: So, what is the advantage of using a generic associated type for `Error` here?

It seems that, with this trait bound, the only thing a caller to `RingSigner` can possibly do with this is convert the given error to `SignerError`. And the callee must implement a way to do that conversion. So why not cut to the chase and always return `SignerError`? It seems like we just have more boilerplate and more conversions pushed onto the caller if we go this route.

cbeck88: It also makes `RingSigner` not object safe, so we cannot have like `Box<dyn RingSigner>` which may be useful.

cbeck88 (changes requested):
left some comments on things that look wrong to me

crypto/ring-signature/signer/src/traits.rs
cbeck88: I could understand if you were trying to like, break a dependency cycle or something. But it seems that, any implementer of RingSigner must depend on this crate, for this trait, and for this error type now -- they still have to have a definition of `SignerError` to implement `Into<SignerError>`. So I can't understand what we're getting by doing this.

transaction/extra/src/tx_summary_unblinding/verifier.rs
cbeck88: the MCIP for tx summary verification is only applicable when the block version is relatively late, at a place where extended message digest is 32 bytes

ryankurte: > It seems much better to fail fast if the extended message digest is not 32 bytes -- this will always be broken and give a garbage output if it is not 32 bytes.

the `heapless::Vec<u8; 32>` i'm using for storing messages doesn't implement `&[u8; 32]`... but i can probably swap this with an `Option<[u8; 32]>` if there is _no case_ where a message / digest will ever be a different length?

transaction/extra/Cargo.toml
ryankurte: oops, good catch!

crypto/ring-signature/signer/src/traits.rs
ryankurte: the concept was that it should be possible to pass up things like transport errors for the ring signer implementation,  for the moment they just get squashed in the transaction signer anyway, but ideally they'd be propagated out.

for an object safe alternate we could have an `Transport<Box<dyn Error>>` so the error type is fixed but we can include transport-specific information?

transaction/extra/src/tx_summary_unblinding/verifier.rs
cbeck88: can we do like `heapless::Vec<u8; 32>::as_ref().try_into()?` or something like this, at the call site?

i feel like it's okay to make the caller responsible for checking this

crypto/ring-signature/signer/src/traits.rs
cbeck88: So I think what I'd suggest is, let's try to think about who is using the `RingSigner`, and how we think they might handle transport errors. Presumably there's like, retriable transport errors, and fatal transport errors? So maybe we can add some possibilities like that in `SignerError`? And otherwise the transport error details are just logged?

And if we find we need more categories later, we can simply add more possibilities to `SignerError`?

cbeck88: it just seems like that's the simplest thing that would work -- the caller probably doesn't need all the details of the transport error, just how they should handle it.

and if we're going to have a "common error" type that all implementations work with, for "common handling", why not just make that the error type.

clearly, we're expecting that the callers are going to squash the implementation specific errors this way anyways, right? so when you say this:

>  but ideally they'd be propagated up (and the error type is consistent with other interfaces, not just the ring signer).

you're not actually coding towards that, right? you're sort of giving the implementation the option to return whatever error they want, but the caller is then going to have to squash it, whatever it is

cbeck88: i'm still thinking about this part:

>  (and the error type is consistent with other interfaces, not just the ring signer)

can you give a concrete example where that is really helpful?

ryankurte: it's probably not helpful, just, consistent. i guess the opposite question is, should we define concrete error types for [other traits](https://github.com/mobilecoinfoundation/mobilecoin/blob/9fbee911bcd183d8afb042d5c9e07bbac7185915/transaction/signer/src/traits.rs#L20)?

i think i actually did look at embedding a boxed type, but there's a serde constraint on [`SignerError`](https://github.com/mobilecoinfoundation/mobilecoin/blob/master/crypto/ring-signature/signer/src/traits.rs#L127) which wasn't so straightforward to work around

cbeck88: Hmm, I wonder why we put `serde` on the `SignerError`, that's somewhat unusual

I'm okay with putting concrete errors for the other traits? I guess they should all accommodate transport error

If you want to go the other way, we could define an additional trait for these errors, and make all of them implement it? We did something like that here:

https://github.com/mobilecoinfoundation/mobilecoin/blob/f106b4fa8a03779ddf287b041687685d386db843/connection/src/traits.rs#L31

cbeck88: I would be okay with generic errors for `RingSigner` and a way to interrogate them as well, and make the current `SignerError` only be the error type for the `NoKeysRingSigner` etc. I just think if we're going to make all errors convertible to `SignerError`, then that should be the error type always.

ryankurte: the conversion bounds trick is just one way to support interrogating generic errors, but it's nbd either way.  have reverted the changes to the error for now / will introduce more states as required ^_^

cbeck88 (approved):

briancorbin (approved):

briancorbin:
* master
  * **PR #3083** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3083" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3083?utm_source=stack-comment).

dorebell (approved):
Looks good to me!

core/types/src/lib.rs
```diff
@@ -9,3 +9,5 @@ pub mod account;
 pub mod keys;
 
 pub mod markers;
+
```
jcape, line 12: Could you open an issue to refactor this to follow [the style guide](https://github.com/mobilecoinfoundation/mobilecoin/blob/master/CONTRIBUTING.md#coding-style) before we release 5.0?

crypto/memo-mac/src/lib.rs
```diff
@@ -0,0 +1,43 @@
+#![no_std]
+// Copyright (c) 2018-2022 The MobileCoin Foundation
+
+//! HMAC code shared by all category 0x01 memos.
+//!
+//! This validation scheme was proposed for standardization in
+//! mobilecoinfoundation/mcips/pull/4
+
```
jcape, line 8:
```suggestion
// Copyright (c) 2018-2022 The MobileCoin Foundation

//! HMAC code shared by all category 0x01 memos.
//!
//! This validation scheme was proposed for standardization in
//! [MCIP #4](https://github.com/mobilecoinfoundation/mcips/blob/main/text/0004-recoverable-transaction-history.md)

#![no_std]

```

crypto/ring-signature/signer/src/traits.rs
```diff
@@ -116,15 +116,15 @@ impl<S: RingSigner> RingSigner for &S {
...
+    ) -> Result<RingMLSAG, SignerError> {
         (*self).sign(message, signable_ring, output_blinding, rng)
     }
 }
 
 /// An error that can occur when using an abstract RingSigner
 #[derive(Clone, Debug, Display, Eq, Hash, Ord, PartialEq, PartialOrd)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-pub enum Error {
+pub enum SignerError {
```
jcape, line 127: Seems like this change is purely cosmetic, is there a reason not to keep it as just `Error`?

core/Cargo.toml
```diff
@@ -28,21 +29,23 @@ mc-core-types = { path = "./types" }
 mc-crypto-dalek = { path = "../crypto/dalek", default-features = false }
 mc-crypto-hashes = { path = "../crypto/hashes", default-features = false }
 mc-crypto-keys = { path = "../crypto/keys", default-features = false }
+mc-crypto-memo-mac = { path = "../crypto/memo-mac", default-features = false }
```
jcape, line 32: Is this actually required? Enclaves are never going to use this...

jcape (approved):
If there's a strong need to merge this now, then we can do so, but I'd really prefer to not add more code to the secure enclaves than is strictly necessary...

core/Cargo.toml
jcape: I suppose a similar question surrounds `AccountId` and `ShortAddressHash`...

crypto/ring-signature/signer/src/traits.rs
ryankurte: it gets renamed to `SignerError` basically everywhere it's used, figured it makes sense to rename it at the root and avoid the collision everywhere else.

core/Cargo.toml
ryankurte: you're right the memo include from `core` can be dropped, that's a remnant from before i minimised this.

`ShortAddressHash` is used in the computation of a TxSummary so this is already in the enclave. previously it was created from a public address when streaming the summary, i needed to lift it so we can [stream](https://mobilecoinofficial.github.io/ledger-mob/ledger_mob_apdu/tx/struct.TxSummaryAddTxOutUnblinding.html) this in for hardware tx summary computation to avoid the need for full `AccountKeys` objects (and the account keys dependency).

`AccountId` is lifted from full-service because we needed a common way to refer to an account between full-service / desktop-wallet / the ledger CI. i'll move this to `mc-transaction-signer` which isn't included in `mc-transaction-core`.

jcape (approved):

crypto/ring-signature/signer/src/traits.rs
jcape: Can we undo this change, then?

PR #3085, opened 2023-02-02
WIP: make a way to store hardening advisory data with sigstructs and load it

cbeck88:
[Rendered README](https://github.com/mobilecoinfoundation/mobilecoin/blob/attest-verifier-config/attest/verifier/config/README.md)

This is meant to address comments on PR #3078, and allow that a client can load multiple MRENCLAVE measurements on startup and also apply different sets of hardening advisories to these as appropriate.

The idea is that there is a directory containing subdirectories with `.css` sigstruct files and associated `.json` files. A client binary can walk this directory and configure `MRENCLAVE` or `MRSIGNER` verifiers using additional data obtained from the json files. We can ship the whole bundle with the client, and it gets installed on the user's filesystem with the app.

There is a lot of ambiguity here. Mostly we are optimizing towards, reducing the amount of new tools we will have to create and use, and particularly, avoiding creating a file format that embeds Intel's sigstruct binary data.

If there is approval we will try to redevelop PR #3078 based on this.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2533,6 +2533,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "mc-attest-verifier-config"
+version = "4.0.2"
+dependencies = [
+ "displaydoc",
+ "mc-attest-verifier",
+ "mc-common",
+ "mc-sgx-css",
+ "serde",
+ "serde_json",
+]
+
+[[package]]
 name = "mc-attest-verifier-types"
 version = "4.0.2"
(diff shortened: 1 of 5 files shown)


jgreat:
Could we accomplish this with just values in a single json file instead of juggling a whole bunch of CSS and metadata? Maybe using the human readable(ish) MRENCLAVE hashes we post in releases.

cbeck88:
Yea that's what James said too, I plan to redevelop this

PR #3097, opened 2023-02-03
Make find_record constant time

samdealy:
###
This PR makes the `e_tx_out_store`'s `find_record` method constant time. This is necessary in order to preserve our obliviousness guraantees.

### DIFF
--- a/fog/view/enclave/impl/src/e_tx_out_store.rs
+++ b/fog/view/enclave/impl/src/e_tx_out_store.rs
@@ -7,7 +7,7 @@
 use alloc::vec;
 
 use aligned_cmov::{
-    subtle::{Choice, ConstantTimeEq},
+    subtle::{Choice, ConditionallySelectable, ConstantTimeEq, ConstantTimeGreater},
     typenum::{Unsigned, U1024, U16, U240, U4096, U64},
     A8Bytes, CMov,
 };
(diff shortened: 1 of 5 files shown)


samdealy:
* feature/fog-view-router
  * **PR #3081** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3081" target="_blank">Graphite</a> 
    * **PR #3082** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3082" target="_blank">Graphite</a> 
      * **PR #3097** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3097" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3097?utm_source=stack-comment).

jcape:
LGTM, but I wrote half these changes, so I don't feel fully confident reviewing them :-)

cbeck88:
I will try to review tomorrow

fog/view/enclave/impl/src/lib.rs
```diff
@@ -4,6 +4,7 @@
 
 #![no_std]
 #![allow(clippy::result_large_err)]
+#![feature(const_cmp)]
```
cbeck88, line 7: i suspect that this is unnecessary:

if we just did 

```
let LENGTH_TO_COPY: usize = core::cmp::min(FIXED_CIPHERTEXT_LENGTH, ValueSize::USIZE - 1);
```

instead of `const`, it would still be subject to constant folding optimizations by llvm. the only reason to make this `const` here is if you want to use it in a generic parameter to a type or something like this

jcape: I take the point that it's "one more nightly feature," but I'd prefer to use a seemingly likely-to-stabilize nightly feature and know it's always going to happen vs. depending on the compiler doing a particular thing in a particular way.

fog/view/enclave/impl/src/e_tx_out_store.rs
```diff
@@ -184,24 +184,24 @@ impl<OSC: ORAMStorageCreator<StorageDataSize, StorageMetaSize>> ETxOutStore<OSC>
...
-        let payload = &value[1..data_end];
-        // Use this instead of payload.len() because the slice `len` method isn't
-        // guaranteed to be constant time.
-        let payload_length = data_end - 1;
-        result.ciphertext[0..payload_length].copy_from_slice(payload);
-        result.payload_length = payload_length as u32;
+        // Always copy the same length to ensure constant time. Note that we are not
+        // copying the payload length because this is variable and would
+        // compromise obliviousness.
+        const LENGTH_TO_COPY: usize = core::cmp::min(FIXED_CIPHERTEXT_LENGTH, ValueSize::USIZE - 1);
```
cbeck88, line 190: I want to make a suggestion:

Instead of taking a `min` of several things with `FIXED_CIPHERTEXT_LENGTH`, let's assert that `Value::USIZE - 1` is less than the `FIXED_CIPHERTEXT_LENGTH`. (Or maybe a `const_assert`.) Since this is all fucked if that assertion isn't right, and then it will make it simpler to see how many bytes are actually copied, and there's no code-path where we silently truncate part of the payload.

cbeck88 (changes requested):
left a comment, instead of having a codepath where we might truncate part of the payload, let's use assertions to ensure that `FIXED_CIPHERTEXT_LENGTH` is large enough relative to the other parameters, so that these `core::cmp::min` calls are unnecessary and truncation will not happen

fog/view/enclave/impl/src/e_tx_out_store.rs
samdealy: That sounds good to me. @jcape what do you think?

jcape: Yep, `const_assert` sounds totally reasonable.

cbeck88 (approved):

Issue #1409, opened 2022-01-31
Set rust-version in all crates


jcape:
Set the `rust-version` of all crates to at least 1.58.1.

PR #3122, opened 2023-02-11
Update MSRV to 1.68.0

samdealy:
### Motivation
This PR makes the MSRV (minimum supported rust version) to `1.68.0` (see #1409). Note that `1.68.0` matches the rust version defined in `rust-toolchain`.

Note that I created a cargo `workspace` for the `sgx` crate. This allows all sub-crates in the `sgx` to access the `rust-version` set in the (new) top-level sgx `Cargo.toml`. 

Also note that to had to implement the `rust-version` field inheritance as follows: 
```
rust-version = { workspace = true }
```
Instead of the recommended [way](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table)
```
rust-version.workspace = true
```
Both ways are valid, but the latter breaks cargo-sort (see DevinR528/cargo-sort#47). Until this issue is resolved, we can just use the former format.

### Future Work
* Once DevinR528/cargo-sort#47 is resolved, update to the `.` format


sgx/Cargo.toml
```diff
@@ -0,0 +1,4 @@
+[workspace]
+
+[workspace.package]
+rust-version = "1.68.0"
```
cbeck88, line 4: just curious, what's the reason for this change? why do we need a new workspace for this?

Cargo.toml
```diff
@@ -179,6 +179,9 @@ exclude = [
     "util/serial",
 ]
 
+[workspace.package]
```
cbeck88, line 182: if we are making `sgx/` it's own workspace, we also need to remove all the `sgx/` paths from the `members` list above, and we likely need to add `sgx/` to the `exclude` list above

cbeck88: hmm, for some reason `sgx` is already in the exclude list. i think i added that to fix a testing issue a long time ago, but i don't remember the details.

cbeck88:
it might be simpler to do the "sgx is it's own workspace" change independently of the "everything declares an MSRV" change -- is there some reason those have to be at the same time?

nick-mobilecoin (approved):
Looks like the following didn't get the update:
- `util/ffi/Cargo.toml`
- `sgx/debug/Cargo.toml`

nick-mobilecoin:
> Looks like the following didn't get the update:
> 
>     * `util/ffi/Cargo.toml`
> 
>     * `sgx/debug/Cargo.toml`

NVM it looks like these fall on their face when linting.

sgx/Cargo.toml
samdealy: It allows all of the sub-packages to access this value. That way, when we update the MSRV again, we don't have to change ~12 values; we only have to change this one value.

Cargo.toml
samdealy: If you or @nick-mobilecoin can confirm that SGX can be part of the top-level workspace, then I can make the change. If not, I think I'll leave the workspace because I don't want each `sgx` sub-package to hardcode the MSRV.

samdealy: For the sake of making this PR have one concern (update MSRV) I'll remove the the workspace delcaration.

fog/view/enclave/trusted/Cargo.toml
```diff
@@ -6,9 +6,13 @@ description = "The MobileCoin Fog user-facing server's enclave entry point."
 edition = "2021"
 license = "GPL-3.0"
 readme = "README.md"
+rust-version = { workspace = true }
 
 [workspace]
 
+[workspace.package]
+rust-version = "1.68.0"
```
jcape, line 14:
```suggestion
rust-version = "1.68.0"

[workspace]
```

jcape (approved):
Generally looks good, the main issues are:
1. Enclaves can just use the version directly (e.g. `mc-foo-enclave-trusted`), the "workspace" entry there is there to prevent cargo invocations in that directory from thinking they're part of the top-level workspace.
2. There are some places where there are multi-paragraph `description` fields, and the version was inserted into the middle of it (these places are noted, but GH wouldn't let me give suggestions for those things).

Approved with the understanding that issues will get fixed :-)

fog/ingest/enclave/trusted/Cargo.toml
```diff
@@ -4,10 +4,14 @@ version = "4.0.2"
...
 edition = "2021"
 license = "GPL-3.0"
+rust-version = { workspace = true }
 
 # Declare as an empty workspace to not confuse cargo.
 [workspace]
 
+[workspace.package]
+rust-version = "1.68.0"
+
```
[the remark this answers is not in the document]
samdealy: Done.

fog/view/enclave/trusted/Cargo.toml
samdealy: Done.

consensus/enclave/trusted/Cargo.toml
```diff
@@ -6,9 +6,13 @@ edition = "2021"
 description = "The MobileCoin Consensus Service's internal enclave entry point."
 license = "GPL-3.0"
 readme = "README.md"
+rust-version = { workspace = true }
 
 [workspace]
 
+[workspace.package]
+rust-version = "1.68.0"
+
```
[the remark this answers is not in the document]
samdealy: Done.

attest/core/Cargo.toml
```diff
@@ -5,6 +5,7 @@ authors = ["MobileCoin"]
 description = '''
 This crate contains necessary functions and utilities to perform remote
 attestation and verification of an SGX enclave.
+rust-version = { workspace = true }
```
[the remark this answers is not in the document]
samdealy: Done.

consensus/enclave/impl/Cargo.toml
```diff
@@ -4,6 +4,7 @@ version = "4.0.2"
 authors = ["MobileCoin"]
 description = '''
 This crate contains the actual implementation of a mobilenode enclave.
+rust-version = { workspace = true }
```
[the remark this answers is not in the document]
samdealy: DOne.

crypto/rand/Cargo.toml
```diff
@@ -4,6 +4,7 @@ version = "4.0.2"
 authors = ["MobileCoin"]
 description = '''
 This crate provides a no-std compatible rng called `McRng`.
+rust-version = { workspace = true }
```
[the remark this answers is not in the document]
samdealy: Done.

sgx/Cargo.toml
samdealy: Ended up implementing your suggestion, see comment below.

PR #3124, opened 2023-02-12
This comments out the "grafana push" part of the prometheus yml template

cbeck88:
AFAIK this grafana push stuff, for the local network script, hasn't been used by anyone in a few years, and it requires you to do additional setup in grafana. If you just want to talk to prometheus web interface, you have to comment this out or you get logspam.

I feel like it makes sense to comment this out, and let people put it back if they actually want to push from local network to grafana, but lmk if you'd rather leave it in, not a big deal.

### DIFF
--- a/tools/local-network/prometheus.yml.template
+++ b/tools/local-network/prometheus.yml.template
@@ -11,11 +11,11 @@ global:
   #    monitor: 'example'
   #
 
-remote_write:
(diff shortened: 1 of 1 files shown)


NotGyro (approved):
Looks good to me - if we're not using it, it makes sense not to incur this extra step.

jcape (approved):

PR #3138, opened 2023-02-15
Porting existing Fog Ledger tests to Router

awygle:
### Future Work

[Soundtrack of this PR]()


fog/ledger/connection/src/router_client.rs
```diff
@@ -0,0 +1,265 @@
...
+use mc_fog_uri::FogLedgerUri;
+use mc_transaction_core::ring_signature::KeyImage;
+use mc_util_grpc::ConnectionUriGrpcioChannel;
+use mc_util_serial::DecodeError;
+use mc_util_uri::{ConnectionUri, UriConversionError};
+use sha2::Sha512;
+use std::sync::Arc;
+
+/// A high-level object mediating requests to the fog ledger router service
+pub struct LedgerGrpcClient {
```
awygle, line 28: This is essentially a copy of the view router client.

fog/ledger/enclave/impl/src/lib.rs
```diff
@@ -354,7 +354,7 @@ where
         &self,
         auth_request: NonceAuthRequest,
     ) -> Result<(NonceAuthResponse, NonceSession)> {
-        self.ake.frontend_accept(auth_request).map_err(|e| e.into())
+        Ok(self.ake.frontend_accept(auth_request)?)
```
awygle, line 357: I don't think this actually changes anything but it matches the view version more closely.

fog/ledger/server/src/key_image_service.rs
```diff
@@ -73,7 +73,36 @@ impl<L: Ledger + Clone, E: LedgerEnclaveProxy> KeyImageService<L, E> {
...
     }
 
-    pub fn auth_impl(
+    pub fn auth_store(
+        &mut self,
+        mut req: AuthMessage,
+        logger: &Logger,
+    ) -> Result<attest::AuthMessage, RpcStatus> {
+        // TODO: Use the prost message directly, once available
+        match self.enclave.frontend_accept(req.take_data().into()) {
```
awygle, line 82: The difference here is that we use `frontend_accept` for the store and `client_accept` for the old-style service. In theory we eventually delete the old-style service.

fog/ledger/server/src/router_handlers.rs
```diff
@@ -332,7 +332,7 @@ async fn authenticate_ledger_store<E: LedgerEnclaveProxy>(
     ledger_store_url: KeyImageStoreUri,
     logger: Logger,
 ) -> Result<(), RouterServerError> {
-    let ledger_store_id = ResponderId::from_str(&ledger_store_url.to_string())?;
+    let ledger_store_id = ledger_store_url.responder_id()?;
```
awygle, line 335: I'd like to discuss this a bit more. The difference in terms of result between these two lines is that the old one includes the protocol specifier, `insecure-key-image-store://`, and the second one doesn't. To me, it seems like either of these should work. However, what we actually got is that the Noise AEAD decryption failed on the response from the Key Image Store to the router.

fog/ledger/connection/src/router_client.rs
```diff
@@ -0,0 +1,256 @@
...
+            .map(|&key_image| KeyImageQuery {
+                key_image,
+                start_block: 0,
+            })
+            .collect();
+        let key_images_request = CheckKeyImagesRequest {
+            queries: key_images_queries,
+        };
+
+        let aad = vec![];
```
samdealy, line 154: Just to confirm- this should be empty? If so, maybe add a comment above explaining why.

fog/ledger/server/src/key_image_service.rs
samdealy: Do you plan on deleting the old-style service as well? Because isn't it true that this service won't ever be contacted by clients anymore? That is, only the router will contact it? if yes, I'd recommend making a follow up issue to get rid of this functionality to make it clear that this is the case + sync with what fog view store does.

fog/ledger/server/src/router_service.rs
```diff
@@ -80,3 +84,94 @@ where
         });
     }
 }
+
+/// Used for the implementation of FogKeyImageApi::check_key_images(),
+/// the legacy unary key-image API, for LedgerRouterService.
+async fn unary_check_key_image_inner<E>(
```
samdealy, line 90: Recommend renaming to `..._impl`. We don't usually use `inner` to denote a function that is exclusively used by another function.

fog/ledger/server/src/router_service.rs
```diff
@@ -80,3 +84,94 @@ where
...
+    .await;
+
+    match result {
+        Ok(mut response) => {
+            if response.has_check_key_image_response() {
+                sink.success(response.take_check_key_image_response()).await
+            } else {
+                let error = rpc_internal_error(
+                    "Inavlid LedgerRequest response",
+                    "No check key image response to client's key image request.".to_string(),
```
samdealy, line 117: Suggest renaming to something like "Cannot provide a check key image response to the client's key image request."

fog/ledger/server/src/router_handlers.rs
samdealy: The `fn responder_id` method grabs the query param from the URL and passes it to the lower-level `from_str` method. There a few methods that grab query params and are named similarly (e.g. `fn consensus_msg_key`). It is a bit confusing, agreed, but I think the solution might involve renaming methods to denote that we are getting the values from a query param. So instead of `fn responder_id` it would be `fn responder_id_from_param` or something less verbose but still indicating that we got it from a query param.

What do you think?

fog/ledger/server/tests/router_connection.rs
```diff
@@ -0,0 +1,910 @@
...
+    WatcherDB::create(db_tmp.path()).unwrap();
+    let watcher = WatcherDB::open_rw(db_tmp.path(), &[url], logger).unwrap();
+    let watcher_dir = db_tmp.path().to_path_buf();
+    (watcher, watcher_dir)
+}
+
+// Test that a fog ledger connection is able to get valid merkle proofs by
+// hitting a fog ledger server
+#[test_with_logger]
+fn fog_ledger_merkle_proofs_test(logger: Logger) {
```
samdealy, line 75: Can we remove the configuration or move some of the set up logic to helper functions / a helper module? Putting all this configuration in the test makes it harder to read and leads to  duplication in tests below.

fog/ledger/server/tests/router_connection.rs
```diff
@@ -0,0 +1,910 @@
...
+        // objects have been destroyed, and hope that those 6 threads see the
+        // shutdown requests within 1 second.
+        sleep(Duration::from_millis(1000));
+    }
+}
+
+// Test that a fog ledger connection is able to check key images by hitting
+// a fog ledger server
+#[test_with_logger]
+fn fog_ledger_key_images_test(logger: Logger) {
```
samdealy, line 285: See above.

fog/ledger/server/tests/router_connection.rs
```diff
@@ -0,0 +1,910 @@
...
+        // objects have been destroyed, and hope that those 6 threads see the
+        // shutdown requests within 1 second.
+        sleep(Duration::from_millis(1000));
+    }
+}
+
+// Test that a fog ledger connection is able to check key images by hitting
+// a fog ledger server
+#[test_with_logger]
+fn fog_ledger_blocks_api_test(logger: Logger) {
```
samdealy, line 555: See above.

fog/ledger/server/tests/router_connection.rs
```diff
@@ -0,0 +1,910 @@
...
+    // in the meantime we can just sleep after grpcio env and all related
+    // objects have been destroyed, and hope that those 6 threads see the
+    // shutdown requests within 1 second.
+    sleep(Duration::from_millis(1000));
+}
+
+// Test that a fog ledger connection is able to check key images by hitting
+// a fog ledger server
+#[test_with_logger]
+fn fog_ledger_untrusted_tx_out_api_test(logger: Logger) {
```
samdealy, line 718: See above.

samdealy (changes requested):

fog/ledger/server/tests/router_connection.rs
```diff
@@ -0,0 +1,908 @@
...
+        // we opened a PR here: https://github.com/tikv/grpc-rs/pull/455
+        // in the meantime we can just sleep after grpcio env and all related
+        // objects have been destroyed, and hope that those 6 threads see the
+        // shutdown requests within 1 second.
+        sleep(Duration::from_millis(1000));
+    }
+}
+
+// Test that a fog ledger connection is able to check key images by hitting
+// a fog ledger server
```
nick-mobilecoin, line 551: :question: : this is the same comment as the previous test, how is this different?

fog/ledger/server/tests/router_connection.rs
```diff
@@ -0,0 +1,908 @@
...
+#[test_with_logger]
+fn fog_ledger_blocks_api_test(logger: Logger) {
+    let mut rng = RngType::from_seed([0u8; 32]);
+
+    let alice = AccountKey::random_with_fog(&mut rng);
+    let bob = AccountKey::random_with_fog(&mut rng);
+    let charlie = AccountKey::random_with_fog(&mut rng);
+
+    let recipients = vec![alice.default_subaddress()];
+
```
nick-mobilecoin, line 561: :wrench: Suggest getting rid of this, since the recipients change per block it would probably be more consistent to use an explicit slice in the last call to `add_block_to_ledger()`
```suggestion
```

fog/ledger/server/tests/router_connection.rs
```diff
@@ -0,0 +1,908 @@
...
+        let queries = [0..1];
+        let result = client.get_blocks(&queries).unwrap();
+        // Check that we got 1 block, as expected
+        assert_eq!(result.blocks.len(), 1);
+        assert_eq!(result.blocks[0].index, 0);
+        assert_eq!(result.blocks[0].outputs.len(), 1);
+        assert_eq!(result.blocks[0].global_txo_count, 1);
+        assert_eq!(
+            result.blocks[0].timestamp_result_code,
+            TimestampResultCode::BlockIndexOutOfBounds as u32
```
nick-mobilecoin, line 676: :question: why is this index out of bounds?

fog/ledger/server/tests/router_connection.rs
```diff
@@ -0,0 +1,908 @@
...
+    // grpcio detaches all its threads and does not join them :(
+    // we opened a PR here: https://github.com/tikv/grpc-rs/pull/455
+    // in the meantime we can just sleep after grpcio env and all related
+    // objects have been destroyed, and hope that those 6 threads see the
+    // shutdown requests within 1 second.
+    sleep(Duration::from_millis(1000));
+}
+
+// Test that a fog ledger connection is able to check key images by hitting
+// a fog ledger server
```
nick-mobilecoin, line 714: :x: Similar to above this doesn't seem to be correct for this function

fog/ledger/server/tests/router_connection.rs
```diff
@@ -0,0 +1,908 @@
...
+#[test_with_logger]
+fn fog_ledger_untrusted_tx_out_api_test(logger: Logger) {
+    let mut rng = RngType::from_seed([0u8; 32]);
+
+    let alice = AccountKey::random_with_fog(&mut rng);
+    let bob = AccountKey::random_with_fog(&mut rng);
+    let charlie = AccountKey::random_with_fog(&mut rng);
+
+    let recipients = vec![alice.default_subaddress()];
+
```
nick-mobilecoin, line 724: :wrench: similar to above consider removing and moving explicitly into the `add_block_to_ledger()`
```suggestion
```

fog/ledger/server/tests/router_connection.rs
```diff
@@ -0,0 +1,908 @@
...
+        assert_eq!(
+            &result.results[1].tx_out_pubkey.clone().unwrap().data[..],
+            &real_tx_out0.public_key.as_bytes()[..]
+        );
+        assert_eq!(result.results[1].result_code, TxOutResultCode::Found);
+        assert_eq!(result.results[1].tx_out_global_index, 0);
+        assert_eq!(result.results[1].block_index, 0);
+        assert_eq!(
+            result.results[1].timestamp_result_code,
+            TimestampResultCode::BlockIndexOutOfBounds as u32
```
nick-mobilecoin, line 854: :question: This may get answered in the previous test, but why is this out of bounds?

fog/ledger/server/tests/router_connection.rs
```diff
@@ -0,0 +1,908 @@
...
+    // grpcio detaches all its threads and does not join them :(
+    // we opened a PR here: https://github.com/tikv/grpc-rs/pull/455
+    // in the meantime we can just sleep after grpcio env and all related
+    // objects have been destroyed, and hope that those 6 threads see the
+    // shutdown requests within 1 second.
+    sleep(Duration::from_millis(1000));
+}
+
+// Infra
+
```
nick-mobilecoin, line 867: :wrench: This feels like a partial thought
```suggestion
```

fog/ledger/server/tests/router_connection.rs
```diff
@@ -0,0 +1,908 @@
...
+/// Adds a block containing one txo for each provided recipient and returns new
+/// block height.
+///
+/// # Arguments
+/// * `ledger_db`
+/// * `recipients` - Recipients of outputs.
+/// * `rng`
```
nick-mobilecoin, line 874: :wrench: Should either fill out all the arguments or remove, having 3 of the 6 arguments listed here with only one documented seems odd.

nick-mobilecoin (changes requested):
Main blockers on questions or needed clarity on test behavior

fog/ledger/connection/src/router_client.rs
awygle: Yes, this should be empty (there's no AAD associated with the ledger query, at least not currently). I'll add a comment.

fog/ledger/server/src/router_service.rs
awygle: Sure. (@NotGyro , for future reference).

fog/ledger/server/src/router_service.rs
awygle: Sounds good.

fog/ledger/server/tests/router_connection.rs
awygle: I didn't write these tests originally, they're just copies of `connection.rs`. You're absolutely right about the structure of the tests not being great. I'll file an issue for fixing it but I don't want to block this PR on it.

fog/ledger/server/tests/router_connection.rs
```diff
@@ -0,0 +1,910 @@
...
+                pick_unused_port().expect("No free ports")
+            ))
+            .unwrap();
+            let config = LedgerRouterConfig {
+                chain_id: "local".to_string(),
+                ledger_db: db_full_path.to_path_buf(),
+                watcher_db: watcher_dir,
+                admin_listen_uri: admin_listen_uri.clone(),
+                client_listen_uri: client_listen_uri.clone(),
+                client_responder_id: ResponderId::from_str(&client_listen_uri.addr()).unwrap(),
```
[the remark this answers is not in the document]
awygle: The update wasn't actually needed - it was when trying to fix tests which were failing due to what turned out to be port conflicts. That being said, I'll change them all to use `.responder_id()` as it seems to be more correct.

fog/ledger/server/src/key_image_service.rs
awygle: Yes, we do intend to delete the old-style service. I'll file the issue.

fog/ledger/server/src/key_image_service.rs
```diff
@@ -73,7 +73,36 @@ impl<L: Ledger + Clone, E: LedgerEnclaveProxy> KeyImageService<L, E> {
...
+            }
+            Err(client_error) => {
+                // There's no requirement on the remote party to trigger this, so it's debug.
+                log::debug!(
+                    logger,
+                    "KeyImageStoreApi::frontend_accept failed: {}",
+                    client_error
+                );
+                let rpc_permissions_error = rpc_permissions_error(
+                    "client_auth",
```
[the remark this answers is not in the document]
awygle: Probably.

fog/ledger/server/src/router_handlers.rs
awygle: Honestly, the issue here is "what is a responder ID" in general, IMO. The name is not very explanatory and since it's only constructed from a string there's not a lot of validation on whether a given ID is acceptable or not.

But James assures me that responder IDs will go away when we move to DCAP, so I'm tempted to just not worry about it.

NotGyro:
* master
  * **PR #3138** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3138" target="_blank">Graphite</a>  👈
    * **PR #3163** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3163" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3138?utm_source=stack-comment).

jcape (changes requested):
This needs to get rebased back onto `feature/fog-ledger-router`

fog/ledger/server/tests/router_connection.rs
```diff
@@ -0,0 +1,910 @@
...
+    time::Duration,
+};
+use tempdir::TempDir;
+use url::Url;
+
+const TEST_URL: &str = "http://www.my_url1.com";
+
+const OMAP_CAPACITY: u64 = 128 * 128;
+
+const GRPC_RETRY_CONFIG: GrpcRetryConfig = GrpcRetryConfig {
```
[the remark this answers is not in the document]
awygle: Constants must be explicitly typed.

fog/ledger/server/tests/router_connection.rs
awygle: #3168

awygle:
OK, all updates have been made and issues filed for followup work.

samdealy (approved):

nick-mobilecoin (approved):

PR #3149, opened 2023-02-18
adjust readiness criteria for fog view

cbeck88:
Jason reported that fog view is now declaring itself ready well before the highest-processed-block-count is close to the block height.

In the past, the main bottleneck was loading the data from Postgres, but now it's actually putting the data into ORAM.

In this commit, we refine the readiness criteria, so that both the db fetcher and the enclave loading thread have to run out of work before the server declares itself ready.

### DIFF
--- a/fog/view/server/src/server.rs
+++ b/fog/view/server/src/server.rs
@@ -391,6 +391,13 @@ where
     /// Keeps track of which blocks we have fed into the enclave.
     enclave_block_tracker: BlockTracker,
 
+    /// Flag which the db fetcher sets to indicate when it has exhausted it's
+    /// initial work.
+    db_fetcher_readiness_indicator: ReadinessIndicator,
+
+    /// Flag which we set to indicate when the server as a whole is read.
+    server_readiness_indicator: ReadinessIndicator,
+
     /// Keeps track how long ago it since we made progress, (or complained about
     /// not making progress) When this gets too distant in the past, we log
     /// a warning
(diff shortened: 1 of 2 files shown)


fog/view/server/src/server.rs
```diff
@@ -391,6 +391,13 @@ where
     /// Keeps track of which blocks we have fed into the enclave.
     enclave_block_tracker: BlockTracker,
 
+    /// Flag which the db fetcher sets to indicate when it has exhausted it's
+    /// initial work.
+    db_fetcher_readiness_indicator: ReadinessIndicator,
+
+    /// Flag which we set to indicate when the server as a whole is read.
```
NotGyro, line 398: Nit: I am assuming "ready" was meant here.
```suggestion
    /// Flag which we set to indicate when the server as a whole is ready.
```

NotGyro (approved):
Looks good to me!

samdealy:
What caused this shift:  "main bottleneck was loading the data from Postgres, but now it's actually putting the data into ORAM."?

cbeck88:
@samdealy i think before, the latency between fog view and postgres was very high, and fog view would only request one block at a time. we made a patch that made it request many more (configurable) blocks at a time and that increased the throughput with the postgres connection considerably. since then i think oram is the bottleneck and not postgres

cbeck88:
planning to retarget to v4.1

samdealy (approved):

PR #3151, opened 2023-02-19
fog-view: Limit number of user events, and tell client this happened

cbeck88:
Jason reported recently that fog test client failed in test-net with a grpc maximum response size error. Eventually it recovered and I'm not sure we fully determined the root cause.

One defect that we noticed during investigation is that, while the client limits the number of ETxOutRecord's it requests, nothing limits the number of user events that fog view returns -- if the client starts from scratch, it always returns all the events.

This patch is meant to fix that. The SQL query is changed so that there is a limit to how many user events it returns, and it is ordered ascending by the user event id. This limit is configurable as a server startup parameter. When the limit is reached, we flag to the client that there may be more user events.

Then we update the `fog-view-protocol` polling module so that when this flag is set, it keeps asking for more seeds until this flag isn't set any more.

I believe that this is a backwards compatible change, because old clients will ignore this flag, and for new clients talking to old servers, this flag will have a default value of false, which matches the status quo behavior of always returning all the results.

### DIFF
--- a/fog/api/proto/view.proto
+++ b/fog/api/proto/view.proto
@@ -145,6 +145,11 @@ message QueryResponse {
     /// This can be used by the client as a hint when choosing cryptonote mixin indices.
     /// This field doesn't have the same "cursor" semantics as the other fields.
     uint64 last_known_block_cumulative_txo_count = 9;
+
+    /// If true, this means that due limits, we could not return all the requested
+    /// user events in one response. Clients cannot compute an accurate balance check
+    /// until they have received all relevant user events.
+    bool may_have_more_user_events = 10;
 }
 
 /// A record of an Rng created by a fog ingest enclave.
--- a/fog/test_infra/src/mock_client.rs
+++ b/fog/test_infra/src/mock_client.rs
@@ -31,8 +31,11 @@ impl<R: RecoveryDb> FogViewConnection for PassThroughViewClient<R> {
         start_from_block_index: u64,
         search_keys: Vec<Vec<u8>>,
     ) -> Result<QueryResponse, Self::Error> {
-        let (user_events, next_start_from_user_event_id) =
-            self.db.search_user_events(start_from_user_event_id)?;
+        const USER_EVENT_LIMIT: usize = 10_000;
+        let (user_events, next_start_from_user_event_id) = self
+            .db
+            .search_user_events(start_from_user_event_id, USER_EVENT_LIMIT)?;
+        let may_have_more_user_events = user_events.len() >= USER_EVENT_LIMIT;
 
         let highest_known_block_count = self
             .db
@@ -75,6 +78,7 @@ impl<R: RecoveryDb> FogViewConnection for PassThroughViewClient<R> {
             tx_out_search_results: Default::default(),
             last_known_block_count: highest_known_block_count,
             last_known_block_cumulative_txo_count: cumulative_txo_count,
+            may_have_more_user_events,
         };
 
         resp.tx_out_search_results = self.db.get_tx_outs(start_from_block_index, &search_keys)?;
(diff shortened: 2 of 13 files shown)


cbeck88:
I still need to add some new test coverage on this feature

fog/api/proto/view.proto
```diff
@@ -145,6 +145,11 @@ message QueryResponse {
     /// This can be used by the client as a hint when choosing cryptonote mixin indices.
     /// This field doesn't have the same "cursor" semantics as the other fields.
     uint64 last_known_block_cumulative_txo_count = 9;
+
+    /// If true, this means that due limits, we could not return all the requested
```
samdealy, line 149:
```suggestion
    /// If true, this means that due to limits, we could not return all the requested
```

fog/api/proto/view.proto
```diff
@@ -145,6 +145,11 @@ message QueryResponse {
...
+    /// user events in one response. Clients cannot compute an accurate balance check
```
samdealy, line 150: In the PR description you write: "I believe that this is a backwards compatible change, because old clients will ignore this flag".

But if it's true that "clients cannot compute an accurate balance check until they have received all relevant user events," then how can an (old) client that ignores this flag accurately compute a balance check?

fog/types/src/view.rs
```diff
@@ -93,6 +93,13 @@ pub struct QueryResponse {
     /// clients sample for mixins.
     #[prost(uint64, tag = "9")]
     pub last_known_block_cumulative_txo_count: u64,
+
+    /// If true, this means that due limits, we could not return all the
```
samdealy, line 97:
```suggestion
    /// If true, this means that due to limits, we could not return all the
```

fog/view/enclave/api/src/lib.rs
```diff
@@ -49,6 +49,9 @@ pub struct UntrustedQueryResponse {
 
     /// The cumulative txo count of the last known block.
     pub last_known_block_cumulative_txo_count: u64,
+
+    /// If we may have more user events than this.
```
samdealy, line 53:
```suggestion
    /// True if we might have more user events than the number of events returned in this response.
```

fog/test_infra/src/mock_client.rs
```diff
@@ -31,8 +31,11 @@ impl<R: RecoveryDb> FogViewConnection for PassThroughViewClient<R> {
         start_from_block_index: u64,
         search_keys: Vec<Vec<u8>>,
     ) -> Result<QueryResponse, Self::Error> {
-        let (user_events, next_start_from_user_event_id) =
-            self.db.search_user_events(start_from_user_event_id)?;
+        const USER_EVENT_LIMIT: usize = 10_000;
+        let (user_events, next_start_from_user_event_id) = self
+            .db
+            .search_user_events(start_from_user_event_id, USER_EVENT_LIMIT)?;
+        let may_have_more_user_events = user_events.len() >= USER_EVENT_LIMIT;
```
samdealy, line 38: It appears that we're returning all the `user_events` in the `query_response` below. If there are `12_000` `user_events`, then we return them all to the client AND we set `may_have_more_user_events` to true. 

However, the logic for `fn search_user_events` has this line: `    .limit(max_num_events as i64)`, and `max_num_events` == `USER_EVENT_LIMIT`. How then can `user_events.len()` ever be > than `USER_EVENT_LIMIT`?

fog/api/proto/view.proto
cbeck88: Good question:

* Today, the amount of user events in postgres is not so much that we can't return all the events, and so all the existing clients are fine, and will be fine with servers before and after this change.
* After this change, we can ship clients that will be okay even when that is no longer the case, and the servers will have to do some sort of pagination.
* The backwards compatible part is, new clients that talk to old servers, will get all the events, and the flag will default to false when they deserialize, so they will correctly decide that they got all the events.

cbeck88:
I still didn't write tests, I'm gonna work on that more soon, sorry

duncp856-a11y (approved):

Issue #2977, opened 2023-01-03
Add session tracking to consensus


jcape:

NotGyro:
Created PR https://github.com/mobilecoinfoundation/mobilecoin/pull/3155 as a first step on this issue.

jcape:
Completed

PR #3155, opened 2023-02-22
Add tracked_sessions field to consensus service

NotGyro:
Adds a `tracked_sessions` field to consensus service and a `_tracked_sessions` field to the client API service in consensus, as well as a `ClientSessionTracking` struct, to fill with tracked session information, which is a stub at present.

### Motivation

Some way of keeping track of how many failed tx proposals have been sent recently needs to be available, so that bad connections can be dropped, to prevent attacks and harmful behavior from buggy clients. See: https://github.com/mobilecoinfoundation/mobilecoin/issues/2977

### Future Work
The rest of https://github.com/mobilecoinfoundation/mobilecoin/issues/2977, as in: 
* Populate the client session tracking structure with metadata about incoming sessions 
* Keep track of how many recent failed tx proposals have been made by each client
* Drop clients which have made too many failed tx proposals in too short of a span of time.


consensus/service/src/api/client_api_service.rs
```diff
@@ -45,6 +49,9 @@ pub struct ClientApiService {
     is_serving_fn: Arc<(dyn Fn() -> bool + Sync + Send)>,
     authenticator: Arc<dyn Authenticator + Send + Sync>,
     logger: Logger,
+    /// Information kept regarding sessions between clients and consensus
+    /// so that we can drop bad sessions.
+    _tracked_sessions: Arc<Mutex<LruCache<ResponderId, ClientSessionTracking>>>,
```
jcape, line 54: This should be a `ClientSession`. `ResponderId` is used to denote a server's externally-visible `fqdn:port`, so a given consensus node will only have one value (set either via `--listen-uri` or `--client-responder-id`).

jcape (changes requested):

consensus/service/src/api/client_api_service.rs
NotGyro: Ahh, thanks for clarifying!

jcape:
There are still clippy lints that need to be fixed.

NotGyro:
* master
  * **PR #3155** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3155" target="_blank">Graphite</a>  👈
    * **PR #3156** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3156" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3155?utm_source=stack-comment).

jcape (approved):

consensus/service/config/src/lib.rs
```diff
@@ -125,6 +125,11 @@ pub struct Config {
     /// The configured block version
     #[clap(long, default_value = "0", value_parser = parse_block_version, env = "MC_BLOCK_VERSION")]
     pub block_version: BlockVersion,
+
+    /// Maximum number of client session tracking structures to retain in
+    /// a least-recently-used cache.
+    #[clap(long, default_value = "4096", env = "MC_CLIENT_TRACKING_CAPACITY")]
```
varsha888, line 131: How was `4096` value derived?

consensus/service/src/api/client_api_service.rs
```diff
@@ -58,6 +66,7 @@ impl ClientApiService {
         is_serving_fn: Arc<(dyn Fn() -> bool + Sync + Send)>,
         authenticator: Arc<dyn Authenticator + Send + Sync>,
         logger: Logger,
+        tracked_sessions: Arc<Mutex<LruCache<ClientSession, ClientSessionTracking>>>,
```
varsha888, line 69: How will LRUCache be used ?

consensus/service/config/src/lib.rs
NotGyro: As discussed in standup - thanks for calling this out, asking about it means James was able to point me towards a similar LRU in the attest code. 
This should be resolved in https://github.com/mobilecoinfoundation/mobilecoin/pull/3155/commits/e67a060c4739e876863dbf1d781240b0e665cdb4

consensus/service/src/api/client_api_service.rs
NotGyro: The intent is to use this to keep track of how many failed transaction proposals a client has sent recently, so that it can drop connections with clients that look broken / malicious.

varsha888: got it, thank you! :)

consensus/service/config/src/lib.rs
varsha888: Thanks Millie for clarifying this :)

varsha888 (approved):

PR #3156, opened 2023-02-23
Start tracking client sessions in consensus

NotGyro:
Pushes sessions to the session LRU cache in ClientAPIService within the `client_tx_propose()`. Also contains a lookup for updating that cached tracking information later - currently it's just a stub.  

### Motivation

This will be used to track how many failed tx proposals have been sent recently, so that bad connections can be dropped, to prevent attacks and harmful behavior from buggy clients. See: https://github.com/mobilecoinfoundation/mobilecoin/issues/2977

### Future Work
The rest of https://github.com/mobilecoinfoundation/mobilecoin/issues/2977, as in: 
* Populate the client session tracking structure with metadata about incoming sessions 
* Keep track of how many recent failed tx proposals have been made by each client
* Drop clients which have made too many failed tx proposals in too short of a span of time.


consensus/service/src/api/client_api_service.rs
```diff
@@ -218,6 +221,24 @@ impl ConsensusClientApi for ClientApiService {
...
+                }
+            }
+            Err(err) => {
+                warn!(
+                    self.logger,
+                    "Unable to acquire mutex on session metadata: {err:?}"
+                );
+            }
+        }
+
```
nick-mobilecoin, line 241: :wrench: Most places with a mutex use `expect()` 
```suggestion
         {
            let mut tracker = self.tracked_sessions.lock().expect("Mutex poisoned");
            if let Some(_session_info) = tracker.get(&session) {
                // TODO: Update fields
            } else {
                // TODO: Populate new session metadata
                tracker.put(session, ClientSessionTracking {});
            }
        }
```

consensus/service/src/api/client_api_service.rs
```diff
@@ -13,7 +13,10 @@ use crate::{
 use grpcio::{RpcContext, RpcStatus, UnarySink};
 use mc_attest_api::attest::Message;
 use mc_attest_enclave_api::ClientSession;
-use mc_common::{logger::Logger, LruCache};
+use mc_common::{
+    logger::{slog::warn, Logger},
+    LruCache,
+};
```
nick-mobilecoin, line 19: :wrench: If the suggestion for the mutex below is acceptable then this will need to be reverted
```suggestion
use mc_common::{logger::Logger, LruCache};
```

nick-mobilecoin (approved):

jcape:
Will approve after the expect changes go in.

jcape (approved):
LGTM, but please commit the suggestions.

consensus/service/src/api/client_api_service.rs
```diff
@@ -218,6 +218,16 @@ impl ConsensusClientApi for ClientApiService {
...
         let _timer = SVC_COUNTERS.req(&ctx);
 
+        {
+            let session = ClientSession::from(msg.channel_id.clone());
+            let mut tracker = self.tracked_sessions.lock().expect("Mutex poisoned");
+            if let Some(_session_info) = tracker.get(&session) {
+                // TODO: Update fields
+            } else {
+                // TODO: Populate new session metadata
+                tracker.put(session, ClientSessionTracking {});
```
jcape, line 228: For future, this should use the `Default` impl, ala `ClientSessionTracking::default()`.

jcape (approved):
LGTM

consensus/service/src/api/client_api_service.rs
NotGyro: Most likely I'll be initializing it with a "last touched" timestamp or something like that in here, so default() would not be ideal but also it would not be an empty initializer like this. The `ClientSessionTracking` struct is a stub deliberately for now.

Is `ClientSessionTracking{foo, bar}` preferred or should I go for more of a `ClientSessionTracking::new(foo, bar)` style?

jcape: I generally prefer the constructor method.

Issue #2833, opened 2022-11-07
Moving APIs from ledger store to router


jcape:
- [ ] #3123 
- [ ] #3163 
- [ ] #3217

PR #3163, opened 2023-02-23
Test key image retrieval via unary API on the router

NotGyro:
Adds the new test `fog_router_unary_key_image_test()`, an equivalent to `fog_ledger_key_images_test()` from `connection.rs` and the single-server implementation of the fog ledger, to `router_connection.rs`. This tests the path for unary requests to the fog ledger router.

### Motivation

In order to consider issue https://github.com/mobilecoinfoundation/mobilecoin/issues/2833 resolved, we need to be sure that the router/store code for the fog ledger implements all of the APIs that the single-server implementation did. This method tests to ensure that the unary key-image API is implemented for the new system.

### Future Work
This test follows the structure of similar tests in `router_connection.rs` and `connection.rs`. Unfortunately, that means a couple of parts of the test are time-dependent and use `sleep()` to ensure proper test behavior. This can contribute to test flakiness and ideally at some point in the future this might no longer be needed.


NotGyro:
* master
  * **PR #3138** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3138" target="_blank">Graphite</a> 
    * **PR #3163** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3163" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3163?utm_source=stack-comment).

fog/ledger/server/tests/router_connection.rs
```diff
@@ -863,6 +863,281 @@ fn fog_ledger_untrusted_tx_out_api_test(logger: Logger) {
     sleep(Duration::from_millis(1000));
 }
 
+// Test that a fog ledger connection is able to check key images by hitting
+// a fog ledger router using the unary API
+#[test_with_logger]
+fn fog_router_unary_key_image_test(logger: Logger) {
+    let mut rng = RngType::from_seed([0u8; 32]);
+
+    for block_version in BlockVersion::iterator() {
```
samdealy, line 872: Can we use the `yare` crate to parametrize this test over `BlockVersion` rather than do a loop? That way, it's more easily configurable and we when one of the tests fails, there will be a more descriptive test name that incorporates the block version. 

See https://github.com/mobilecoinfoundation/mobilecoin/blob/feature/fog-view-router/fog/view/server/tests/streaming_smoke_tests.rs#L14

fog/ledger/server/tests/router_connection.rs
```diff
@@ -863,6 +863,281 @@ fn fog_ledger_untrusted_tx_out_api_test(logger: Logger) {
...
+            .add_block_signature(&url1, 1, signed_block_a1, filename.clone())
+            .unwrap();
+
+        // Update last synced to block 2, to indicate that this URL did not participate
+        // in consensus for block 2.
+        watcher.update_last_synced(&url1, 2).unwrap();
+
+        {
+            // Make Key Image Store
+            let store_uri = KeyImageStoreUri::from_str(&format!(
```
samdealy, line 940: It appears that this test is only involving one `KeyImageStore` acting as a shard. We should test this scenario but we also need to test the scenario in which there are multiple shards. Are there plans to do so in a follow up PR?

fog/ledger/server/tests/router_connection.rs
```diff
@@ -863,6 +863,281 @@ fn fog_ledger_untrusted_tx_out_api_test(logger: Logger) {
     sleep(Duration::from_millis(1000));
 }
 
+// Test that a fog ledger connection is able to check key images by hitting
+// a fog ledger router using the unary API
+#[test_with_logger]
+fn fog_router_unary_key_image_test(logger: Logger) {
```
samdealy, line 869: We'll need to remove the test set up logic from this test because it makes it hard to read. Can you make an issue to track this work?

samdealy (changes requested):

fog/ledger/server/tests/router_connection.rs
awygle: I think this falls under #3168

fog/ledger/server/tests/router_connection.rs
awygle: Also #3168

fog/ledger/server/tests/router_connection.rs
NotGyro: Yes - I suspect this would fit better in a separate PR and is out-of-scope for this one, but it's very necessary to test that behavior because only having one store is so different from what the router would actually be doing in production. Is it safe to say this is out of scope for this PR, you think? Should I make an issue for this?

NotGyro: Per discussion this looks safe to bring forward into a separate PR and move out of this one.

awygle (approved):
LGTM, this is what I had in mind. This test will also get reworked as part of #3168 .

nick-mobilecoin (approved):
approving knowing that #3168 exists.

I don't think we should be adding many more tests following this pattern.
This added 250 lines of code for a meaningful difference of
```diff
-             let mut client =
-                 LedgerGrpcClient::new(client_listen_uri, verifier, grpc_env, logger.clone());
+             let mut client = FogKeyImageGrpcClient::new(
+                 String::default(),
+                 router_client_listen_uri,
+                 GRPC_RETRY_CONFIG,
+                 verifier,
+                 grpc_env,
+                 logger.clone(),
+             );
```

PR #3171, opened 2023-02-24
Update FVR readiness check to incorporate store readiness statuses

samdealy:
### Motivation
Makes FVR's gRPC readiness depend on the readiness of it's shards. If one store from each shard is ready then FVR is also ready. 

We decided to implement this because the router should only start serving client requests when all of its shards are ready, which, in the case of fog view, means that the shard has loaded enough TxOuts int oORAM, making it possible to give a client a complete `QueryResponse`.

### Future Work
* Consider implementing the callback as a custom `trait` rather than as an `Fn` object.


samdealy:
* sam/fix-readiness
  * **PR #3171** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3171" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3171?utm_source=stack-comment).

util/grpc/src/health_service.rs
```diff
@@ -146,3 +147,24 @@ impl From<ReadinessIndicator> for ServiceHealthCheckCallback {
...
+/// Returns a `ServiceHealthCheckCallback` that is used in the gRPC
+/// readiness check for routers.
+pub fn get_router_callback(
+    shard_health_clients: Vec<health_api_grpc::HealthClient>,
+) -> ServiceHealthCheckCallback {
+    Arc::new(move |_| -> HealthCheckStatus {
+        let is_ready = shard_health_clients.iter().all(|client| {
+            let mut request = HealthCheckRequest::new();
+            request.set_service("fog-view".to_string());
+            client.check(&request).map_or(false, |response| {
```
samdealy, line 160: @aweis89 Can you describe how the shard (i.e. the Kubernetes load balancer) handles this request? Will it `SERVING` if at least one of its stores returns a `SERVING` response? And does the shard handle retries (if not, I might consider putting some retry logic here).

util/grpc/src/health_service.rs
```diff
@@ -146,3 +147,24 @@ impl From<ReadinessIndicator> for ServiceHealthCheckCallback {
         })
     }
 }
+
+/// Returns a `ServiceHealthCheckCallback` that is used in the gRPC
+/// readiness check for routers.
+pub fn get_router_callback(
```
awygle, line 153: I don't really like having this in `util/grpc` rather than something under `fog`... but then again we might want it to be used for non-fog routers someday, based on @jcape 's general plans, so I guess it's fine.

awygle (approved):
This looks good as a first implementation - someday we should consider having the router only go down if it can't cover all possible queries with its available shards (i.e. a `[0, 3)`, `[1, 4)`, and `[2, MAX)` set of shards, if the `[1, 4)` shard goes down every block is still covered), but that is very much future work.

aweis89:
I'm questioning whether this is actually needed. When all replicas for a given shard fail the health check, the LB for that shard won't allow any connections to those downstream shards. And so I wonder if the router really needs to be taken offline rather than just returning an error to the client when it can't reach a backend for a given shard. Keeping the router online would have the advantage of allowing the router to control the error message and response format being sent to the end client

util/grpc/src/health_service.rs
aweis89: The default LB is fairly simple and will just direct traffic only to healthy backends. So if at least one shard is healthy, it should be sending traffic to that shard and getting a `SERVING` response. Notably, for a brief period the health check could be failing before the LB picks up on it so it's possible to get a `NOT_SERVING` response until that happens. E.x. the current config for that is using the following params for backend health checks, implying that within a ~10 second period, it could be failing before removed from the LB:
` yaml
periodSeconds: 10
 successThreshold: 1
 timeoutSeconds: 1
`

Regarding retries, the default k8s LB has no logic for retries. So I'd say that's a good idea to implement client side

aweis89:
> This looks good as a first implementation - someday we should consider having the router only go down if it can't cover all possible queries with its available shards (i.e. a `[0, 3)`, `[1, 4)`, and `[2, MAX)` set of shards, if the `[1, 4)` shard goes down every block is still covered), but that is very much future work.

Interesting had to implement something [similar](https://github.com/mobilecoinofficial/fog-operator/blob/10fbc4b8a149e8232f4dc893cf7c4f50a5fab313/api/v1/fogreplicastore_types.go#L103) in the operator as well

util/grpc/src/health_service.rs
aweis89: > In terms of `SERVING` vs. `NOT_SERVING`, we're asking if the load balancer itself has an "up/down" state that we can query, and you're saying "no, but we can pass this through to a store, which is equivalent". Is that correct?

Yes that's correct. If the LB's theoretical own health check would report health if any replicas from the shard are up, having it proxy to any healthy backends would have the same behavior I think.

> The caveat is that the shard may trigger false positives if a given store has died but the LB/shard doesn't know about it, which leads to the question "should we retry?"
> 
> I'm kind of ambivalent, but I'm leaning towards "no more than we're willing to retry client queries, and possibly not at all."

I think the issue of false positives is generally a valid concern. But also in general I think it's a best for resiliency to design for common network failures and these types of intermittent errors. Health checks will always have gaps as you mentioned, and they don't either cover every intermitten failure scenario (e.x. network connectivity).

In general health checks I think should be viewed more as a self-healing mechanism than a replacement for retries. This is more of my general bias though and not really specific to this feature. Istio, the most popular service mesh, by default does retries up to twice I believe.

That said maybe this isn't the place to be doing it and a generic shared client or (better yet imo) a service mesh should be the one responsible for resiliency patterns like retries (circuit breaking, anomouly detection..)

util/grpc/src/health_service.rs
```diff
@@ -155,6 +155,7 @@ pub fn get_router_callback(
     logger: Logger
 ) -> ServiceHealthCheckCallback {
     Arc::new(move |_| -> HealthCheckStatus {
+        log::info!("Health check callback executing...");
```
jcape: Suggest debug

util/grpc/src/health_service.rs
jcape: I'm saying that in this particular case, where the response the router gives to k8s depends on the result of a different k8s check, the router's health check should not attempt to do anything to "work around" the behavior of the "inner" k8s check.

If it does, then the router is attempting to intercept and modify the behavior of one k8s check (which ops controls) in order to respond to another k8s check (which ops controls).

jcape:
> This looks good as a first implementation - someday we should consider having the router only go down if it can't cover all possible queries with its available shards (i.e. a `[0, 3)`, `[1, 4)`, and `[2, MAX)` set of shards, if the `[1, 4)` shard goes down every block is still covered), but that is very much future work.

Unfortunately, "all possible queries" starts at whatever block fog was started at. The cursoring feature exists to make that decision on a per-phone basis, but the the router has no way to know what the union of all phone's cursor would be---and, of course, when restoring from backup (PIN/mnemonic/futuretech), your phone's cursor is always going to be at `fog_zero`.

jcape:
> I'm questioning whether this is actually needed. When all replicas for a given shard fail the health check, the LB for that shard won't allow any connections to those downstream shards. And so I wonder if the router really needs to be taken offline rather than just returning an error to the client when it can't reach a backend for a given shard. Keeping the router online would have the advantage of allowing the router to control the error message and response format being sent to the end client. It also simplifies the logic for the router's health check (which we're having some trouble getting to work using this implementation)

All of this is done to maintain the behavior for mobile clients that we current exhibit for _Fog View Legacy_ nodes. The promise made is that clients are able to assume when they call `Query` they will get a consistent, up-to-date view that is ahead of wherever the phone is, particularly during restarts.

If we want to remove this proxied health check behavior, should define a new `SparseQuery` API on the router which has the behavior (and requisite list-of-ranges), and deprecate the old API---i.e. at best, we could add that API to 6.0, and remove the existing `Query` API in 7.0. I think it's worth doing, because it's not like phones are compute bound on view retrieval.

But that's also a hell of a lot of churn to simplify a health check.

fog/view/server/src/fog_view_router_server.rs
```diff
@@ -108,11 +107,25 @@ where
...
+                    .connect_to_uri(&shard.uri, &logger);
+                HealthClient::new(channel)
+            })
+            .collect::<Vec<_>>();
+
+        let health_service = mc_util_grpc::HealthService::new(
+            Some(get_router_callback(shard_health_clients, logger.clone())),
+            logger.clone(),
+        )
+        .into_service();
```
jcape, line 128: It's not currently possible for this to work with dynamic shards. You'll need to make a "shards" cache that's shared between the router service and the health service.

Aside: I'm a little confused about `RouterClientListenUri` below, can we chat about it tomorrow?

jcape (changes requested):
This is only taking the static startup configuration into account, and will need to be worked significantly for dynamic shards.

samdealy:
@jcape, @aweis89, and I met today and decided that we will forgo implementing this type of shard-aware health check in the code. Rather, we'll use K8 operator logic to perform the same type of health check. I'll introduce a new PR that offers a simple health check that signifies that the router is able to respond to requests / hasn't crashed.

PR #3180, opened 2023-02-27
Update version of builder image to fix permissions issues in /opt/rustup. New builder image thanks to Jason Greathouse.

christian-oudard:
Change .mobconf to point to 0.0.23

### Motivation

The current builder image hits permission issues when doing builds, because it can't write to /opt/rustup.

### DIFF
--- a/.mobconf
+++ b/.mobconf
@@ -1,9 +1,9 @@
 [image]
 url = mobilecoin/builder-install
-tag = v0.0.21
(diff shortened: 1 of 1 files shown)


nick-mobilecoin (approved):

cbeck88:
Thank you!

cbeck88 (approved):

PR #3190, opened 2023-03-01
Remove client-facing API from store

NotGyro:
Removes the FogKeyImageApi trait, and any initialization of the associated 
GRPC server, from `KeyImageService` (the service for a Key Image Store). 

### Motivation

After much discussion, the team involved in the Fog Ledger Router project settled on moving directly to the ledger/store configuration. Rather than adding it as an alternative mode that the Fog Ledger could be hosted in for an incremental rollout, it was decided that we should move directly to the router/store configuration in order to avoid additional maintenance and any possible bugs from supporting both designs at the same time. So, this PR is about removing support for direct interaction with clients (i.e. without a router) from the key image store. 

### Future Work

* As KeyImageService will always function as a store from now on, there should be no need for the KeyImageClientListenUri enum. Instead, a KeyImageStoreUri should be passed to the service in every case. *This may be needed within this pull request, instead of as future work, which is part of the reason it is a draft.*


NotGyro:
* feature/fog-ledger-router
  * **PR #3190** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3190" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3190?utm_source=stack-comment).

NotGyro:
This is failing `test_get_ingress_key_records()`, from `fog/sql_recovery_db`, for me locally. I'm looking into it.

PR #3192, opened 2023-03-01
Make Fog Ledger Store return externally-addressable URI to Router

awygle:
### Motivation
Pulling in changes discovered by Fog View Router leading to PR #3081. Returning remotely-accessible Responder ID URI from Key Image Stores rather than local bind address.


samdealy (approved):

NotGyro (approved):
Looks good to me! This is a simpler change than I realized it would be. Thank you for taking care of this, I might pull it in to some of my branches to attempt to fix issues present in there.

PR #3195, opened 2023-03-02
Update misleading comment

wjuan-mob:
Update a single comment to add clarification

### Motivation

The comment for the function claims it does something it specifically does not do.

### Future Work

None

[Soundtrack of this PR]()


cbeck88 (approved):
Thanks

nick-mobilecoin (approved):
nice find!

PR #3211, opened 2023-03-08
Chore/remove temp dir

wjuan-mob:
This is removing references to tempdir and replacing with tempfile. 
All the changes are in test code or in things like test_utils. 
This is a creation of https://github.com/mobilecoinfoundation/mobilecoin/pull/3209 not on an external fork to resolve an issue with CI.

### Motivation

Tempdir is deprecated. And it is needed in order to mitigate a vulnerability in a dependency in tempdir on remove_dir_all: https://rustsec.org/advisories/RUSTSEC-2023-0018 

### Future Work
There will likely need to be a follow up pr to address the use of slog-gelf which also has a dependency on tempdir.

[Soundtrack of this PR]()


cbeck88 (approved):

eranrund (approved):
LGTM once CI is green

PR #3214, opened 2023-03-09
mobilecoind: add support for mixed transactions, including those with SCIs in them

cbeck88:
This adds a new endpoint to mobliecoind called `generate_mixed_transaction`.

Unlike the `generate_transaction` endpoint, it does not assume that every outlay has the token id of the fee. Instead the request has a `fee_token_id` and each outlay is now an `OutlayV2` which has a token id parameter.

`Outlay` is used both in the requests and responses for `generate_transaction`, so what I decided to do is extend `Outlay` to `OutlayV2`, which has a token id parameter, and `TxProposal` now contains `OutlayV2`. That protobuf extension was backwards compatible. All the historical requests, though, still use `Outlay`.

Clients that only know about `Outlay` will still deserialize `TxProposal` correctly, ignoring the new field. Clients that know about `OutlayV2` will be able to see the token id of each outlay.

The `generate_transaction` request type is not changed, it still uses the old outlay type. This is because I think it would be ambiguous -- old clients expect to set that field using the `token_id` parameter which governs the fee. If the server would default that to zero, it could change the meaning of requests coming from old clients.

The `generate_mixed_transaction` request type uses `OutlayV2` as the parameter.

Let me know if you see a problem with this compatibility strategy, happy to revisit it.

---

This PR increased in scope from what I originally planned -- to implement this stuff correctly, we need to be able to track balance changes coming from SCIs across several tokens, and then aggregate all those changes across the whole transaction to figure out the sizes of change outputs. So I added some functionality to `SignedContingentInput` to help with this.

Additionally, I found myself doing a lot of manual `u64` and `u128` arithmetic which is a bit subtle. In earlier PRs Eran had suggested that we should figure out a way to DRY all that up, but I didn't see a good way and it was only used in two places at the time. This time around, now that I have to do all that yet again, I decided to make such a way to reuse code, which is now a small utility class `U64Ratio`.

I'm happy to try to split out the lower level changes into a separate PR that can be reviewed first, but I feel like it's helpful to see what I'm trying to accomplish in order to scope those changes.

---

In latest commits, we have also removed the `FALLBACK_FEE` constant, which was unexpected and thought to be potentially hazardous during testing of what happens if you try to pay a fee in eusd and you don't specify a fee. If the user doesn't specify a fee, and the blockchain doesn't give a minimum fee for a particular token, then an error is returned that this token cannot be used to pay fees.


mobilecoind/src/payments.rs
```diff
@@ -324,16 +365,184 @@ impl<T: BlockchainConnection + UserTxConnection + 'static, FPR: FogPubkeyResolve
...
+    /// * `inputs` - UTXOs that will be spent by the transaction.
+    /// * `scis` - SCIs to incorporate into the transaction
+    /// * `outlays` - Output amounts and recipients.
+    /// * `last_block_infos` - Last block info responses from the network, for
+    ///   determining fees. This should normally come from polling_network_state
+    /// * `opt_fee` - Transaction fee in picoMOB. If zero, defaults to MIN_FEE.
+    /// * `opt_tombstone` - Tombstone block. If zero, sets to default.
+    /// * `opt_memo_builder` - Optional memo builder to use instead of the
+    ///   default one (EmptyMemoBuilder).
+    pub fn build_mixed_transaction(
```
cbeck88, line 417: One place we could make an executive decision here is, instead of having `build_transaction` and `build_mixed_transaction` in this file, we could just have `build_mixed_transaction`, and the current callers of `build_transaction` could call this one instead, since it's a superset of the functionality.

Such a change has a larger code footprint on the existing APIs, but since `build_tx_proposal` has also been extended in this way, likely we should just do the same thing here I guess.

I will likely do this in the next commit.

mobilecoind/src/service.rs
```diff
@@ -2173,6 +2276,7 @@ build_api! {
 
     // Signed contingent inputs
     generate_swap GenerateSwapRequest GenerateSwapResponse generate_swap_impl,
+    generate_mixed_tx GenerateMixedTxRequest GenerateMixedTxResponse generate_mixed_tx_impl,
```
cbeck88, line 2279: i still need to write some tests for this API

transaction/extra/src/signed_contingent_input.rs
```diff
@@ -82,11 +83,16 @@ impl SignedContingentInput {
     /// Note: This does check any other rules like tombstone block, or
     /// confirm proofs of membership, which are normally added only when this
     /// is incorporated into a transaction
-    pub fn validate(&self) -> Result<(), SignedContingentInputError> {
+    pub fn validate(&self) -> Result<SignedContingentInputAmounts, SignedContingentInputError> {
```
cbeck88, line 86: I decided to change this function so that validate returns the list of all the amounts of things it unmasked.

This makes it less complicated to find all those amounts after validating the SCI, and avoids the overhead of unmasking the TxOut's again and again.

transaction/extra/src/signed_contingent_input.rs
```diff
@@ -215,8 +316,10 @@ pub enum SignedContingentInputError {
     MissingPartialFillChange,
     /// Index out of bounds
     IndexOutOfBounds,
-    /// Fractional change amount was zero
+    /// Partial fill change amount was zero
     ZeroPartialFillChange,
+    /// Partial fill output amount was zero
+    ZeroPartialFillOutput,
```
cbeck88, line 322: this is not currently an error from the consensus network's point of view, but if we change input rules validation to use `U64Ratio`, then it probably will be.

I think that's okay, there's not any interesting use-case for having zero-value partial fill outputs.

mobilecoind/src/service.rs
cbeck88: okay, have a test, but should make more tests still

mobilecoind/src/service.rs
```diff
@@ -4184,6 +4292,126 @@ mod test {
...
+            client.generate_mixed_tx(&request).unwrap()
+        };
+
+        assert_eq!(
+            generate_mixed_tx_response
+                .get_tx_proposal()
+                .get_scis()
+                .len(),
+            1
+        );
```
cbeck88, line 4412: i would like to try to unmask the outputs of the Tx and see that they make sense, going to extend this test

mobilecoind/src/service.rs
cbeck88: I think I should also try to make it so that specifying the utxo list is not required to use this function, and the server will figure that out for you if you don't specify

mobilecoind/api/src/lib.rs
```diff
@@ -21,6 +21,15 @@ mod autogenerated_code {
 
 pub use autogenerated_code::{mobilecoind_api::*, *};
 
+// This is used for some tests, we want to test that the fields that appeared
+// in an outlay in a request match the corresponding fields in the outlay in the
+// response.
+impl PartialEq<Outlay> for OutlayV2 {
+    fn eq(&self, other: &Outlay) -> bool {
+        self.value == other.value && self.receiver == other.receiver
```
eranrund, line 29: I think this also must check for self.token_id == 0

mobilecoind/src/payments.rs
```diff
@@ -1078,62 +1232,91 @@ impl<T: BlockchainConnection + UserTxConnection + 'static, FPR: FogPubkeyResolve
...
                 .map_err(|_| Error::TxBuild("failed creating InputCredentials".into()))?,
             );
+            *balance_sheet.entry(utxo.token_id.into()).or_default() -= utxo.value as i128;
+        }
+
+        for (sci_for_tx, sci_amounts) in scis {
+            if sci_for_tx.partial_fill_value == 0 {
+                tx_builder.add_presigned_input(sci_for_tx.sci.clone())?;
+            } else {
+                // The tx_builder expects to be given the partial fill change value,
```
eranrund, line 1242: Maybe it makes sense to change the transaction builder api (in a separate pr)?
My thinking is that since this API is always used by the claimer of an SCI, wouldn't it make sense that they always specify how much they want to get? I think they'll always start with that so they'll end up doing this exact subtraction every time this is used.

mobilecoind/src/payments.rs
```diff
@@ -1078,62 +1232,91 @@ impl<T: BlockchainConnection + UserTxConnection + 'static, FPR: FogPubkeyResolve
...
+                return Err(Error::InsufficientFunds);
+            }
+            // Note: RTH normally requires that we write change even if it's zero, but we
+            // haven't done that here yet, and we're always writing empty memos.
+            // Note: RTH doesn't tolerate having multiple change outputs right now IIRC,
+            // so we would have to extend the scope of RTH to work with complex mixed
+            // transactions that have multiple change outputs, if we want that
+            // to work.
+            if *val < 0 {
+                let change_amount = Amount::new(-*val as u64, *token_id);
```
eranrund, line 1312: Do we need to make sure the change fits in a u64? I know the worst case is the tx wont get accepted, but it might make debugging easier if we check for that.

mobilecoind/src/service.rs
```diff
@@ -4184,6 +4292,226 @@ mod test {
...
+        // Get list of unspent tx outs
+        let utxos = mobilecoind_db
+            .get_utxos_for_subaddress(&counterparty_monitor_id, 0)
+            .unwrap();
+        assert!(!utxos.is_empty());
+
+        // We will try to take half of the offered input.
+        let mut sci_for_tx = api::SciForTx::new();
+        sci_for_tx.set_sci(generate_swap_response.get_sci().clone());
+        sci_for_tx.set_partial_fill_value(offered_value / 4);
```
eranrund, line 4394: Doesn't this imply taking a quarter?

mobilecoind/src/service.rs
```diff
@@ -4184,6 +4292,226 @@ mod test {
...
+            utxos
+                .iter()
+                .filter(|utxo| utxo.token_id == 1)
+                .map(Into::into)
+                .collect(),
+        );
+        request.set_scis(vec![sci_for_tx].into());
+        let generate_mixed_tx_response = client.generate_mixed_tx(&request).unwrap();
+
+        assert_eq!(
```
eranrund, line 4411: could be nice to compare the actual scis

mobilecoind/src/service.rs
```diff
@@ -4184,6 +4292,226 @@ mod test {
...
+        for output in tx.prefix.outputs.iter() {
+            if let Ok((amount, _scalar)) =
+                output.view_key_match(swap_counterparty.view_private_key())
+            {
+                if amount.token_id == Mob::ID {
+                    assert_eq!(amount.value, offered_value / 4 - Mob::MINIMUM_FEE);
+                    found_counterparty_output = true;
+                } else {
+                    assert_eq!(*amount.token_id, 1);
+                    // It's not very easy to figure out what the value of this change output should
```
eranrund, line 4436: Isn't it going to be the offered amount minus claimed amount (which is hardcoded below)?

mobilecoind/src/service.rs
```diff
@@ -4184,6 +4292,226 @@ mod test {
...
+            }
+        }
+        assert!(found_counterparty_output);
+        assert!(found_counterparty_change);
+        assert!(found_originator_output);
+        assert!(found_originator_change);
+
+        // Changing the fee token id to 1 should work, and slightly adjust the output
+        // values.
+        request.set_fee_token_id(1);
```
eranrund, line 4459: Checking my understanding here: does this result in zero fee since there's no fee configured in the default fee map?
If so, maybe we want to change the fee map to include a fee for some non-mob tokens? I think this can be done [here](https://github.com/mobilecoinfoundation/mobilecoin/blob/mobilecoind-mixed-transactions/mobilecoind/src/test_utils.rs#L135-L138)

mobilecoind/src/service.rs
```diff
@@ -4184,6 +4292,226 @@ mod test {
         }
     }
 
+    #[test_with_logger]
+    fn test_generate_mixed_tx(logger: Logger) {
```
eranrund, line 4296: This is a really nice, easy to follow test.

transaction/extra/src/signed_contingent_input.rs
eranrund: Nice, I think this can be useful in the DEQS

eranrund (changes requested):
Very nice work, got some small comments but over all looks great.

mobilecoind/src/service.rs
cbeck88: yeah i have to update the comment, thanks

mobilecoind/src/service.rs
cbeck88: I think that's right -- will look into this

mobilecoind/api/src/lib.rs
cbeck88: So I think in some cases, there are tests that are like, call `generate_tx`, where `token_id = 1`, and `OutlayV1` is used.

Then, in the `TxProposal` that results, the outlays that are recorded are now an `OutlayV2`, and they do have all their token id's stored with them, so they will have `token_id = 1`.

So if I also check that `self.token_id == 0`, then that test will be failing.

---

Probably the right thing to do is, not create this PartialEq impl, and instead go fix up those tests so that when comparing Outlay and OutlayV2, we convert Outlay to OutlayV2 first providing appropriate context. I did it this way first because it was faster to develop.

mobilecoind/src/service.rs
cbeck88: I think what happens is, we need to figure out what is the value of the token id 1 input that `swap_counterparty` will select. Then we can figure out what the size of the token id 1 change output should be for them.

I think I could likely figure this out by filtering in the `utxo` set and getting the value of the first one, I can follow up with that.

mobilecoind/src/payments.rs
cbeck88: gonna make an issue

cbeck88: #3230

mobilecoind/api/src/lib.rs
eranrund: Thats also acceptable (not creating it)

mobilecoind/src/service.rs
cbeck88: When I tried to actually test it, it's selecting an unexpectedly large fee, with just the default map, still trying to figure out why

mobilecoind/src/service.rs
```diff
@@ -4184,6 +4306,249 @@ mod test {
...
+                    assert_eq!(*amount.token_id, 1);
+                    // The quote asks for 123 eusd, counterparty fulfills to 1/4,
+                    // rounding up so they give away 31.
+                    // Change is their eusd utxo value minus that, minus the eusd transaction fee.
+                    // FIXME: I'm not sure where the 10 billion number comes from, it is not the
+                    // same as the value of Mob::MINIMUM_FEE, and doesn't appear to be in
+                    // FeeMap::default()
+                    assert_eq!(
+                        amount.value,
+                        counterparty_eusd_utxo_value - 31 - 10_000_000_000
```
cbeck88, line 4515: this is the fee it's actually selecting right now, with the default map...

eranrund: Looks like there's a `FALLBACK_FEE` const...

eranrund: seems questionable to me to allow this, it could result in people paying much higher fees than expected if something is misconfigured

cbeck88: okay, i got rid of fallback fee and extended the testing fee map as you suggested in the latest commit, and there are more tests of what happens when the user specifies a fee value

mobilecoind/src/payments.rs
```diff
@@ -179,17 +216,17 @@ impl<T: BlockchainConnection + UserTxConnection + 'static, FPR: FogPubkeyResolve
...
+            (opt_fee, block_version)
         } else {
             let fee = last_block_info
                 .minimum_fee_or_none(&token_id)
-                .unwrap_or(FALLBACK_FEE);
-            let block_version = max(
-                candidate_block_version,
-                last_block_info.network_block_version,
-            );
+                .ok_or_else(|| Error::TxBuild("Token cannot be used to pay fees".into()))?;
```
cbeck88, line 229: this is what we do now instead of the `FALLBACK_FEE`

it's possible that we should give this error if the token id is not in the last_block_info, even if the user provided a fee value

mobilecoind/src/payments.rs
```diff
@@ -179,17 +216,17 @@ impl<T: BlockchainConnection + UserTxConnection + 'static, FPR: FogPubkeyResolve
...
         // nonzero
-        let candidate_block_version = self.ledger_db.get_latest_block()?.version;
+        let block_version = max(
+            self.ledger_db.get_latest_block()?.version,
+            last_block_info.network_block_version,
+        );
+
         Ok(if opt_fee != 0 {
-            (opt_fee, candidate_block_version)
+            (opt_fee, block_version)
```
cbeck88, line 225: i think it was slightly buggy that we were not considering `last_block_info.network_block_version` here before

mobilecoind/src/service.rs
```diff
@@ -4184,6 +4306,330 @@ mod test {
...
+        let fee_override = 500_000;
+        request.set_fee(fee_override);
+
+        let generate_mixed_tx_response = client.generate_mixed_tx(&request).unwrap();
+        let mut response_sci = SignedContingentInput::try_from(
+            generate_mixed_tx_response.get_tx_proposal().get_scis()[0].get_sci(),
+        )
+        .unwrap();
+        // The request sci doesn't have proofs, so to test equal, we have to delete
+        // those in resp
```
cbeck88, line 4568: maybe this is the wrong approach, maybe the response shouldn't contain these proofs either. they are already elsewhere in the Tx anyways, the server is duplicating them I guess

transaction/extra/src/signed_contingent_input.rs
```diff
@@ -142,17 +149,111 @@ impl SignedContingentInput {
...
+    /// Computes the hypothetical change in balances that will occur if we fill
+    /// this SCI to a certain degree
+    ///
+    /// Add the outputs and inputs to a BTreemap which functions as a balance
+    /// sheet. Outputs from the SCI are positive, and the value of the input
+    /// is negative.
+    ///
+    /// Arguments:
+    /// partial_fill_value: The amount of the partial_fill_change we want to
+    /// keep. This should be zero if this is not a partial fill SCI.
```
nick-mobilecoin, line 197: :wrench: The use of "change" and "we" is a bit confusing.
I'm thinking this is saying that the `partial_fill_value` is the amount being taken/claimed by the counterparty?
```suggestion
    /// * `partial_fill_value` - The amount being taken by the counter party. 
    ///   This should be zero if this is not a partial fill SCI.
```

It feels like this argument should be an Option<> like the partial change on the SCI.

nick-mobilecoin (approved):
The main blocker is making sure the proto changes are correct and that falling back to MOB is what we want to do if these protos happen to talk to an older API.

mobilecoind-json/src/data_types.rs
```diff
@@ -486,6 +486,39 @@ impl TryFrom<&JsonOutlay> for api::Outlay {
     }
 }
 
+#[derive(Deserialize, Serialize, Default, Debug, Clone)]
+pub struct JsonOutlayV2 {
+    pub value: JsonU64,
+    pub receiver: JsonPublicAddress,
+    pub token_id: JsonU64,
+}
```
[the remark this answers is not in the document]
cbeck88: It is used in all of the historical request objects, see `src/bin/main.rs`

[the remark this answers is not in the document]
cbeck88: I think it's more error prone if we don't mirror the proto types into json separately. There's also issues around, if we have conversions from `JsonOutlay` to both `Outlay` and `OutlayV2` it can be a lossy conversion and hide a bug that causes people to send the wrong token.

We do have to have Outlay and OutlayV2 in proto or I'm think we will break backwards compat.

mobilecoind/api/proto/mobilecoind_api.proto
```diff
@@ -493,12 +518,58 @@ message GenerateTxRequest {
...
+    repeated SciForTx scis = 8;
 }
 
 message GenerateTxResponse {
     TxProposal tx_proposal = 1;
 }
 
+// Generate a transaction proposal object with mixed token types.
+// Notes:
+// - Sum of inputs needs to be greater than sum of outlays and fee.
```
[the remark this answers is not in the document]
cbeck88: yes -- i copied this comment, we should fix it for the `build_transaction` call also

mobilecoind/api/proto/mobilecoind_api.proto
```diff
@@ -493,12 +518,58 @@ message GenerateTxRequest {
...
+// - The set of inputs to use would be chosen automatically by mobilecoind.
+// - The fee field could be set to zero, in which case mobilecoind would choose a fee.
+// Right now that fee is hardcoded.
```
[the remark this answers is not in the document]
cbeck88: We actually changed this and deleted the hard coded fee, it's not hard-coded anymore.
It will choose based on the network reported minimum fee, and if that doesn't exist, we have an error.

mobilecoind-json/src/data_types.rs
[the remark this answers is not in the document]
wjuan-mob: Shouldn't the proto definition be extendable? Normally optional fields can be added I think. And then we can have the default behavior when token_id isn't defined. Which is what we want.

mobilecoind/api/proto/mobilecoind_api.proto
```diff
@@ -108,11 +109,20 @@ enum TxStatus {
...
 // Structure used in specifying the list of outputs when generating a transaction.
+// Here the token id is implied from context, and matches the fee token id.
 message Outlay {
     uint64 value = 1;
     external.PublicAddress receiver = 2;
 }
 
+// Structure used in specifying the list of outputs in a transaction.
+// Here the token id is explicit.
+message OutlayV2 {
```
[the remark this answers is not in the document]
eranrund: You can't tell if its missing or of it is intentionally set to 0 to represent MOB

mobilecoind/api/proto/mobilecoind_api.proto
```diff
@@ -493,12 +518,58 @@ message GenerateTxRequest {
...
+    // All UnspentTxOuts must belong to the same sender_monitor_id.
+    // mobilecoind would choose a subset of these inputs to construct the transaction.
+    // Total input amount must be >= sum of outlays + fees, for each token id involved.
+    repeated UnspentTxOut input_list = 3;
+
+    // List of SCIs to be added to the transaction
+    repeated SciForTx scis = 4;
+
+    // Outputs to be generated by the transaction. This excludes change and fee.
+    repeated OutlayV2 outlay_list = 5;
```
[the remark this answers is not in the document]
cbeck88: The fee is not considered an outlay. That's how this API works. Regardless of if you think it maybe should be considered an outlay, we can't change that now.

mobilecoind/src/payments.rs
```diff
@@ -90,6 +114,19 @@ impl TxProposal {
...
+/// data about what degree to fill it, if it is a partial fill SCI.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct SciForTx {
+    /// The signed contingent input to add to the transaction
+    pub sci: SignedContingentInput,
+    /// The amount to take from the maximum allowed volume of this SCI.
+    /// The remainder is returned to the originator as change.
+    /// For partial fill SCIs, this
+    /// must be nonzero. For non-partial fill SCIs, this must be zero.
+    pub partial_fill_value: u64,
```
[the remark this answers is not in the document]
cbeck88: Right now it's just a pass-through from proto, proto doesn't have `Option<u64>`.

My bias is that having a pass-through from proto is simpler and easier to maintain.

mobilecoind/api/proto/mobilecoind_api.proto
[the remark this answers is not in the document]
wjuan-mob: Oh I didn't realize that they didn't serialize default values. That's kind of surprising.

mobilecoind/api/proto/mobilecoind_api.proto
```diff
@@ -493,12 +518,58 @@ message GenerateTxRequest {
...
+
+    // Fee value, in smallest representable units (u64).
+    // (For MOB this is picoMOB.)
+    // Setting to 0 causes mobilecoind to choose a value.
+    // The value used can be checked (but not changed) in tx_proposal.tx.prefix.fee
+    uint64 fee = 6;
+
+    // Token id to use for the transaction.
```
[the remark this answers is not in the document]
cbeck88: thank you

mobilecoind/api/proto/mobilecoind_api.proto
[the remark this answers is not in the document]
wjuan-mob: I think a work around that might be interesting is to use a oneof for the token id field, so it defaults to a pseudo null value, instead of to 0.

transaction/extra/src/signed_contingent_input.rs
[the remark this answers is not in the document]
cbeck88: > I can understand avoiding the unmasking, but I feel like someone isn't going to think to look to validate() to get the SignedContingentInputAmounts.

Well, they can look in the documentation, or the code examples that will be all around them after this.

> than to have a secondary output with validate().

What is actually wrong with this? I've seen plenty of validate functions that return data and I don't see a good reason not to do this.

The `try_from` suggestion would work without unnecessary overhead, but it is less KISS, so my bias is against it.

transaction/extra/src/signed_contingent_input.rs
cbeck88: in this documentation, it's better to use terms like "us" than "the counter party" because it's more clear to the caller, IMO. I will try to improve this though

mobilecoind/src/payments.rs
cbeck88: yes, in the scenario that there is network congestion, users have to pay higher fees to get their transactions through

eranrund: Yes, it will happily take a larger fee. Which is why this FALLBACK_FEE was problematic - users might end up unexpectedly paying higher fees than necessary

mobilecoind-json/src/data_types.rs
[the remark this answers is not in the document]
cbeck88: See my earlier comments on the PR description:

> Outlay is used both in the requests and responses for generate_transaction, so what I decided to do is extend Outlay to OutlayV2, which has a token id parameter, and TxProposal now contains OutlayV2. That protobuf extension was backwards compatible. All the historical requests, though, still use Outlay.

Extending outlay in the responses is fine -- extending outlay in the requests is not fine. The current semantics of all the requests in mobilecoind is that token id is a single top-level field that specifies the token id for everything in the entire transaction. If we change all of those to OutlayV2, we break old clients unless we ignore the new token id field in OutlayV2. That's confusing to the client and a hazard, so that's not an acceptable approach.

That means that we cannot completely get rid of the unextended version of Outlay. We replace it with OutlayV2 in responses, and use OutlayV2 in any new requests, but we don't migrate to the new Outlay in any pre-existing request objects, because we can't do so without breaking things.

mobilecoind/src/payments.rs
```diff
@@ -263,50 +300,150 @@ impl<T: BlockchainConnection + UserTxConnection + 'static, FPR: FogPubkeyResolve
...
+        last_block_infos: &[BlockInfo],
+        opt_fee: u64,
+        opt_tombstone: u64,
+        opt_memo_builder: Option<Box<dyn MemoBuilder + 'static + Send + Sync>>,
+    ) -> Result<TxProposal, Error> {
+        let logger = self.logger.new(o!("sender_monitor_id" => sender_monitor_id.to_string(), "outlays" => format!("{outlays:?}")));
+        log::trace!(logger, "Building pending transaction...");
+
+        // Must have at least one output
+        if outlays.is_empty() && scis.is_empty() {
```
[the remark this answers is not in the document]
cbeck88: it does not

mobilecoind/api/proto/mobilecoind_api.proto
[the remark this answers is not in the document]
cbeck88: I think that's a more complicated approach than what we did here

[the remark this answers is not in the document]
cbeck88: I don't want all the clients that use this to have to deal with enums and "pseudo-null values" and one ofs and so on.

I like that with OutlayV2, you can be 100% sure that the token id is present in a meaningful way. And with Outlay v1, you can be 100% sure that the token id is not present and has to be inferred from context. And we can use the appropriate one in APIs as appropriate.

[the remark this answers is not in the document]
wjuan-mob: I guess that's fair. The only thing I'm really trying to avoid is having to distinguish outlay and outlayv2 scenarios and make sure that the default behaviour is consistent, but at the end of the day, having to check a null value is similar.

mobilecoind/src/payments.rs
```diff
@@ -1181,6 +1365,15 @@ impl<T: BlockchainConnection + UserTxConnection + 'static, FPR: FogPubkeyResolve
             tx,
             outlay_index_to_tx_out_index,
             outlay_confirmation_numbers,
+            scis: scis
+                .iter()
+                .cloned()
+                .map(|(mut sci_for_tx, _sci_amount)| {
+                    // clear out proofs here since they were not part of the request
+                    sci_for_tx.sci.tx_in.proofs = Vec::default();
```
eranrund, line 1373: I'm not sure this is necessary, maybe its better to include the proofs that were attached to each SCI? The end user can always clear them if they want to and we can document that the proofs are included in the response.

wjuan-mob (approved):

mobilecoind/api/proto/mobilecoind_api.proto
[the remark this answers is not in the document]
cbeck88: > I think a work around that might be interesting is to use a oneof for the token id field, so it defaults to a pseudo null value, instead of to 0.

thinking about this more, I don't think this would work either.

Protobuf serializers are allowed to omit defaulted values, even if the user explicitly set token id to 0 (MOB).

If we change what the deserializer thinks that means, from 0 to "null", that's a breaking change. So I don't think the oneof idea really helps anything -- the default value, when the user omitted this field, has to be zero.

mobilecoind/src/payments.rs
cbeck88: I think they can also find the proofs in the `Tx` itself, right? they could find the TxIn that matches this SCI and then the proofs will be there.

It will make the response considerably smaller if we can avoid putting the proofs twice, and it makes it match what the user actually submitted -- that was why i went back and did it this way.

eranrund: Cool, sounds good

mobilecoind/src/payments.rs
```diff
@@ -63,14 +62,36 @@ pub struct Outlay {
...
+    /// Amount being sent.
+    pub amount: Amount,
+
+    /// Destination.
+    pub receiver: PublicAddress,
+}
+
+impl Outlay {
+    // This is used for some tests
+    pub fn to_outlay_v2(&self, token_id: TokenId) -> OutlayV2 {
```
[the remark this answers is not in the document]
cbeck88: thanks -- not sure how that happened, i must have lost all the uses when i refactored at some point.

maybe we should just delete it, the only place it seems to be usable is like line 306

mobilecoind/api/proto/mobilecoind_api.proto
```diff
@@ -142,14 +152,24 @@ message UnspentTxOut {
...
+
 // Structure used to refer to a prepared transaction
 message TxProposal {
     // List of inputs being spent.
     repeated UnspentTxOut input_list = 1;
 
     // List of outputs being created.
     // This excludes the fee output.
-    repeated Outlay outlay_list = 2;
+    repeated OutlayV2 outlay_list = 2;
```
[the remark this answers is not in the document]
nick-mobilecoin: I think that makes sense, it seems like the main place we might see this is in client 1 -> mobilecoind -> client 2.
But it's the same issue if client 1 -(new endpoint)-> mobilecoind -(old endpoint)-> client 2

mobilecoind/api/proto/mobilecoind_api.proto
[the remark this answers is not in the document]
nick-mobilecoin: > if you think it maybe should be considered an outlay for stylistic reasons, we can't change that now

My concern isn't so much that fee isn't in outlay here, it's that there doesn't seem to be consistent use of the term.  It sounds like term usage in this instance is outside the scope of this PR. I'll stew on how we might get and maintain n'sync for usage.

cbeck88:
> The main blocker is making sure the proto changes are correct and that falling back to MOB is what we want to do if these protos happen to talk to an older API.

What I've tried to do is, not change any of the request types of the older APIs -- falling back to MOB is a hazard, so that's why there's an `OutlayV2` and we didn't just add `token_id` to all `Outlay`.

Only new APIs use `OutlayV2` in requests, and in the response type `TxProposal`, I think it's safe to use `OutlayV2`, which can be deserialized by old clients correctly as an `Outlay`.

transaction/extra/src/signed_contingent_input.rs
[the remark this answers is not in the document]
nick-mobilecoin: I'm fine with this, I don't think the try_from suggestion was much, if any, better.

mobilecoind/api/proto/mobilecoind_api.proto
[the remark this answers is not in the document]
cbeck88: yeah i'm not sure what the ideal term is -- i think in terms of implementation, `Outlay` is "all the stuff you pass to `TransactionBuilder.add_output` right now. But IMHO this seems like an adequate term for this -- an amount of money to be sent to some address

PR #3217, opened 2023-03-10
Remove client-facing API from key-image store

NotGyro:
Removes the FogKeyImageApi trait, and any initialization of the associated 
GRPC server, from `KeyImageService` (the service for a Key Image Store). 

### Motivation

After much discussion, the team involved in the Fog Ledger Router project settled on moving directly to the ledger/store configuration. Rather than adding it as an alternative mode that the Fog Ledger could be hosted in for an incremental rollout, it was decided that we should move directly to the router/store configuration in order to avoid additional maintenance and any possible bugs from supporting both designs at the same time. So, this PR is about removing support for direct interaction with clients (i.e. without a router) from the key image store.


NotGyro:
* feature/fog-ledger-router
  * **PR #3217** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3217" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3217?utm_source=stack-comment).

NotGyro:
This branch's previous incarnation was a bit polluted with commits which should not have been part of this branch. So, I re-created it, to avoid an overcomplicated rebase, because the actual changes present in this branch were pretty minimal.

samdealy (approved):

awygle (approved):
This looks good to me. It also removes the old ledger server, but that's fine.

PR #3224, opened 2023-03-13
make transaction core input rules validation use U64Ratio

cbeck88:
This extends the changes in #3214 around DRY'ing up the u64 / u128 arithmetic stuff.

I don't think we should merge this in release/v4.1 since we can't modify the enclave in 4.1 anyways, I think we should just do this in master later. But this at least shows the direction of the U64Ratio stuff.

### DIFF
--- a/util/u64-ratio/README.md
+++ b/util/u64-ratio/README.md
@@ -5,7 +5,7 @@ This crate provides a type `U64Ratio` which represents a ratio of two `u64`
 integers, and supports some basic functionality using these:
 
 * Comparing two ratios
-* Multiplying a `u64` by the ratio, rounding up or down as appropriate.
+* Multiplying a `u64` by the ratio, rounding up or down as the caller requests.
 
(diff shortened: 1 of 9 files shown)


util/u64-ratio/README.md
```diff
@@ -0,0 +1,12 @@
+mc-util-u64-ratio
+=================
+
+This crate provides a type `U64Ratio` which represents a ratio of two `u64`
+integers, and supports some basic functionality using these:
+
+* Comparing two ratios
+* Multiplying a `u64` by the ratio, rounding up or down as appropriate.
```
wjuan-mob, line 8: It might help to clarify what you mean by as appropriate here.

wjuan-mob (approved):

cbeck88:
rebased on master

cbeck88:
I think there are still some cargo issues in the various enclaves to fix, will do that next

awygle (approved):
LGTM.

PR #3232, opened 2023-03-13
mobilecoind: fix a bug in swap generation

cbeck88:
The `SignedContingentInputBuilder` does not know the TxOut global
    indices of the ring elements, they usually aren't supplied to it
    unless merkle proofs of membership were supplied to it.
    It usually ends up setting `tx_out_global_indices` to a list of zeroes.
    Then the originator is supposed to set these correctly,
    and it falls to the counterparty to find proofs of membership.
    
    I forgot to do this in `generate_swap_impl`, so `mobilecoind` was
    leaving `tx_out_global_indices` with an incorrect setting. This
    commit fixes it and adds a test.

this bug was introduced a few days ago here: https://github.com/mobilecoinfoundation/mobilecoin/pull/3212

### DIFF
--- a/mobilecoind/src/payments.rs
+++ b/mobilecoind/src/payments.rs
@@ -1378,9 +1378,12 @@ impl<T: BlockchainConnection + UserTxConnection + 'static, FPR: FogPubkeyResolve
         }
 
         // Build sci.
-        sci_builder
+        let mut result = sci_builder
             .build(&NoKeysRingSigner {}, rng)
-            .map_err(|err| Error::TxBuild(format!("build tx failed: {err}")))
+            .map_err(|err| Error::TxBuild(format!("build tx failed: {err}")))?;
+
+        result.tx_out_global_indices = global_indices;
+        Ok(result)
     }
 }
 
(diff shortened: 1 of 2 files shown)


eranrund (approved):

mobilecoind/src/payments.rs
```diff
@@ -1378,9 +1378,12 @@ impl<T: BlockchainConnection + UserTxConnection + 'static, FPR: FogPubkeyResolve
         }
 
         // Build sci.
-        sci_builder
+        let mut result = sci_builder
             .build(&NoKeysRingSigner {}, rng)
-            .map_err(|err| Error::TxBuild(format!("build tx failed: {err}")))
+            .map_err(|err| Error::TxBuild(format!("build tx failed: {err}")))?;
+
+        result.tx_out_global_indices = global_indices;
```
wjuan-mob, line 1385: Do we need to sort global_indices first?

cbeck88: i think they are supposed to correspond in order to the TxOut's in the ring

samdealy (approved):

PR #3243, opened 2023-03-15
make the copyright hook smarter

cbeck88:
It frequently happens that I modify a file but the copyright line is wrong (from last year). Then the hook inserts a new copyright line, instead of fixing the old one, and I go back and fix it manually.

I have done this enough that I would rather make the tool fix the copyright line.

I can use `git commit -n` but it feels dirty and sometimes I need both formatting and copyright fixes.

### DIFF
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -155,11 +155,9 @@ for file in $(git diff --name-only --cached --diff-filter=d | egrep -v '^sgx/sgx
 do
 	if [[ ${file: -3} == ".rs" ]]; then
 	    FIRST_LINE=$(head -n 1 ${file})
-	    if ! [[ "${FIRST_LINE}" =~ "// Copyright (c)" ]]
-	    then
+	    if ! [[ "${FIRST_LINE}" =~ "// Copyright (c)" ]]; then
 	        missing_copyright="$file\n$missing_copyright"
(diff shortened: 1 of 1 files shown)


hooks/pre-commit
```diff
@@ -149,30 +149,46 @@ fi
...
-		if ! [[ $(head -n 1 ${file}) == "${COPYRIGHT_LINE}" ]]
-		then
-		    diff="$file\n$diff"
-		fi
+	    FIRST_LINE=$(head -n 1 ${file})
+	    if ! [[ "${FIRST_LINE}" =~ "// Copyright (c)" ]]
+	    then
+	        missing_copyright="$file\n$missing_copyright"
+	    elif ! [[ "${FIRST_LINE}" == "${COPYRIGHT_LINE}" ]]
+	    then
```
jcape, line 162:
```suggestion
	    if ! [[ "${FIRST_LINE}" =~ "// Copyright (c)" ]]; then
	        missing_copyright="$file\n$missing_copyright"
	    elif [[ "${FIRST_LINE}" != "${COPYRIGHT_LINE}" ]]; then
```

jcape (approved):
LGTM, minor nit about

```
if; then
```

vs.

```
if
then
```

PR #3261, opened 2023-03-20
expose Schnorrkel Error from mc-crypto-keys

cbeck88:
This type is already exposed in public APIs of RistrettoPublic and various conversions, but it is not actually exported from the crate, so there is no way for downstream to name the type.

This commit re-exports the type at crate level, which is already part of the public API anyways.

### DIFF
--- a/crypto/keys/src/lib.rs
+++ b/crypto/keys/src/lib.rs
@@ -76,7 +76,7 @@ pub(crate) use base64::engine::general_purpose::STANDARD as BASE64_ENGINE;
 
 pub use digest::Digest;
 pub use mc_util_repr_bytes::{typenum::Unsigned, GenericArray, LengthMismatch, ReprBytes};
-pub use schnorrkel_og::{SignatureError as SchnorrkelError};
+pub use schnorrkel_og::SignatureError as SchnorrkelError;
 pub use signature::{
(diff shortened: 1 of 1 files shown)


nick-mobilecoin (commented):
Curious what the use case is?

I may  have missed other places, but this looks to be used in 3 places:
- [RistrettoPublic::verify_digestable()](https://github.com/mobilecoinfoundation/mobilecoin/blob/c96e1b3f3840e629f00d6503dc71a42d70c66f65/crypto/keys/src/ristretto.rs#L390) which maps the error
-  [RistrettoPublic::veirfy_schnorrkel()](https://github.com/mobilecoinfoundation/mobilecoin/blob/c96e1b3f3840e629f00d6503dc71a42d70c66f65/fog/sig/authority/src/ristretto.rs#L27) which maps the error
- [verify()](https://github.com/mobilecoinfoundation/mobilecoin/blob/c96e1b3f3840e629f00d6503dc71a42d70c66f65/crypto/sig/src/lib.rs#L50) which is exposed as `SignatureError`

cbeck88:
I have a service that calls `verify_schnorrkel` and I need to do things like implement `From<SchnorrkelError> for Error`.

It seems to me that there's no good reason that I shouldn't be able to do this.

nick-mobilecoin (approved):

Issue #3200, opened 2023-03-06
More flexible memo builder


cbeck88:
One request we've had is that, we'd like to make it easier from a software interface point of view for teams like Moby to start using a new memo type in their app.

This ticket is for a more flexible memo builder for the type 1 authenticated memos.

Recall that, memos have 2 type bytes, and the first byte is a category, and the second byte is a type within the category.

The usecase we want to enable is that:
* For example, Moby team wants to have a new authenticated sender memo, but the 32 free bytes, they want to use in some novel way.
* They would write an MCIP for their new memo type, but they want to be able to start using it without waiting for a mobilecoin.git or SDK release, for faster iteration.

The proposed resolution is:
* Make a more flexible version of RTH memo builder (or possible just extend that builder) to allow that they specify the second byte as a run-time parameter, and inject custom functionality that writes the 32 byte memo.
* The SDK will then wrap and expose this, perhaps with some documentation around this being an advanced feature and that they should write an MCIP before using new memo types on mainnet
* Apps would then be able to make app-specific memos without needing to coordinate engineering work with core team and sdk team.

There was earlier discussion around this in [MCIP 60](https://github.com/mobilecoinfoundation/mcips/pull/60)

wjuan-mob:
> 

I think that our current design uses the mcips as the source of truth. You can reference https://github.com/mobilecoinfoundation/mobilecoin/blob/f8be8e7efd936de3eba85a9568005f7fbb5e9e30/transaction/extra/src/memo/mod.rs#L40 for types that have been provided inside mobilecoin, but as types are added by the moby team for example, those would be added in mcips.

PR #3269, opened 2023-03-22
Feature/flexible memo

wjuan-mob:
Add a flexible memo generator option to RTHbuilder
Change set payment intent and set payment request to return a result, which will error when one or the other has already been set.
Add tests for flexible memo generation
Add compute_authenticated_sender_memo and compute_destination_memo
### Motivation
This is intended to address issue [3200](
https://github.com/mobilecoinfoundation/mobilecoin/issues/3200
)
This should help provide a way for users to quickly add custom memo types which adhere to the authenticatedsender and destination format while taking advantage of the additional unused 32 bytes.

### Future Work

Add tests to sample paykit for the flexible memo generation


wjuan-mob:
* **#3269** <a href="https://app.graphite.com/github/pr/mobilecoinfoundation/mobilecoin/3269?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/mobilecoinfoundation/mobilecoin/3269?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3275** <a href="https://app.graphite.com/github/pr/mobilecoinfoundation/mobilecoin/3275?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#3276](https://github.com/mobilecoinfoundation/mobilecoin/pull/3276) <a href="https://app.graphite.com/github/pr/mobilecoinfoundation/mobilecoin/3276?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#3279** <a href="https://app.graphite.com/github/pr/mobilecoinfoundation/mobilecoin/3279?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

dolanbernard (approved):
LGTM

samdealy (changes requested):

transaction/builder/src/memo_builder/rth_memo_builder.rs
```diff
@@ -82,17 +82,41 @@ pub enum CustomMemoType {
...
+    ) -> Result<FlexibleMemoPayload, NewMemoError>;
+}
+
+/// This is a struct that contains a generator used for creating flexible output
+/// memos and flexible change memos
+#[derive(Clone)]
+pub struct FlexibleMemoGenerator {
+    /// This is a reference to the actual generator that will be called.
+    pub generator: Arc<Box<dyn FlexibleMemoGeneratorTrait>>,
+}
```
samdealy, line 108: Traits are meant to be implemented by structs via impl blocks. This `FlexibleMemoGenerator` struct adds a layer of indirection- it holds a dynamically erased type that implements the trait, which is not commonly done in rust.

Remove this struct and see below for how to remove this indirection.

transaction/builder/src/memo_builder/rth_memo_builder.rs
```diff
@@ -82,17 +82,41 @@ pub enum CustomMemoType {
...
+
+impl Debug for FlexibleMemoGenerator {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        f.debug_struct("FlexibleMemoGenerator")
+            .field("generator", &self.generator)
+            .finish()
+    }
+}
```
samdealy, line 116: Since we're removing this struct, we'll want to implement `Debug` for the concrete struct (`FlexibleMemoGeneratorPaymentRequest`) that implement  the trait. So I recommend removing this here and adding it below.

transaction/builder/src/memo_builder/rth_memo_builder.rs
```diff
@@ -532,6 +536,46 @@ mod tests {
...
+            (self.output_memo_generator)(context)
+        }
+
+        fn create_change_memo(
+            &self,
+            context: FlexibleMemoChangeContext,
+        ) -> Result<FlexibleMemoPayload, NewMemoError> {
+            (self.change_memo_generator)(context)
+        }
+    }
```
samdealy, line 578: Remove the use of closures and just implement the methods directly on a new struct called something like `TestFlexibleMemoGenerator`. We shouldn't be using `Fn` trait or closures in this new pattern.

I also don't think we need to test the usage of closures since this is an implementation detail and tests shouldn't test implementation details. We can trust that the rust language's guarantees about composing methods from closures should just work.

transaction/builder/src/memo_builder/rth_memo_builder.rs
```diff
@@ -553,12 +597,15 @@ mod tests {
                 })
             });
 
+        let generator_closure = FlexibleMemoGeneratorClosure {
+            output_memo_generator: flexible_output_memo_generator_closure,
+            change_memo_generator: flexible_change_memo_generator_closure,
+        };
+
```
samdealy, line 604: Remove, see above.

transaction/builder/src/transaction_builder.rs
```diff
@@ -963,34 +962,44 @@ pub mod transaction_builder_tests {
             (BlockVersion::try_from(2).unwrap(), TokenId::from(2)),
         ]
     }
+    // Test memo generator that creates a payload which corresponds to a payment
+    // request.
+    #[derive(Debug)]
+    struct FlexibleMemoGeneratorPaymentRequest {}
+
+    impl FlexibleMemoGeneratorTrait for FlexibleMemoGeneratorPaymentRequest {
```
samdealy, line 970: This is the conventional way to implement a trait, and looks good to me. Just rename to `FlexibleMemoGenerator` and add the `Debug` impl discussed above.

samdealy (changes requested):

transaction/builder/src/memo_builder/rth_memo_builder.rs
wjuan-mob: I'm not sure I understand how to do this. Do you mind helping me out? I ended up using this approach because it seemed like the way forward to get clone on the containing struct which I want.

transaction/builder/src/memo_builder/rth_memo_builder.rs
wjuan-mob: I actually wanted to keep this because I think it's helpful to have an example of creating a struct that implements the trait using closures. I agree that it's a little inelegant, but I think it should be equivalent for the purpose of the test. I can get rid of it if you feel strongly about it, but I do think it isn't introducing extra complexity entirely for no reason.

transaction/builder/src/memo_builder/rth_memo_builder.rs
```diff
@@ -82,17 +82,41 @@ pub enum CustomMemoType {
...
-pub type FlexibleOutputMemoGenerator = Box<
-    dyn Fn(FlexibleMemoOutputContext) -> Result<FlexibleMemoPayload, NewMemoError> + Sync + Send,
->;
-/// A function that returns a flexible memopayload for change
-pub type FlexibleChangeMemoGenerator = Box<
-    dyn Fn(FlexibleMemoChangeContext) -> Result<FlexibleMemoPayload, NewMemoError> + Sync + Send,
->;
+/// This is a trait that must be implemented by generators used for creating
+/// flexible output memos and flexible change memos
+pub trait FlexibleMemoGeneratorTrait: Sync + Send + Debug {
```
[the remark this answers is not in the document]
wjuan-mob: You're right. That's a bad naming convention. I'll fix that.

transaction/builder/src/memo_builder/rth_memo_builder.rs
cbeck88: I agree, we shouldn't introduce this struct, this is needlessly complicated

wjuan-mob: I'm working on it. Should be updated shortly.

transaction/builder/src/memo_builder/rth_memo_builder.rs
```diff
@@ -76,12 +75,107 @@ pub struct RTHMemoBuilder {
...
+        f.debug_struct("FlexibleMemoGenerator")
+            .field("generator", &self.generator)
+            .finish()
+    }
+}
+
+/// This is the context provided to the flexible memo generator which is set on
+/// the rth memo builder. Each of these fields has a direct correspondence to a
+/// field in the RTHMemoBuilder
+pub struct FlexibleMemoBuilderContext {
```
cbeck88, line 121: i think this is too complicated, this code will be very hard to follow

instead of implementing a new generator object, can we just add things to the RTHMemoBuilder?

transaction/builder/src/memo_builder/rth_memo_builder.rs
cbeck88: I think I agree with Sam, this code is simpler and cleaner if you just implement functions on a struct as normal. 

> I can get rid of it if you feel strongly about it, but I do think it isn't introducing extra complexity entirely for no reason.

What is the reason? 

> helpful to have an example of creating a struct that implements the trait using closures

This is not a good reason to code things in a convoluted way. Creating examples like this is not directly related to any of mobilecoin's business or engineering goals afaict.

transaction/builder/src/memo_builder/rth_memo_builder.rs
```diff
@@ -250,46 +337,71 @@ impl MemoBuilder for RTHMemoBuilder {
...
+                    let flexible_memo_payload = &flexible_memos.change_memo_payload;
+                    if flexible_memo_payload.memo_type_bytes[0] != 0x02 {
+                        return Err(NewMemoError::FlexibleMemo(format!("The flexible change memo has a memopayload of the incorrect memo type: {:?}", flexible_memo_payload.memo_type_bytes)));
+                    }
+                    let memo_data = compute_destination_memo(
+                        self.last_recipient.clone(),
+                        self.fee.value,
+                        self.num_recipients,
+                        self.total_outlay,
+                        flexible_memo_payload.memo_data,
```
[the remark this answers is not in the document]
wjuan-mob: I think it would make sense. The other memos are also paired.

transaction/builder/src/memo_builder/rth_memo_builder.rs
```diff
@@ -93,6 +124,20 @@ impl Default for RTHMemoBuilder {
...
+/// An error that occurs when setting up a memo builder
+///
+/// These errors are usually created setting invalid field combinations on the
+/// memo builder. We have included error codes for some known useful error
+/// conditions. For a custom MemoBuilder, you can try to reuse those, or use the
+/// Other error code.
+#[derive(Clone, Debug, Display, Eq, PartialEq)]
+pub enum MemoBuilderError {
+    /// Invalid state change
+    StateChange(String),
```
samdealy, line 136: Can we rename this to something more descriptive? I"m not sure what a `StateChange` error is...

transaction/builder/src/memo_builder/rth_memo_builder.rs
```diff
@@ -186,26 +255,44 @@ impl MemoBuilder for RTHMemoBuilder {
...
+                        let tx_public_key = memo_context.tx_public_key;
+                        let flexible_memo_payload = &flexible_memos.output_memo_payload;
+                        let memo_data = compute_authenticated_sender_memo(
+                            flexible_memo_payload.memo_type_bytes,
+                            cred,
+                            recipient.view_public_key(),
+                            &tx_public_key.into(),
+                            &flexible_memo_payload.memo_data,
+                        );
+                        if flexible_memo_payload.memo_type_bytes[0] != 0x01 {
```
samdealy, line 272: Can you make the `0x01` a named constant?

transaction/builder/src/memo_builder/rth_memo_builder.rs
```diff
@@ -250,46 +337,71 @@ impl MemoBuilder for RTHMemoBuilder {
...
-                self.total_outlay,
-                self.fee.value,
-                payment_request_id,
-            ) {
-                Ok(mut d_memo) => {
+        if let Some(custom_memo_type) = &self.custom_memo_type {
+            match custom_memo_type {
+                CustomMemoType::FlexibleMemos(flexible_memos) => {
+                    let flexible_memo_payload = &flexible_memos.change_memo_payload;
+                    if flexible_memo_payload.memo_type_bytes[0] != 0x02 {
```
samdealy, line 352: Can we make `0x02` a named constant as well?

transaction/builder/src/memo_builder/rth_memo_builder.rs
```diff
@@ -250,46 +337,71 @@ impl MemoBuilder for RTHMemoBuilder {
...
-                    d_memo.set_num_recipients(self.num_recipients);
-                    Ok(d_memo.into())
+                CustomMemoType::PaymentRequestId(payment_request_id) => {
+                    match DestinationWithPaymentRequestIdMemo::new(
+                        self.last_recipient.clone(),
+                        self.total_outlay,
+                        self.fee.value,
+                        *payment_request_id,
+                    ) {
+                        Ok(mut d_memo) => {
```
samdealy, line 374: nit: Rename to `destination_memo`. It's more readable that way.

transaction/builder/src/memo_builder/rth_memo_builder.rs
```diff
@@ -76,12 +75,44 @@ pub struct RTHMemoBuilder {
...
+    /// This is used when generating change memos.
+    /// It must contain a memo type with the first byte being 0x02
+    pub change_memo_payload: FlexibleMemoPayload,
+}
+
+/// This is the payload data used for creating custom MemoPayloads with types
+/// which are defined in MCIPs.
+#[derive(Clone, Debug)]
+pub struct FlexibleMemoPayload {
+    /// memo_type_bytes corresponds to the returned memo type. This should be
```
samdealy, line 101:
```suggestion
    /// Corresponds to the returned memo type. This should be
```

transaction/builder/src/memo_builder/rth_memo_builder.rs
```diff
@@ -76,12 +75,44 @@ pub struct RTHMemoBuilder {
...
+    /// listed in an MCIP.
+    pub memo_type_bytes: [u8; 2],
+    /// memo_data: corresponds to some 32 byte encoding of data used for
```
samdealy, line 104:
```suggestion
    /// Corresponds to some 32 byte encoding of data used for
```

transaction/builder/src/memo_builder/rth_memo_builder.rs
```diff
@@ -93,6 +124,20 @@ impl Default for RTHMemoBuilder {
     }
 }
 
+/// An error that occurs when setting up a memo builder
+///
+/// These errors are usually created setting invalid field combinations on the
+/// memo builder. We have included error codes for some known useful error
+/// conditions. For a custom MemoBuilder, you can try to reuse those, or use the
+/// Other error code.
```
samdealy, line 132:
```suggestion
/// other error code.
```

transaction/builder/src/memo_builder/rth_memo_builder.rs
```diff
@@ -309,3 +421,276 @@ impl MemoBuilder for RTHMemoBuilder {
...
+        let change_memo_payload = FlexibleMemoPayload {
+            memo_type_bytes,
+            memo_data,
+        };
+        FlexibleMemoPayloads {
+            output_memo_payload,
+            change_memo_payload,
+        }
+    }
+    fn get_invalid_flexible_memos() -> FlexibleMemoPayloads {
```
samdealy, line 466:
```suggestion
    }
    
    fn get_invalid_flexible_memos() -> FlexibleMemoPayloads {
```

samdealy (approved):

transaction/builder/src/memo_builder/rth_memo_builder.rs
```diff
@@ -76,12 +75,44 @@ pub struct RTHMemoBuilder {
...
+/// This contains both the output memo payload and the change memo payload which
+/// will be used when generating custom output memos and change memos
+#[derive(Clone, Debug)]
+pub struct FlexibleMemoPayloads {
+    /// This is used when generating output memos.
+    /// It must contain a memo type with the first byte being 0x01
+    pub output_memo_payload: FlexibleMemoPayload,
+    /// This is used when generating change memos.
+    /// It must contain a memo type with the first byte being 0x02
+    pub change_memo_payload: FlexibleMemoPayload,
```
cbeck88, line 94: 1. This is too hard to use as an API. This was not even asked for by Moby team, so they probably don't even want this but now they have to construct it with the right type bytes. At the very least it should be an option that they can default to None.
2. There was not part of what was discussed in #3200 , I think it's just a scope creep and we shouldn't even do it.

transaction/builder/src/memo_builder/rth_memo_builder.rs
```diff
@@ -76,12 +75,44 @@ pub struct RTHMemoBuilder {
...
+}
+
+/// This is the payload data used for creating custom MemoPayloads with types
+/// which are defined in MCIPs.
+#[derive(Clone, Debug)]
+pub struct FlexibleMemoPayload {
+    /// Corresponds to the returned memo type. This should be
+    /// listed in an MCIP.
+    pub memo_type_bytes: [u8; 2],
```
cbeck88, line 103: this is wrong, it needs to be just `second_memo_type_byte: u8`, and the only allowed first byte is 0x01, for the authenticated sender category, as described in #3200

cbeck88 (changes requested):
Remove the custom destination memos, don't allow the caller to set custom first type byte.

cbeck88:
William has pointed out that this MCIP has paired the payment request id and payment intent id memo types:

PR #3291, opened 2023-03-25
use mc-rand, delete mc-crypto-rand

cbeck88:


common/Cargo.toml
```diff
@@ -42,7 +42,7 @@ loggers = [
 mc-crypto-dalek = { path = "../crypto/dalek", default_features = false, features = [ "serde", "alloc" ] }
 mc-crypto-digestible = { path = "../crypto/digestible" }
 mc-crypto-keys = { path = "../crypto/keys", default-features = false, features = [ "serde", "alloc", "prost" ] }
-mc-crypto-rand = { path = "../crypto/rand" }
+mc-rand = "1"
```
varsha888, line 45: Nit: "1.0" to be consistent with other Cargo.toml files
```suggestion
mc-rand = "1.0"
```

varsha888 (approved):

PR #3208, opened 2023-03-08
Track failed tx proposals in consensus

NotGyro:
* Adds a field, `tx_proposal_failures` to `ClientSessionTracking`
* Implements the method `fail_tx_proposal()` on `ClientSessionTracking`
* Invokes `session.fail_tx_proposal(now)` when a tx proposal is found to be invalid.

## Motivation

This will be used to track how many failed tx proposals have been sent recently, so that bad connections can be dropped, to prevent attacks and harmful behavior from buggy clients. See: https://github.com/mobilecoinfoundation/mobilecoin/issues/2977

### Future Work
* Actually drop clients
* Prometheus metrics related to this system.
* Code to determine if a particular client / IP address needs to be blocked temporarily or indefinitely.

PR #3296, opened 2023-03-28
Failure limit on tx proposals

NotGyro:
* Detects when a client has submitted excessive invalid tx proposals within a configurable rolling window
* Adds `tx_failure_window` and `tx_failure_limit` fields to the consensus service config Rust structure (for use with `clap`). 
* Adds `tx_failure_window_seconds` and `tx_failure_limit`, as well as previously-missed `client_tracking_capacity` to the `ConsensusNodeConfig` protobuf, and amends `get_node_config` to add those values to that configuration export.

### Motivation

This will be used to track how many failed tx proposals have been sent recently, so that bad connections can be dropped, to prevent attacks and harmful behavior from buggy clients. See: https://github.com/mobilecoinfoundation/mobilecoin/issues/2977

### Future Work
* Slam-testing to determine reasonable defaults for `tx_failure_window` and `tx_failure_limit`

### DIFF
--- a/consensus/service/src/api/client_api_service.rs
+++ b/consensus/service/src/api/client_api_service.rs
@@ -13,7 +13,7 @@ use crate::{
 use grpcio::{RpcContext, RpcStatus, UnarySink};
 use mc_attest_api::attest::Message;
 use mc_attest_enclave_api::ClientSession;
-use mc_common::{logger::Logger, LruCache};
+use mc_common::{logger::{Logger, log}, LruCache};
 use mc_consensus_api::{
     consensus_client::{ProposeMintConfigTxResponse, ProposeMintTxResponse},
     consensus_client_grpc::ConsensusClientApi,
@@ -140,7 +140,6 @@ impl ClientApiService {
             if let TxManagerError::TransactionValidation(cause) = &err {
                 counters::TX_VALIDATION_ERROR_COUNTER.inc(&format!("{cause:?}"));
 
-                let tracking_window = Duration::from_secs(5); // TODO, placeholder before draft PR gets published.
                 let mut tracker = self.tracked_sessions.lock().expect("Mutex poisoned");
                 let record = if let Some(record) = tracker.get_mut(&session_id) {
                     record
@@ -150,10 +149,17 @@ impl ClientApiService {
                         .get_mut(&session_id)
                         .expect("Adding session-tracking record should be atomic.")
                 };
-                let _recent_failure_count =
-                    record.fail_tx_proposal(&Instant::now(), &tracking_window);
-                // TODO: drop session when recent_failure_count reaches some
-                // number
+                let recent_failure_count =
+                    record.fail_tx_proposal(&Instant::now(), &self.config.tx_failure_window);
+
+                if (recent_failure_count as u32) >= self.config.tx_failure_limit { 
+                    // TODO: Some action to take to counter the harmful traffic
+                    log::warn!(self.logger,
+                        "Client has {} recent failed tx proposals within the last {}",
+                        recent_failure_count,
+                        self.config.tx_failure_window.as_secs_f32()
+                    );
+                }
             }
             err
         })?;
@@ -260,6 +266,10 @@ impl ClientApiService {
         response.set_block_signing_key((&self.enclave.get_signer()?).into());
         response.set_block_version(*self.config.block_version);
         response.set_scp_message_signing_key((&self.config.msg_signer_key.public_key()).into());
+
+        response.set_client_tracking_capacity(self.config.client_tracking_capacity as u64);
+        response.set_tx_failure_window_seconds(self.config.tx_failure_window.as_secs());
+        response.set_tx_failure_limit(self.config.tx_failure_limit);
 
         Ok(response)
     }
--- a/consensus/api/proto/consensus_config.proto
+++ b/consensus/api/proto/consensus_config.proto
@@ -69,4 +69,24 @@ message ConsensusNodeConfig {
 
     // SCP message signing key.
     external.Ed25519Public scp_message_signing_key = 8;
+
+    // Maximum number of client session tracking structures to retain in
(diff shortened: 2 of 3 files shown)


NotGyro:
* **#3296** <a href="https://app.graphite.com/github/pr/mobilecoinfoundation/mobilecoin/3296?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/mobilecoinfoundation/mobilecoin/3296?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3208** <a href="https://app.graphite.com/github/pr/mobilecoinfoundation/mobilecoin/3208?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

nick-mobilecoin (changes requested):
:x: this looks like it might have a merge issue, it seems to be diffing against other changes that are present in the mainline branch `master`.

consensus/service/src/api/client_api_service.rs
```diff
@@ -274,13 +292,54 @@ impl ConsensusClientApi for ClientApiService {
...
+        let session_id = ClientSession::from(msg.channel_id.clone());
+
         {
-            let session = ClientSession::from(msg.channel_id.clone());
             let mut tracker = self.tracked_sessions.lock().expect("Mutex poisoned");
             // Calling get() on the LRU bumps the entry to show up as more
             // recently-used.
-            if tracker.get(&session).is_none() {
-                tracker.put(session, ClientSessionTracking::new());
+            if tracker.get(&session_id).is_none() {
```
nick-mobilecoin, line 301: :wrench: This feels like this block could be its own method similar to the `check_request_chain_id()` guard clause below

consensus/service/src/api/client_api_service.rs
```diff
@@ -308,8 +367,21 @@ impl ConsensusClientApi for ClientApiService {
                     ConsensusGrpcError::NotServing.into()
                 }
             } else {
-                self.handle_proposed_tx(msg)
-                    .or_else(ConsensusGrpcError::into)
+                let result = self.handle_proposed_tx(msg);
+                // The block present below rate-limits suspicious behavior.
+                if let Err(_err) = &result {
+                    let mut tracker = self.tracked_sessions.lock().expect("Mutex poisoned");
```
nick-mobilecoin, line 373: :wrench: I usually lean on the 3 strike rule:
1. you're writing code
2. It might be coincidence that the logic is duplicate
3. You're out, refactor.

This is the 3rd time this patter of `if get() else put() && get()` has been present. This should have method which does this for you. 
Unfortunately due to the mutex this will be a little bit more work as it will need to return a session guard that implements drop, to free up the mutex.

consensus/service/src/api/client_api_service.rs
```diff
@@ -308,8 +367,21 @@ impl ConsensusClientApi for ClientApiService {
...
+                        record
+                    } else {
+                        tracker.put(session_id.clone(), ClientSessionTracking::new());
+                        tracker
+                            .get_mut(&session_id)
+                            .expect("Adding session-tracking record should be atomic.")
+                    };
+                    record.fail_tx_proposal(Instant::now(), self.config.tx_failure_window);
+                }
+                result.or_else(ConsensusGrpcError::into)
```
nick-mobilecoin, line 384: :question: Thoughts on using a match statement here?
Something about seeing is_err logic and then in a separate block modifying that error.

consensus/service/src/api/client_api_service.rs
```diff
@@ -308,8 +367,21 @@ impl ConsensusClientApi for ClientApiService {
                     ConsensusGrpcError::NotServing.into()
                 }
             } else {
-                self.handle_proposed_tx(msg)
-                    .or_else(ConsensusGrpcError::into)
+                let result = self.handle_proposed_tx(msg);
+                // The block present below rate-limits suspicious behavior.
+                if let Err(_err) = &result {
```
nick-mobilecoin, line 372: :wrench: Thinking `is_err()` would work here since the contents don't appear to be looked at
```suggestion
                if result.is_err() {
```

consensus/service/src/api/client_api_service.rs
```diff
@@ -1825,4 +1897,112 @@ mod client_api_tests {
...
+
+        match propose_tx_response {
+            Err(grpcio::Error::RpcFailure(rpc_status)) => {
+                assert_eq!(rpc_status.code(), RpcStatusCode::RESOURCE_EXHAUSTED);
+            }
+            _ => panic!(
+                "Unexpected response upon continuing to use\
+                        a rate-limited session: {propose_tx_response:?}"
+            ),
+        }
```
nick-mobilecoin, line 1997: :wrench: it might lose a little bit of info on failures but perhaps
```suggestion
        assert!(
            matches!(propose_tx_response, Err(grpcio::Error::RpcFailure(failure)) if failure.code() == RpcStatusCode::RESOURCE_EXHAUSTED)
        );

```

consensus/service/src/api/client_api_service.rs
```diff
@@ -1825,4 +1897,112 @@ mod client_api_tests {
...
+
+        let tracker = tracked_sessions
+            .lock()
+            .expect("Attempt to lock session-tracking mutex failed.");
+        // This session should have been dropped at this point.
+        assert_eq!(tracker.len(), 0);
+
+        // Because of the behavior of Mockall, if this returns without calling
+        // client_close() exactly once, it will panic and the test will fail.
```
nick-mobilecoin, line 2006: :wrench: it doesn't seem like this comment is needed since the tracker will close it.
```suggestion
```

Issue #2405, opened 2022-08-15
Fog/Mobilecoind More efficient logging for upgrades.


jgreat:
Right now fog services that use mobilecoind to sync the ledger generate log entries for each block processed. Since the ledger is approaching 1M+ blocks this creates several million log entries on each application instance when we do upgrades or blue/green style deployments.

This puts additional load and on the infrastructure and adds cost to ship all these logs to our logging provider. 

No idea on what the level of effort is for these but, some ideas to reduce volume of logs during startup.

- Reduce level to ERROR during initial sync and reset to LOG_LEVEL when service is healthy.
- Don't log successful archive retrieved for each node, just log once for each block.
- Don't log each node/block when doing roll up blocks just log success on the batch.

jgreat:
The main offender is `Archive block retrieved for Url`. We get a message for each block * 10 nodes.

```
2023-02-15 23:54:55.603855312 UTC INFO Archive block retrieved for Url { scheme: "https", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("bd-mobilecoin-ledger.s3.amazonaws.com")), port: None, path: "/binance.mobilecoin.bdnodes.net/", query: None, fragment: None } 1282651, mc.app: mobilecoind, mc.module: mc_watcher::watcher, mc.src: watcher/src/watcher.rs:191
```

On start up, if we're back filling the ledger, logging messages by batch would be nice.  Maybe 1 message for each 10k blocks processed until we catch up?

jgreat:
These "error" logs from something deeper GRPC/HTTP stack are also unnecessary, any advice on how to mute these would be awesome.

```
E0324 19:08:30.384842702      34 hpack_parser.cc:1216]       Error parsing metadata: error=invalid value key=content-type value=text/html
```

jcape:
@jgreat gRPC errors need to be set using `GRPC_VERBOSITY`, see https://github.com/grpc/grpc/blob/master/TROUBLESHOOTING.md for more details. Otherwise, fixed in #3301

PR #3301, opened 2023-03-31
Update log messages

varsha888:
https://github.com/mobilecoinfoundation/mobilecoin/issues/2405
=> Changed log levels
=> Changed log messages
=> Set get block contents max sleep time to 60 seconds

### DIFF
--- a/ledger/sync/src/ledger_sync/ledger_sync_service.rs
+++ b/ledger/sync/src/ledger_sync/ledger_sync_service.rs
@@ -25,6 +25,7 @@ use mc_util_telemetry::{
 use mc_util_uri::ConnectionUri;
 use retry::delay::Fibonacci;
 use std::{
+    cmp::min,
     collections::{BTreeMap, HashMap, HashSet},
     sync::{Arc, Condvar, Mutex},
     thread,
@@ -40,6 +41,8 @@ const MAX_CONCURRENT_GET_BLOCK_CONTENTS_CALLS: usize = 50;
 
 /// Telemetry metadata: number of blocks appended to the local ledger.
 const TELEMETRY_NUM_BLOCKS_APPENDED: Key = telemetry_static_key!("num-blocks-appended");
+
+const MAX_SLEEP_INTERVAL: Duration = Duration::from_secs(60);
 
 pub struct LedgerSyncService<
     L: Ledger,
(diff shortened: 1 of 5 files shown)


watcher/src/watcher_db.rs
```diff
@@ -456,7 +456,11 @@ impl WatcherDB {
                 Ok((ts, res)) => match res {
                     TimestampResultCode::WatcherBehind => {
                         if watcher_behind_timer.elapsed() > watcher_timeout {
-                            log::warn!(self.logger, "watcher is still behind on block index = {} after waiting {} seconds, caller will be blocked", block_index, watcher_timeout.as_secs());
+                            log::warn!(
+                                self.logger,
+                                "Caller will be blocked in {} seconds",
```
jgreat, line 461: Do we lose some context there about why "Caller will be blocked?"  I think the original message seems to have more human context.

jgreat (commented):
I don't see a change for the "Archive block retrieved for Url" message?  Is that going to be in a different PR?

ledger/sync/src/ledger_sync/ledger_sync_service.rs
```diff
@@ -719,7 +721,12 @@ fn get_block_contents<TF: TransactionsFetcher + 'static>(
                                     // Sleep, with a linearly increasing delay. This prevents
                                     // endless retries
                                     // as long as the deadline is not exceeded.
-                                    thread::sleep(Duration::from_secs(num_attempts + 1));
+                                    let interval = if num_attempts + 1 < MAX_SLEEP_INTERVAL {
+                                        num_attempts + 1
+                                    } else {
+                                        MAX_SLEEP_INTERVAL
+                                    };
+                                    thread::sleep(Duration::from_secs(interval));
```
samdealy, line 729:
```suggestion
let next_interval = Duration::secs_from(num_attempts + 1);
let interval = std::cmp::min(next_interval, &MAX_SLEEP_INTERVAL)
thread::sleep(interval);
```

samdealy (changes requested):

watcher/src/watcher_db.rs
varsha888: I'll change it back

jcape (approved):

PR #3307, opened 2023-04-09
refactor mobilecoind api to not depend on consensus-api

cbeck88:
I am having a lot of build problems in another project due to mc-mobilecoind-api pulling in mbedtls, which exports some symbols that conflicts with another C library that some rust code depends on.

(A grpcio thread crashes, causing the next thread to make a grpc request to hang:

```
023-04-09T22:21:35.457700Z  INFO mobilecoind_buddy::worker: check balance: 1
2023-04-09T22:21:35.458247Z  INFO mobilecoind_buddy::worker: check balance: 0
[New Thread 0x7ffff4c30640 (LWP 86787)]
thread '<unnamed>' panicked at 'Called self-test rand without enabling self-test', /home/chris/.cargo/git/checkouts/rust-mbedtls-12ed44423be981f0/ac6ee17/mbedtls/src/self_test.rs:52:20
stack backtrace:
[New Thread 0x7fffdca49640 (LWP 86788)]
[New Thread 0x7fffd100d640 (LWP 86789)]
[New Thread 0x7fffd080c640 (LWP 86790)]
[New Thread 0x7fffca690640 (LWP 86791)]
[New Thread 0x7fffc9e8f640 (LWP 86792)]
   0: rust_begin_unwind
             at /rustc/a266f11990d9544ee408e213e1eec8cc9eb032b7/library/std/src/panicking.rs:579:5
   1: core::panicking::panic_fmt
             at /rustc/a266f11990d9544ee408e213e1eec8cc9eb032b7/library/core/src/panicking.rs:64:14
   2: core::panicking::panic_display
             at /rustc/a266f11990d9544ee408e213e1eec8cc9eb032b7/library/core/src/panicking.rs:147:5
   3: core::panicking::panic_str
             at /rustc/a266f11990d9544ee408e213e1eec8cc9eb032b7/library/core/src/panicking.rs:131:5
   4: core::option::expect_failed
             at /rustc/a266f11990d9544ee408e213e1eec8cc9eb032b7/library/core/src/option.rs:2055:5
   5: rand
   6: _ZN9grpc_core12BdpEstimator12CompletePingEv
   7: _ZL22finish_bdp_ping_lockedPvP10grpc_error
   8: _Z31grpc_combiner_continue_exec_ctxv
   9: _ZN9grpc_core7ExecCtx5FlushEv
  10: _ZL10end_workerP12grpc_pollsetP19grpc_pollset_workerPS2_
  11: _ZL12pollset_workP12grpc_pollsetPP19grpc_pollset_workerN9grpc_core9TimestampE
  12: _ZL7cq_nextP21grpc_completion_queue12gpr_timespecPv
  13: grpcio::env::poll_queue
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
[New Thread 0x7fffc968e640 (LWP 86793)]
[Thread 0x7ffff6242640 (LWP 86784) exited]
[New Thread 0x7fffc8e8d640 (LWP 86794)]
2023-04-09T22:21:35.558976Z  INFO mobilecoind_buddy::worker: worker polling loop
2023-04-09T22:21:35.559023Z  INFO mobilecoind_buddy::worker: check ledger status
```
)

I determined that the only reason I have mbedtls in my build is that `mobilecoind-api` depends on `consensus-api` and this pulls in `mc-attest-core` etc.

This change allows me to break this dependency.

---

I think this change is good from a separation of concerns point of view anyway. Consensus should be able to evolve its API without worrying if it's breaking mobilecoind users.


mobilecoind/api/proto/mobilecoind_api.proto
```diff
@@ -961,10 +960,28 @@ message GetNetworkStatusResponse {
...
     bool is_behind = 4;
 
-    // The latest block info reported by a consensus node
-    consensus_common.LastBlockInfoResponse last_block_info = 5;
+    // The latest block info data reported by a consensus node
+    LastBlockInfo last_block_info = 5;
 }
 
+// Data about the network state and last block processed by the consensus network
+message LastBlockInfo {
```
nick-mobilecoin, line 968: :question: Double checking my understanding from this comment
> I think this change is good from a separation of concerns point of view anyway. Consensus should be able to evolve its API without worrying if it's breaking mobilecoind users.

The `LastBlockInfo` was being reused from consensus due to convenience and not because they needed to stay n'sync?
This suggests there is little to no benefit to trying to place the `LastBlockInfo` in a more common proto like `blockchain`.

mobilecoind/api/proto/mobilecoind_api.proto
```diff
@@ -961,10 +960,28 @@ message GetNetworkStatusResponse {
...
+    // Block index
```
nick-mobilecoin, line 969: :wrench: It looks like `2` was intentionally dropped when this got brought in here. I'm thinking it should be marked reserved to prevent accidental adding back in, https://protobuf.dev/programming-guides/proto3/#reserved
```suggestion
    reserved 2;
    
    // Block index
```

mobilecoind/src/service.rs
```diff
@@ -2158,6 +2158,17 @@ impl<T: BlockchainConnection + UserTxConnection + 'static, FPR: FogPubkeyResolve
                 RpcStatus::with_message(RpcStatusCode::INTERNAL, "no peers reachable".to_owned())
             })?;
 
+        let mut mcd_last_block_info = api::LastBlockInfo::new();
+        mcd_last_block_info.set_index(last_block_info.block_index);
```
nick-mobilecoin, line 2162: :thought_balloon: The way this `last_block_info` was derived seems to deviate from the previous use of `LastBlockInfo` being what consensus said it was, thus further justifying a separate one for mobilecoind

nick-mobilecoin (approved):
Looks good as long as we're OK with the two `LastBlockInfo` versions possibly deviating.

This _specific_ dependency issue will hopefully go away with https://github.com/mobilecoinfoundation/mcips/pull/64, but that's a ways out for this current problem

mobilecoind/api/proto/mobilecoind_api.proto
cbeck88: I think this would be a breaking change to mobilecoind api, so we shouldn't do it. There are python scripts that use last block info to try to figure out what the network block version is

cbeck88:
I dont think this struct is exposed through mobilecoind-json, we generally dont add new features to that.

jcape:
Deleting deprecated fields and methods is always a breaking change :-)

wjuan-mob (approved):

mobilecoind/api/proto/mobilecoind_api.proto
cbeck88: i decided to just add the deprecated field back as it was, there's no good reason to remove it in this change.

jcape (approved):

mobilecoind/api/proto/mobilecoind_api.proto
cbeck88: yeah i think that's right, it was just a pass through, but i wasn't sure if that was a good idea at the time.

PR #3313, opened 2023-04-11
make mobilecoind expose the chain-id via rpc

cbeck88:


nick-mobilecoin (approved):
LGTM

I was expecting the chain id to be some kind of network query, but it looks like the client sends the chain id they want to use in the message(s) and the network rejects the messages if the chain id doesn't match.

cbeck88:
we want to do chain ids on chain, but we haven't gotten there yet

eranrund (approved):

wjuan-mob (approved):

PR #3315, opened 2023-04-17
Add par_iter to try to improve performance

wjuan-mob:
Use a par_iter to filter the tx_outs to attempt to improve performance.
Add comments explaining the code.

### Motivation

Mobilecoind is slow to sync and is inhibiting development on it.
A similar change has been made in full service and should be ported over.

### Future Work
Make the unit test into a proper benchmark.
Port other full service related improvements to mobilecoind, centralizing the code as much as possible to make it easier to keep them in sync.


cbeck88:
Can you try to measure the actual performance difference? Like, how long does it take to scan testnet or mainnet on a new account?

cbeck88:
Even a rough estimate, like, 1 hour vs. > 3 hours, would be helpful. If there's no difference then that would be surprising and may mean there's something wrong with the PR

cbeck88 (approved):

mobilecoind/src/sync.rs
```diff
@@ -603,6 +615,193 @@ mod test {
...
+            .get_utxos_for_subaddress(&monitor_id, DEFAULT_SUBADDRESS_INDEX)
+            .unwrap();
+        assert_eq!(utxos.len(), 0);
+
+        let monitor_data = mobilecoind_db.get_monitor_data(&monitor_id).unwrap();
+        assert_eq!(monitor_data.next_block, 0);
+
+        // Process the first MAX_BLOCKS_PROCESSING_CHUNK_SIZE blocks.
+        let result = sync_monitor(&ledger_db, &mobilecoind_db, &monitor_id, &logger).unwrap();
+        assert_eq!(result, SyncMonitorOk::MoreBlocksPotentiallyAvailable);
```
samdealy, line 684: This logic is reused multiple times. It might make sense to factor it out into a helper method in a future PR.

mobilecoind/src/sync.rs
```diff
@@ -603,6 +615,193 @@ mod test {
...
+                .add_output(Amount::new(10, Mob::ID), &receiver.subaddress(i), &mut rng)
+                .unwrap();
+            tx_outs.push(tx_out);
+        }
+        let start = Instant::now();
+
+        add_txos_to_ledger(&mut ledger_db, BlockVersion::MAX, &tx_outs, &mut rng).unwrap();
+
+        let add_txos_time = start.elapsed();
+        log::info!(logger, "add_txos_to_ledger took {:?}", add_txos_time);
```
samdealy, line 794:
```suggestion
        log::info!(logger, "add_txos_to_ledger took {add_txos_time:?}");
```
(I'm surprised clippy doesn't catch this)

mobilecoind/src/sync.rs
```diff
@@ -603,6 +615,193 @@ mod test {
...
+
+        let start = Instant::now();
+
+        let result = sync_monitor(&ledger_db, &mobilecoind_db, &monitor_id, &logger).unwrap();
+
+        let sync_monitor_time = start.elapsed();
+        log::info!(logger, "sync_monitor took {:?}", sync_monitor_time);
```
samdealy, line 801:
```suggestion
        log::info!(logger, "sync_monitor took {sync_monitor_time:?}");
```

mobilecoind/src/sync.rs
```diff
@@ -345,69 +346,77 @@ fn match_tx_outs_into_utxos(
...
+            // Generate the shared secret between the subaddress and output public key.
+            let shared_secret =
+                get_tx_out_shared_secret(account_key.view_private_key(), &tx_public_key);
+
+            // Get the amount and blinding factor for the output.
+            let (amount, _blinding) = tx_out
+                .get_masked_amount()
+                .expect("missing masked amount")
+                .get_value(&shared_secret)
+                .expect("Malformed amount"); // TODO
```
samdealy, line 393: What's the TODO here? Can we also put it above the line rather than to the right if it's still needed?

mobilecoind/src/sync.rs
```diff
@@ -603,6 +615,193 @@ mod test {
...
+        let account_keys: Vec<_> = (0..5).map(|_i| AccountKey::random(&mut rng)).collect();
+
+        let data = MonitorData::new(
+            account_keys[0].clone(),
+            DEFAULT_SUBADDRESS_INDEX, // first subaddress
+            5,                        // number of subaddresses
+            0,                        // first block
+            "",                       // name
+        )
+        .unwrap();
```
samdealy, line 632:
```suggestion
        let first_subaddress = DEFAULT_SUBADDRESS_INDEX;
        let subaddress_count = 5;
        let first_block_index = 0;
        let name = "";
        let data = MonitorData::new(
            account_keys[0].clone(),
            first_subaddress,
            subaddress_count,
            first_block_index,
            name,
        )
        .unwrap();
```
I think it's usually better to use descriptive variable names rather than comments because the latter become stale more easily.

samdealy (approved):

Issue #3281, opened 2023-03-24
V4.1 - fog missing Prometheus metric duration buckets for ledger and view paths


jgreat:
Looks like duration_count buckets are missing some entries in v4.1. I see a `ledger_server_grpc_duration_count`, but its only showing admin and healthcheck paths.

v4.0 shows timing buckets for:

- fog_ledger.FogBlockAPI.GetBlocks
- fog_ledger.FogMerkleProofAPI.GetOputputs
- fog_ledger.FogKeyImageAPI.Auth
- fog_ledger.FogUntrustedTxOutApi.GetTxOuts
- fog_ledger.FogKeyImageAPI.CheckKeyImages
- fog_ledger.FogMerkleProofAPI.Auth
- fog_view.FogViewAPI.Auth
- fog_view.FogViewAPI.Query 

v4.0 example of ledger in grafana:

![image](https://user-images.githubusercontent.com/5324639/227390439-56640708-93c1-44de-96e0-389b34a6666c.png)

varsha888:
https://github.com/mobilecoinfoundation/mobilecoin/pull/3322

PR #3322, opened 2023-04-20
Prometheus edit service names

varsha888:
[Issue](https://github.com/mobilecoinfoundation/mobilecoin/issues/3281)

This issue was because fog service names were same for OpMetrics and ServiceMetrics. 
Consensus service has different names for the two metrics. 
This PR edits fog service names to be consistent with "consensus_service" in Prometheus. 

Screen Shot 2023-04-20 at 1 02 29 PM
Screen Shot 2023-04-20 at 1 09 18 PM

### DIFF
--- a/fog/ingest/server/src/lib.rs
+++ b/fog/ingest/server/src/lib.rs
@@ -36,7 +36,7 @@ use mc_util_metrics::ServiceMetrics;
 
 lazy_static::lazy_static! {
     /// Generates service metrics for tracking
-    pub static ref SVC_COUNTERS: ServiceMetrics = ServiceMetrics::new_and_registered("fog_ingest");
+    pub static ref SVC_COUNTERS: ServiceMetrics = ServiceMetrics::new_and_registered("fog_ingest_service");
 }
 
 // Helper to format a sequence as a comma-separated list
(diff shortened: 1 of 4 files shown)


fog/ingest/server/src/lib.rs
```diff
@@ -36,7 +36,7 @@ use mc_util_metrics::ServiceMetrics;
 
 lazy_static::lazy_static! {
     /// Generates service metrics for tracking
-    pub static ref SVC_COUNTERS: ServiceMetrics = ServiceMetrics::new_and_registered("fog_ingest");
+    pub static ref SVC_COUNTERS: ServiceMetrics = ServiceMetrics::new_and_registered("fog_ingest_service");
```
nick-mobilecoin, line 39: :question: Making sure I understand this change. The previous name "fog_ingest" name collided with `fog/ingest/server/src/counters.rs`
```rust
    pub static ref OP_COUNTERS: OpMetrics = OpMetrics::new_and_registered("fog_ingest");
```
resulting in the same counter being used for both and thus both metrics in the `fog_ingest` namespace.

nick-mobilecoin (approved):

fog/ingest/server/src/lib.rs
varsha888: When the name collided the OpMetrics would show up but not the ServiceMetrics

jgreat (approved):
I see duration_bucket metrics for the view and ledger API "methods". LGTM

awygle (approved):

PR #3328, opened 2023-04-21
Change log level to debug for watcher

varsha888:
[Issue](https://github.com/mobilecoinfoundation/mobilecoin/issues/2405 ) 

Changed the log level to debug because of https://github.com/mobilecoinfoundation/mobilecoin/issues/2405#issuecomment-1432279302

### DIFF
--- a/watcher/src/watcher.rs
+++ b/watcher/src/watcher.rs
@@ -197,7 +197,7 @@ impl Watcher {
             for (src_url, (block_index, block_data_result)) in url_to_block_data_result {
                 match block_data_result {
                     Ok(block_data) => {
-                        log::info!(
+                        log::debug!(
                             self.logger,
                             "Archive block retrieved for {} {}",
                             src_url,


watcher/src/watcher.rs
```diff
@@ -197,7 +197,7 @@ impl Watcher {
             for (src_url, (block_index, block_data_result)) in url_to_block_data_result {
                 match block_data_result {
                     Ok(block_data) => {
-                        log::info!(
+                        log::debug!(
```
nick-mobilecoin, line 200: :question: Want to make sure we don't want to batch up like mentioned in https://github.com/mobilecoinfoundation/mobilecoin/issues/2405#issuecomment-1432279302 and turning these _off_ for normal use case is desired.

nick-mobilecoin (approved):

watcher/src/watcher.rs
varsha888: Batching is for later and this PR is for 5.0 release

PR #3332, opened 2023-04-25
Feature/router deployment cleanup

joekottke:
* Update names and labels to reference fog-{view,ledger}-{router,store}
* Cleaned up port names
* on-disk filenames of supervisord config files
* Manifest filenames
* blockHeightRetrieval for dev networks

### Motivation

We needed some cleanup after merging the routers into the master branch.  This solves that.

### Future Work

There will probably be more work as we test and verify the systems.


.internal-ci/helm/fog-services/templates/fog-view-grpc-ingress.yaml
```diff
@@ -25,6 +25,6 @@ spec:
         pathType: Prefix
         backend:
           service:
-            name: {{ include "fogServices.fullname" . }}-fog-view
+            name: {{ include "fogServices.fullname" . }}-fog-view-router
```
jgreat, line 28: FWIW, I think this will break the ingress connection to the existing "loaded" view instances when applied.

.internal-ci/helm/fog-services/templates/fog-view-rollout-cr.yaml
```diff
@@ -192,21 +193,22 @@ spec:
...
           topologySpreadConstraints:
           - topologyKey: topology.kubernetes.io/zone
             maxSkew: 1
             whenUnsatisfiable: DoNotSchedule
             labelSelector:
               matchLabels:
-                app: fog-view
+                app: fog-view-store
                 helm.sh/chart: {{ include "fogServices.chart" . }}
                 {{- include "fogServices.selectorLabels" . | nindent 16 }}
```
jgreat, line 205: I think this is correct, we want to make sure the store replicas are distributed across zones for availability.

jgreat (commented):
Some gotchas on moving the ingress to point to fog-view-router and fog-ledger-router. 

### Follow ups: 

Standardize and fix binary names
- use `-` not `_` in binary names.
- rename `key-image-store` to `ledger-store` or change all `ledger` references to `key-image`

.internal-ci/helm/fog-services/templates/fog-view-grpc-ingress.yaml
joekottke: Yes, this is where we will have to deploy the stand-alone PREVIOUS ledger and view to match the selector within that service name.  That will allow access during the upgrade, with no downtime.

.internal-ci/helm/fog-services/templates/fog-view-rollout-cr.yaml
joekottke: Those specifics are more for review by @aweis89 . I'm just cleaning up ports, names, etc.

nick-mobilecoin (approved):

PR #3335, opened 2023-04-27
Add disk persistence for fog-ledger-router, shard size default updates

joekottke:
* Add disk persistence for fog-ledger-store
* Double default shard sizes for ledger and view stores
* Change default of fog-ledger-router replicaCount to 2, matchine fog-view-router

### Motivation

The fog-ledger-router needs disk persistence for storing ledger and watcher databases.  Otherwise a pod restart causes a re-download of the full blockchain.

The shard size defaults were increased.

### Future Work

* Possible further adjustments to shard sizes

### DIFF
--- a/.internal-ci/helm/fog-services/templates/fog-ledger-rollout-cr.yaml
+++ b/.internal-ci/helm/fog-services/templates/fog-ledger-rollout-cr.yaml
@@ -118,6 +118,8 @@ spec:
                 failureThreshold: 2
                 periodSeconds: 10
               volumeMounts:
+              - name: fog-data
+                mountPath: /fog-data
               - name: aesm-socket-dir
                 mountPath: /var/run/aesmd
               - name: supervisor-conf
@@ -182,6 +184,10 @@ spec:
             affinity:
               {{- toYaml .Values.fogLedgerRouter.affinity | nindent 14 }}
             volumes:
+            {{- if eq .Values.fogLedger.persistence.enabled false }}
+            - name: fog-data
+              emptyDir: {}
+            {{- end }}
             - name: aesm-socket-dir
               emptyDir: {}
             - name: supervisor-conf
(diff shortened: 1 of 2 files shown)


.internal-ci/helm/fog-services/templates/fog-ledger-rollout-cr.yaml
```diff
@@ -182,6 +184,10 @@ spec:
             affinity:
               {{- toYaml .Values.fogLedgerRouter.affinity | nindent 14 }}
             volumes:
+            {{- if eq .Values.fogLedger.persistence.enabled false }}
+            - name: fog-data
+              emptyDir: {}
+            {{- end }}
```
nick-mobilecoin, line 190: :question: I'm understanding this to be an emptyDir when persistence isn't enabled, and per the docs, https://kubernetes.io/docs/concepts/storage/volumes/#emptydir it will go away on pod deletion.

When persistence is enabled then the VolumeClaimTemplate is used, https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#volume-claim-templates

nick-mobilecoin (approved):

.internal-ci/helm/fog-services/templates/fog-ledger-rollout-cr.yaml
joekottke: Thats correct.  There is a slight variation that with an `emptyDir {}`, if a container restarts on the same machine, the data from that `emptyDir` will remain.  If the data was just written inside the container (without and `emptyDir` volume), a restart would blow away that data.

jgreat: correct.  emptyDir will create the temp volume so the app can write to the directory but the data will not survive a pod delete or host move.

jgreat (approved):

joekottke:
Successful CD was run here:  https://github.com/mobilecoinfoundation/mobilecoin/actions/runs/4815541639

PR #3337, opened 2023-04-27
Update `local_network.py` to have a default launch

nick-mobilecoin:
Previously the `local_network.py` script required manually bootstrapping
a ledger and specifying a network type. Now the `local_network.py`
script will bootstrap the ledger, unless the `--no-bootstrap-ledger`
option is provided. The network type defaults to the `dense5` network.

### Future Work

[Soundtrack of this PR]()

### DIFF
--- a/mint-auditor/README.md
+++ b/mint-auditor/README.md
@@ -55,9 +55,7 @@ To run the integration tests, which live in `mint-auditor/tests/`, there are fou
 ```
 $ cargo build --release
 $ export MC_LOG=info
-$ export LEDGER_BASE=$PWD/target/sample_data/ledger
-$ ./tools/local-network/bootstrap.sh
-$ ./tools/local-network/local_network.py --network-type dense5 --skip-build &
+$ ./tools/local-network/local_network.py --no-build &
 ```
 
 2. Generate and authorize minters.
--- a/mobilecoind-dev-faucet/README.md
+++ b/mobilecoind-dev-faucet/README.md
@@ -108,9 +108,7 @@ It is relatively straightforward to test the faucet locally using the `tools/loc
 ```
 $ cargo build --release
 $ export MC_LOG=info,rocket=error
-$ export LEDGER_BASE=$PWD/target/sample_data/ledger
-$ ./tools/local-network/bootstrap.sh
-$ ./tools/local-network/local_network.py --network-type dense5 --skip-build &
+$ ./tools/local-network/local_network.py --no-build &
 ```
 
 Then, start a faucet and set it to also work in the background:
--- a/tools/fog-local-network/README.md
+++ b/tools/fog-local-network/README.md
@@ -55,7 +55,7 @@ In order to use it, the following steps are necessary.
     LEDGER_BASE=$(pwd)/ledger \
     IAS_SPID="..." \
     IAS_API_KEY="..." \
-    python3 ../tools/fog-local-network/fog_local_network.py --network-type dense5
+    python3 ../tools/fog-local-network/fog_local_network.py
     ```
     Note that all of the above arguments are identical to the mobilecoin local_network.py script.
     The script is known to work with python 3.6 or later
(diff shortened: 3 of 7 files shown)


nick-mobilecoin:
* master
  * **PR #3337** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3337" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3337?utm_source=stack-comment).

varsha888 (commented):
Looks good. This would also need the README to be updated for fog_local_network and local_network.

nick-mobilecoin:
> Looks good. This would also need the README to be updated for fog_local_network and local_network.

Updated the READMEs in https://github.com/mobilecoinfoundation/mobilecoin/pull/3337/commits/9e47ab2b51df6b694db9c5ba10a936022c0edc79

varsha888 (approved):

cbeck88:
@nick-mobilecoin i am worried that this can break some workflows where you take down a network and bring it back up, maybe we can make this more intelligent? like if no ledger base files are found, then it auto bootstraps, otherwise it does not auto bootstrap?

nick-mobilecoin:
> @nick-mobilecoin i am worried that this can break some workflows where you take down a network and bring it back up, maybe we can make this more intelligent? like if no ledger base files are found, then it auto bootstraps, otherwise it does not auto bootstrap?

I may need to understand the bootstrap better. I did not consider that case, but read this comment https://github.com/mobilecoinfoundation/mobilecoin/blob/nick/local-network/tools/local-network/bootstrap.sh#LL13C1-L16C59
> The tool attempts to avoid rebootstrapping if it doesn't appear necessary,
> which means:
> - target/sample_data already exists (and conf.json exists)
> - conf.json has the same parameters as the user requests

cbeck88:
> I may need to understand the bootstrap better. I did not consider that case, but read this comment https://github.com/mobilecoinfoundation/mobilecoin/blob/nick/local-network/tools/local-network/bootstrap.sh#LL13C1-L16C59

Right I forgot I added that a long time ago.

So I want to understand I guess, exactly how the workflow is going to work.

So an example workflow I've used has been like, bootstrap a network, run the local network at one block version, do some transactions, then take down the network and start it again at a later block version and see what happens.

Or, doing this using the mint auditor tool to try to test that that thing is working.

It's important for workflows like this that the ledger being used by the local network script is not being wiped out each time.

---

The other thing is, I believe that `bootstrap.sh` does this whole `cargo run -p bootstrap` business, and this is fairly annoying because it causes a bunch of common packages to get rebuilt with different cargo features, and then the local network script will, by default build all the software for the network as well. So making bootstrap run every time will cause a bunch of stuff to get built and then rebuilt unnecessarily, even if the bootstrap binary itself will know by the time it is built not to actually wipe out the network.

---

I guess my larger question is, what is wrong with making the user of the script explicitly bootstrap? Given that it's a one-time thing it's not that onerous and it makes it easier to reason about what is happening.

nick-mobilecoin:
I think with our current desire to reduce scope and maintain focus this should PR should probably get closed. 
Prior to our refocusing efforts the ROI was already questionable.

The main intent of this change was to improve the ergonomics of these utilities for developers new to the code base.

> I guess my larger question is, what is wrong with making the user of the script explicitly bootstrap? Given that it's a one-time thing it's not that onerous and it makes it easier to reason about what is happening.

When one is a new developer, bootstrapping is one more step that the new developer has to stop and try to figure out the what and why of. There are a few assumptions about the environment this scripts are run in. When things don't work due to these assumptions the developer is having to look at each step and try to figure out where they might have done it wrong.

> I see no reason to rename things like --skip-build to --no-build, what's the motivation given that it may break existing scripts and instructions?

That was probably a feature creep when I added the default for network type and the bootstrap arg. In general CLI's generally use "--some-thing" and "--no-some-thing" for the args, python 3.9 and above has a dedicated action for the pair [BooleanOptionalAction](https://docs.python.org/3/library/argparse.html#action).

nick-mobilecoin:
close in favor of higher priority tasks

PR #3340, opened 2023-04-28
fix: use 2 replicas for shards

aweis89:
### Future Work

[Soundtrack of this PR]()


joekottke:
Can you remove the settings for the values in the `mobilecoin-workflow-dev-deploy.yaml` that sets things to 2 replicas, now that we have the defaults set to 2 as well?

joekottke (approved):

nick-mobilecoin (approved):

PR #3341, opened 2023-05-01
Add Payment ID to request protobuf

the-real-adammork:
### Motivation

Asked for by Alex implicitly in this design doc - https://docs.google.com/document/d/1pfA5GJPqpP1idJFAqeI2wz5tYWQ74k1VUUskN-r4PU4/edit#heading=h.h9ex6368ccr4  - to help w/ Moby Payment Requests.

### DIFF
--- a/api/proto/printable.proto
+++ b/api/proto/printable.proto
@@ -23,6 +23,9 @@ message PaymentRequest {
 
     /// Token id to transact in.
     uint64 token_id = 4;
+
+    /// Payment Id
+    uint64 payment_id = 5;
 }
 
 /// Message encoding a private key and a UTXO, for the purpose of


the-real-adammork:
small change, tests now passing

api/proto/printable.proto
```diff
@@ -23,6 +23,9 @@ message PaymentRequest {
 
     /// Token id to transact in.
     uint64 token_id = 4;
+
+    /// Payment Id
+    uint64 payment_id = 5;
```
nick-mobilecoin, line 28: :question: I'm guessing from the rest of this proto that using `uint64` over `fixed64` isn't a concern of leaking information via message size?

nick-mobilecoin (approved):

api/proto/printable.proto
the-real-adammork: whats the reason to use fixed64 ? Im not familiar with that type.

nick-mobilecoin: fixed 64 will always be eight bytes in the proto message, while uint64 will use as few bytes as possible to represent the value.
My understanding is that the consistent(fixed) sizes can be used inside of encrypted messages to avoid leaking possible size ranges, if the information should be kept private.
Based on the other fields in the message using `fixed` here probably won't gain anything.

the-real-adammork: ok, fixed sounds better, but we should change all of them, and then thats a higher impact change. Lets keep as UInt64.

voloshyn (approved):

PR #3348, opened 2023-05-03
bootstrap mobilecoind with preloaded ledger and watcher db

jgreat:
### Motivation

Bootstrap fresh mobilecoind instances by downloading a preloaded ledger and watcher db from a local url instead of individual blocks and making millions of consensus calls. 

This saves quite a bit of network transit costs (5GB ledger, 6GB watcher) and speeds up loading to 2-3 minutes vs 10-20 minutes from S3+consensus.

- Set `MC_LEDGER_DB_URL` and `MC_WATCHER_DB_URL` env variables local network urls that store the data.mdb files.
- Add entrypoints and startup scripts to mobilecoind, fogingest and fogledger containers.
- Add configmap and notes in helm chart values for fog-ingest, fog-services and mobilecoind.

### Future Work

- GHA Scheduled cron job to update stored copies of the ledgers.
- Azure Blob Storage accounts with replication to testnet and mainnet fog subscriptions has already been set up.


.internal-ci/docker/entrypoints/fog-ledger.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+
+# Copyright (c) 2018-2022 The MobileCoin Foundation
+
+set -eo pipefail
+
+data="/fog-data"
+
+if [[ -n "${MC_LEDGER_DB_URL}" ]]
```
nick-mobilecoin, line 9: :shrug: I generally lean on the 3 strike rule. first implementation is needed, second instance of the same implementation you may not know what's the same, third instance it's probably duplicate.

The current implementation may just be the nature of bash and docker, but I wonder if there should be a function(s) defined in a common bash script that the entrypoints call into. `restore_ledger_db()` and `restore_watcher_db()`.

IDK know this needs to change mainly just thinking/asking out loud

.internal-ci/helm/fog-ingest-config/values.yaml
```diff
@@ -27,3 +27,6 @@ fogIngest:
   POSTGRES_CONNECTION_TIMEOUT: '5'
   POSTGRES_MAX_CONNECTIONS: '3'
   FOG_PUBKEY_EXPIRY_WINDOW: '10'
+  # Bootstrap ledger/watcher db from public bucket
+  # MC_LEDGER_DB_URL: ''
+  # MC_WATCHER_DB_URL: ''
```
nick-mobilecoin, line 32: It may be how you all are evolving the helm stuff, but in general I lean towards not committing commented out code, and not adding commented out code. 
```suggestion
```

nick-mobilecoin (approved):

.internal-ci/docker/entrypoints/fog-ledger.sh
jgreat: I agree with the general principal, but each one of these entrypoints is for a separate instance that gets installed in a separate container. There's also the difference of how mobilecoind and the storage is implemented in each instance. So creating a standardized module (sourced file) is a much bigger project.

The whole set of containers and charts needs a refactor to clean up, standardize and utilize lessons learned.  

Also each time I try to set up abstractions for stuff like this, I get complaints about complexity... so its hard to win this one :shrug:

.internal-ci/helm/fog-ingest-config/values.yaml
jgreat: Normally I would agree, but the values.yaml file tends to serve as documentation and examples.  See something like https://github.com/goharbor/harbor-helm/blob/master/values.yaml on how the values are commented and explained.

.internal-ci/docker/entrypoints/fogingest.sh
```diff
@@ -0,0 +1,35 @@
...
+then
+    ### CBB: these should use ENV vars for configuration.
+    #   Need to fix .internal-ci/helm/mc-core-common-config/templates/mobilecoind-supervisord-mobilecoind-configmap.yaml
+    echo "MC_LEDGER_DB_URL set, restoring ${data}/ledger/data.mdb from backup"
+    if [[ -f "${data}/ledger/data.mdb" ]]
+    then
+        echo "Found existing ledger database, skipping download"
+    else
+        echo "Downloading ledger data.mdb"
+        curl -L "${MC_LEDGER_DB_URL}" -o "${data}/ledger/data.mdb"
```
joekottke, line 19: What happens if this `curl` fails?  Does the entrypoint crash?  I guess a way to test it is to set a known invalid `${MC_LEDGER_DB_URL}` and confirm what happens.

.internal-ci/docker/entrypoints/fogingest.sh
```diff
@@ -0,0 +1,35 @@
...
+
+if [[ -n "${MC_WATCHER_DB_URL}" ]]
+then
+    echo "MC_WATCHER_DB_URL set, restoring ${data}/watcher/data.mdb from backup"
+    if [[ -f "${data}/watcher/data.mdb" ]]
+    then
+        echo "Found existing watcher database, skipping download"
+    else
+        echo "Downloading watcher data.mdb"
+        curl -L "${MC_WATCHER_DB_URL}" -o "${data}/watcher/data.mdb"
```
joekottke, line 31: Same question about this `curl` as above.

joekottke (changes requested):

.internal-ci/docker/entrypoints/fogingest.sh
jgreat: If the curl fails the entrypoint exits error and the pod will error.  I think this is the behavior we want.  At this point we would want a human to fix the url var or remove the value.

If the pod already has a database, the curl doesn't happen.  So the error condition should only happen on a fresh new deployment without an existing .mdb file if the URL var is set.

joekottke (approved):

PR #3375, opened 2023-06-02
Change from passing verifier to passing measurements

nick-mobilecoin:
Previously many interfaces that wished to attest to an enclave would
create and pass a verifier down to the logic that would eventually try
and verify the enclave's report.
Now these interfaces pass in a `Vec<TrustedMeasurement>` with all of the
measurements they wish to verify against. This isolates the logic for
how an enclave report is verified to support different verification
methods based on EPID vs DCAP.


nick-mobilecoin:
* master
  * **PR #3369** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3369" target="_blank">Graphite</a> 
    * **PR #3375** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3375" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3375?utm_source=stack-comment).

fog/ledger/server/tests/store.rs
```diff
@@ -242,7 +241,7 @@ pub fn direct_key_image_store_check(logger: Logger) {
     // AuthResponseOutput
     let auth_message = attest::AuthMessage::from(client_auth_response);
     // Initiator accepts responder's message.
-    let auth_response_event = AuthResponseInput::new(auth_message.into(), Verifier::default());
```
eranrund, line 245: I understand why this works, but I find it somewhat counter-intuitive. A default verifier letting everything verify makes sense to me, however my expectation would be that if an empty array of measurements results in the measurement obtained from the enclave not matching anything in the (empty) array and the verification will fail. I guess it's just weird to me that some methods, when called, force a verification of some sort (all of them, I think?) but the one that receives an array essentially doesn't do anything if the array is empty.
I am not sure it's worth addressing or what a good solution would be (other than `Option<Vec<...>>` which is going to be annoying, maybe made a bit better by `into Option<...>`).

watcher/src/verification_reports_collector.rs
```diff
@@ -73,12 +72,11 @@ impl NodeClient for ConsensusNodeClient {
             };
 
         // Contact node and get a VerificationReport.
-        let verifier = Verifier::default();
         let mut client = ThickClient::new(
             // TODO: Supply a chain-id to watcher?
             String::default(),
             node_url.clone(),
-            verifier,
+            [],
```
eranrund, line 79: This is pretty much the only place where we have non-test code that wants to not verify anything.

fog/ledger/server/tests/store.rs
eranrund: I guess another way to phrase this is that maybe it makes it a bit too easy for users to implicitly skip verification? `&[]` is perhaps a bit less explicit than a default Verifier that is not configured to verify anything?

I'm fine leaving it as is, but would like to hear @cbeck88 's opinion on this as well.

nick-mobilecoin: I agree an empty array meaning "don't check measurements" seems like a bad API.

I created #3374, It's title was "Handle report only interfaces", I updated name to emphasize the potential footgun here rather than the current non test usage of no measurements.

eranrund (approved):

cbeck88:
I find it a little funky that we pass `Vec<TrustedMeasurements>` rather than just `TrustedMeasurements` ? This is already a collection, right?

cbeck88:
I feel like it might be simpler to pass just `TrustedMeasurements`, but fine either way

nick-mobilecoin:
> I find it a little funky that we pass `Vec<TrustedMeasurements>` rather than just `TrustedMeasurements` ? This is already a collection, right?

`TrustedMeasurement` is one measurement, a MrEnclave or a MrSigner version, not a collection.

cbeck88:
I see, I got confused and thought it was this: https://github.com/mobilecoinfoundation/mobilecoin/blob/2eb944aed82d050be2209fa78a2358d4c095866f/attest/verifier/config/src/lib.rs#L114

cbeck88:
I think one concern i have with passing sets of measurements around is that different advisories may apply to them but there will be no way to express that in the api. Part of the motivation for the attest verifier config crate

cbeck88:
Misclick sorry

cbeck88 (changes requested):
Lets pass TrustedMeasurementSet or a list of StatusVerifierConfig or something that associates hardening advisories to individual MRENCLAVE etc.

nick-mobilecoin:
> Lets pass TrustedMeasurementSet or a list of StatusVerifierConfig or something that associates hardening advisories to individual MRENCLAVE etc.

It may be a naming thing with with the ambiguity of "measurement"
In this PR the `TrustedMrEnclaveMeasurement` and the `TrustedMrSignerMeasurement` includes the advisories:
- https://github.com/mobilecoinfoundation/mobilecoin/blob/21f3dedaa533e4a71b02e6cee330afecaea29339/attest/verifier/config/src/lib.rs#L31
- https://github.com/mobilecoinfoundation/mobilecoin/blob/21f3dedaa533e4a71b02e6cee330afecaea29339/attest/verifier/config/src/lib.rs#L105

```rust
pub struct TrustedMrEnclaveMeasurement {
    /// The MRENCLAVE measurement
    ///
    /// For JSON this will be hex-encoded bytes.
    #[serde(with = "hex", rename = "MRENCLAVE")]
    mr_enclave: [u8; 32],
    /// The list of config advisories that are known to be mitigated in
    /// software at this enclave revision.
    #[serde(default)]
    mitigated_config_advisories: Vec<String>,
    /// The list of hardening advisories that are known to be mitigated in
    /// software at this enclave revision.
    #[serde(default)]
    mitigated_hardening_advisories: Vec<String>,
}
```

nick-mobilecoin:
I wonder what a good name would for these
"measurement" was already a bit overloaded as it's use in the SGX SDK  only captured the mrsigner key value and not the ISVSVN and product ID.

cbeck88 (approved):
Gonna go drink my coffee 😮‍💨 😁☕

nick-mobilecoin:
Kicking for re-review since I was waiting for the `mc-attestation-verifier` crate to come in and this resulted in more than minor changes once it rebased.

eranrund (approved):

dolanbernard (approved):

PR #3390, opened 2023-06-14
light client verifier

cbeck88:
Still need to add tests, but I think this is a reasonable approach.

I chose not to use the built-in SCP "is blocking or quorum set" because I think this implementation is simpler and will do what we want here.


light-client/verifier/src/verifier.rs
```diff
@@ -0,0 +1,117 @@
...
+    /// Validate that one or more TxOut's appeared in a particular block that
+    /// was externalized
+    pub fn validate_txos_in_block(
+        &self,
+        txos: &[TxOut],
+        block: &Block,
+        block_contents: &BlockContents,
+        block_metadata: &[BlockMetadata],
+    ) -> Result<(), Error> {
+        // Validate the block and block contents
```
nick-mobilecoin, line 101: :upside_down_face: Suggest omitting this comment it seems to be reiterating the function name below, communicating "what" and not providing any extra insight for "why"
```suggestion
```

light-client/verifier/src/verifier.rs
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    /// Validate that one or more TxOut's appeared in a particular block that
+    /// was externalized
+    pub fn validate_txos_in_block(
+        &self,
+        txos: &[TxOut],
+        block: &Block,
+        block_contents: &BlockContents,
+        block_metadata: &[BlockMetadata],
```
nick-mobilecoin, line 99: :question: Guessing there is no larger container for these block types?
It feels like all of these should be what is put on the block chain for each block, but I can understand the difficulty in clearly naming all the sub pieces

nick-mobilecoin (commented):
I'm probably a bit scared from "verify" and "validate" in SW processes, https://en.wikipedia.org/wiki/Verification_and_validation

I think we may want to come to consensus or understanding of our uses of "verify" and "validate".

IFF we can trust a random internet site, https://www.precisely.com/blog/data-quality/data-validation-vs-data-verification
My understanding of that description is that "validate" is to ensure the structural shape of the data, while "verify" is to insure the integrity of the values.

That just happened to be the first site I found, I have no vested interest in which term is what, just that we ensure we're all on a common understanding of those terms

light-client/verifier/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "mc-light-client-verifier"
+version = "4.1.0-pre0"
+authors = ["MobileCoin"]
+edition = "2018"
```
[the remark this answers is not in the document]
cbeck88: i copied from another crate, I guess we still have a few 2018 crates floating around? will try to find it

light-client/verifier/src/verifier.rs
```diff
@@ -0,0 +1,117 @@
...
+
+        for (range, validator_set) in self.historical_validator_sets.iter() {
+            if range.contains(&block.index) {
+                let metadata_block_id = validator_set.validate_block_id_signatures(metadata)?;
+                if metadata_block_id != block.id {
+                    return Err(Error::BlockIdMismatch);
+                }
+                return Ok(());
+            }
+        }
```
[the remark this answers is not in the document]
cbeck88: yeah okay, i agree

light-client/verifier/src/verifier.rs
```diff
@@ -0,0 +1,117 @@
...
+        for txo in txos {
+            if !block_contents
+                .outputs
+                .iter()
+                .any(|block_contents_txo| block_contents_txo == txo)
+            {
+                return Err(Error::TxOutNotFound);
+            }
+        }
+        Ok(())
```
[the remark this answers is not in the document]
cbeck88: I am thinking maybe the error should say which TxOut was not found, but if we don't go that route then this is a fine alternative

light-client/verifier/src/verifier.rs
cbeck88: There is likely an ArchiveBlock object or something that contains this, but I would prefer not to tie the ArchiveBlock format to this code, because I expect it might evolve somehow. That's my thinking anyways.

What I'm thinking is likely to happen is, there will be some C-FFI here so we can pass an object like this to go code. I am thinking to make a proto schema that contains these 4 things and have the go code pass a pointer to proto bytes, which rust code deserializes. But I'm thinking that's specific to the C-FFI and not necessarily the rust validator object.

cbeck88:
I think ive been using validate and verify interchangeably, but i am probably not doing it correctly then perhaps. Thanks for pointing this out.

We have a bunch of other code that mostly used `validate` everywhere and not `verify`, maybe we can just continue with that?

nick-mobilecoin:
With the common term "validator node", it seems that in the domain of blockchains the terms "validate" and "verify" are more or less interchangeable.

Which is fine by me, the disparate usage of the terms in SW process was always confusing.

light-client/verifier/src/verifier.rs
```diff
@@ -0,0 +1,111 @@
...
+        }
+        if block.index >= self.trusted_validator_set_start_block {
+            return self
+                .trusted_validator_set
+                .verify_block_id_signatures(&block.id, metadata);
+        }
+
+        for (range, validator_set) in self.historical_validator_sets.iter() {
+            if range.contains(&block.index) {
+                return validator_set.verify_block_id_signatures(&block.id, metadata);
```
remoun, line 64: Might the ranges overlap? If so, we should try all matching ranges, and return Ok if any of them succeed

remoun (commented):
Looking good so far!

light-client/verifier/src/verifier.rs
cbeck88: I don't think it makes sense for the ranges to overlap, that would mean that there are two different authoritative validator sets for some block.

I think it might be okay to test all remaining ranges and return an error if there was overlap, but since I don't think this is going to happen, I don't think it's super important. That could also happen during input validation instead.

light-client/verifier/src/verifier.rs
cbeck88: > It feels like all of these should be what is put on the block chain for each block, but I can understand the difficulty in clearly naming all the sub pieces

Another difficulty is, BlockMetadata is not "on the chain", it's rather stored next to the chain. It can't be under the block hash, because the enclave computes and emits the block and its hash before the node signs the block metadata with its message signing key.

light-client/verifier/src/verifier.rs
```diff
@@ -0,0 +1,111 @@
...
+        // Note: for big enough blocks, it's probably faster to throw them in a hash set
+        // first, but it has to be very big for this to be noticeable.
+        for (idx, txo) in txos.iter().enumerate() {
+            if !block_contents
+                .outputs
+                .iter()
+                .any(|block_contents_txo| block_contents_txo == txo)
+            {
+                return Err(Error::TxOutNotFound(idx));
+            }
```
nick-mobilecoin, line 107: :question: Are there any guarantees TXO order?
Thinking if the TXO's happened to get sorted or filtered prior to this the "idx" may lose it's meaning.

Though perhaps that's up to the caller to handle or normalize

nick-mobilecoin (approved):

light-client/verifier/src/verifier.rs
remoun: How about we return the TXO pub key?

eranrund (approved):

cbeck88:
okay, i'm no longer working on this, gonna leave it up until some time monday in case anyone wants to review the test coverage or something

eranrund (approved):

PR #3407, opened 2023-07-07
fix (chart): watcher backups

jgreat:
- Update watcher backup process to only fire off once per day.
- Backup sidecar to use infra image so we have aws and mdb tools available.
- Fix backup script to use mdb tools to gracefully copy lmdb files.
- remove /target from .dockerignore (can't build docker containers/find binaries if we're just ignoring the default build dir)

### Motivation

AVR Watcher backup storage is a large infra expense.  This will continue to grow while we work on integrating AVRs into the chain.  In the mean time, we discussed and we're going to scale back the granularity and duplication of data. Complete backups will still continue to be available, we're just moving the timing from hourly to daily.

Note: since we're not changing binaries, I don't believe another 5.0.x release is necessary.  We will out-of-band build and publish updated charts for this change with 4.x and 5.x binaries.  This should be added to the 5.0 release branch and rolled up to master just in case we do need to make later 5.x releases.
 

### Future Work
- Clean up redundant historical copies.
- Complete AVRs on chain and retire the Watcher services.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,3 @@
 .git/
 cargo/
-target/
+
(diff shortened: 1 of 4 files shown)


joekottke (approved):
LGTM

joekottke (approved):

PR #3421, opened 2023-07-19
Use `TargetInfo` from `mc-sgx-core-types`

nick-mobilecoin:
Change to the  `TargetInfo` from `mc-sgx-core-types` in order to be able
to create a DCAP quote.


nick-mobilecoin:
* master
  * **PR #3421** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3421" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3421?utm_source=stack-comment).

attest/untrusted/src/lib.rs
```diff
@@ -82,10 +81,35 @@ impl QuotingEnclave {
...
+}
+
+impl From<mc_sgx_dcap_ql::Error> for TargetInfoError {
+    fn from(src: mc_sgx_dcap_ql::Error) -> Self {
+        match src {
+            mc_sgx_dcap_ql::Error::QuoteLibrary(QlError::Busy) => TargetInfoError::QeBusy,
+            e => TargetInfoError::Ql(e),
+        }
+    }
+}
```
nick-mobilecoin, line 103: In particular the `mc-sgx-dcap-ql` crate is meant to be used in a `std` environment, it calls into functions that are available via `.so`s.
Because of this the `TargetInfoError` was brought out of the core types and moved here to the untrusted side.

attest/untrusted/src/lib.rs
```diff
@@ -55,12 +59,18 @@ impl QuotingEnclave {
...
+        {
+            // The Intel QE and PCE provided with `libsgx-dcap-ql` only work on SGX
+            // hardware. For EPID there is a simulator implementation of
+            // [sgx_init_quote()](https://github.com/intel/linux-sgx/blob/1efe23c20e37f868498f8287921eedfbcecdc216/sdk/simulation/uae_service_sim/quoting_sim.cpp#L138)
+            // Unfortunately there doesn't seem to be a DCAP equivalent.
+            Ok((TargetInfo::default(), gid))
+        }
+        #[cfg(not(feature = "sgx-sim"))]
+        {
+            Ok((TargetInfo::for_quoting_enclave()?, gid))
```
nick-mobilecoin, line 73: For this I think the logic belongs longer term down in `mc-sgx-dcap-ql`. However the way that the `sgx-sim` feature is enabled here in the `MobileCoin` repo it can't propagate to dependency repos:
```mermaid
sequenceDiagram
    cargo build->>+mc-attest-untrusted: Build crate With features
    mc-attest-untrusted->>mc-attest-untrusted: Resolve features
    mc-attest-untrusted->>+mc-sgx-dcap-ql: Build crate With features
    mc-sgx-dcap-ql->>-mc-attest-untrusted: Propagate `CARGO_DEP` values
    mc-attest-untrusted->>mc-attest-untrusted: Invoke build.rs
    Note over mc-attest-untrusted: build.rs adds rustc-cfg=feature="sgx-sim"
    mc-attest-untrusted->>-mc-attest-untrusted: Compile with rustc
```
In the above diagram one can see that the `mc-attest-untrusted` crate doesn't enable the `sgx-sim` feature until it's `build.rs` is run, which is after `mc-sgx-dcap-ql` is built.

I punted for now on trying to hook these 2 up nicer. We can either try and key off the same environment variable as the MobileCoin repo or we could look at a more explicit enabling of the `sgx-sim` feature via a cargo build command line or an alias with `.cargo/config.toml`

light-client/verifier/Cargo.toml
```diff
@@ -11,7 +11,7 @@ base64 = "0.21"
 displaydoc = "0.2"
 prost = { version = "0.11", default-features = false, features = ["prost-derive"] }
 serde = { version = "1.0", default-features = false, features = ["alloc", "derive"] }
-serde_with = { version = "2.3", features = ["hex"] }
+serde_with = { version = "3.1", default-features = false, features = ["hex"] }
```
nick-mobilecoin, line 14: Updated these serde-with's to keep them all n'sync

dolanbernard (approved):

.github/workflows/dependent-repos.yml
```diff
@@ -10,7 +10,7 @@ env:
 jobs:
   android-bindings:
     runs-on: [self-hosted, Linux, large]
-    container: mobilecoin/builder-install:v0.0.25
+    container: mobilecoin/builder-install:v0.0.26
```
eranrund, line 13: I guess the dependent repos don't yet build with 0.0.27?
If that's the case then whats causing it?

attest/untrusted/src/lib.rs
eranrund: Personally I think the env var solution is fine, explicitly specifying features in cargo build commands feels more annoying to me since you have to keep remembering to do it, and I'm not sure how you would selectively enable it in `.cargo/config.toml`

.github/workflows/dependent-repos.yml
nick-mobilecoin: nice catch!, that's a fail on my part when I bumped during development of this feature from 26 to 27

nick-mobilecoin: updated now

attest/untrusted/src/lib.rs
nick-mobilecoin: Sounds good i'll look to making the sgx n'sync with the main repo

eranrund (approved):

PR #3430, opened 2023-07-20
fix (helm): Fog routers deployed in stacks.

jgreat:
### Motivation

Since we found out that we can't scale fog-stores by adding more replicas (routers don't load-balance), This redesigns the fog-router/stores into "stacks". Each router has a dedicated set of stores with 1 replica. We run multiple stacks to provide availability and more capacity.

Also included is a change to fog-test-client to make the polling interval configurable.  The previous default of 50ms was a bit aggressive, considering we were measuring the result in 1s buckets.


joekottke (changes requested):
@jgreat Do you need to update the in-line values file that is created in the dev deploy GitHub action(s)?

joekottke (approved):
Discussed things with @jgreat.  LGTM!

aweis89 (approved):

PR #3434, opened 2023-07-21
Use `IsvSvn` from `mc-sgx-core-types`

nick-mobilecoin:
Previously a `SecurityVersion` type from `mc-attest-core` was used. Now
the `IsvSvn` from `mc-sgx-core-types` is used.

### Motiviation

This is the first on a series of PRs to update the `Report` type to the `mc-sgx-core-types` version. Any transitive types used by `Report` will need to be updated accordingly


attest/core/src/types/report_body.rs
```diff
@@ -164,9 +165,9 @@ impl ReportBody {
 
         // Check if the security version is high enough
         let svn = self.security_version();
-        if minimum_security_version > svn {
+        if minimum_security_version.as_ref() > svn.as_ref() {
```
nick-mobilecoin, line 168: The `as_ref()` is currently needed until https://github.com/mobilecoinfoundation/sgx/issues/374 gets done.

attest/verifier/src/avr.rs
```diff
@@ -14,7 +14,7 @@ use mc_attest_core::{IasNonce, VerificationReportData};
 use serde::{Deserialize, Serialize};
 
 /// An enumeration of AVR data verifiers
-#[derive(Clone, Debug, Deserialize, Eq, Hash, Ord, PartialEq, PartialOrd, Serialize)]
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
```
nick-mobilecoin, line 17: `IsvSvn` doesn't implement `Deserialize, Ord, PartialOrd, Serialize`.  Removing these seemed to have no effect, so I went that route for now. 

I'm kind of curious why these traits were here:
- `Ord` and `PartialOrd` don't seem to make much sense to compare verifier structs to each other, but I think they're required if one is using `BtreeMap`. 
- `Serialize` and `Deserialize` seems interesting, I'm trying to figure out when you'd want to receive a verifier over the wire versus build your own. i.e. can you trust the one over the wire is correct?

dolanbernard (approved):

eranrund (approved):

PR #3463, opened 2023-07-29
fix (view-router): shardng strategy should not assume 50% overlap

jgreat:
### Motivation

View stores assumed a 50% overlap on shards and required half the shard total range to be loaded before returning data to routers. Having a full 50% overlap is resource heavy. This change will makes sure at least 1 block is loaded but will allow for smaller or no overlap configurations. 

- Add some debug logging for view router/store behavior troubleshooting.
- Change logic for returning NOT_READY or SUCCESS to router.

### Future Work

- Revisit logic for when to start serving responses to router.
- This NOT_READY/SERVING behavior should set the gRPC Health Check endpoint, which will enable/disable routing to stores on the k8s/system level.

### DIFF
--- a/fog/view/server/src/shard_responses_processor.rs
+++ b/fog/view/server/src/shard_responses_processor.rs
@@ -45,10 +45,18 @@ pub fn process_shard_responses(
     let mut view_store_uris_for_authentication = Vec::new();
     let mut new_query_responses = Vec::new();
 
-    for (shard, response) in shards_and_responses {
+    log::debug!(
+        logger,
+        "process_shard_responses called on {} shards_and_responses",
+        shards_and_responses.len()
+    );
+
+    for (i, (shard, response)) in shards_and_responses.into_iter().enumerate() {
         if response.block_range != shard.block_range {
             return Err(RouterServerError::ViewStoreError(format!("The shard response's block range {} does not match the shard's configured block range {}.", response.block_range, shard.block_range)));
         }
+        log::debug!(logger, "Shard {} status: {:?}", i, response.status);
+
         match response.status {
             mc_fog_types::view::MultiViewStoreQueryResponseStatus::Unknown => {
                 log::error!(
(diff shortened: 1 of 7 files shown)


eranrund (approved):

fog/view/server/src/sharding_strategy.rs
```diff
@@ -95,11 +95,18 @@ impl EpochShardingStrategy {
...
+        // for cost. We should revisit at some point and create
+        // alternative strategies. This logic should drive the health
+        // check endpoint that k8s can consume more than a response to
+        // the router.
+        // let epoch_block_range_length =
+        //     self.epoch_block_range.end_block -
+        // self.epoch_block_range.start_block;
+        // let minimum_processed_block_count = epoch_block_range_length / 2;
+
+        u64::from(processed_block_count) >= 1
```
nick-mobilecoin, line 109: :question: Should a similar change be made in https://github.com/mobilecoinfoundation/mobilecoin/blob/master/fog/ledger/server/src/sharding_strategy.rs ?

nick-mobilecoin (commented):
LGTM pending the build fixes and if ledger should also be adjusted

fog/view/server/src/sharding_strategy.rs
jgreat: I'll fix that.  

For some reason ledger seems to work, but why would that be the case? 

Maybe because fog-test-client is only hitting the parts of ledger that are integrated with the router service? 

I don't think that's true because I do see cpu load on the stores.

fog/view/server/src/shard_responses_processor.rs
```diff
@@ -45,10 +45,17 @@ pub fn process_shard_responses(
...
     let mut new_query_responses = Vec::new();
 
-    for (shard, response) in shards_and_responses {
+    log::debug!(
+        logger,
+        "process_shard_responses called on {} shards_and_responses",
+        shards_and_responses.len()
+    );
+
+    for (_i, (shard, response)) in shards_and_responses.into_iter().enumerate() {
```
nick-mobilecoin, line 54: :wrench: If `_i` is ignored it seems like this can probably go back to the way it was without using enumerate
```suggestion
    for (shard, response) in shards_and_responses {
```

jgreat: I just followed the linter suggestion, but that makes more sense.

Fixed.

nick-mobilecoin (approved):

PR #3480, opened 2023-08-07
Change sim signing certs to use ECDSA

nick-mobilecoin:
Previously the sim signing certificates used RSA keys. Now they use
ECDSA keys. Changing to ECDSA allows for re-use with DCAP quote signing
which happen to use ECDSA signatures

### DIFF
--- a/attest/net/src/sim.rs
+++ b/attest/net/src/sim.rs
@@ -90,8 +90,10 @@ impl RaClient for SimClient {
         let entropy = OsEntropy::new();
         let mut csprng = CtrDrbg::new(Arc::new(entropy), None).expect("Could not create CtrDrbg");
 
-        let mut signer = Pk::from_private_key(IAS_SIM_SIGNING_KEY.as_bytes(), None)
-            .expect("Could not load signing key.");
+        let mut null_terminated_key = IAS_SIM_SIGNING_KEY.as_bytes().to_vec();
+        null_terminated_key.push(0);
+        let mut signer =
+            Pk::from_private_key(&null_terminated_key, None).expect("Could not load signing key.");
         let mut signature = vec![0u8; 1024];
         let bytes_signed = signer
             .sign(
--- a/attest/verifier/build.rs
+++ b/attest/verifier/build.rs
@@ -20,6 +20,10 @@ use mbedtls_sys::types::{
 };
 use mc_util_build_script::Environment;
 use mc_util_build_sgx::{IasMode, SgxEnvironment, SgxMode};
+use p256::{
+    ecdsa::SigningKey,
+    pkcs8::{EncodePrivateKey, LineEnding},
+};
 use rand::{RngCore, SeedableRng};
 use rand_hc::Hc128Rng;
 use std::{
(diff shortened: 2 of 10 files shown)


nick-mobilecoin:
* master
  * **PR #3480** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3480" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3480?utm_source=stack-comment).

attest/net/src/sim.rs
```diff
@@ -90,8 +90,10 @@ impl RaClient for SimClient {
         let entropy = OsEntropy::new();
         let mut csprng = CtrDrbg::new(Arc::new(entropy), None).expect("Could not create CtrDrbg");
 
-        let mut signer = Pk::from_private_key(IAS_SIM_SIGNING_KEY.as_bytes(), None)
-            .expect("Could not load signing key.");
+        let mut null_terminated_key = IAS_SIM_SIGNING_KEY.as_bytes().to_vec();
+        null_terminated_key.push(0);
+        let mut signer =
+            Pk::from_private_key(&null_terminated_key, None).expect("Could not load signing key.");
```
nick-mobilecoin, line 96: MbedTls expects a null terminated string, but strict PEM parsers will bark on extraneous trailing characters.  So decided to remove the null byte and update the key string for mbedtls only. This entire file will go away once we're on DCAP.

attest/verifier/build.rs
```diff
@@ -164,23 +168,17 @@ fn main() {
...
-                .write_private_der_vec()
-                .expect("Could not export privkey to DER"),
-            None,
-        )
-        .expect("Could not parse privkey from DER");
+        let private_key = SigningKey::random(&mut *RNG.lock().expect("mutex poisoned"));
+        let der_private_key = private_key
+            .to_pkcs8_der()
+            .expect("Could not export privkey to DER");
+        let mut root_subject_key = Pk::from_private_key(der_private_key.as_bytes(), None)
```
nick-mobilecoin, line 175: Note the previous implementation had the `root_subject_key` as the private key, I think this was incorrect it was also placed in the certificate as the subject key.

attest/verifier/src/ias.rs
```diff
@@ -104,7 +104,7 @@ impl IasReportVerifier {
         // will support
         let profile = Profile::new(
             vec![HashType::Sha256, HashType::Sha384, HashType::Sha512],
-            vec![PkType::Rsa, PkType::Ecdsa],
+            vec![PkType::Rsa, PkType::Eckey, PkType::Ecdsa],
```
nick-mobilecoin, line 107: This had to be updated in order for MbedTls to handle the ECDSA certificates. The previous `PkType::Ecdsa` didn't fulfill the full need in order to work with elliptical curves.

eranrund:
Since I'm lacking a lot of context here, can you please elaborate on how these sim keys are going to be used in DCAP data structures?

nick-mobilecoin:
> Since I'm lacking a lot of context here, can you please elaborate on how these sim keys are going to be used in DCAP data structures?

I changed description to say "quote signing" instead of the generic data structures.

Normally we use the root CA provided by Intel to check a certificate chain created by the PCE (provisioning certificate enclave), this chain signs the quote generated by the QE (quoting enclave).
These enclaves are built and signed by Intel and only work on actual SGX hardware. These certificate chains use ECDSA some other signatures inside the DCAP quote also use EDCDSA.

In order to work in a simulated SGX environment we provide a fake certificate chain that creates and signs the DCAP quote, #3482. This PR is to update the certificates to use ECDSA so that all the sim logic in the DCAP quote is ECDSA signing.

eranrund (approved):

nick-mobilecoin:
@nick-mobilecoin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3480).

nick-mobilecoin:
@nick-mobilecoin merged this pull request with [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3480).

PR #3495, opened 2023-08-11
Refactor logic for generating sim files in `mc-attest-verifier`

nick-mobilecoin:
The logic for (re)generating sim files has been re-worked to make it
easier to add the generation of more sim files.

This is _smaller_ refactor to prepare for the need to generate more sim files.

### DIFF
--- a/attest/verifier/build.rs
+++ b/attest/verifier/build.rs
@@ -25,6 +25,7 @@ use std::{
     fs::write,
     path::{Path, PathBuf},
     sync::{Arc, Mutex},
+    time::SystemTime,
 };
 
 lazy_static::lazy_static! {
@@ -68,6 +69,10 @@ impl RngCallback for RngForMbedTls {
 }
 
 fn main() {
+    // This path is inside of the repo and not the normal output directory.
+    // This is to reuse the generated files between the main build and the enclave
+    // builds. There is not a good way to communicate a common build directory
+    // between the different builds.
     let base_dir = env::var("CARGO_MANIFEST_DIR").expect("Could not read manifest dir");
     let data_path = PathBuf::from(base_dir).join("data").join("sim");
 
@@ -86,12 +91,19 @@ fn main() {
         generate_sim_files(&data_path);
     }
 }
+
+const ROOT_ANCHOR_FILENAME: &str = "root_anchor.pem";
+const SIGNER_KEY_FILENAME: &str = "signer.key";
+const CHAIN_FILENAME: &str = "chain.pem";
+
+const GENERATED_FILENAMES: &[&str] = &[ROOT_ANCHOR_FILENAME, SIGNER_KEY_FILENAME, CHAIN_FILENAME];
+
 fn generate_sim_files(data_path: impl AsRef<Path>) {
     let data_path = data_path.as_ref();
 
-    let root_anchor_path = data_path.join("root_anchor.pem");
-    let signer_key_path = data_path.join("signer.key");
-    let chain_path = data_path.join("chain.pem");
+    let root_anchor_path = data_path.join(ROOT_ANCHOR_FILENAME);
+    let signer_key_path = data_path.join(SIGNER_KEY_FILENAME);
+    let chain_path = data_path.join(CHAIN_FILENAME);
 
     const ROOT_SUBJECT: &str = "C=US,ST=CA,L=Santa Clara,O=Intel Corporation,CN=Simulation Intel SGX Attestation Report Signing CA\0";
     const SIGNER_SUBJECT: &str = "C=US,ST=CA,L=Santa Clara,O=Intel Corporation,CN=Simulation Intel SGX Attestation Report Signer\0";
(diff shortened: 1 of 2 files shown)


nick-mobilecoin:
* master
  * **PR #3495** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3495" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3495?utm_source=stack-comment).

attest/verifier/build.rs
```diff
@@ -290,4 +81,170 @@ fn main() {
     if sgx.ias_mode() == IasMode::Development {
         cargo_emit::rustc_cfg!("feature=\"ias-dev\"");
     }
+
+    if should_generating_sim_files(&data_path) {
+        generate_sim_files(&data_path);
+    }
+}
+fn generate_sim_files(data_path: impl AsRef<Path>) {
```
nick-mobilecoin, line 89: :( sorry I was hoping github would wee that this code was moved, diffing locally might show it better.

attest/verifier/build.rs
```diff
@@ -72,213 +66,10 @@ impl RngCallback for RngForMbedTls {
...
-        .expect("Could not stringify signer key path"));
-    cargo_emit::rerun_if_changed!(chain_path
-        .to_str()
-        .expect("Could not stringify signer chain path"));
-
-    purge_expired_cert(&root_anchor_path);
-    purge_expired_cert(&chain_path);
-
-    if !(root_anchor_path.exists() && signer_key_path.exists() && chain_path.exists())
-        || env::var("MC_SEED").is_ok()
```
nick-mobilecoin, line 154: I may misunderstanding what the intent here was.

My understanding is that Cargo will not invoke this build.rs file unless one of the "rerun_if_changed" files has been touched, or the MC_SEED variable has changed.

This means that the "purge_expired_cert" won't normally be hit.

Based on the expiry time below it seems like one would need to somehow manually kick this every week, I know the comment says 30 days, but the code is one week.

```rust
        // Good for 30 days
        let end_time = end_now + Duration::weeks(1);
```
I haven't been playing with enclave verification specifically to know what people have been doing because I'm pretty sure that after a week a manual intervention of some kind is necessary.

eranrund: That might've just been an oversight. I will note that I have seen the `mc-attest-verifier` crate being rebuilt way more often than necessary in multiple occasions, so perhaps that masked this issue. I never dug into what was causing it (and it's possible it has since been resolved - it's been awhile since I spent too much time building this repo to notice the build behavior)

attest/verifier/build.rs
```diff
@@ -290,4 +81,170 @@ fn main() {
...
+        .write_pem_string(&mut csprng)
+        .expect("Could not create PEM string of certificate");
+
+    write(chain_path, root_cert_pem + &signer_cert_pem).expect("Unable to write cert chain");
+}
+
+/// Returns true of the build script should generate the sim files
+fn should_generating_sim_files(data_path: impl AsRef<Path>) -> bool {
+    let data_path = data_path.as_ref();
+    let signal_file = data_path.join("delete_me_to_regenerate_sim_files.empty");
```
eranrund, line 226: Can't we regenerate if one of the generated sim files are deleted?

attest/verifier/build.rs
```diff
@@ -290,4 +81,170 @@ fn main() {
...
+    // the signal file.
+    if !signal_file.exists() {
+        write(&signal_file, "").expect("Could not write signal file");
+        return true;
+    }
+
+    // If we got this far we assume that the MC_SEED environment variable was
+    // changed The downside to this approach is that if the MC_SEED is set
+    // initially there will always be one incremental build:
+    // - first build the `signal_file` is created, which tells cargo to re-run this
```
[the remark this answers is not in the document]
nick-mobilecoin: I was thinking about this some more over the weekend.
I'm wondering why these files are generated into the repo itself.  I think it's so that the untrusted and the enclaves can use the same files.
But I think it would be better to place these in `$CARGO_TARGET_DIR/sim` like how the enclaves create dedicated directories.

As in general tany of these files could inadvertently be commited to github.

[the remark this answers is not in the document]
eranrund: I'm not sure why that is. This code has been here for a very very long time and `cargo` went through multiple changes and improvements since it was first written...
It might no longer be the case that there's a good reason for it, or maybe it encodes some learnings that did not get properly documented. Only one way to find out :/

attest/verifier/build.rs
```diff
@@ -290,4 +81,170 @@ fn main() {
...
+    // However, the creation/updating of the signal file by this build script is
+    // itself a modification. So we only concern ourselves with the existence of
+    // the signal file.
+    if !signal_file.exists() {
+        write(&signal_file, "").expect("Could not write signal file");
+        return true;
+    }
+
+    // If we got this far we assume that the MC_SEED environment variable was
+    // changed The downside to this approach is that if the MC_SEED is set
```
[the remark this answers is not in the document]
nick-mobilecoin: fixed, thanks!

attest/verifier/build.rs
nick-mobilecoin: Reverted back to regenerating if any of the sim files are missing, or older than 1 week

attest/verifier/build.rs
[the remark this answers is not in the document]
nick-mobilecoin: I churned more than I should have. I'm pretty sure that it's to share the certs with the main workspace build and the enclaves.

The problem of placing the data files down in the build dir, arises in that there is no common build dir form the enclave's perspective.  The enclaves only know about their version of `CARGO_TARGET_DIR` and `OUT_DIR` and there isn't an easy way for them to correlate to the the main repo's `CARGO_TARGET_DIR` and `OUT_DIR`. It would require either updating the enclave build utility to propagate the main root dir or the logic in this module would have to know about the enclave build directories.  IFF we wanted to invest I would suggest the main root dir is given to the enclave builds, but at this point I don't think we get much benefit updating and can continue on the current path.

attest/verifier/build.rs
```diff
@@ -86,12 +91,19 @@ fn main() {
         generate_sim_files(&data_path);
     }
 }
+
+const ROOT_ANCHOR_FILENAME: &str = "root_anchor.pem";
+const SIGNER_KEY_FILENAME: &str = "signer.key";
+const CHAIN_FILENAME: &str = "chain.pem";
+
+const GENERATED_FILENAMES: &[&str] = &[ROOT_ANCHOR_FILENAME, SIGNER_KEY_FILENAME, CHAIN_FILENAME];
```
nick-mobilecoin, line 99: This is me knowing about a couple of more files coming down the pipe and it seemed like the least error prone way to know about the file names in 2 locations.

attest/verifier/build.rs
```diff
@@ -290,4 +86,185 @@ fn main() {
...
+    const SIGNER_SUBJECT: &str = "C=US,ST=CA,L=Santa Clara,O=Intel Corporation,CN=Simulation Intel SGX Attestation Report Signer\0";
+
+    let mut serial: [u8; 1] = [1u8];
+
+    let now = Utc::now();
+    let end_now = now;
+    // Starting 1 hour ago
+    let start_time = now - Duration::hours(1);
+    // Good for 30 days
+    let end_time = end_now + Duration::weeks(1);
```
eranrund, line 118: Might be a good time to fix this

eranrund (approved):

attest/verifier/build.rs
nick-mobilecoin: oh yah, I was going to bump it to 30 days, unless you know the history of 1 week?

nick-mobilecoin:
@nick-mobilecoin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3495).

nick-mobilecoin:
@nick-mobilecoin merged this pull request with [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3495).

PR #3500, opened 2023-08-12
Remove "-D warnings" for development builds

nick-mobilecoin:
The "-D warnings" is only enforced during clippy or lint runs. If one
pushes to CI or runs `tools/lint.sh` the warnings will show up as
errors.

PR #3501, opened 2023-08-13
Update lint script to lint root workspace once

nick-mobilecoin:
Previously clippy would run twice on the root workspace.
Now the root workspace is only linted once.

### DIFF
--- a/tools/lint.sh
+++ b/tools/lint.sh
@@ -35,14 +35,18 @@ cargo install --version 1.0.9 --locked cargo-sort
 
 # We want to check with --all-targets since it checks test code, but that flag
 # leads to build errors in enclave workspaces, so check it here.
+cargo sort --workspace --grouped $CHECK
 cargo clippy --all --all-features --all-targets -- -D warnings
+cargo fmt -- --unstable-features $CHECK
 
-for toml in $(grep --exclude-dir cargo --exclude-dir rust-mbedtls --include=Cargo.toml -r . -e '\[workspace\]' | cut -d: -f1); do
+# Run on all the nested workspaces.
+# We exclude `cargo` paths as these are crates from `[patch.crates-io]`
+for toml in $(find . -mindepth 2 -type f -not -path '*/cargo/*' -name Cargo.toml -exec grep -l -e '\[workspace\]' {} +); do
   pushd $(dirname $toml) >/dev/null
   echo "Linting in $PWD"
   cargo sort --workspace --grouped $CHECK
-  cargo fmt -- --unstable-features $CHECK
   cargo clippy --all --all-features -- -D warnings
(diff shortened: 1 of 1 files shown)


nick-mobilecoin:
* master
  * **PR #3500** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3500" target="_blank">Graphite</a> 
    * **PR #3501** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3501" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3501?utm_source=stack-comment).

tools/lint.sh
```diff
@@ -35,14 +35,18 @@ cargo install --version 1.0.9 --locked cargo-sort
...
 # We want to check with --all-targets since it checks test code, but that flag
 # leads to build errors in enclave workspaces, so check it here.
+cargo sort --workspace --grouped $CHECK
 cargo clippy --all --all-features --all-targets -- -D warnings
+cargo fmt -- --unstable-features $CHECK
 
-for toml in $(grep --exclude-dir cargo --exclude-dir rust-mbedtls --include=Cargo.toml -r . -e '\[workspace\]' | cut -d: -f1); do
+# Run on all the nested workspaces.
+# We exclude `cargo` paths as these are crates from `[patch.crates-io]`
+for toml in $(find . -mindepth 2 -type f -not -path '*/cargo/*' -name Cargo.toml -exec grep -l -e '\[workspace\]' {} +); do
```
nick-mobilecoin, line 44: Limiting this logic to only the nested enclaves changed incremental runs of this script from ~1min to ~40sec on my machine. Or ~33% speed up.

tools/lint.sh
```diff
@@ -35,14 +35,18 @@ cargo install --version 1.0.9 --locked cargo-sort
...
   pushd $(dirname $toml) >/dev/null
   echo "Linting in $PWD"
   cargo sort --workspace --grouped $CHECK
-  cargo fmt -- --unstable-features $CHECK
   cargo clippy --all --all-features -- -D warnings
+  cargo fmt -- --unstable-features $CHECK
```
eranrund, line 49: I wonder why you moved the fmt after clippy? Its usually faster so to me it would make sense to have it run before the slower test

nick-mobilecoin: I'll move back. I think this was left over from me trying to get the `--fix` flag of clippy working. When fixing with clippy it will leave files in a state that still need formatting.

nick-mobilecoin: fmt is now prior to clippy

eranrund (approved):

PR #3508, opened 2023-08-16
Bump mc-attestation-verifier to 0.3.1

nick-mobilecoin:
### Future Work

[Soundtrack of this PR]()

PR #3509, opened 2023-08-16
Add struct to be returned back with an enclave Report

nick-mobilecoin:
When an enclave generates a `Report` it has a need to provide a nonce a
public key, and custom identity. The fog enclave would use the custom
identity to store the ingress key.
Previously the nonce was provided to the quoting enclave during quote
genration. This would then be returned in the quoting enclave's
`Report`. The public key and custom identity was stored in the app
enclave's `ReportBody.report_data`.

For DCAP, the quoting enclave's no longer have a means to store a nonce.
The app enclave's `ReportBody.report_data` is only 64 bytes and is thus
at capacity with no room for the nonce. To work with the nonce, the DCAP
app enclave's will store a hash over the nonce, public key, and custom
identity. When a DCAP app enclave is queried for a `Report` it will
return the `Report` as well as the structure holding the nonce, public
key and custom identifier. During quote verification this structure will
be hashed and compared to the nonce stored in the quote's app enclave.

This change brings in the containing struct `EncalveReportDataContents`
and the quote verification logic needed for it.

### Future Work

[Soundtrack of this PR]()


nick-mobilecoin:
* master
  * **PR #3508** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3508" target="_blank">Graphite</a> 
    * **PR #3509** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3509" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3509?utm_source=stack-comment).

attest/verifier/types/src/verification.rs
```diff
@@ -151,6 +154,63 @@ impl Message for VerificationSignature {
     }
 }
 
+/// Structure for holding the contents of the Enclave's Report Data.
+/// The Enclave's ReportData member contains a SHA256 hash of this structure's
+/// contents.
+#[derive(Debug, Clone)]
+pub struct EnclaveReportDataContents {
```
nick-mobilecoin, line 161: I'm up for suggestions on this name.  
I went a little verbose because `ReportData` is a `mc-sgx-core-type` that is specific to the `Report.report_data` field.  I'm guessing the field was named early on in Intel's sgx development as they seem to store hashes or signatures in this field as well. So it's more of a digest than the actual data. This struct here is meant to the actual data.

I'm generally not fond of "metadata"

Maybe `ReportDataForReals`

attest/verifier/types/src/verification.rs
```diff
@@ -151,6 +154,63 @@ impl Message for VerificationSignature {
...
+    nonce: QuoteNonce,
+    key: X25519Public,
+    custom_identity: [u8; 32],
+}
+
+impl EnclaveReportDataContents {
```
nick-mobilecoin, line 167: I should mention this is in this module because the newer DCAP version of the `VerificationReport` will be coming into this file soonish and that new type will eventually contain an instance of this type.

attest/verifier/types/src/verification.rs
eranrund: I think this name is fine

eranrund (approved):
Nice. And I suppose in theory this can be extended to support having arbitrary data that can be "vouched for" by the report?

nick-mobilecoin:
> Nice. And I suppose in theory this can be extended to support having arbitrary data that can be "vouched for" by the report?

Yes, the only caveat there is that, it will change the underlying structure so we'll need to do one of:
- bump the protobuf tag on each change
- ensure empty defaults for new fields, to work with serde serialization 
- do prost all the way down

I forgot to add serde serialzation to this new struct, i'll do it later when it's actually needed

nick-mobilecoin:
@nick-mobilecoin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3509).

nick-mobilecoin:
[Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3509) rebased this pull request as part of a merge.

PR #3512, opened 2023-08-17
Use a MRSIGNER identity for watcher.

nick-mobilecoin:
Previously watcher didn't verify VerificationReports against a known
identity. This was because watcher isn't trying to securely communicate
with an enclave, it's leveraging the common attestation API to get
access to the VerificationReports. The VerificationReports are stored
for historical proof of the enclaves. The new DCAP verifier requires an
identity in order to succeed. Now the watcher will provide a MRSIGNER
identity with a 0 SVN so that any signed consensus node will succeed in
verifying, providing watcher with the verification report.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6293,6 +6293,7 @@ dependencies = [
  "mc-blockchain-types",
  "mc-common",
  "mc-connection",
+ "mc-consensus-enclave-measurement",
  "mc-crypto-digestible",
  "mc-crypto-keys",
  "mc-ledger-db",
--- a/watcher/Cargo.toml
+++ b/watcher/Cargo.toml
@@ -21,6 +21,7 @@ mc-attest-core = { path = "../attest/core" }
 mc-blockchain-types = { path = "../blockchain/types" }
 mc-common = { path = "../common", features = ["log"] }
 mc-connection = { path = "../connection" }
+mc-consensus-enclave-measurement = { path = "../consensus/enclave/measurement" }
 mc-crypto-digestible = { path = "../crypto/digestible" }
 mc-crypto-keys = { path = "../crypto/keys" }
 mc-ledger-db = { path = "../ledger/db" }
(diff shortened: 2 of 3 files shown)


nick-mobilecoin:
* master
  * **PR #3512** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3512" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3512?utm_source=stack-comment).

eranrund:
This is not ideal... it might be fine as a stopgap, but we do want the watcher to capture reports that were signed by other signers if that ever shows up somehow.

nick-mobilecoin:
> This is not ideal... it might be fine as a stopgap, but we do want the watcher to capture reports that were signed by other signers if that ever shows up somehow. Do you think you can come up with a plan to make that possible? Might require passing a verifier to ThickClient or having an option to override the default one?

Ok good to know, previously watcher would only look at the public signing key, https://github.com/mobilecoinfoundation/mobilecoin/blob/beaa1e74bea2045eb776931977bb73555de3a2df/attest/ake/src/initiator.rs#L165
which was probably ok as it means it's an enclave whose data may be malformed.

It looks like one of my other changes added a check for debug or not, when it wasn't there previously, https://github.com/mobilecoinfoundation/mobilecoin/blob/a882be4b4fdfc359a2b4bfded1fc430e9064ba3f/attest/ake/src/initiator.rs#L168

I'll have to do some more digging, I tried removing  the thickclient and using the `ClientInitiate` directly, but ran into issues getting access to the necessary bytes. The `AuthPending.state` is not public, and I don't think it should be.

nick-mobilecoin:
Moving to draft to reconsider how watcher can get access to the DCAP verification report equivalent

eranrund:
Thanks!

nick-mobilecoin:
Closing this as we're going to try to expand the attest state machine to support a state for the use case of getting the verification report without verifying it.

PR #3516, opened 2023-08-21
Add UnverifiedReport state for attestation

nick-mobilecoin:
A new state `UnverifiedReport` as been added to the attest key exchange
state machine. The `UnverifiedReport` allows clients like watcher to
get the attestation report from the key exchange handshake without
requiring the report to pass verification. This is useful when the
client doesn't know or need to know the identity of the enclave.

### DIFF
--- a/attest/ake/src/initiator.rs
+++ b/attest/ake/src/initiator.rs
@@ -4,7 +4,7 @@
 
 use crate::{
     AuthPending, AuthRequestOutput, AuthResponseInput, ClientInitiate, Error, NodeInitiate, Ready,
-    Start, Transition, UnverifiedReport,
+    Start, Terminated, Transition, UnverifiedReport,
 };
 use alloc::vec::Vec;
 use mc_attest_core::{ReportDataMask, VerificationReport};
@@ -197,8 +197,8 @@ where
     }
 }
 
-/// AuthPending + UnverifiedReport => Ready + VerificationReport
-impl<KexAlgo, Cipher, DigestAlgo> Transition<Ready<Cipher>, UnverifiedReport, VerificationReport>
+/// AuthPending + UnverifiedReport => Terminated + VerificationReport
+impl<KexAlgo, Cipher, DigestAlgo> Transition<Terminated, UnverifiedReport, VerificationReport>
     for AuthPending<KexAlgo, Cipher, DigestAlgo>
 where
     KexAlgo: Kex,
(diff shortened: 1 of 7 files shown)


nick-mobilecoin:
* master
  * **PR #3516** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3516" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3516?utm_source=stack-comment).

watcher/src/verification_reports_collector.rs
```diff
@@ -117,6 +109,75 @@ impl NodeClient for ConsensusNodeClient {
     }
 }
 
+fn verification_report_from_node_url(
```
nick-mobilecoin, line 112: Most of this logic is taken from https://github.com/mobilecoinfoundation/mobilecoin/blob/7b68eaae98e390bcca638769079b89cec0c476c2/connection/src/thick.rs#L302

Unfortunately I didn't see a test for the report behavior of the `ConsensusNodeClient` or one for the `ThickClient::attest()`.

The test in this module tests the `VerificationReportsCollector` with a fake `NodeClient`.

To test this out I ran it against the testnet nodes as documented, https://github.com/mobilecoinfoundation/mobilecoin/blob/7b68eaae98e390bcca638769079b89cec0c476c2/watcher/README.md?plain=1#L4.

attest/ake/src/initiator.rs
```diff
@@ -196,3 +196,41 @@ where
         }
     }
 }
+
+/// AuthPending + UnverifiedReport => Ready + VerificationReport
```
eranrund, line 200: Is `Ready` really desirable here? Maybe it is better to introduce some other `State` for this that does not actually expose the cipher?

nick-mobilecoin: Added a new `Terminated` state.

attest/ake/src/state.rs
```diff
@@ -110,3 +110,10 @@ where
 }
 
 impl<Cipher> State for Ready<Cipher> where Cipher: NoiseCipher {}
+
+/// The state after an auth response has been sent by a responder/received by
+/// an initiator, but no further communication can occur. A new handshake must
+/// be created in order to continue communication.
+pub struct Terminated;
```
nick-mobilecoin, line 117: Definitely open to naming suggestions here. `Terminated` was one of the first things that came to mind.

dolanbernard: `Terminated` sounds OK, but I think it's helpful to think about why this would be the case.

nick-mobilecoin: Can you elaborate on 
> I think it's helpful to think about why this would be the case.

I'm not quite following

eranrund (approved):

PR #3520, opened 2023-08-22
fix: (helm) update shard values and set default

jgreat:
### Motivation

- set better router and shard values for test/main. 

I moved the shard range generator count into where the stores were configured for each environment and changed the name of the `store` section to `shardRangeGenerator` now that the settings cover more than just the stores.


.internal-ci/helm/fog-services/values.yaml
```diff
@@ -41,18 +41,24 @@ ingress:
   enabled: true
 
 #### Fog View
-fogViewStores:
+fogViewShardRangeGenerator:
+  # The dev options are set in the github workflow deployment.
+  # This is just for reference.
+  # dev:
+  #   shardSize: 25_000
```
nick-mobilecoin, line 48: Looking at .github/workflows/mobilecoin-workflow-dev-deploy.yaml
the values seem to be 10_000 for size and 5_000 for exceed by

.internal-ci/helm/fog-services/values.yaml
```diff
@@ -152,18 +158,24 @@ fogView:
...
 #### Fog Ledger
-fogLedgerStores:
+fogLedgerShardRangeGenerator:
+  # The dev options are set in the github workflow deployment.
+  # This is just for reference.
+  # dev:
+  #   shardSize: 25_000
+  #   exceedBlockHeightBy: 10_000
+  #   shardOverlap: 0
+  #   count: 2
```
nick-mobilecoin, line 168: I don't think I understand how the override is working in .github/workflows/mobilecoin-workflow-dev-deploy.yaml, but I get the feeling it's the same as the view l
the values seem to be 10_000 for size and 5_000 for exceed by

nick-mobilecoin (approved):

.internal-ci/helm/fog-services/values.yaml
```diff
@@ -152,18 +158,24 @@ fogView:
...
   main:
-    shardSize: 400_000
-    exceedBlockHeightBy: 200_000
-    shardOverlap: 200_000
+    shardSize: 800_000
+    exceedBlockHeightBy: 10_000
```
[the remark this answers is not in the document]
jgreat: Good catch. should be 400_000

.internal-ci/helm/fog-services/values.yaml
jgreat: I'll fix that reference

aweis89 (approved):
LGTM

PR #3523, opened 2023-08-23
Use `EnclaveReportDataContents` for quoting

nick-mobilecoin:
Previously a `QuoteNonce` was passed back from generating an enclave
report and that nonce was used in generating a quote as well as in
verifying the quote. Now an `EnclaveReportDataContents` is passed back
from an enclave report and used to quote and verify the quote. The
reason for this is to support storing more data than the 64 byte
`ReportData` member of the Quote.


nick-mobilecoin:
* master
  * **PR #3523** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3523" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3523?utm_source=stack-comment).

crypto/ake/enclave/src/lib.rs
```diff
@@ -645,77 +644,61 @@ impl<EI: EnclaveIdentity> AkeEnclaveState<EI> {
...
         let mut report_data = ReportData::default();
         let report_data_bytes: &mut [u8] = report_data.as_mut();
-        let identity = self.get_kex_identity();
-        let identity_bytes: &[u8] = identity.as_ref();
-        let custom_identity_bytes = self.custom_identity.get_bytes_for_report();
+        let identity_bytes: &[u8] = report_contents.key().as_ref();
         report_data_bytes[..identity_bytes.len()].copy_from_slice(identity_bytes);
-        report_data_bytes[identity_bytes.len()..].copy_from_slice(custom_identity_bytes.as_ref());
+        report_data_bytes[identity_bytes.len()..]
+            .copy_from_slice(report_contents.custom_identity());
```
nick-mobilecoin, line 658: DCAP will be storing a hash of the `report_contents` in the `ReportData`, which means DCAP will need to provide the `report_contents` as part of it's version of the `VerificationReport`.

This EPID implementation was not changed to prevent breaking its `VerificationReport`.

.cargo/config
```diff
@@ -19,3 +19,11 @@ rustflags = ["-C", "target-cpu=skylake", '--cfg=curve25519_dalek_backend="simd"'
 [cargo-new]
 name = "MobileCoin"
 email = ""
+
```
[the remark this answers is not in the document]
nick-mobilecoin: This has been reverted now

attest/core/src/nonce.rs
```diff
@@ -2,36 +2,31 @@
...
 
 /// A trait used to define common operations on nonce values
 pub trait Nonce:
-    AsRef<[u8]>
-    + PartialEq
-    + Sized
-    + for<'bytes> TryFrom<&'bytes [u8]>
-    + TryFrom<Vec<u8>>
-    + ConstantTimeEq
+    AsRef<[u8]> + PartialEq + Sized + for<'bytes> TryFrom<&'bytes [u8]> + ConstantTimeEq
```
nick-mobilecoin, line 20: The `TryFrom<Vec<u8>>` was unecessary since it created a new vec anyway.
This is overall more ergonomic with  `TryFrom<&'bytes [u8]>`.

dolanbernard (approved):

PR #3525, opened 2023-08-23
fix: (helm) bump fog-Ingest startup probe

jgreat:
### Motivation

Bump startup probe to 300 attempts (3000 seconds) to compensate for long lmdb bootstrap download at initial container startup.


nick-mobilecoin (commented):
Just to double check, we want to bump this from 5 minutes to 50 minutes.

jgreat:
> Just to double check, we want to bump this from 5 minutes to 50 minutes.

Yes.  We ran into an issue where the lmdb download takes longer than 5 min to finish. We need the startup phase of the checks to wait longer than the download.  

30 tries to 300 was just a nice round bump, but I guess its a bit odd on the amount of time.

PR #3526, opened 2023-08-23
fix: (docker) download lmdb bootstrap files to same volume as destination

jgreat:
### Motivation

We download the lmdb database from block storage to a temporary location so we don't end up with a corrupted database in case the start up process is interrupted.  The problem is moving the files from /tmp to the data volume was taking longer than the original curl. 

Download to the same volume/filesystem, then the `mv` is just changing pointers rather than transferring GBs of data.

Bonus:
- consolidate logic into 1 script used by the 3 containers. 
- clean up paths for external installed binaries.

### DIFF
--- a/.internal-ci/helm/mobilecoind/values.yaml
+++ b/.internal-ci/helm/mobilecoind/values.yaml
@@ -43,7 +43,7 @@ mobilecoind:
     # MC_LEDGER_DB_URL: ''
     # MC_WATCHER_DB_URL: ''
 
-  initContainers: {}
+  initContainers: []
 
   nodeSelector: {}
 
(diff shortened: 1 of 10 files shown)


nick-mobilecoin (approved):

.internal-ci/helm/mobilecoind/values.yaml
```diff
@@ -38,37 +38,12 @@ mobilecoind:
...
-    - -c
-    - |
-      set -x
-      if [ -f /mobilecoin/ledger/data.mdb ]; then
-        /usr/bin/mc-ledger-migration --ledger-db /data/ledger
-      fi
-    volumeMounts:
-    - name: data
-      mountPath: /data
+  initContainers: []
```
joekottke, line 46: Where does the migrate happen after this change?  Is it happening in the main process?

joekottke: Offline discussion:  mobilecoind does the migration itself.

joekottke (approved):

PR #3530, opened 2023-08-31
Fog Ledger Enclave Health Check

dolanbernard:
* Use regular `get_identity` calls to monitor health of Fog Ledger enclave

### Motivation

There is an issue that is likely in the enclave code that is causing the enclave to crash in rare cases. This should be fixed, but when this crash happens, the enclave never recovers. This PR adds a simple check to ensure the enclave is running and makes sure it gets restarted if it isn't. While the root cause of the issue is also important, it is very rare and difficult to re-create. The fact that there is no recovery from this is also a problem. This PR fixes the fact that there is no recovery from the enclave either crashing completely or not being in a healthy state.

### Future Work

Monitor network logs for this error. Figure out what the cause of this particular crash is and fix it.

### DIFF
--- a/fog/ledger/server/src/bin/router.rs
+++ b/fog/ledger/server/src/bin/router.rs
@@ -66,7 +66,7 @@ fn main() {
     loop {
         std::thread::sleep(std::time::Duration::from_millis(1000));
         if enclave.get_identity().is_err() {
-            mc_common::logger::log::error!(logger, "get_identity call to ledger enclave failed. Enclave may not be running or is not in a healthy state.");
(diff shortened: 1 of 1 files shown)


joekottke:
@dolanbernard Shouldn't this be a PR against the `release/v5.0` branch?  I think this is causing an issue with the container, as what is in master is DCAP focused, I think?  And therefor the consensus nodes are coming up with this error:  `error while loading shared libraries: libsgx_dcap_ql.so.1: cannot open shared object file: No such file or directory`

PR #3537, opened 2023-09-06
Add `EnclaveReportDataContents` to `DcapEvidence`

nick-mobilecoin:
The `DcapEvidence` now contains a member `report_data`, which holds
`EnclaveReportDataContents`. This allows the `DcapEvidence` to provide
the report data used by clients for things like public key identities.

### DIFF
--- a/attest/verifier/types/src/verification.rs
+++ b/attest/verifier/types/src/verification.rs
@@ -30,7 +30,12 @@ const TAG_DCAP_EVIDENCE_QUOTE3: u32 = 1;
 const TAG_DCAP_EVIDENCE_COLLATERAL: u32 = 2;
 const TAG_DCAP_EVIDENCE_REPORT_DATA: u32 = 3;
 
-// Quote3 and Collateral cannot trivially be made to implement prost::Message.
+// Quote3 and Collateral cannot trivially be made to implement prost::Message:
+//    - Quote3 is a already an opaque byte array which implements serde. Prost
+//      isn't brought in to mc-sgx-dcap-types to minimize dependencies.
+//    - Collateral is composed of X509-cert types which do not implement
+//      prost::Message.
+//
 // Since they implement serde Serialize and Deserialize though, we can manually
 // implement it for DcapEvidence. To do this, we use serde to serialize and
 // deserialize them to/from Vec<u8>
--- a/attest/verifier/types/src/verification.rs
+++ b/attest/verifier/types/src/verification.rs
@@ -23,10 +23,12 @@ use sha2::{Digest, Sha256};
 pub struct DcapEvidence {
     pub quote: Option<Quote3<Vec<u8>>>,
     pub collateral: Option<Collateral>,
+    pub report_data: Option<EnclaveReportDataContents>,
 }
 
 const TAG_DCAP_EVIDENCE_QUOTE3: u32 = 1;
 const TAG_DCAP_EVIDENCE_COLLATERAL: u32 = 2;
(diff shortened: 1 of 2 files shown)


nick-mobilecoin:
* master
  * **PR #3537** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3537" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3537?utm_source=stack-comment).

attest/verifier/types/src/verification.rs
```diff
@@ -44,6 +46,9 @@ impl Message for DcapEvidence {
         let collateral_bytes: Vec<u8> =
             mc_util_serial::serialize(&self.collateral).expect("Failed to serialize Collateral");
         encoding::bytes::encode(TAG_DCAP_EVIDENCE_COLLATERAL, &collateral_bytes, buf);
+        let report_data_bytes: Vec<u8> = mc_util_serial::serialize(&self.report_data)
+            .expect("Failed to serialize EnclaveReportDataContents");
+        encoding::bytes::encode(TAG_DCAP_EVIDENCE_REPORT_DATA, &report_data_bytes, buf);
```
nick-mobilecoin, line 51: I think this will likely be the last field added to the DcapEvidence.
While I'm a fan of the 3 strike rule:
1. it needs to be there.
2. it might be duplicate logic but ok.
3. your out and should probably dry up the code.

I don't think trying to macroizing this for 3 fields will gain us much and will likely make it harder to understand what's going on.
At the same time I'm not opposed to it if we feel it's worth it.

attest/verifier/types/src/verification.rs
```diff
@@ -23,10 +23,12 @@ use sha2::{Digest, Sha256};
...
     pub quote: Option<Quote3<Vec<u8>>>,
     pub collateral: Option<Collateral>,
+    pub report_data: Option<EnclaveReportDataContents>,
 }
 
 const TAG_DCAP_EVIDENCE_QUOTE3: u32 = 1;
 const TAG_DCAP_EVIDENCE_COLLATERAL: u32 = 2;
+const TAG_DCAP_EVIDENCE_REPORT_DATA: u32 = 3;
 
 // Quote3 and Collateral cannot trivially be made to implement prost::Message.
```
eranrund, line 33: I know this existed before this PR but I think mentioning why that is the case would be helpful for people looking at this in the future.

attest/verifier/types/src/verification.rs
eranrund: I think macroizing this is unnecessary and will just make reading it harder.

attest/verifier/types/src/verification.rs
eranrund: Might also be worth mentioning why this needs Prost at all

eranrund (approved):

attest/verifier/types/src/verification.rs
nick-mobilecoin: > I know this existed before this PR but I think mentioning why that is the case would be helpful for people looking at this in the future.

I'm not sure I follow:
Are you saying add some comments better describing the protobuf tag numbering, or are you saying we want to better describe the usage of protobof to serialize the `DcapEvidence` between endpoints?

eranrund: I'm requesting to mention why Quote3 and Collateral cannot trivially be made to implement prost::Message.

eranrund: Thanks!

PR #3545, opened 2023-09-11
Helm chart upgrades

aweis89:
- Fixes shardOverlap templating to use correct value
- Adds name override to reduce name & label length of shards and routers
- Adds a Delete retention policy to the ledger router and ledger shards PVC cleanup

### Future Work

[Soundtrack of this PR]()


jgreat (approved):
LGTM

PR #3554, opened 2023-09-13
Add `Digestible` impl to `EnclaveReportDataContents`

nick-mobilecoin:
The `EnclaveReportDataContents` is a member of `DcapEvidence` which will
need to be part of `BlockMetadataContents` which derives `Digestible`.
In order to support this future need a `Digestible` implementation has
been added to the `EnclaveReportDataContents`.

Also add `EnclaveReportDataContents` to the external protobufs. To
accomodate the need of it being part of the `BlockMetadataContents` in
the protobufs.

### DIFF
--- a/attest/verifier/types/src/verification.rs
+++ b/attest/verifier/types/src/verification.rs
@@ -6,7 +6,7 @@ use alloc::{string::String, vec::Vec};
 use base64::{engine::general_purpose::STANDARD as BASE64_ENGINE, Engine};
 use core::fmt::{Debug, Display};
 use hex_fmt::{HexFmt, HexList};
-use mc_crypto_digestible::Digestible;
+use mc_crypto_digestible::{DigestTranscript, Digestible};
 use mc_crypto_keys::X25519Public;
 use mc_sgx_core_types::QuoteNonce;
 use mc_sgx_dcap_types::{Collateral, Quote3};
@@ -39,6 +39,8 @@ const TAG_DCAP_EVIDENCE_REPORT_DATA: u32 = 3;
 // Since they implement serde Serialize and Deserialize though, we can manually
 // implement it for DcapEvidence. To do this, we use serde to serialize and
 // deserialize them to/from Vec<u8>
+// This is implementation is tested in `attest/untrusted/src/sim.rs` as the
+// quote and collateral are not trivial to construct.
 impl Message for DcapEvidence {
     fn encode_raw<B>(&self, buf: &mut B)
     where
@@ -270,13 +272,13 @@ impl Message for VerificationSignature {
 }
 
 /// Structure for holding the contents of the Enclave's Report Data.
-/// The Enclave's ReportData member contains a SHA256 hash of this structure's
-/// contents.
+/// The Enclave Quote's ReportData member contains a SHA256 hash of this
+/// structure's contents.
 #[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq)]
 pub struct EnclaveReportDataContents {
     nonce: QuoteNonce,
     key: X25519Public,
-    custom_identity: [u8; 32],
+    custom_identity: Option<[u8; 32]>,
 }
 
 impl EnclaveReportDataContents {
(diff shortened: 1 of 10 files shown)


nick-mobilecoin:
* master
  * **PR #3554** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3554" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3554?utm_source=stack-comment).

attest/verifier/types/src/verification.rs
```diff
@@ -270,13 +272,13 @@ impl Message for VerificationSignature {
...
-/// The Enclave's ReportData member contains a SHA256 hash of this structure's
-/// contents.
+/// The Enclave Quote's ReportData member contains a SHA256 hash of this
+/// structure's contents.
 #[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq)]
 pub struct EnclaveReportDataContents {
     nonce: QuoteNonce,
     key: X25519Public,
-    custom_identity: [u8; 32],
+    custom_identity: Option<[u8; 32]>,
```
nick-mobilecoin, line 281: I updated this to an option as it's only actually used by the fog enclaves, the other ones currently use a "null" identity that's all zeros, but we could change this in the future to be clearer that there is no custom id.

attest/verifier/types/src/verification.rs
```diff
@@ -309,8 +315,8 @@ impl EnclaveReportDataContents {
     }
 
     ///  Get the custom identity
-    pub fn custom_identity(&self) -> &[u8; 32] {
-        &self.custom_identity
+    pub fn custom_identity(&self) -> Option<&[u8; 32]> {
+        self.custom_identity.as_ref()
     }
 
     /// Returns a SHA256 hash of the contents of this structure.
```
nick-mobilecoin, line 322: The choice of sha256 was more or less arbitrary, if desired we could remove this and leverage the `Digestible` implementation. As it feels like a bit of duplicate work determining what gets hashed/digested in both implementations.

At the same time I don't fully understand when one would prefer a `MerlinTranscript` over `sha256`.

attest/verifier/types/src/verification.rs
```diff
@@ -349,4 +379,72 @@ mod tests {
...
+        assert_eq!(digest_1, digest_2);
+
+        let mut modified_nonce = nonce.clone();
+        let nonce_bytes: &mut [u8] = modified_nonce.as_mut();
+        nonce_bytes[0] += 1;
+        let modified_nonce_report_data =
+            EnclaveReportDataContents::new(modified_nonce, key.clone(), custom_identity);
+
+        let modified_nonce_digest = modified_nonce_report_data.digest32::<MerlinTranscript>(b"");
+        assert_ne!(digest_1, modified_nonce_digest);
```
nick-mobilecoin, line 406: I'm not a fan of workflow or recipe tests, where one sets up the base case, then toggles a lever and re-asserts something. I used a workflow test here as it felt a bit heavy to repeat the 10 line setup in mutliple test cases and I'm not sure functionalizing it would be much better.

That said if one can think of a clearer or more concise way to test, I'm open for suggestions.

attest/verifier/types/src/verification.rs
eranrund: I think when trying to hash structs, the digestible stuff is considered safer since it also includes a domain separator, field names, lengths of arrays, etc.

attest/verifier/types/src/verification.rs
eranrund: I think it's fine in this case

eranrund (approved):

attest/verifier/types/src/verification.rs
nick-mobilecoin: I created https://mobilecoin-eng.atlassian.net/browse/CORE-59 to re-visit consolidating the hash/digest usage

PR #3555, opened 2023-09-13
Add `Digestible` impl for `DcapEvidence`

nick-mobilecoin:
The `DcapEvidence` will need to be part of BlockMetadataContents which
derives Digestible.
In order to support this future need a Digestible implementation has
been added to the `DcapEvidence`.

Also add `DcapEvidence` to the external protobufs. To
accommodate the need of it being part of the `BlockMetadataContents` in
the protobufs.


nick-mobilecoin:
* master
  * **PR #3554** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3554" target="_blank">Graphite</a> 
    * **PR #3555** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3555" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3555?utm_source=stack-comment).

api/src/convert/dcap_evidence.rs
```diff
@@ -0,0 +1,66 @@
...
+}
+
+impl TryFrom<&external::DcapEvidence> for DcapEvidence {
+    type Error = ConversionError;
+    fn try_from(src: &external::DcapEvidence) -> Result<Self, Self::Error> {
+        let quote: Option<Quote3<Vec<u8>>> = mc_util_serial::deserialize(src.get_quote())
+            .map_err(|_| ConversionError::InvalidContents)?;
+        if quote.is_none() {
+            return Err(ConversionError::MissingField("quote".to_string()));
+        }
```
nick-mobilecoin, line 31: :question: I'm curious why we have `prost` implementations and these `protobuf` ones.
The only reason that `quote` is an option on `DcapEvidence` is to appease the prost derive. It seems like it would be better to use `external::DcapEvidence` everwhere that we want to serialize `DcapEvidence` and then use this `TryFrom` to fail for missing fields. As it currently stands one could construct and pass around a `DcapEvidence` without a `quote` which should never happen in a well behaving system.

attest/untrusted/src/sim.rs
```diff
@@ -53,37 +53,51 @@ impl SimQuotingEnclave {
...
+        collateral_from_from_raw_data(
+            SIM_TCB_INFO,
+            SIM_QE_IDENTITY,
+            SIM_CRL,
+            IAS_SIM_SIGNING_CHAIN,
+        )
     }
 }
 
+fn collateral_from_from_raw_data(
```
nick-mobilecoin, line 65: Up for naming suggestion here, I wanted to make it easier to construct one off `Collateral`s for the tests below

attest/untrusted/src/sim.rs
```diff
@@ -302,4 +318,125 @@ mod test {
...
+        let modified_quote =
+            Quote3::try_from(quote_bytes).expect("Failed to create modified quote");
+        let modified_quote_evidence = DcapEvidence {
+            quote: Some(modified_quote),
+            collateral: Some(collateral.clone()),
+            report_data: Some(report_data.clone()),
+        };
+
+        let modified_quote_digest = modified_quote_evidence.digest32::<MerlinTranscript>(b"");
+        assert_ne!(digest_1, modified_quote_digest);
```
nick-mobilecoin, line 375: Similar to #3554 not fond of this workflow style test, but couldn't come up with a nicer way.

api/src/convert/dcap_evidence.rs
```diff
@@ -0,0 +1,66 @@
...
+
+use crate::{external, ConversionError};
+use mc_attest_verifier_types::DcapEvidence;
+use mc_sgx_dcap_types::{Collateral, Quote3};
+
+impl From<&DcapEvidence> for external::DcapEvidence {
+    fn from(src: &DcapEvidence) -> Self {
+        let mut dst = Self::new();
+
+        let quote = mc_util_serial::serialize(&src.quote).expect("Invalid Dcap Evidence");
```
eranrund, line 13: This feels a bit funky to me, but I'm not sure what else we can do here without getting all the SGX DCAP structs to have a proto version...
Doing it like this pretty much means only Rust code can ever do anything with this field. Not necessarily a problem, but I think we don't have any places where that's case (I might be wrong).

api/src/convert/dcap_evidence.rs
eranrund: Can't `quote` be made required in the prost derive and then it will no longer be an option? 

And on whether why we have both `prost` and `protobuf` - I think we use `prost` for protobuf messages we want to use inside the enclave (since `prost` is no_std compatible), but in some cases we also want to use the message in the GRPC api (`grpcio` stuff), which currently requires `protobuf` (they have a `prost` feature but we haven't tried switching to it).

api/src/convert/dcap_evidence.rs
eranrund: Having all the SGX stuff implement prost (maybe with a feature flag) would be nice... I know it's a non-trivial amount of work but that's what we've done in many other places :/

api/src/convert/dcap_evidence.rs
eranrund: You mean the issue is a missing `Default` implementation?

api/src/convert/dcap_evidence.rs
nick-mobilecoin: I think I'll pull it in and do it for quote and collateral.  We can come back and fill any others we need.

api/src/convert/dcap_evidence.rs
nick-mobilecoin: > You mean the issue is a missing `Default` implementation?
Yes sorry, partial thought in the original comment.
Prost uses `Self::default()` because everything is optional in protobufs so when an item is missing you need some kind of default either the `Option<T>` where None is a default or some kind of default value.

I may be trying to leverage the type system too much, but I feel if we allow creating a `DcapEvidence` without a structurally sound quote, collateral, and report data, then we force all consumers to both check for the evidence being in a valid format as well as having to verify (the signatures). It seems like we should limit when and where potentially "corrupt" data is passed around.

eranrund: I think consumers are going to have to check that anyway since there's no guarantees about anything inside a protobuf message...

api/src/convert/dcap_evidence.rs
nick-mobilecoin: @eranrund I created, https://mobilecoin-eng.atlassian.net/browse/CORE-58
My thought being that while this implementation isn't ideal as it stands, it can move us forward to getting more of the DCAP implementation in while `CORE-58` can be done anytime before releasing 6.0.0

Not sure your thoughts there?  I don't like putting something in we know we don't want in the final code, but one thing we've struggled with is ensuring we have tasks that others can work on. By not blocking subsequent work we ensure whoever is available can take up the prost/protobuf re-work.

eranrund: I'm okay with it if you trust us to not forget it. In many cases this is an easy decision (to postpone so we can keep making progress), but in this particular case the cost of forgetting to fix this would mean another block version and more annoying code to keep around forever. So I guess I'm fine with this approach if you trust the release process/your own process to ensure this is addressed before we go live.

nick-mobilecoin: I'm going to sit on this and hope I have an epiphany over the weekend.
I agree that forgetting this fix in the 6.0.0 release would be bad, I want to see how far I get on my task to convert over to DCAP quotes as I think that will inform what other outstanding tasks I've missed, thus indicating if where people should focus.

nick-mobilecoin:
Closing this as it's being re-worked in #3569

PR #3556, opened 2023-09-13
Add `DcapEvidence` to `BlockMetadataContents`

nick-mobilecoin:
The `BlockMetadataContents` has been updated to accomodate the new
`DcapEvidence` that will be provided for attesting with the DCAP api.

Due to the change in `BlockMetadataContents` the block version has been
bumped.


nick-mobilecoin:
* master
  * **PR #3554** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3554" target="_blank">Graphite</a> 
    * **PR #3555** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3555" target="_blank">Graphite</a> 
      * **PR #3556** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3556" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3556?utm_source=stack-comment).

api/src/convert/block_metadata.rs
```diff
@@ -24,13 +29,27 @@ impl TryFrom<&blockchain::BlockMetadataContents> for BlockMetadataContents {
...
+            }
+            (false, true) => {
+                let report = src.get_verification_report().try_into()?;
+                AttestationEvidence::VerificationReport(report)
+            }
+            _ => {
+                return Err(ConversionError::MissingField(
+                    "attestation_evidence".to_string(),
+                ))
+            }
```
nick-mobilecoin, line 45: While technically this captures `(true, true)` that shouldn't happen with a one of

blockchain/types/src/block_metadata.rs
```diff
@@ -22,9 +47,10 @@ pub struct BlockMetadataContents {
     #[prost(message, required, tag = 2)]
     quorum_set: QuorumSet,
 
-    /// IAS report for the enclave which generated the signature.
-    #[prost(message, required, tag = 3)]
-    verification_report: VerificationReport,
+    /// Attestation evidence for the enclave which generated the signature.
+    #[prost(oneof = "AttestationEvidence", tags = "3, 5")]
+    #[digestible(name = "verification_report")]
+    attestation_evidence: Option<AttestationEvidence>,
```
nick-mobilecoin, line 53: I think we need to make sure this is the path we want to go.

Another option is to create a new  enum that contains `BlockMetadataContents` and `BlockMetadataContentsDcap` or similar and then during decode we key off of the block version for which variant to decode.

nick-mobilecoin, line 53: The use of `Option` here is similar to the comment on #3555 on questioning if we should be using prost up here at this level. This really isn't optional, there should always be one of the variants, but to work with prost this has to be an option.

mobilecoind/src/service.rs
```diff
@@ -2609,7 +2609,7 @@ mod test {
         // Add a block with a non-MOB token ID.
         add_block_to_ledger(
             &mut ledger_db,
-            BlockVersion::THREE,
+            BLOCK_VERSION,
```
nick-mobilecoin, line 2612: This used to be adding a block version 3 after block version 4 was already in the chain. Ledger will error out in these instances, https://github.com/mobilecoinfoundation/mobilecoin/blob/90560096c7cb306758b2d22bd4c5eb9891d20b66/ledger/db/src/ledger_db.rs#L695

api/src/convert/block_metadata.rs
```diff
@@ -24,13 +29,27 @@ impl TryFrom<&blockchain::BlockMetadataContents> for BlockMetadataContents {
     fn try_from(src: &blockchain::BlockMetadataContents) -> Result<Self, Self::Error> {
         let block_id = src.get_block_id().try_into()?;
         let quorum_set = src.get_quorum_set().try_into()?;
-        let report = src.get_verification_report().try_into()?;
+        let attestation_evidence = match (src.has_dcap_evidence(), src.has_verification_report()) {
```
eranrund, line 32: This shouldn't be necessary - I suggest looking at the generated code (search for `blockchain.rs` under `target`).
You'll find an enum called `BlockMetadataContents_oneof_attestation_evidence`, and you will be able to match against `src.attestation_evidence` which is an `Option<BlockMetadataContents_oneof_attestation_evidence>`

blockchain/types/src/block_metadata.rs
eranrund: I like this better, but it needs to be tested carefully. We must make sure that the digestible output of an existing BlockMetadataContents remains the same.

EDIT: saw the test for this, nice!

blockchain/types/src/block_metadata.rs
```diff
@@ -129,3 +157,70 @@ impl BlockMetadata {
...
+                QuorumSetMember::Node(test_node_id(9)),
+                QuorumSetMember::Node(test_node_id(8)),
+                QuorumSetMember::Node(test_node_id(7)),
+            ],
+        );
+
+        let block_v3 = BlockMetadataContentsV3 {
+            block_id: BlockID([1; 32]),
+            quorum_set: quorum_set.clone(),
+            verification_report: VerificationReport::default(),
```
eranrund, line 204: Would be nice to test with something that is not all zeros, to represent a more real-world scenario.

eranrund (changes requested):
I think this is the right approach. The conversion stuff can be cleaned up a bit (no need to the `match (src.has....)` stuff), and the test case can be improved to be more representative of real world (I am very cautious when touching anything that involves maintaining digestive-backwards-compatible behavior)

I fully agree we  should review this particular change very carefully, so more eyes on it would be good.

api/src/convert/block_metadata.rs
nick-mobilecoin: Logic has been updated to look at the `attestation_evidence` field of the `blockchain::BlockMetadataContents`.

I didn't realize they were public fields, thanks!

blockchain/types/src/block_metadata.rs
nick-mobilecoin: Test case has been updated to use `run_with_several_seeds` and `make_verification_report()`

nick-mobilecoin:
> I think this is the right approach. The conversion stuff can be cleaned up a bit (no need to the `match (src.has....)` stuff), and the test case can be improved to be more representative of real world (I am very cautious when touching anything that involves maintaining digestive-backwards-compatible behavior)
> 
> I fully agree we should review this particular change very carefully, so more eyes on it would be good.

@eranrund I think I've addressed your comments, but didn't bump for re-review as I'm not sure if you blocked for the changes or to ensure more eyes happen here before merging

eranrund:
> > I think this is the right approach. The conversion stuff can be cleaned up a bit (no need to the `match (src.has....)` stuff), and the test case can be improved to be more representative of real world (I am very cautious when touching anything that involves maintaining digestive-backwards-compatible behavior)
> > I fully agree we should review this particular change very carefully, so more eyes on it would be good.
> 
> @eranrund I think I've addressed your comments, but didn't bump for re-review as I'm not sure if you blocked for the changes or to ensure more eyes happen here before merging

I'll take a look. I blocked for the changes but more eyes would be good given the impact of this particular PR.

api/src/convert/block_metadata.rs
```diff
@@ -24,13 +29,31 @@ impl TryFrom<&blockchain::BlockMetadataContents> for BlockMetadataContents {
     fn try_from(src: &blockchain::BlockMetadataContents) -> Result<Self, Self::Error> {
         let block_id = src.get_block_id().try_into()?;
         let quorum_set = src.get_quorum_set().try_into()?;
-        let report = src.get_verification_report().try_into()?;
+        let attestation_evidence = match &src.attestation_evidence {
+            Some(evidence) => {
```
eranrund, line 33: Nit: You can merge the match statements and match on `Some(blockchain::BlockMetadataContents_oneof_attestation_evidence::dcap_evidence(evidence))` etc

blockchain/types/src/block_metadata.rs
eranrund: I'm not sure its similar since this is a oneof field...

eranrund (approved):

api/src/convert/block_metadata.rs
nick-mobilecoin: oh that would be better

blockchain/types/src/block_metadata.rs
nick-mobilecoin: true, it's a prost limitation in that there is no default enum variant when there isn't a simple enum, so to get a default, `Option<T>` was used.

api/src/convert/block_metadata.rs
nick-mobilecoin: updated, note I brought in `BlockMetadataContents_oneof_attestation_evidence` to use directly as it allowed keeping the match expression on one line for cargo fmt

nick-mobilecoin:
closing in favor of #3577

PR #3565, opened 2023-09-19
Add prost implementations for `Quote3` and `Collateral`

nick-mobilecoin:
Add prost implementations for `Quote3` and `Collateral`. The
implementations are done in this repository instead of
`mc-sgx-dcap-types` to prevent having to publish proto files from the
`mc-stx-***` crates.

The `prost::Message` implementations use `prost-build` and a `*.proto`
file, which deviates from other `prost::Message` implementations in the
repository. The reasons for this deviation are:
- Prevent changing `Quote3` and `Collateral` to use `Option<T>`
  fields. `Option<T>` or defaults need to be implemented to work
  directly with `::prost::Message`.
- Ensure that the `::prost::Message` implementations use the same tags
  as the gRPCio protobuf implementations.

### DIFF
--- a/attest/api/proto/attest.proto
+++ b/attest/api/proto/attest.proto
@@ -52,7 +52,7 @@ message Message {
 /// An AEAD message with an explicit nonce.
 ///
 /// This message is technically compatible with [`Message`], but exists to
-// ensure generated code doesn't use Message.
+/// ensure generated code doesn't use Message.
 message NonceMessage {
         /// A byte array containing plaintext authenticated data.
         bytes aad = 1;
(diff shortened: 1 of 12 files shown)


nick-mobilecoin:
* master
  * **PR #3565** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3565" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3565?utm_source=stack-comment).

attest/verifier/types/src/convert/collateral.rs
```diff
@@ -0,0 +1,266 @@
...
+    Certificate,
+};
+
+impl TryFrom<prost::Collateral> for Collateral {
+    type Error = ConversionError;
+
+    fn try_from(value: prost::Collateral) -> Result<Self, Self::Error> {
+        // Note: sgx_collateral uses null bytes at the end of most arrays.
+
+        let mut sgx_collateral = version_3_1_empty_collateral();
```
nick-mobilecoin, line 23: I'm not fond of the low level building up of the collateral via the `sgx_ql_qve_collateral_t`. I think if this lived in `mc-sgx-dcap-types` it wouldn't be as questionable. At the same time I think building up the `Collateral` via a normal constructor maybe error prone as all 7 of the members should exist to be valid.

As the commit message says It felt like doing the prost types and conversion here in the main MobileCoin repo feels like the quickest way to get it done, while creating a binary representation that should be stable and ergonomic for the future.

attest/api/proto/attest.proto
```diff
@@ -65,3 +65,62 @@ message NonceMessage {
         /// The explicit nonce.
         fixed64 nonce = 4;
 }
+
```
nick-mobilecoin, line 68: My plan is to move the DcapEvidence and EncalveReportDataContents into this proto file as well so that they're available for the `mc-attest-verifier-types` in the enclave.

attest/verifier/types/build.rs
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) 2023 The MobileCoin Foundation
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    prost_build::compile_protos(&["attest.proto"], &["../../api/proto"])?;
```
nick-mobilecoin, line 4: I know this deviates from other prost usages, but I think we should be moving towards this, where the prost type is built from the *.proto file, instead of relying on maintainers to ensure the prost attribute tag matches that in the *.proto file. This also allows for the prost types to be made up of simple byte vectors while leaving the _common_ type to have typed fields

eranrund (approved):
I think this is actually quite elegant, and I agree its a good shift from the current situation where the proto and Rust struct can get out of sync.

PR #3566, opened 2023-09-19
Convert to/from protobuf types for `Quote3` and `Collateral`

nick-mobilecoin:
Add to/from implementations for the `Quote3` and `Collateral` in
`mc-attest-api`. This allows converting between the protobuf
representation and the common rust types.


nick-mobilecoin:
* master
    * **PR #3566** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3566" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3566?utm_source=stack-comment).

attest/api/src/convert/collateral.rs
```diff
@@ -0,0 +1,90 @@
...
+
+impl TryFrom<&Collateral> for attest::Collateral {
+    type Error = ConversionError;
+    fn try_from(src: &Collateral) -> Result<Self, Self::Error> {
+        let prost = prost::Collateral::try_from(src)?;
+        let bytes = prost.encode_to_vec();
+        let mut proto = Self::default();
+        proto
+            .merge_from_bytes(&bytes)
+            .expect("failure to merge means prost and protobuf are out of sync");
```
nick-mobilecoin, line 19: This is a try_from so I'm fine if we think an error is better here, It just feels like if this fails then the prost and gRPC protobuf implementations are broken.

attest/api/src/convert/collateral.rs
```diff
@@ -0,0 +1,90 @@
...
+use mc_attest_verifier_types::prost;
+use mc_sgx_dcap_types::Collateral;
+use mc_util_serial::Message;
+use protobuf::{CodedOutputStream, Message as ProtoMessage};
+
+impl TryFrom<&Collateral> for attest::Collateral {
+    type Error = ConversionError;
+    fn try_from(src: &Collateral) -> Result<Self, Self::Error> {
+        let prost = prost::Collateral::try_from(src)?;
+        let bytes = prost.encode_to_vec();
```
nick-mobilecoin, line 15: Like the comment in `encode_to_protobuf_vec` streaming to an actual bytestream and then decoding back seemed like the best way to make sure the prost and protobuf conversion logic was the same.  I know it takes a bit more memory and time to convert, but I'm guessing actually sending over the network will greatly overshadow this conversion time

attest/api/src/convert/quote3.rs
```diff
@@ -0,0 +1,51 @@
...
+use mc_sgx_dcap_types::Quote3;
+
+impl<T: AsRef<[u8]>> From<&Quote3<T>> for attest::Quote3 {
+    fn from(src: &Quote3<T>) -> Self {
+        let mut dst = Self::new();
+        dst.set_data(src.as_ref().to_vec());
+
+        dst
+    }
+}
```
nick-mobilecoin, line 15: Since quote as as_ref and can convert directly from bytes, I figured doing it directly instead of streaming like `Collateral` does would be fine and the quote is not likely to change

attest/api/src/convert.rs
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) 2018-2022 The MobileCoin Foundation
 
-//! Conversions from gRPC message types into consensus_enclave_api types.
+//! Conversions from gRPC message types into attest api types.
+pub mod collateral;
+pub mod quote3;
```
nick-mobilecoin, line 5: Note this file was renamed

dolanbernard (approved):

PR #3568, opened 2023-09-20
Move `EnclaveReportDataContents` protbuf to `mc-attest-api`

nick-mobilecoin:
Previously the `EnclaveReportDataContents` protobuf definition was in
`mc-api`, now it's in `mc-attest-api`.

This also adds a prost implementation for `EnclaveReportDataContents`.

PR #3569, opened 2023-09-20
Add post and protobuf implementations for `DcapEvidence`

nick-mobilecoin:
This removes the use of `Option<T>` in `DcapEvidence` delegating that to
the `prost::DcapEvidence` type.


nick-mobilecoin:
* master
    * **PR #3566** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3566" target="_blank">Graphite</a> 
      * **PR #3568** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3568" target="_blank">Graphite</a> 
        * **PR #3569** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3569" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3569?utm_source=stack-comment).

attest/api/src/convert.rs
```diff
@@ -181,12 +182,8 @@ impl From<EnclaveMessage<NonceSession>> for NonceMessage {
...
 /// ```
 pub(crate) fn encode_to_protobuf_vec<T: ProtoMessage>(msg: &T) -> Result<Vec<u8>, ConversionError> {
-    let mut bytes = vec![];
-    let mut stream = CodedOutputStream::vec(&mut bytes);
-    msg.write_to_with_cached_sizes(&mut stream)
-        .map_err(|e| ConversionError::Other(e.to_string()))?;
-    stream
-        .flush()
+    let bytes = msg
+        .write_to_bytes()
```
nick-mobilecoin, line 186: Note this change. I missed the `write_to_bytes()` function originally and the `write_to_with_cached_sizes()` specifically calls out
> Sizes of this messages and nested messages must be cached by calling `compute_size` prior to this call.

https://github.com/stepancheg/rust-protobuf/blob/f7259280e1b64a8bafd505b688d35f93552d0a38/protobuf/src/message.rs#L40

While wrapping this one function call `write_to_bytes()` with `encode_to_protobuf_vec()` is a bit questionable normally, I think since alot of the Message API is C like with temporally dependent functions keeping this wrapped is probably best.

attest/verifier/types/src/verification.rs
```diff
@@ -11,125 +17,23 @@ use mc_crypto_keys::X25519Public;
...
+    pub quote: Quote3<Vec<u8>>,
+    pub collateral: Collateral,
+    pub report_data: EnclaveReportDataContents,
 }
 
 #[derive(Clone, Oneof)]
 pub enum EvidenceKind {
     #[prost(message, tag = "4")]
-    Dcap(DcapEvidence),
+    Dcap(prost::DcapEvidence),
```
nick-mobilecoin, line 33: Note that this now uses the `prost::DcapEvidence`, this will force end users to convert to the standard `DcapEvidence`, but it's about the same as them having to check all 3 options that used to be there.

attest/verifier/types/src/verification.rs
```diff
@@ -11,125 +17,23 @@ use mc_crypto_keys::X25519Public;
...
-use prost::{
-    bytes::{Buf, BufMut},
-    encoding::{self, DecodeContext, WireType},
-    DecodeError, Message, Oneof,
-};
 use serde::{Deserialize, Serialize};
 use sha2::{Digest, Sha256};
 
-#[derive(Clone, Debug, Default, Deserialize, Eq, PartialEq, Serialize)]
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
```
nick-mobilecoin, line 23: If so desired we could use the `prost::DcapEvidence` everywhere and not need this type. It would put us more or less where we were.  I like having the type with non optional fields, but I'm fine seeing the other approach too.

eranrund: I like the non-optional fields as well.

eranrund (approved):

PR #3570, opened 2023-09-20
Move `Digestible` implementation to `prost::EnclaveReportDataContents`

nick-mobilecoin:
Previously the `Digestible` implementation was on the
`EnclaveReportDataContents`. Now the `Digestible` implementation is on
the `prost::EnclaveReportDataContents`. This is because the prost
version is what will be needed in the `BlockMetadataContents` which
requires `Digestible` on all decedents.

### DIFF
--- a/attest/verifier/types/src/convert/enclave_report_data_contents.rs
+++ b/attest/verifier/types/src/convert/enclave_report_data_contents.rs
@@ -62,16 +62,19 @@ impl Digestible for prost::EnclaveReportDataContents {
         let typename = b"EnclaveReportDataContents";
         transcript.append_agg_header(context, typename);
 
-        let Self {nonce, key, custom_identity} = self;
-        transcript.append_primitive(context, b"nonce", &nonce);
-        transcript.append_primitive(context, b"key", &key);
+        let Self {
+            nonce,
+            key,
+            custom_identity,
+        } = self;
+        transcript.append_primitive(context, b"nonce", nonce);
+        transcript.append_primitive(context, b"key", key);
         // Since custom identity is optional we only include it if it has data.
         if !self.custom_identity.is_empty() {
-            transcript.append_primitive(context, b"custom_identity", &custom_identity);
+            transcript.append_primitive(context, b"custom_identity", custom_identity);
         }
 
         transcript.append_agg_closer(context, typename);
-
     }
 }
 
--- a/attest/verifier/types/src/convert/enclave_report_data_contents.rs
+++ b/attest/verifier/types/src/convert/enclave_report_data_contents.rs
@@ -61,13 +61,17 @@ impl Digestible for prost::EnclaveReportDataContents {
     ) {
         let typename = b"EnclaveReportDataContents";
         transcript.append_agg_header(context, typename);
-        transcript.append_primitive(context, b"nonce", &self.nonce);
-        transcript.append_primitive(context, b"key", &self.key);
+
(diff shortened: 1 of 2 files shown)


nick-mobilecoin:
* master
    * **PR #3566** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3566" target="_blank">Graphite</a> 
      * **PR #3568** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3568" target="_blank">Graphite</a> 
        * **PR #3569** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3569" target="_blank">Graphite</a> 
          * **PR #3570** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3570" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3570?utm_source=stack-comment).

attest/verifier/types/src/convert/enclave_report_data_contents.rs
```diff
@@ -51,11 +51,32 @@ impl From<&EnclaveReportDataContents> for prost::EnclaveReportDataContents {
...
+        context: &'static [u8],
+        transcript: &mut DT,
+    ) {
+        let typename = b"EnclaveReportDataContents";
+        transcript.append_agg_header(context, typename);
+        transcript.append_primitive(context, b"nonce", &self.nonce);
+        transcript.append_primitive(context, b"key", &self.key);
+        // Since custom identity is optional we only include it if it has data.
+        if !self.custom_identity.is_empty() {
+            transcript.append_primitive(context, b"custom_identity", &self.custom_identity);
```
nick-mobilecoin, line 68: There is an `append_to_transcript` impl for `Vec<u8>` but it always uses `bytes` as the typename, which seems like a less than ideal way to ensure consistency. I.e. if prost happens to re-order the members while keeping the tags.

attest/verifier/types/src/convert/enclave_report_data_contents.rs
```diff
@@ -51,11 +51,32 @@ impl From<&EnclaveReportDataContents> for prost::EnclaveReportDataContents {
...
+        }
+        transcript.append_agg_closer(context, typename);
```
eranrund, line 70: I was wondering if there's a way to protect us from forgetting to add new fields here if we ever evolve the protobuf, and came up with this with seems to work and will complain if a new field got added.

```suggestion
        let typename = b"EnclaveReportDataContents";
        transcript.append_agg_header(context, typename);

        let Self {nonce, key, custom_identity} = self;
        transcript.append_primitive(context, b"nonce", &nonce);
        transcript.append_primitive(context, b"key", &key);
        // Since custom identity is optional we only include it if it has data.
        if !self.custom_identity.is_empty() {
            transcript.append_primitive(context, b"custom_identity", &custom_identity);
        }

        transcript.append_agg_closer(context, typename);

```

eranrund (approved):

attest/verifier/types/src/convert/enclave_report_data_contents.rs
nick-mobilecoin: oh, that's a cool use of destructuring.  and would definitly help ensure it, especially for the upcoming `Collateral` with 7 fields that would be good to know we got them all.

PR #3572, opened 2023-09-21
Add `Digestible` for `Collateral` and `Quote3`

nick-mobilecoin:
The prost implementations of `Collateral` and `Quote3` now support
`Digestible`.

### Future Work

[Soundtrack of this PR]()


nick-mobilecoin:
* master
  * **PR #3572** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3572" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3572?utm_source=stack-comment).

attest/verifier/types/src/convert/enclave_report_data_contents.rs
```diff
@@ -67,12 +67,9 @@ impl Digestible for prost::EnclaveReportDataContents {
...
         } = self;
-        transcript.append_primitive(context, b"nonce", nonce);
-        transcript.append_primitive(context, b"key", key);
-        // Since custom identity is optional we only include it if it has data.
-        if !self.custom_identity.is_empty() {
-            transcript.append_primitive(context, b"custom_identity", custom_identity);
-        }
+        nonce.append_to_transcript(context, transcript);
+        key.append_to_transcript(context, transcript);
+        custom_identity.append_to_transcript_allow_omit(context, transcript);
```
nick-mobilecoin, line 72: I had previously mentioned the "hesitation" about using `Vec<u8>`'s Digestible implementation, but I think it's a false assurance to manually call `append_primitive`
While we're using `Self{ a, b, c}` nothing guarantees that `a` is really an `A` type it could be any type since the compiler will elide it.  That's why I updated the test below to manually build up a digest and compare those, to help ensure that the order is really correct.
That's also why I'm no longer testing the difference of each field as the test building the digest ensures there are 2 places adding each field.

attest/verifier/types/src/convert/collateral.rs
```diff
@@ -138,12 +139,44 @@ fn version_3_1_empty_collateral() -> sgx_ql_qve_collateral_t {
...
+            qe_identity_issuer_chain,
+            qe_identity,
+        } = self;
+        pck_crl_issuer_chain.append_to_transcript(context, transcript);
+        root_ca_crl.append_to_transcript(context, transcript);
+        pck_crl.append_to_transcript(context, transcript);
+        tcb_info_issuer_chain.append_to_transcript(context, transcript);
+        tcb_info.append_to_transcript(context, transcript);
+        qe_identity_issuer_chain.append_to_transcript(context, transcript);
+        qe_identity.append_to_transcript(context, transcript);
```
eranrund, line 166: The context should be the field name
e.g. see https://github.com/mobilecoinfoundation/mobilecoin/blob/master/blockchain/types/src/block.rs#L188

eranrund: Looks like I missed this in previous reviews so that should also get addressed

eranrund (changes requested):

attest/verifier/types/src/convert/enclave_report_data_contents.rs
nick-mobilecoin: NVM, the context should have the field name, I misunderstood the usage, fixed now.

attest/verifier/types/build.rs
```diff
@@ -1,6 +1,24 @@
...
+    // `Digestible` should be stable. There are tests in each of the
+    // `convert/<type_name>.rs` files that help ensure the `Digestible` field
+    // order stability.
+    for t in ["EnclaveReportDataContents", "Quote3", "Collateral"].iter() {
+        config.type_attribute(
+            t,
+            "#[derive(serde::Serialize, serde::Deserialize, Digestible, Eq)]",
+        );
+    }
+    config.compile_protos(&["attest.proto"], &["../../api/proto"])?;
```
nick-mobilecoin, line 22: It looks like this is another instance of Nick didn't RTFM.

I think deriving is probably a better way to go. I like having the tests for ensuring the backwards compatibility.

Also I added serde and Eq here as they'll be needed soonish and figured it would keep the noise down on future PRs

attest/verifier/types/src/convert/enclave_report_data_contents.rs
```diff
@@ -162,80 +136,42 @@ mod tests {
...
-            nonce: nonce_bytes.to_vec(),
-            key: key_bytes.to_vec(),
-            custom_identity: vec![],
-        };
-
-        let report_data_with_zeroed_custom_id = prost::EnclaveReportDataContents {
-            nonce: nonce_bytes.to_vec(),
-            key: key_bytes.to_vec(),
-            custom_identity: zeroed_custom_identity.to_vec(),
+        // We manually build up the digest here, to help ensure that the digest
```
nick-mobilecoin, line 139: Note I removed the "without_custom_id" version as deriving Digestible defaults to Strings and Vecs being omittable.
Having an `EnclaveReportDataContents` in flight missing the key or nonce is a failure case that doesn't matter too much what the digest ends up being, so it seems unnecessary to test for presence or absence of all the fields.

attest/verifier/types/src/convert/collateral.rs
nick-mobilecoin: Good catch!

I've updated the implementation to derive `Digestible`, but kept the tests (with correct contexts).

attest/verifier/types/build.rs
eranrund: Woo! Super cool that this is possible.

eranrund (approved):

nick-mobilecoin:
### Merge Activity

* **Sep 21, 5:31 PM**: @nick-mobilecoin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3572).

PR #3577, opened 2023-09-22
Add `AttestationEvidence` to `BlockchainMetadata`

nick-mobilecoin:
Previously the `BlockchainMetadata` contained a filed
`VerificationReport`. With DCAP there is a new type of attestation
evidence `DcapEvidence`. The `BlockchainMetadata` has been updated to
take an `AttestationEvidence` which is a oneof the older
`VerificationReport` or the newer `DcapEvidence`

PR #3579, opened 2023-09-25
Moving the `DcapEvidence` protobufs to `mc-api`

nick-mobilecoin:
Previously the protobufs for `DcapEvidence` and its sub-structs were in
`mc-attest-api`. Due to dependencies in `mc-attest-api` having the
`DcapEvidence` type defined there would result in the wasm builds failing
once types in `mc-api` need to depend on `mc-attest-api`. To prevent
this future failure the `DcapEvidence` is brought into the `mc-api`
crate. This is the crate that houses the `VerificationReport` which the
`DcapEvidence` is meant to replace.


nick-mobilecoin:
* master
  * **PR #3579** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3579" target="_blank">Graphite</a>  👈
    * **PR #3577** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3577" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3579?utm_source=stack-comment).

attest/verifier/types/build.rs
```diff
@@ -26,6 +26,6 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
             "#[derive(serde::Serialize, serde::Deserialize, Digestible, Eq)]",
         );
     }
-    config.compile_protos(&["attest.proto"], &["../../api/proto"])?;
+    config.compile_protos(&["external.proto"], &["../../../api/proto"])?;
```
nick-mobilecoin, line 29: I had initially put the `DcapEvidence` and sub types in `mc-attest-api` so that this wasn't going so far up the stack to get to the proto file.
However due to protobufs need for `std` and the enclaves need for `no_std` this quickly became a dependency issue when I tried to bring in the changes for #3577 .

Ideally I think the conversions, https://github.com/mobilecoinfoundation/mobilecoin/blob/b66be56ef34d1b7ce4e8dfdc6d33117e522f1ab2/api/src/convert/mod.rs#L10 should go next to the Rust types. As right now if one wants the protobufs for `foo` they pull in `mc-api` and also end up pulling in the protobufs and backend crates for `bar`, `baz`, etc.  
It wasn't immediately apparent and I chased my tail a bit, I believe the reason for this less than ideal dependency setup is because `protobufs` requires `std` so you can't put the `To/From` definitions next to the rust types as many of these types need to be used in an enclave. Due to the orphan rule, this means the only other place you can put the `To/From` implementations is next to the protobuf definitions.

eranrund: Thanks for explaining whats going on here. It makes sense and is indeed not ideal but I am not sure if there are many options here, the only other solution I can think of is perhaps putting the protobuf(std) stuff behind a feature flag.

eranrund (approved):

attest/verifier/types/build.rs
nick-mobilecoin: I think I'll leave this as is for now. It follows what the other types are doing and doesn't bring in any extra dependencies.
Ideally we replace grpc with tonic. grpc is one of our build/compilation bottlenecks, the protobuf stuff is `std` and it looks to change between 2 and 3. Prost seems to be more stable and simpler.  I'm just not sure we can justify the cost to move from grpc for a while.

nick-mobilecoin:
### Merge Activity

* **Sep 26, 10:56 AM**: @nick-mobilecoin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3579).

PR #3583, opened 2023-09-26
Chant type of `mc-attest-verifier::Error::Verification`

nick-mobilecoin:
Previously the `mc-attest-verifier::Error::Verification` type wrapped a
`VerificationReportData`. Now it wraps a `String` type. This allows for
a more generic interface to provide back the error from attempting to
verify the legacy `VerificationReportData` or the newer `DcapEvidence`.

Verification of the newer `DcapEvidence` provides a string output that
more clearly states where verification failed and as such passing back
the evidence itself isn't very beneficial.

### DIFF
--- a/sgx/report-cache/untrusted/src/lib.rs
+++ b/sgx/report-cache/untrusted/src/lib.rs
@@ -8,8 +8,10 @@ use mc_attest_core::{
     PibError, ProviderId, QuoteError, QuoteSignType, VerificationReport, VerificationReportData,
     VerifyError,
 };
+use mc_attest_enclave_api::Error as AttestEnclaveError;
 use mc_attest_net::{Error as RaError, RaClient};
 use mc_attest_untrusted::{QuotingEnclave, TargetInfoError};
+use mc_attest_verifier::Error as VerifierError;
 use mc_common::logger::{log, o, Logger};
 use mc_sgx_report_cache_api::{Error as ReportableEnclaveError, ReportableEnclave};
 use mc_util_metrics::IntGauge;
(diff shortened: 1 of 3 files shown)


nick-mobilecoin:
* master
  * **PR #3583** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3583" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3583?utm_source=stack-comment).

sgx/report-cache/untrusted/src/lib.rs
```diff
@@ -208,39 +206,7 @@ impl<E: ReportableEnclave, R: RaClient> ReportCache<E, R> {
...
-                } else {
-                    Err(Error::ReportableEnclave(
-                        ReportableEnclaveError::AttestEnclave(AttestEnclaveError::Verify(
-                            VerifierError::Verification(report_data),
-                        )),
-                    ))
-                }
-            }
-            Err(other) => Err(other.into()),
+            Err(e) => Err(e.into()),
```
nick-mobilecoin, line 209: The newer DCAP attestation interface doesn't have this TCB update mechanism. It was the only place that the `VerificationReport` from the verification failure was looked at. Since it won't be used in the future just removed it instead of trying to re-implement it without the failed `VerificationReport`.

eranrund: How do you know that this isn't needed for normal operation while we are still not fully transitioned to DCAP?
I'm completely unfamiliar with this but there's probably a reason why this was here in the first place... and I have no idea what's going to break if its no longer there and we're still not switched to DCAP.

nick-mobilecoin: Thank you!
That's a fair concern, and it's actually not that hard to maintain the same behavior.

I've added back the TCB update and retry behavior. Previously the `VerificationReportData` was decoded here, https://github.com/mobilecoinfoundation/mobilecoin/blob/6ccad2f67de22547c43398728a176a8770f82e41/attest/verifier/src/ias.rs#L225 and passed back in the error. Now the decoding happens in the error handler if it has the potential for a TCB update.

eranrund (approved):

PR #3588, opened 2023-09-28
Add `VerificationMessage` to `DcapVerifier`

nick-mobilecoin:
The `VerificationMessage` allows one to pretty print the result of a
failure to verify.

### DIFF
--- a/attest/untrusted/src/sim.rs
+++ b/attest/untrusted/src/sim.rs
@@ -206,12 +206,8 @@ fn c_struct_as_bytes<T>(c_struct: &T) -> &[u8] {
 #[cfg(test)]
 mod test {
     use super::*;
-    use mc_attest_verifier::DcapVerifier;
     use mc_attest_verifier_types::EnclaveReportDataContents;
-    use mc_attestation_verifier::{Evidence, TrustedMrEnclaveIdentity};
     use mc_sgx_dcap_types::{CertificationData, TcbInfo};
-    use p256::pkcs8::der::DateTime;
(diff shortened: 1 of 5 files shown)


attest/untrusted/src/sim.rs
```diff
@@ -249,26 +245,4 @@ mod test {
         let result = TcbInfo::try_from(&leaf_cert);
         assert!(result.is_ok());
     }
```
nick-mobilecoin, line 247: Originally I had the test for `DcapVerifier` in this `sim.rs` file to prevent an module meant for the enclave from depending on a file outside of the enclave, but that's only for tests, so I think it's ok and it's easier to see when `DcapVerifier` is being tested where it's defined

attest/verifier/src/dcap.rs
```diff
@@ -55,14 +58,26 @@ impl DcapVerifier {
...
+    fn fmt_padded(
+        &self,
+        f: &mut Formatter<'_>,
+        pad: usize,
+        result: &VerificationOutput<DcapVerifierOutput>,
+    ) -> core::fmt::Result {
+        let is_success = result.is_success();
+        let status = choice_to_status_message(is_success);
+
+        writeln!(f, "{:pad$}{status} DCAP evidence:", "")?;
```
nick-mobilecoin, line 76: I don't have a strong preference on what this header line should be, so up for suggestions on making it clearer if needed.

eranrund (approved):

PR #3592, opened 2023-09-29
`From` instead of `TryFrom` for proto to prost

nick-mobilecoin:
Previously some conversions in `mc-api` that were from protobuf types to
prost types were implemented as `TryFrom`. Now these conversions are
implemented as `From`. This allows for a more ergonomic API preventing
callers from having to check for errors that really shouldn't happen in
this conversion step.

### DIFF
--- a/api/src/convert/collateral.rs
+++ b/api/src/convert/collateral.rs
@@ -2,7 +2,7 @@
 
 //! Convert to/from external::Collateral
 
-use crate::{convert::encode_to_protobuf_vec, external, ConversionError};
+use crate::{external, ConversionError};
 use mc_attest_verifier_types::prost;
 use mc_sgx_dcap_types::Collateral;
 use mc_util_serial::Message;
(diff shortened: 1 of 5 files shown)


nick-mobilecoin:
* master
  * **PR #3592** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3592" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3592?utm_source=stack-comment).

api/src/convert/collateral.rs
```diff
@@ -24,12 +24,25 @@ impl TryFrom<&Collateral> for external::Collateral {
...
-        let bytes = encode_to_protobuf_vec(src)?;
-        let prost = prost::Collateral::decode(bytes.as_slice())?;
+        let prost = prost::Collateral::from(src);
         Ok((&prost).try_into()?)
     }
 }
 
+impl From<&external::Collateral> for prost::Collateral {
+    fn from(src: &external::Collateral) -> Self {
+        Self {
```
nick-mobilecoin, line 34: I think using the `Self{}` here will ensure we don't omit a field and prost only has the fields unlike protobufs so this shouldn't be too bad to maintain.

Implementation wise these are both structures of nested byte vectors, or strings, so there really shouldn't be any issue conerting between the two, it's when one wants the `mc-sgx-dcap-types::Collateral` that it can fail.

dolanbernard (approved):

nick-mobilecoin:
### Merge Activity

* **Sep 29, 4:35 PM**: @nick-mobilecoin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3592).

PR #3605, opened 2023-10-05
Implement `Message` for  `EvidenceKind`

nick-mobilecoin:
Previously `EvidenceMessage` was a proxy to decode the `EvidenceKind`
enum. This resulted in clients needing to first attempt to decode as
`EvidenceMessage` and then fall back to `VerificationReport`. Now the
logic of decoding `VerificationReport` versus `DcapEvidence` is wrapped
in the `Message` implementation of `EvidenceKind`.

### DIFF
--- a/attest/verifier/types/src/verification.rs
+++ b/attest/verifier/types/src/verification.rs
@@ -33,14 +33,6 @@ pub enum EvidenceKind {
     Dcap(prost::DcapEvidence),
 }
 
-// We default to `EvidenceKind::Epid` as that was the original kind of
-// `VerificationReport`.
-impl Default for EvidenceKind {
-    fn default() -> Self {
-        EvidenceKind::Epid(Default::default())
-    }
-}
-
 impl From<VerificationReport> for EvidenceKind {
     fn from(report: VerificationReport) -> Self {
         EvidenceKind::Epid(report)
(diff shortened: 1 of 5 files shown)


nick-mobilecoin:
* master
  * **PR #3605** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3605" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3605?utm_source=stack-comment).

attest/verifier/types/src/verification.rs
```diff
@@ -27,16 +27,108 @@ pub struct DcapEvidence {
...
+// The tag to indicate a `DcapEvidence` variant. For backwards compatibility
+// this must be outside the range of tags used by `VerificationReport`.
+const TAG_DCAP_EVIDENCE: u32 = 4;
+
+/// In order to make `EvidenceKind` backwards compatible with the previous
+/// logic which would send `VerificationReport`'s, the protobuf tags that
+/// correspond to tags defined in `VerificationReport` will decode to the
+/// `EvidenceKind::Epid()` variant. Tags outside of a `VerificationReport`'s
+/// will be treated as other variants.
+impl Message for EvidenceKind {
```
nick-mobilecoin, line 73: Curious on thoughts to this approach. I think the previous plan of:
1. Try decoding as `EvidenceMessage`
2. If it's `None`, then try decoding as `VerificationReport`
This allows consumers of `EvidenceKind` and the protobufs that may come along, to always try to decode as `EvidenceKind` and not worry as much about the nitty gritty.

attest/verifier/types/src/verification.rs
```diff
@@ -27,16 +27,108 @@ pub struct DcapEvidence {
     pub report_data: EnclaveReportDataContents,
 }
 
-#[derive(Clone, Oneof)]
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
 pub enum EvidenceKind {
-    #[prost(message, tag = "4")]
+    Epid(VerificationReport),
```
nick-mobilecoin, line 32: Other places I've used `AttestationEvidence` with a variant of `VerificationReport`
https://github.com/mobilecoinfoundation/mobilecoin/blob/c709b85bd47b31402003f09b84a1e9a9496a3356/blockchain/types/src/block_metadata.rs#L18

We probably need to revisit the naming here and in other places to be consistent

attest/verifier/types/src/verification.rs
```diff
@@ -27,16 +27,108 @@ pub struct DcapEvidence {
...
+/// will be treated as other variants.
+impl Message for EvidenceKind {
+    fn encode_raw<B>(&self, buf: &mut B)
+    where
+        B: BufMut,
+        Self: Sized,
+    {
+        match self {
+            EvidenceKind::Dcap(evidence) => {
+                encoding::message::encode(4, evidence, buf);
```
eranrund, line 81:
```suggestion
                encoding::message::encode(TAG_DCAP_EVIDENCE, evidence, buf);
```

eranrund (commented):
I don't think I have a super clear understanding of how this gets used.
How does this look in a `.proto` file or non-Rust code that wants to look at the report?

nick-mobilecoin:
> I don't think I have a super clear understanding of how this gets used. How does this look in a `.proto` file or non-Rust code that wants to look at the report?

Sorry, that context wasn't really provided.
This wouldn't work in a `.proto` file. Fortunately all places where the `VerificationReport` leaves a service via protobufs it's a field inside of another type so we're able to leverage `oneof` in those cases.

Where this proposed change comes into play, is in 2 places:
1. Sending the `VerificationReport` around in the kex handshake, 
  - https://github.com/mobilecoinfoundation/mobilecoin/blob/f48e2267a4f269097d4459ab17b874e021250171/attest/ake/src/responder.rs#L93
  - https://github.com/mobilecoinfoundation/mobilecoin/blob/f48e2267a4f269097d4459ab17b874e021250171/attest/ake/src/initiator.rs#L163
  - https://github.com/mobilecoinfoundation/mobilecoin/blob/f48e2267a4f269097d4459ab17b874e021250171/attest/ake/src/initiator.rs#L222
2. Watcher being able to retrieve previously stored `VerificationReport` and store new `DcapEvidence` unambiguosly
  - https://github.com/mobilecoinfoundation/mobilecoin/blob/f48e2267a4f269097d4459ab17b874e021250171/watcher/src/watcher_db.rs#L722
  - https://github.com/mobilecoinfoundation/mobilecoin/blob/f48e2267a4f269097d4459ab17b874e021250171/watcher/src/watcher_db.rs#L786
  
The first use case is temporary and will only need to persist for clients to ease the transition from `VerificationReport` to `DcapEvidence`.

For watcher, I think it will need to persist indefinitely. It may be that we could migrate the entry in the database similar to what Fog Report is doing, https://github.com/mobilecoinfoundation/mobilecoin/blob/f48e2267a4f269097d4459ab17b874e021250171/fog/sql_recovery_db/src/lib.rs#L1222. 
Fog report didn't need special logic as it's an ephemeral entry that will be changed on service restart.

eranrund:
> > I don't think I have a super clear understanding of how this gets used. How does this look in a `.proto` file or non-Rust code that wants to look at the report?
> 
> Sorry, that context wasn't really provided. This wouldn't work in a `.proto` file. Fortunately all places where the `VerificationReport` leaves a service via protobufs it's a field inside of another type so we're able to leverage `oneof` in those cases.
> 
> Where this proposed change comes into play, is in 2 places:
> 
>     1. Sending the `VerificationReport` around in the kex handshake,
> 
> 
>     * https://github.com/mobilecoinfoundation/mobilecoin/blob/f48e2267a4f269097d4459ab17b874e021250171/attest/ake/src/responder.rs#L93
> 
>     * https://github.com/mobilecoinfoundation/mobilecoin/blob/f48e2267a4f269097d4459ab17b874e021250171/attest/ake/src/initiator.rs#L163
> 
>     * https://github.com/mobilecoinfoundation/mobilecoin/blob/f48e2267a4f269097d4459ab17b874e021250171/attest/ake/src/initiator.rs#L222
> 
> 
>     2. Watcher being able to retrieve previously stored `VerificationReport` and store new `DcapEvidence` unambiguosly
> 
> 
>     * https://github.com/mobilecoinfoundation/mobilecoin/blob/f48e2267a4f269097d4459ab17b874e021250171/watcher/src/watcher_db.rs#L722
> 
>     * https://github.com/mobilecoinfoundation/mobilecoin/blob/f48e2267a4f269097d4459ab17b874e021250171/watcher/src/watcher_db.rs#L786
> 
> 
> The first use case is temporary and will only need to persist for clients to ease the transition from `VerificationReport` to `DcapEvidence`.
> 
> For watcher, I think it will need to persist indefinitely. It may be that we could migrate the entry in the database similar to what Fog Report is doing,
> 
> https://github.com/mobilecoinfoundation/mobilecoin/blob/f48e2267a4f269097d4459ab17b874e021250171/fog/sql_recovery_db/src/lib.rs#L1222
> .
> Fog report didn't need special logic as it's an ephemeral entry that will be changed on service restart.

Thanks for the detailed explanation! I like the thought of not having this kind of hack stick around forever, so I think a migration for the watcher to use a oneof field would be beneficial.

attest/verifier/types/src/verification.rs
```diff
@@ -27,16 +27,68 @@ pub struct DcapEvidence {
...
+}
+
+// A local only struct to make it nest a `prost::DcapEvidence` behind a tag.
+// This allows `DcapEvidence` to be used in prior locations where
+// `VerificationReport` was used. The tag is `4` to avoid collisions with the
+// `VerificationReport` tags.
+#[derive(Message)]
+struct DcapEvidenceDecoder {
+    #[prost(message, tag = 4)]
+    pub dcap: Option<prost::DcapEvidence>,
```
nick-mobilecoin, line 91: I realized this will also need to stay indefinitely in the ake handshake, or we'll need to do a similar transition in the 7.0 release so that the DcapEvidence is sent directly in the ake handshake without the "4" tag.  Either way this special logic is hidden behind the `EvidenceKind` so should mitigate most issues around it.

attest/verifier/types/src/verification.rs
nick-mobilecoin: changed implementation, to have dedicated methods on the `EvidenceKind` so that it's less likely someone tries to use this serialization method in other places where it's not intended

eranrund (approved):

PR #3607, opened 2023-10-06
Use general term "attestation evidence" in watcher

nick-mobilecoin:
Previously `mc-watcher` used "verification report" or similar to refer
to the `VerificationReport`. With the anticipated use of `DcapEvidence`
`mc-watcher` has been updated to use a more generic term of "attestation
evidence" where appropriate.

I was making changes to watcher to support handling `DcapEvidence` and it seemed that many places where `verification_report` (or similar) was used would be ambiguous to future maintainers. This renaming may be a bit aggressive or perhaps we should completely avoid it as this PR is larger than ideal.

### DIFF
--- a/watcher/src/watcher_db.rs
+++ b/watcher/src/watcher_db.rs
@@ -56,16 +56,22 @@ impl MetadataStoreSettings for WatcherDbMetadataStoreSettings {
 /// Block signatures database name.
 pub const BLOCK_SIGNATURES_DB_NAME: &str = "watcher_db:block_signatures";
 
-/// VerificationReports database name.
-pub const VERIFICATION_REPORTS_BY_BLOCK_SIGNER_DB_NAME: &str =
+/// Attestation evidence database name.
+/// Called `verification_reports_by_block_signer` due to original attestation
+/// evidence being `VerificationReport`
+pub const ATTESTATION_EVIDENCE_BY_BLOCK_SIGNER_DB_NAME: &str =
     "watcher_db:verification_reports_by_block_signer";
 
-/// Verification reports poll queue database name.
-pub const VERIFICATION_REPORTS_POLL_QUEUE_DB_NAME: &str =
+/// Attestation evidence poll queue database name.
+/// Called `verification_reports_poll_queue` due to original attestation
+/// evidence being `VerificationReport`
+pub const ATTESTATION_EVIDENCE_POLL_QUEUE_DB_NAME: &str =
     "watcher_db:verification_reports_poll_queue";
 
-/// Verification reports by report hash database name.
-pub const VERIFICATION_REPORTS_BY_HASH_DB_NAME: &str = "watcher_db:verification_reports_by_hash";
+/// Attestation evidence by hash database name.
+/// Called `verification_reports_by_hash` due to original attestation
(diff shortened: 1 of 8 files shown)


nick-mobilecoin:
* master
  * **PR #3607** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3607" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3607?utm_source=stack-comment).

watcher/src/watcher_db.rs
```diff
@@ -672,70 +674,66 @@ impl WatcherDB {
...
             hex::encode(signer.to_bytes()),
-            verification_report.is_some(),
+            attestation_evidence.is_some(),
             value_bytes.len(),
         );
 
-        // First, write the hash -> verification report entry.
-        let hash: [u8; 32] = value_bytes.digest32::<MerlinTranscript>(b"verification_report");
+        // First, write the hash -> attestation evidence entry.
+        let hash: [u8; 32] = value_bytes.digest32::<MerlinTranscript>(b"attestation_evidence");
```
nick-mobilecoin, line 734: My understanding is that updating this value to the hash will only affect if we restart Watcher DB without restarting consensus. This creates a possibility we have 2 entries after a restart. One for `"verification_report"` and one for `"attestation_evidence"`.  It appears we have logic to handle duplicate entries.

The hash only seems to be used to see if the report is already in the DB, so don't think this will affect historical look ups.

watcher/src/watcher_db.rs
```diff
@@ -56,16 +56,22 @@ impl MetadataStoreSettings for WatcherDbMetadataStoreSettings {
 /// Block signatures database name.
 pub const BLOCK_SIGNATURES_DB_NAME: &str = "watcher_db:block_signatures";
 
-/// VerificationReports database name.
-pub const VERIFICATION_REPORTS_BY_BLOCK_SIGNER_DB_NAME: &str =
+/// Attestation evidence database name.
+/// Called `verification_reports_by_block_signer` due to original attestation
+/// evidence being `VerificationReport`
+pub const ATTESTATION_EVIDENCE_BY_BLOCK_SIGNER_DB_NAME: &str =
```
nick-mobilecoin, line 62: My understanding is that changing these strings would result in issues getting the historical data, unless some kind of migration happened. So left the names the same here.

eranrund: Correct, the strings cannot be changed

watcher/src/watcher_db.rs
eranrund: The hash is also used here (but is not being calculated there): https://github.com/mobilecoinfoundation/mobilecoin/blob/944ababc13831fd57cb91dce0ba959733610b3f6/watcher/src/watcher_db.rs#L833C45-L833C73

but I don't think this will get affected by the hashing being changed, so we're good.

eranrund (approved):
I think this is fine even if it contains many changes since it'll make grepping for attestation-evidence-related code easier in the future. Thank you!

PR #3608, opened 2023-10-07
Update watcher to handle `EvidenceKind`

nick-mobilecoin:
Previously watcher would only work with `VerificationReport`s. Now
watcher will work with `EvidenceKind` allowing it to work with both the
legacy `VerificationReport` and the newer `DcapEvidence`.

### DIFF
--- a/watcher/src/attestation_evidence_collector.rs
+++ b/watcher/src/attestation_evidence_collector.rs
@@ -9,7 +9,8 @@ use mc_attest_ake::{
     AuthRequestOutput, ClientInitiate, Start, Transition, UnverifiedAttestationEvidence,
 };
 use mc_attest_api::{attest::AuthMessage, attest_grpc::AttestedApiClient};
-use mc_attest_core::{VerificationReport, VerificationReportData};
+use mc_attest_core::{EvidenceKind, VerificationReport, VerificationReportData};
+use mc_attest_verifier_types::prost;
 use mc_common::{
     logger::{log, Logger},
     time::SystemTimeProvider,
@@ -38,7 +39,7 @@ use std::{
 use url::Url;
 
 /// A trait that specifies the functionality AttestationEvidenceCollector needs
-/// in order to go from a ConsensusClientUri into a VerificationReport, and the
+/// in order to go from a ConsensusClientUri into a EvidenceKind, and the
 /// associated signer key.
 pub trait NodeClient {
     /// Get attestation evidence for a given client.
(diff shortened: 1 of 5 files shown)


nick-mobilecoin:
* master
  * **PR #3607** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3607" target="_blank">Graphite</a> 
    * **PR #3608** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3608" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3608?utm_source=stack-comment).

watcher/src/attestation_evidence_collector.rs
```diff
@@ -84,42 +84,64 @@ impl NodeClient for ConsensusNodeClient {
...
+        .map_err(|err| format!("Unable to construct key: {err}"))?;
 
-        let signer_bytes = &custom_data_bytes[32..];
+    Ok(signer_public_key)
+}
 
-        let signer_public_key = Ed25519Public::try_from(signer_bytes)
-            .map_err(|err| format!("Unable to construct key: {err}"))?;
+/// Get the block signer key from a [`prost::DcapEvidence`].
+pub fn get_block_signer_from_dcap_evidence(
```
nick-mobilecoin, line 125: Made public to be used in the db_dump.rs file

watcher/src/attestation_evidence_collector.rs
```diff
@@ -738,4 +757,54 @@ mod tests {
             sleep(Duration::from_millis(100));
         }
     }
+
+    #[test]
+    fn consensus_node_block_signer_from_dcap_evidence() {
+        // For DCAP evidence the signer key is the custom_identity.
```
nick-mobilecoin, line 763: The test up above uses a `TestNodeClient`, which doesn't exercise the `get_block_signature()` so did these isolating unit tests here. Even though this module updated to use `EvidenceKind` in many places, the behavior only matters for the `get_block_signature()`

watcher/src/attestation_evidence_collector.rs
```diff
@@ -143,7 +165,7 @@ fn attestation_evidence_from_node_url(
         .try_next(&mut csprng, unverified_evidence_event)
         .map_err(|err| format!("Failed decoding attestation evidence from {node_url}: {err}"))?;
 
-    Ok(attestation_evidence)
+    Ok(attestation_evidence.into())
```
nick-mobilecoin, line 168: Once #3602 comes in this will be getting an `Evidence` kind from the `Initiator`

watcher/src/bin/db-dump.rs
```diff
@@ -121,14 +121,21 @@ fn display_report_status(
...
+                                    .expect("failed constructing verification report data");
+                                format!(
+                                    "report available, id {} generated at {}",
+                                    report_data.id, report_data.timestamp
+                                )
+                            }
+                            EvidenceKind::Dcap(evidence) => {
+                                let signer = attestation_evidence_collector::get_block_signer_from_dcap_evidence(evidence)
+                                    .expect("failed getting signer from dcap evidence");
+                                format!("DCAP evidence available, signer {signer}")
```
nick-mobilecoin, line 136: Not really sure what data we want to dump for DCAP evidence, there is no ID field or time like in the `VerificationReport`

eranrund: I think this is sufficient for now, we can update this if we ever use it again.

eranrund (approved):

PR #3619, opened 2023-10-17
Remove use of `default()` for unit structs

nick-mobilecoin:
This fixes clippy lints which will be coming in
<https://rust-lang.github.io/rust-clippy/master/index.html#/default_constructed_unit_structs>

### DIFF
--- a/api/src/convert/input_ring.rs
+++ b/api/src/convert/input_ring.rs
@@ -130,7 +130,7 @@ mod tests {
                 block_version,
                 Amount::new(Mob::MINIMUM_FEE, Mob::ID),
                 fpr.clone(),
-                EmptyMemoBuilder::default(),
+                EmptyMemoBuilder,
             )
             .unwrap();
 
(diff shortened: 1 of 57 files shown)


nick-mobilecoin:
* master
  * **PR #3619** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3619" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3619?utm_source=stack-comment).

nick-mobilecoin:
Converting to draft, may need to push the rust version up to work with wasm :(

nick-mobilecoin:
> Converting to draft, may need to push the rust version up to work with wasm :(

Worked around the wasm issue by not changing it from `McRng::default()` yet. The failure is that `mc-rand` uses a type alias for `McRng` when using wasm, https://github.com/mobilecoinfoundation/rand/blob/437dc69f4e852df20756b46542572c7bc29f3575/src/wasm.rs#L5 this prevents one from doing `let foo = McRng` in wasm builds. Thinking longer term we may want to to have a facade in front of the wasm type like fallback does, https://github.com/mobilecoinfoundation/rand/blob/main/src/fallback.rs

eranrund (approved):

PR #3621, opened 2023-10-17
Place test modules at end of files

nick-mobilecoin:
Newer clippy versions will warn if the test module is not the last one
in the file,
https://rust-lang.github.io/rust-clippy/master/index.html#/items_after_test_module

Also update some comment formatting for newer `cargo fmt` behavior.

PR #3622, opened 2023-10-17
Remove use of `&mut` that didn't mutate

nick-mobilecoin:
The removal of mutable references that don't need to be mutable is to
address the coming clippy lint of
https://rust-lang.github.io/rust-clippy/master/index.html#/needless_pass_by_ref_mut

### DIFF
--- a/consensus/enclave/impl/src/lib.rs
+++ b/consensus/enclave/impl/src/lib.rs
@@ -1167,7 +1167,7 @@ mod tests {
 
             let tx = create_transaction(
                 block_version,
-                &mut ledger,
+                &ledger,
                 &tx_out,
                 &sender,
                 &recipient.default_subaddress(),
@@ -1254,7 +1254,7 @@ mod tests {
 
             let tx = create_transaction(
                 block_version,
-                &mut ledger,
+                &ledger,
                 &tx_out,
                 &sender,
                 &recipient.default_subaddress(),
@@ -1396,7 +1396,7 @@ mod tests {
 
                     create_transaction(
                         block_version,
-                        &mut ledger,
+                        &ledger,
                         &tx_out,
                         &sender,
                         &recipient.default_subaddress(),
@@ -1535,7 +1535,7 @@ mod tests {
 
                     create_transaction(
                         block_version,
-                        &mut ledger,
+                        &ledger,
                         &tx_out,
                         &sender,
                         &recipient.default_subaddress(),
@@ -1567,7 +1567,7 @@ mod tests {
 
                     let tx = create_transaction(
                         block_version,
-                        &mut ledger,
+                        &ledger,
                         &spendable_output,
                         &sender,
                         &recipient.default_subaddress(),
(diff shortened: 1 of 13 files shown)


nick-mobilecoin:
* master
  * **PR #3621** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3621" target="_blank">Graphite</a> 
    * **PR #3622** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3622" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3622?utm_source=stack-comment).

consensus/enclave/impl/src/lib.rs
```diff
@@ -1167,7 +1167,7 @@ mod tests {
 
             let tx = create_transaction(
                 block_version,
-                &mut ledger,
+                &ledger,
```
nick-mobilecoin, line 1170: These are tests, and they only read from the DB don't actually write to it.

fog/ingest/server/src/controller.rs
```diff
@@ -265,7 +265,7 @@ where
     /// but with checks for idle state, and updates to sealed backups
     pub fn set_ingress_private_key(&self, msg: EnclaveMessage<PeerSession>) -> Result<(), Error> {
         // Lock our state for this entire call
-        let mut state = self.get_state();
+        let state = self.get_state();
```
nick-mobilecoin, line 268: I'm not sure if `mut` was here by accident or if someone wanted to limit to only one reference at a time.
If it's the latter, I don't think it's necessary. My understanding is the MutexGuard is not Sync as such the compiler will make sure that only one thread would be accessing the `state`, after the `lock()` call.  So even if multiple locations, in this code was using the reference to the state, they shouldn't be fighting each other as they're on the same thread.

fog/ledger/server/tests/router_connection.rs
```diff
@@ -88,23 +88,23 @@ fn fog_ledger_merkle_proofs_test(logger: Logger) {
             &recipients,
             &[],
             &mut rng,
-            &mut watcher,
+            &watcher,
```
nick-mobilecoin, line 91: The watcher db `add` methods aren't mut...? https://github.com/mobilecoinfoundation/mobilecoin/blob/f957c956e67ce28a6cc5f571d28e6e364e3f7f2e/watcher/src/watcher_db.rs#L269

It kind of feels like they should be, but they're interacting with a state outside of rust where the one mutable instance may not make as much sense.

fog/ingest/server/src/controller.rs
eranrund: It's probably just because `write_state_file_inner` wanted a `mut`, even if it didn't actually need one.

fog/ledger/server/tests/router_connection.rs
eranrund: I don't think the lack of a`mut` necessarily implies no state is being manipulated (e.g. anything that locks a mutex can then manipulate state and the mutex doesn't need a `mut` reference to be locked).

eranrund (approved):

PR #3643, opened 2023-10-23
Fog-Report - Helm chart improvements for consolidating fog instances.

jgreat:
- Break out fog-report chart from fog-services monolith.
- Update GHA and integration tests to support multiple fog-report instances.

### Motivation

We need to be able to build fog instances that have multiple copies of fog-report, fog-view-router and fog-ledger-router with appropriate ingress and secret alignment.

### DIFF
--- a/.github/workflows/mobilecoin-dev-cd.yaml
+++ b/.github/workflows/mobilecoin-dev-cd.yaml
@@ -35,7 +35,7 @@ jobs:
 # Generate environment information
 ############################################
   generate-metadata:
-    if: "! startsWith(github.head_ref, 'dependabot/')"
+    if: ${{ ! startsWith(github.head_ref, 'dependabot/') }}
     name: 👾 Environment Info 👾
     runs-on: [self-hosted, Linux, small]
     outputs:
@@ -80,11 +80,11 @@ jobs:
       MINTING_TRUST_ROOT_PUBLIC_KEY_PEM: ${{ github.workspace }}/.tmp/minting_trust_root.public.pem
     steps:
     - name: Checkout
-      if: "! contains(github.event.head_commit.message, '[skip build]')"
+      if: ${{ ! contains(github.event.head_commit.message, '[skip build]') }}
       uses: actions/checkout@v3
 
     - name: Write environment values
-      if: "! contains(github.event.head_commit.message, '[skip build]')"
+      if: ${{ ! contains(github.event.head_commit.message, '[skip build]') }}
       env:
         ENCLAVE_SIGNING_KEY: ${{ secrets.DEV_ENCLAVE_SIGNING_KEY }}
         MINTING_TRUST_ROOT_PUBLIC: ${{ secrets.DEV_MINTING_TRUST_ROOT_PUBLIC }}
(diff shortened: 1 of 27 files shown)


.github/workflows/mobilecoin-dev-cd.yaml
```diff
@@ -35,7 +35,7 @@ jobs:
 # Generate environment information
 ############################################
   generate-metadata:
-    if: "! startsWith(github.head_ref, 'dependabot/')"
```
jgreat, line 38: Cleaning up syntax so actionlint doesn't complain.

.github/workflows/mobilecoin-dev-cd.yaml
```diff
@@ -367,13 +367,14 @@ jobs:
         - mc-core-dev-env-setup
         - mobilecoind
         - watcher
+        - fog-report
```
jgreat, line 370: Add fog-report to the list

.github/workflows/mobilecoin-workflow-dev-deploy.yaml
```diff
@@ -43,10 +43,17 @@ on:
       DEV_RANCHER_TOKEN:
         description: "Rancher access token"
         required: true
+      DEV_FOG_REPORT_B_SIGNING_CERT:
```
jgreat, line 46: "a" cert values are populated by the mobilecoin-workflow-dev-setup-environment.yaml action.  I'll make a second pass and see if I can make this more consistent.  For now I'm looking to minimize changes.

.internal-ci/helm/fog-services-config/templates/fog-report-configmap.yaml
joekottke, line 1: Why is this file deleted?

.internal-ci/helm/fog-services-config/values.yaml
```diff
@@ -32,12 +32,6 @@ fogReport:
   signingCert:
     key: ''
     crt: ''
-  configMap:
-    # https://docs.diesel.rs/diesel/r2d2/struct.Builder.html
-    POSTGRES_IDLE_TIMEOUT: '60'
-    POSTGRES_MAX_LIFETIME: '120'
-    POSTGRES_CONNECTION_TIMEOUT: '5'
-    POSTGRES_MAX_CONNECTIONS: '3'
```
joekottke, line 40: Are these deletions pertinent?  Or are they just unused?

joekottke (approved):
Just curious about the deletions, and a few comments.  Non-blocking.

.internal-ci/helm/fog-services-config/values.yaml
jgreat: This moved in to the fog-report chart.  Removed this value and configmap so we don't have conflicts when applying the new fog-report chart.

.internal-ci/helm/fog-services-config/templates/fog-report-configmap.yaml
jgreat: This moved in to the fog-report chart.  Removed this value and configmap so we don't have conflicts when applying the new fog-report chart.

PR #3706, opened 2023-11-10
Remove newlines from enclave panic message

nick-mobilecoin:
The enclave panic message is passed to a logger that doesn't work with newlines. In order to keep the contents of the panic message from the enclave a space is used to separate the "Enclave panic:" prefix and the actual message.


nick-mobilecoin:
This change is to support better error reporting than what we're getting which is
image

jgreat (approved):

PR #3711, opened 2023-11-12
Add timestamps to `Block`s

nick-mobilecoin:
Add a `ConsensusValue` to contain timestamps. This timestamp is
populated when a node initially proposes a set of values. Each node that
proposes values will propose a time. The largest time that is validated
by all the nodes during consensus will be used in the block.


nick-mobilecoin:
* main
  * **PR #3711** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3711?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3711?utm_source=stack-comment).

blockchain/types/src/block.rs
```diff
@@ -192,6 +207,11 @@ pub fn compute_block_id(
     root_element.append_to_transcript(b"root_element", &mut transcript);
     contents_hash.append_to_transcript(b"contents_hash", &mut transcript);
 
+    // Earlier blocks didn't have a timestamp so it will come back as 0
+    if timestamp != 0 {
```
eranrund, line 211: Might be better to rely on the block version for this

consensus/enclave/impl/src/lib.rs
```diff
@@ -322,6 +322,20 @@ impl SgxConsensusEnclave {
         Ok(transactions)
     }
 
+    /// Validate the provided timestamp is valid.
+    ///
+    /// # Arguments
+    /// * `timestamp` - The timestamp to validate
+    /// * `parent_block_timestamp` - The timestamp of the parent block.
+    fn validate_timestamp(timestamp: u64, parent_block_timestamp: u64) -> Result<u64> {
```
eranrund, line 330: might want to require timestamp=0 if the block version does not support timestamps, and only if it does ensure that the timestamp is greater than (or greater-than or equal) to the parent.

consensus/service/src/byzantine_ledger/worker.rs
```diff
@@ -609,6 +617,7 @@ impl<
                 mint_config_tx.prefix.tombstone_block > expired_block_index
             }
             ConsensusValue::MintTx(mint_tx) => mint_tx.prefix.tombstone_block > expired_block_index,
+            ConsensusValue::TimeStamp(_) => true,
```
eranrund, line 620: Why retain the timestamp for the next block?

consensus/service/src/byzantine_ledger/worker.rs
```diff
@@ -835,6 +845,7 @@ impl<
                 ConsensusValue::MintTx(mint_tx) => {
                     mint_txs.push(mint_tx);
                 }
+                ConsensusValue::TimeStamp(timestamp) => timestamps.push(timestamp),
```
eranrund, line 848: Do you think we can enforce that there is only a single externalized timestamp? If not, then when/why would we see multiple?

consensus/service/src/timestamp_validator.rs
```diff
@@ -0,0 +1,154 @@
...
+
+const MAX_TIMESTAMP_AGE: u64 = 30 * 1000; // 30 seconds
+
+/// Errors related to validating timestamps
+#[derive(Debug, Display, Clone, PartialEq)]
+pub enum Error {
+    /// The timestamp is in the future now{0}, timestamp{1}
+    InFuture(u64, u64),
+    /** The timestamp is older than the last bock: last_bock timestamp{0},
+    timestamp{1} */
```
eranrund, line 16: `timestamp{0}` seems off, should it just be `last_block:{0} cur_block:{1}` or something like that?

consensus/service/src/timestamp_validator.rs
```diff
@@ -0,0 +1,154 @@
...
+    OlderThanLastBlock(u64, u64),
+    /// The timestamp is too far in the past now{0}, timestamp{1}
```
eranrund, line 18: Similarly, combining the timestamp with the word before it seems a bit weird to me

consensus/service/src/timestamp_validator.rs
```diff
@@ -0,0 +1,154 @@
...
+        .duration_since(std::time::UNIX_EPOCH)
+        .unwrap()
+        .as_millis() as u64;
+
+    if timestamp <= last_block_timestamp {
+        return Err(Error::OlderThanLastBlock(last_block_timestamp, timestamp));
+    }
+
+    if timestamp > now {
+        return Err(Error::InFuture(now, timestamp));
```
eranrund, line 35: Is it possible that this will create issues if a timestamp comes from a remote node who is a few millisecond ahead of the current node for some reason?

peers/src/consensus_msg.rs
```diff
@@ -29,6 +29,9 @@ pub enum ConsensusValue {
 
     /// MintTx({0})
     MintTx(MintTx),
+
+    /// Timestamp({0})ms
+    TimeStamp(u64),
```
eranrund, line 34: Super nit but `TimeStamp` or `Timestamp`? I don't know if I've encountered it as two separate words

blockchain/types/src/block.rs
nick-mobilecoin: Fixed in #3733

consensus/enclave/impl/src/lib.rs
nick-mobilecoin: Conversation brought over to, https://github.com/mobilecoinfoundation/mobilecoin/pull/3733/files#r1399799957

consensus/service/src/byzantine_ledger/worker.rs
nick-mobilecoin: no longer present in new work of #3733

consensus/service/src/byzantine_ledger/worker.rs
nick-mobilecoin: This is taken care of in #3733 now during block externalizing

consensus/service/src/timestamp_validator.rs
nick-mobilecoin: conversation brought over to, https://github.com/mobilecoinfoundation/mobilecoin/pull/3733/files#r1399804585

consensus/service/src/timestamp_validator.rs
nick-mobilecoin: conversation brought over to, https://github.com/mobilecoinfoundation/mobilecoin/pull/3733/files#r1399806093

consensus/service/src/timestamp_validator.rs
nick-mobilecoin: Conversation brought over to https://github.com/mobilecoinfoundation/mobilecoin/pull/3733/files#r1399786804

peers/src/consensus_msg.rs
nick-mobilecoin: fixed in 3733

nick-mobilecoin:
Closing in favor of #3733 
This approach doesn't work because all of the `ConsensusValue`s are in a common pool that SCP uses to nominate and ballot values.
If SCP pulls the timestamp value from the pool and nomination or balloting fails, then when SCP tries to propose a second time there will be no timestamp in it's proposal. If this happens for all nodes, then there will be no timestamp to externalize to the block.

SCP treats the `ConsensusValues` as opaque so it would need to call back to the byzantine ledger worker to get/ensure there is a timestamp in every nomination/ballot.

PR #3719, opened 2023-11-14
Reduce the static startup memory of enclaves

nick-mobilecoin:
Previously the enclaves would reserve memory for 10_000 backend
connections. An enclave would never need that many connections as it
represents each shard that a router would connect to. This value has
been reduced to 100, which reduces enclave static startup memory from
~32MB to ~320KB.

### DIFF
--- a/crypto/ake/enclave/src/lib.rs
+++ b/crypto/ake/enclave/src/lib.rs
@@ -41,16 +41,20 @@ const MAX_AUTH_PENDING_REQUESTS: usize = 64;
 const MAX_PEER_SESSIONS: usize = 64;
 
 /// Maximum number of concurrent sessions to this enclave from router enclaves.
-const MAX_FRONTEND_SESSIONS: usize = 500;
+const MAX_FRONTEND_SESSIONS: usize = 64;
 
 /// Max number of backends that this enclave can connect to as a client.
-const MAX_BACKEND_SESSIONS: usize = 10_000;
+/// Current shard size is 400_000 blocks, with 2 replicas per shard.
+/// At 2 million blocks this would require 10 connections. Using 100 connections
+/// allows for 10x increase of 2 million blocks.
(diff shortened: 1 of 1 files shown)


crypto/ake/enclave/src/lib.rs
```diff
@@ -41,16 +41,20 @@ const MAX_AUTH_PENDING_REQUESTS: usize = 64;
 const MAX_PEER_SESSIONS: usize = 64;
 
 /// Maximum number of concurrent sessions to this enclave from router enclaves.
-const MAX_FRONTEND_SESSIONS: usize = 500;
+const MAX_FRONTEND_SESSIONS: usize = 64;
```
nick-mobilecoin, line 44: This should be a many to one where there are Many backends to one frontend, as such each shard should only be talking to a handful of routers. I used 64 since that was the same as the peer sessions.
Each one of these connections is 2096. So this reduces ~1MB to 128KB

crypto/ake/enclave/src/lib.rs
```diff
@@ -41,16 +41,20 @@ const MAX_AUTH_PENDING_REQUESTS: usize = 64;
...
 
 /// Max number of backends that this enclave can connect to as a client.
-const MAX_BACKEND_SESSIONS: usize = 10_000;
+/// Current shard size is 400_000 blocks, with 2 instances per shard.
+/// At 2 million blocks this would require 10 connections. Using 100 connections
+/// allows for 10x increase of 2 million blocks.
+const MAX_BACKEND_SESSIONS: usize = 100;
```
nick-mobilecoin, line 50: This and the `MAX_BACKEND_AUTH_PENDING_REQUESTS` were the heavy hitters.
At 10,000 this used 20,960,000B and `MAX_BACKEND_AUTH_PENDING_REQUESTS` used 12,480,000B. 
These values represent how many shard replicas a router needs to talk to.

eranrund (approved):

PR #3721, opened 2023-11-14
Remove fog view router omap

nick-mobilecoin:
Previously the fog view router allocated 30MB of OMAP. This was a hold over from when fog view router was fog view. Only the view stores need OMAP and fog view router does not need OMAP.


eranrund (approved):

holtzman (approved):
lgtm

PR #3733, opened 2023-11-20
Provide proposed timestamp along with tx

nick-mobilecoin:
When values are proposed to SCP, a timestamp is provided with the
proposed value. These timestamps are validated and consolidated to
populate the timestamp field of the `Block`.

### Future Work

[Soundtrack of this PR]()


nick-mobilecoin:
* main
  * **PR #3733** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3733?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3733?utm_source=stack-comment).

consensus/enclave/impl/src/lib.rs
```diff
@@ -3564,4 +3603,158 @@ mod tests {
...
+        );
+    }
+
+    #[test_with_logger]
+    fn test_form_block_fails_for_invalid_timestamp(logger: Logger) {
+        let mut rng = Hc128Rng::from_seed([77u8; 32]);
+
+        for block_version in BlockVersion::iterator() {
+            let enclave = SgxConsensusEnclave::new(logger.clone());
+            let blockchain_config = BlockchainConfig {
```
nick-mobilecoin, line 3687: I'm not sure when we're worried about consensus working on previous block versions.
Even though previous versions wouldn't read the blocks timestamp this would still populate it for previous versions.

consensus/service/src/mint_tx_manager/mod.rs
```diff
@@ -176,48 +167,39 @@ impl<L: Ledger> MintTxManager for MintTxManagerImpl<L> {
...
-            let active_mint_config = match self.ledger_db.get_active_mint_config_for_mint_tx(tx) {
-                Ok(active_mint_config) => active_mint_config,
-                Err(err) => {
-                    log::warn!(
-                        self.logger,
-                        "failed finding an active mint config for mint tx {}: {}",
-                        tx,
-                        err
-                    );
+        let allowed_txs = candidates
```
nick-mobilecoin, line 176: Similar to MintTxConfig, notice how this looks at the entire `MintTx` for deduplication over only the nonce and token id

consensus/service/src/timestamp_validator.rs
```diff
@@ -0,0 +1,236 @@
...
+    TooOld(u64, u64),
+}
+
+pub fn validate(timestamp: u64, latest_block: &Block) -> Result<(), Error> {
+    let now = std::time::SystemTime::now()
+        .duration_since(std::time::UNIX_EPOCH)
+        .expect("Failed to get system time")
+        .as_millis() as u64;
+
+    if timestamp > now {
```
nick-mobilecoin, line 28: Up for suggestions here, we could allow for timestamps in the future.
i.e. node A's clock is a bit behind node B's so when Node A validates a value from NodeB it could seem like it's in the future. We might consider allowing 1-3 seconds ahead.
Based on looking at the historical signed at timestamps, if we throw out significant outliers all the node times are within 3 seconds of each other, https://github.com/nick-mobilecoin/timestamp-metrics/blob/main/timestamps.ipynb
Search for "We can see now that the nodes are all within 3 seconds of each other.", Using a 3 second window would throw out 80,000 blocks of the 2 million that were analyzed.

nick-mobilecoin:
The approach in #3711 didn't pan out.
The reason being that adding a timestamp as a `ConsensusValue` means that it is in a common pool of the other `ConsensusValue`s. 
When SCP process nominations and ballots, it's possible that it only takes a subset of the values in the pool. When this happens the timestamp could be taken initially and then later on the next attempt there is no timestamp in the pool of values SCP pulls from.  SCP treats the `ConsensusValues` opaquely so it would have required updating SCP to have some kind of understanding that it needs one of the timestamp variant, and then it would need to callback somehow to re-populate with a timestamp.  This re-population would need to be  synchronized across the nodes.

consensus/enclave/impl/src/lib.rs
```diff
@@ -322,6 +322,21 @@ impl SgxConsensusEnclave {
         Ok(transactions)
     }
 
+    /// Validate the provided timestamp is valid for use in building a block.
+    ///
+    /// # Arguments
+    /// * `timestamp` - The timestamp to validate
+    /// * `parent_block` - The parent block of the block being built.
+    fn validate_timestamp(timestamp: u64, parent_block: &Block) -> Result<u64> {
```
nick-mobilecoin, line 330: bringing up the comment from, https://github.com/mobilecoinfoundation/mobilecoin/pull/3711#discussion_r1393478335
> might want to require timestamp=0 if the block version does not support timestamps, and only if it does ensure that the timestamp is greater than (or greater-than or equal) to the parent.

Do we want to keep supporting older block versions in consenus? Should we have some kind of logic that says consensus will only consense on a block version 4 or higher?
We have logic to ensure the block version doesn't go down.

consensus/service/src/timestamp_validator.rs
```diff
@@ -0,0 +1,236 @@
...
+/// Provides logic for validating a timestamp used in consensus
+
+const MAX_TIMESTAMP_AGE: u64 = 30 * 1000; // 30 seconds
+
+/// Errors related to validating timestamps
+#[derive(Clone, Debug, Display, Eq, PartialEq)]
+pub enum Error {
+    /// The timestamp is in the future now: {0}, timestamp: {1}
+    InFuture(u64, u64),
+    /** The timestamp is not newer than the last bock. last_bock timestamp: {0},
```
nick-mobilecoin, line 15: Bringing in comment from, https://github.com/mobilecoinfoundation/mobilecoin/pull/3711#discussion_r1393486572
> timestamp{0} seems off, should it just be last_block:{0} cur_block:{1} or something like that?

Are you, @eranrund, thinking that the wording of this module and errors should be focusing on the current block being built, rather than a standalone timestamp?
I think especially with the refactor that `validate()` looks at `latest_block` directly it doesn't seem to far off.

consensus/service/src/timestamp_validator.rs
```diff
@@ -0,0 +1,236 @@
...
+    timestamp: {1} */
+    NotNewerThanLastBlock(u64, u64),
+    /// The timestamp is too far in the past now: {0}, timestamp: {1}
```
nick-mobilecoin, line 18: From, https://github.com/mobilecoinfoundation/mobilecoin/pull/3711#discussion_r1393486807
> Similarly, combining the timestamp with the word before it seems a bit weird to me

perhaps
"The current block time is too old, current time {0}, block time: {1}"
IDk the "current block" and "current time" is confusing, I think "now" feels a bit off in the error message.

consensus/service/src/mint_tx_manager/mod.rs
```diff
@@ -108,32 +109,25 @@ impl<L: Ledger> MintTxManager for MintTxManagerImpl<L> {
...
-            }
-            seen_nonces.insert(nonce_with_token_id);
-            true
-        });
-
-        Ok(allowed_txs)
+    ) -> MintTxManagerResult<Vec<(MintConfigTx, u64)>> {
+        let mut candidates = timestamp_validator::sort_and_dedup(txs.iter());
+        candidates.truncate(max_elements);
+        Ok(candidates)
```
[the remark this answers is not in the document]
eranrund: So the idea here is that the nonce needs to be unique per token id - which does mean you can see the same nonce across different token ids. We don't want two instances of the same `(nonce, token_id)` which previously the logic here enforced, but with this change that will no longer be the case since you will be comparing against the entire `MintConfigTx`.
Perhaps `nonce` is a wrong name and it should've been `unique id` or something like that. Does that make sense?

eranrund (changes requested):
Overall this LGTM but there's an issue with the nonce de-duping stuff in the mint/mint config txs, and maybe a need to allow for timestamps in the near future. I made comments with more details in the relevant parts.
After those are addressed I'll take a second pass at this since while on the surface this is fairly easy to follow, I do want to increase the chances of not missing anything since this does change the blockchain in a meaningful way.
Thanks for iterating on this!

consensus/enclave/impl/src/lib.rs
nick-mobilecoin: Should be addressed in https://github.com/mobilecoinfoundation/mobilecoin/pull/3733/commits/5a107ab9d09b3a93e4354940ea8e61b998adbd29

consensus/enclave/impl/src/lib.rs
nick-mobilecoin: Older block versions are supported in https://github.com/mobilecoinfoundation/mobilecoin/pull/3733/commits/5a107ab9d09b3a93e4354940ea8e61b998adbd29

consensus/enclave/impl/src/lib.rs
```diff
@@ -3564,4 +3603,158 @@ mod tests {
...
+            timestamp: parent_timestamp,
+        };
+        assert_eq!(
+            SgxConsensusEnclave::validate_timestamp(timestamp, &parent_block),
+            Ok(timestamp)
+        );
+    }
+
+    #[test_with_logger]
+    fn test_form_block_fails_for_invalid_timestamp(logger: Logger) {
```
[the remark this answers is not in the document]
nick-mobilecoin: good catch, I had to look for a bit, it was the default zero but it wasn't clear
This should be explicitly set in https://github.com/mobilecoinfoundation/mobilecoin/pull/3733/commits/5a107ab9d09b3a93e4354940ea8e61b998adbd29

consensus/enclave/impl/src/lib.rs
```diff
@@ -3564,4 +3603,158 @@ mod tests {
...
+            assert_matches!(
+                enclave.form_block(
+                    &parent_block,
+                    FormBlockInputs {
+                        well_formed_encrypted_txs_with_proofs,
+                        ..Default::default()
+                    },
+                    &root_element,
+                ),
+                Err(Error::FormBlock(_))
```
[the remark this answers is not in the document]
nick-mobilecoin: This looks for the word "timestamp" in the error string, done in https://github.com/mobilecoinfoundation/mobilecoin/pull/3733/commits/5a107ab9d09b3a93e4354940ea8e61b998adbd29
I don't want to be too strict on the error string, but think that should suffice to know it's the timestamp and not something else misconfigured

consensus/enclave/impl/src/lib.rs
```diff
@@ -938,13 +953,21 @@ impl ConsensusEnclave for SgxConsensusEnclave {
             mint_txs,
         };
 
+        let timestamp = if config.block_version.timestamps_are_supported() {
+            Self::validate_timestamp(inputs.timestamp, parent_block)?
+        } else {
+            // This should be zero in these instances, Block::new_with_parent()
+            // will panic if it's not.
+            inputs.timestamp
+        };
```
nick-mobilecoin, line 962: We may need to rethink this here.
The byzantine_ledger/worker.rs fills out the timestamps and provides them here, but it doesn't know what the block version the enclave is going to be using.
It may be that we should use zero instead of inputs.timestamp here.

consensus/service/src/byzantine_ledger/worker.rs
```diff
@@ -858,6 +880,8 @@ impl<
             .get_root_tx_out_membership_element()
             .expect("Failed getting root tx out membership element");
 
+        let timestamp = timestamps.into_iter().max().unwrap_or(0);
```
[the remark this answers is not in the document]
nick-mobilecoin: added comment

consensus/service/src/timestamp_validator.rs
```diff
@@ -0,0 +1,274 @@
...
+pub fn validate_with_logger(
+    timestamp: u64,
+    latest_block: &Block,
+    logger: &Logger,
+) -> Result<(), Error> {
+    validate(timestamp, latest_block).map_err(|e| {
+        log::warn!(logger, "Consensus Value timestamp invalid: {e}");
+        e
+    })
+}
```
nick-mobilecoin, line 44: not sure on the thoughts here, I could keep the logging logic fully in mint tx manager if we don't want a function down this low, it just seemed odd to have mint tx manager to explicitly only log the bad time and nothing else.

consensus/enclave/mock/src/lib.rs
```diff
@@ -351,7 +351,7 @@ impl ConsensusEnclave for ConsensusServiceMockEnclave {
         };
 
         let timestamp = if block_version.timestamps_are_supported() {
-            parent_block.timestamp + 1
+            inputs.timestamp
         } else {
             0
         };
```
nick-mobilecoin, line 357: Note this is ignoring the inputs.timestamp for non timestamp version, but it's similar to the enclave impl, the worker.rs populates the value and it doesn't have knowledge of the supported block version.

consensus/enclave/impl/src/lib.rs
eranrund: Hmm, yeah, that seems reasonable to me (unless we plumb the block version into the byzantine ledger worker but it seems too heavy handed to do that just for this.

consensus/service/src/timestamp_validator.rs
eranrund: I think this is fine

consensus/service/src/tx_manager/mod.rs
```diff
@@ -265,11 +273,14 @@ impl<E: ConsensusEnclave + Send, UI: UntrustedInterfaces + Send> TxManager
...
-        let cache_entries = Self::get_cache_entries(&cache, tx_hashes.iter())?;
+        // We don't have a timestamp for each transaction and we aren't going
+        // to return one, so we use 0 for the timestamp used in `get_cache_entries`
+        let hashes = tx_hashes.iter().cloned().zip(repeat(0)).collect::<Vec<_>>();
+        let cache_entries = Self::get_cache_entries(&cache, hashes.iter())?;
 
         cache_entries
             .into_iter()
-            .map(|entry| {
+            .map(|(entry, _)| {
```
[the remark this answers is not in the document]
nick-mobilecoin: fixed in https://github.com/mobilecoinfoundation/mobilecoin/pull/3733/commits/d95e1d2fd6a803a2ac8675a199e19f20a93e9ada

consensus/enclave/impl/src/lib.rs
nick-mobilecoin: Using 0 now, added a comment in this location as well as in the field description of `FormBlockInputs`

consensus/service/src/mint_tx_manager/mod.rs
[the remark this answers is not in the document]
nick-mobilecoin: Added back logic to prevent having multiple `(nonce, token_id)` in the Mint Config or Txs

consensus/service/src/mint_tx_manager/mod.rs
nick-mobilecoin: brought back logic to prevent multiple `(nonce, token_id)`s in the mint transactions.

nick-mobilecoin:
@eranrund I believe I've addressed the timestamp locking up consensus in https://github.com/mobilecoinfoundation/mobilecoin/pull/3733/commits/6bcb229d9c10cbe6992babfd4e25fd9e15746df2

The failure case:
consensus was not able to proceed because a value's timestamp had exceeded the age of 30 seconds.

The slam test threw enough transactions at a node to exceed the [PENDING_LIMIT](https://github.com/mobilecoinfoundation/mobilecoin/blob/f5907cf3374b1e7b57a173af72950f247b789864/consensus/service/src/api/client_api_service.rs#L38) of transactions.
While this node was bogged down handling receiving these, it wasn't able to get a consensus message to all the other nodes within the 30 second time. Thus in this instance nodes `3302` and `3303` saw the values within the 30 second time, but nodes `3300`, `3301` and `3304` did not see these values within the time period. For this test transactions were only being sent to one node, `3302` or `3303`. Since this was the only node proposing txs the network got stuck since this node felt those values were all valid and per SCP a node can't change it's mind on voting to nominate a value.  Currently the only slot reset mechanism is if a node sees that the other nodes are moving forward in the blockchain without it, then that node will abort a slot attempt and catch up.
In this test scenario only one node was receiving transactions so there was no way for the block to move forward. While less likely there is potential for this in the real network if one were able to flood transactions to enough nodes and stall them enough such that the necessary quorum consider all the values invalid.

We could increase the age out time to something like 5 or 10 minutes, to reduce the likely hood of this event but still allow manual restarts. However, in general I'm thinking we shouldn't allow values to become invalid outside of the blockchain moving forward.

nick-mobilecoin:
I'm currently looking at the failure on https://github.com/mobilecoinfoundation/mobilecoin/actions/runs/7049871021
I haven't seen that one before so it may be related to this change :(.

eranrund:
> @eranrund I believe I've addressed the timestamp locking up consensus in [6bcb229](https://github.com/mobilecoinfoundation/mobilecoin/commit/6bcb229d9c10cbe6992babfd4e25fd9e15746df2)
> 
> The failure case: consensus was not able to proceed because a value's timestamp had exceeded the age of 30 seconds.
> 
> The slam test threw enough transactions at a node to exceed the [PENDING_LIMIT](https://github.com/mobilecoinfoundation/mobilecoin/blob/f5907cf3374b1e7b57a173af72950f247b789864/consensus/service/src/api/client_api_service.rs#L38) of transactions. While this node was bogged down handling these, it wasn't able to get a consensus message to all the other nodes within the 30 second time. Thus, in this instance, nodes `3302` and `3303` saw the values within the 30 second time, but nodes `3300`, `3301` and `3304` did not see these values within the time period. For this test, transactions were only being sent to one node, `3302` or `3303`. Since this was the only node proposing txs, the network got stuck, because this node felt those values were all valid and per SCP a node can't change it's mind on voting to nominate a value. Currently the only slot reset mechanism is if a node sees that the other nodes are moving forward in the blockchain without it, then that node will abort a slot attempt and catch up. In this test scenario, only one node was receiving transactions so there was no way for the block to move forward. While less likely, there is potential for this in the real network if one were able to flood transactions to enough nodes and stall them enough such that the necessary quorum considered all other the values invalid due to time out.
> 
> We could increase the age out time to something like 5 or 10 minutes, to reduce the likely hood of this event but still allow manual restarts. However, in general I'm thinking we shouldn't allow values to become invalid outside of the blockchain moving forward.

Thanks for digging into this - this makes sense. I think removing that check is acceptable.

The [current CI failure](https://github.com/mobilecoinfoundation/mobilecoin/actions/runs/7049871021/job/19189327703?pr=3733) is also an interesting one but since it's fog I somehow doubt its related to the changes in this PR. I've re-ran it to see what happens.

consensus/service/src/tx_manager/mod.rs
```diff
@@ -216,15 +224,18 @@ impl<E: ConsensusEnclave + Send, UI: UntrustedInterfaces + Send> TxManager
...
         let context_opt = {
             let cache = self.lock_cache();
             cache.get(tx_hash).map(|entry| entry.context.clone())
         };
 
         if let Some(context) = context_opt {
             let _timer = counters::VALIDATE_TX_TIME.start_timer();
-            self.untrusted.is_valid(context)?;
+            self.untrusted.is_valid(context, timestamp).map_err(|e| {
+                log::warn!(self.logger, "Failed validated tx_hash: {e}");
```
eranrund, line 236:
```suggestion
                log::warn!(self.logger, "Failed validating tx_hash {tx_hash}: {e}");
```

eranrund (approved):
LGTM, but we should probably wait for an accepted MCIP before merging.
Thanks for working through this!

PR #3734, opened 2023-11-20
Add `timestamp` field to `Block`

nick-mobilecoin:
The `Block` now has a `timestamp` field. This field is meant to
represent a consensed timestamp for the block that all the nodes will
sign.

Currently the timestamp is *not* being populated. The logic to properly
set the timestamp is expected to come in a follow on PR.

### Future Work

[Soundtrack of this PR]()


nick-mobilecoin:
* main
  * **PR #3734** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3734?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #3733** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3733?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3734?utm_source=stack-comment).

consensus/service/src/byzantine_ledger/pending_values.rs
```diff
@@ -22,15 +22,16 @@ pub struct PendingValues<TXM: TxManager, MTXM: MintTxManager> {
...
+    /// duplicates
+    ///     2. Track how long each transaction took to externalize.
     ///
     /// To accomplish these goals we store, in addition to the queue of pending
     /// values, a map that maps a value to when we first encountered it.
     /// This essentially gives us an ordered HashMap.
     ///
-    /// Note that we only store a timestamp for values that were handed to us
-    /// directly from a client. That behavior is enforced by
+    /// Note that we only store a received time for values that were handed to
```
nick-mobilecoin, line 33: Renamed `timestamp` in this module to `received_time` to avoid the ambiguity with the block timestamp versus the performance logging of the value processing

ledger/db/src/ledger_db.rs
```diff
@@ -2400,6 +2401,7 @@ mod ledger_db_test {
                 &last_block,
                 &Default::default(),
                 &block_contents,
+                last_block.timestamp + 1,
```
nick-mobilecoin, line 2404: Where possible increase previous timestamp by 1 to conform with logic coming down the line to ensure timestamps are valid.

ledger/db/src/test_utils/mod.rs
```diff
@@ -429,7 +429,13 @@ pub fn add_block_contents_to_ledger(
     let new_block = if num_blocks > 0 {
         let parent = ledger_db.get_block(num_blocks - 1)?;
 
-        Block::new_with_parent(block_version, &parent, &Default::default(), &block_contents)
+        Block::new_with_parent(
+            block_version,
+            &parent,
+            &Default::default(),
+            &block_contents,
+            parent.timestamp + 1,
```
eranrund, line 437: Not sure if this causes any problems (or will cause) but this will set a timestamp even for block versions that do not support it. Perhaps worth setting to 0 based on the block_version?

eranrund (approved):

ledger/db/src/test_utils/mod.rs
nick-mobilecoin: Let me know what you think, I pushed the logic into `Block::new()` as well as `api/src/convert/block.rs`

api/src/convert/block.rs
```diff
@@ -31,6 +38,12 @@ impl TryFrom<&blockchain::Block> for Block {
         let root_element = TxOutMembershipElement::try_from(value.get_root_element())?;
         let contents_hash = BlockContentsHash::try_from(value.get_contents_hash())?;
 
+        let timestamp = if value.get_version().timestamps_are_supported() {
+            value.get_timestamp()
+        } else {
+            0
+        };
```
nick-mobilecoin, line 45: Should we error out if we got a timestamp in a version less than 3, or is zeroing an ok thing to do here?

blockchain/types/src/block.rs
```diff
@@ -123,14 +136,24 @@ impl Block {
...
         version: BlockVersion,
         parent_id: &BlockID,
         index: BlockIndex,
         cumulative_txo_count: u64,
         root_element: &TxOutMembershipElement,
         block_contents: &BlockContents,
+        timestamp: u64,
     ) -> Self {
+
+        let timestamp = if version.timestamps_are_supported() {
```
nick-mobilecoin, line 151: If desired we could panic here if someone provides a timestamp for block version 3 or lower, throwing it away is kind of nice ergonomically, but is also a bit hidden

api/src/convert/block.rs
eranrund: I actually am not sure we should have the check here at all, since its not like we validate the block_id is correct either, but if we were to keep the check then I think erroring is the right move. I don't think we should have logic in conversion methods that modify the data.

blockchain/types/src/block.rs
eranrund: I think panicing in the form of an assert seems reasonable, it's not something we expect to see ever happen and we be good to catch it if it does.

ledger/db/src/test_utils/mod.rs
eranrund: I think having that particular function have subtle behavior were it zeros out an unexpected timestamp is not ideal since that could mask some misbehavior up the chain that we didn't expect, where somehow a timestamp appeared where it shouldn't have, so as I suggested in the other comment I think panicing there would be my preference. But open for other ideas/suggestions.

api/src/convert/block.rs
nick-mobilecoin: Are you thinking remove these and place a panic where we create the block?

Then the test utilities would have the logic to ensure tests populate timestamps appropriatly?

blockchain/types/src/block.rs
```diff
@@ -123,14 +135,19 @@ impl Block {
...
         version: BlockVersion,
         parent_id: &BlockID,
         index: BlockIndex,
         cumulative_txo_count: u64,
         root_element: &TxOutMembershipElement,
         block_contents: &BlockContents,
+        timestamp: u64,
     ) -> Self {
+        assert!(timestamp == 0 && version.timestamps_are_supported());
+
```
nick-mobilecoin, line 150: I'll leave it to you @eranrund on keeping the assert.

I'm a bit scarred from being in code bases where people were too aggressive on asserts and it prevented one from quickly getting started.
At the same time having the error surface as soon as possible in the pipeline is nice.

api/src/convert/block.rs
eranrund: > Are you thinking remove these and place a panic where we create the block?
> 
> Then the test utilities would have the logic to ensure tests populate timestamps appropriatly?

Yes thats what I was thinking

consensus/enclave/impl/src/lib.rs
```diff
@@ -937,13 +937,14 @@ impl ConsensusEnclave for SgxConsensusEnclave {
...
         };
-        //
+
         // Form the block.
         let block = Block::new_with_parent(
             config.block_version,
             parent_block,
             root_element,
             &block_contents,
+            0,
```
nick-mobilecoin, line 947: Because we don't have a timestamp value here yet we can't assert that it's 0 or supported and not zero, moving that change into #3733

eranrund (approved):

PR #3735, opened 2023-11-21
Use `EvidenceKind` in sql recovery db

nick-mobilecoin:
Previously `AttestationEvidenceOneOf` was used in sql recovery db. This
prevented decoding the legacy EPID based fog reports. Now `EvidenceKind`
is used which allows the decoding of legacy EPID fog reports.

### DIFF
--- a/fog/sql_recovery_db/src/lib.rs
+++ b/fog/sql_recovery_db/src/lib.rs
@@ -18,7 +18,7 @@ mod schema;
 mod sql_types;
 
 use crate::sql_types::{SqlCompressedRistrettoPublic, UserEventType};
-use ::prost::{DecodeError, Message};
+use ::prost::Message;
 use chrono::NaiveDateTime;
 use clap::Parser;
 use diesel::{
@@ -26,6 +26,7 @@ use diesel::{
     prelude::*,
     r2d2::{ConnectionManager, Pool},
 };
+use mc_attest_verifier_types::EvidenceKind;
 use mc_blockchain_types::Block;
 use mc_common::{
     logger::{log, Logger},
@@ -38,7 +39,6 @@ use mc_fog_recovery_db_iface::{
     IngressPublicKeyRecord, IngressPublicKeyRecordFilters, IngressPublicKeyStatus, RecoveryDb,
     RecoveryDbError, ReportData, ReportDb,
 };
-use mc_fog_report_types::AttestationEvidenceOneOf;
 use mc_fog_types::{
     common::BlockRange,
     view::{FixedTxOutSearchResult, TxOutSearchResultCode},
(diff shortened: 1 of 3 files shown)


nick-mobilecoin:
* main
  * **PR #3735** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3735?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3735?utm_source=stack-comment).

fog/sql_recovery_db/src/lib.rs
```diff
@@ -1147,14 +1147,12 @@ impl SqlRecoveryDb {
             .load::<(Option<i64>, String, Vec<u8>, i64)>(conn)?
             .into_iter()
             .map(|(ingest_invocation_id, report_id, report, pubkey_expiry)| {
-                let attestation_evidence = AttestationEvidenceOneOf::decode(&*report)?.evidence;
+                let attestation_evidence = EvidenceKind::from_bytes(report)?;
```
nick-mobilecoin, line 1150: Since this decodes here and encodes on line 1221 our tests ensure the round trip logic works, but they don't actually check the historically encoded EPID report could be decoded.

I'm not sure it's worth it, updating the tests to support it as EPID should go away after the 6.0 release. We only saw this issue as we tried to re-use a database that had a pre-existing fog report in it.  If we had removed the table entries like the one fog migration it would have been a non issue.

The way this was tracked down was that I ran the `test_reports_db()` test case and replaced the `report` bytes here with the `hex::decode()` output of the attached txt file.

For reference the start of this text file is
```
0a83020a80020
```
while the previous behavior would result in hex bytes of
```
12bd1e0a83020a8
```
Notice the `12bd1e` which represents the proto tag `2` and the size of the subsequent bytes.
[old_report.txt](https://github.com/mobilecoinfoundation/mobilecoin/files/13421914/old_report.txt)

dolanbernard (approved):

eranrund (approved):
We should probably start a document (if we haven't already) with notes/things to remember when releasing 6.0. I don't think that right now we have a `clear out the reports table` as part of our new release deployment playbook.

nick-mobilecoin:
> We should probably start a document (if we haven't already) with notes/things to remember when releasing 6.0. I don't think that right now we have a `clear out the reports table` as part of our new release deployment playbook.

This PR should eliminate the need to clear out the reports table. I think this instance was a misunderstanding on my part.
However I agree we should find a way to capture gotchas or steps to do.  Currently we're a bit fragmented on tracking.  Internally we have asana, jira and notion. Many of the previous notion release steps have been brought into asana. I would vote we create these kind of tasks there. At the same time it may not work well for a procedure checklist.

PR #3740, opened 2023-11-22
Use `serial` backend for curve25519-dalek

nick-mobilecoin:
Previously the `curve25519-dalek` was using the `simd` backend. This
backend performed slower than the `serila` in debug builds. Looking at
the release builds there doesn't seem to be any difference in
performance. Of note is that the release build removed all of the rust
intrinsic function calls, while the debug builds kept them around. Now
the `serial` backend is used for curve25519-dalek to allow for the debug
builds to run better.

With `simd` debug test runs of
`transaction_builder::transaction_builder_tests::test_max_transaction_size`
would take ~80 seconds. Using the `serial` backend this test takes ~15
seconds.

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -5,16 +5,16 @@
 # ...so instead we list all target triples (Tier 1 64-bit platforms)
 
 [target.x86_64-unknown-linux-gnu]
-rustflags = ["-C", "target-cpu=skylake", '--cfg=curve25519_dalek_backend="simd"', '--cfg=feature="precomputed-tables"']
+rustflags = ["-C", "target-cpu=skylake", '--cfg=curve25519_dalek_backend="serial"', '--cfg=feature="precomputed-tables"']
 
 [target.x86_64-pc-windows-gnu]
-rustflags = ["-C", "target-cpu=skylake", '--cfg=curve25519_dalek_backend="simd"', '--cfg=feature="precomputed-tables"']
+rustflags = ["-C", "target-cpu=skylake", '--cfg=curve25519_dalek_backend="serial"', '--cfg=feature="precomputed-tables"']
 
 [target.x86_64-pc-windows-msvc]
-rustflags = ["-C", "target-cpu=skylake", '--cfg=curve25519_dalek_backend="simd"', '--cfg=feature="precomputed-tables"']
+rustflags = ["-C", "target-cpu=skylake", '--cfg=curve25519_dalek_backend="serial"', '--cfg=feature="precomputed-tables"']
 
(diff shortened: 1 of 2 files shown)


nick-mobilecoin:
* main
  * **PR #3740** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3740?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3740?utm_source=stack-comment).

nick-mobilecoin:
For reference attached here are before and after times running locally

[before.txt](https://github.com/mobilecoinfoundation/mobilecoin/files/13444970/before.txt)
[after.txt](https://github.com/mobilecoinfoundation/mobilecoin/files/13444971/after.txt)

nick-mobilecoin:
This looks to be a debug only performance gain.

Doing some digging and some more thinking on this [_mm256_set1_epi64x](https://doc.rust-lang.org/beta/src/core/stdarch/crates/core_arch/src/x86/avx.rs.html#2503-2505) and the like is implemented as a function in rust.
In particular [_mm256_set1_epi64x](https://doc.rust-lang.org/beta/src/core/stdarch/crates/core_arch/src/x86/avx.rs.html#2503-2505) calls another function, [_mm256_setr_epi64x](https://doc.rust-lang.org/beta/src/core/stdarch/crates/core_arch/src/x86/avx.rs.html#2421-2423) which calls [transmute](https://doc.rust-lang.org/src/core/intrinsics.rs.html#1380), though transmute looks to be  a built in.
The _mm256* functions are marked inline, but this in-lining only happens in opt-level 1 or higher.
The results seem similar to https://internals.rust-lang.org/t/a-bunch-of-core-functions-should-be-inline-always/14173
From the previous link
> I thought #[inline(always)] didn't actually have an impact in debug mode, [but it does 19](https://rust.godbolt.org/z/EzGMq3).
Looking at [inline-threshold](https://doc.rust-lang.org/rustc/codegen-options/index.html#inline-threshold) one can see that it calls out inline always for opt-level 0. Though it says opt-level 1 should only add lifetime intrinsics, it seems that it is inlining more than that.

Since this looks to be only a debug performance gain it may be best not to make any change at this time.
In general I don't think we should downplay or sacrifice debug performance, but not seeing any immediate gains at this time. The performance issue here came onto my radar as I was looking at overall unit test performance. Not because these tests in particular were noticeably slowing me down.

PR #3749, opened 2023-11-28
Enables dcap

aweis89:
### Future Work

[Soundtrack of this PR]()


.github/workflows/mobilecoin-workflow-dev-deploy.yaml
```diff
@@ -331,11 +326,6 @@ jobs:
         image:
           org: ${{ inputs.docker_image_org }}
         fogIngest:
-          resources:
-            limits:
-              intel.com/sgx: 2000
-            requests:
-              intel.com/sgx: 2000
```
joekottke, line 338: Wouldn't we want these to be updated to the new resource definition?  We have lower limits in Dev than we do in Prod.

joekottke (changes requested):
The `ledger-distribution` check in the `node_hw` entrypoint definitely needs to be added back in, and I'm curious about the resource definitions/amounts.

.internal-ci/helm/fog-ledger/values.yaml
```diff
@@ -62,21 +62,21 @@ fogLedger:
 
     resources:
       limits:
-        intel.com/sgx: 5000
+        sgx.intel.com/epc: "512Ki"
```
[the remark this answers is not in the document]
jgreat: so 5000 should be 20Mi size. Probably `20MiB` if it supports proper bytes.

.internal-ci/helm/fog-ingest/values.yaml
```diff
@@ -30,25 +30,26 @@ fogIngest:
   podAnnotations:
     fluentbit.io/include: 'true' # collect logs with fluentbit
     fluentbit.io/exclude-jaeger-agent: 'true'
+    sgx.intel.com/quote-provider: fog-ingest
```
jgreat, line 33: What does this do? 

`quote-provider` makes me think its the source for getting quotes, so Azure, or Intel endpoint.

.internal-ci/docker/Dockerfile.bootstrap-tools
```diff
@@ -16,17 +16,17 @@ ARG DEBIAN_FRONTEND=noninteractive
...
-        gettext \
-        libssl1.1 \
-        jq \
-        vim \
-        python3-pip \
-        zstd \
-        openssl \
- && rm -rf /var/cache/apt \
- && rm -rf /var/lib/apt/lists/*
+  && apt-get install -y \
```
jgreat, line 19: The indents here were purposeful to show that these were options under the `apt-get install` command. It would be nice to have them back.

jgreat (changes requested):
Some minor clean up on formatting.

Side note  I would love if we could find a automatic yaml linting tool with something like .rc rules so we can standardize.

aweis89:
> Some minor clean up on formatting.
> 
> Side note I would love if we could find a automatic yaml linting tool with something like .rc rules so we can standardize.

Ironically I've been using yamlls which does formatting as well, but it actually seems to prefer to double quotes so I needed to turn it off to revert the `"true"` to `'true'` to single quotes

jgreat (approved):
Will follow up with sgx.intel.com/epc values.

PR #3770, opened 2023-12-06
Remove newlines from enclave panic message

nick-mobilecoin:
The enclave panic message is passed to a logger that doesn't work with
newlines. In order to keep the contents of the panic message from the
enclave a space is used to separate the "Enclave panic:" prefix and the
actual message.

### Future Work

[Soundtrack of this PR]()


nick-mobilecoin:
This was originally done in `release/v5.2` under #3706, however it was reverted in order to prevent unecessary changes to the enclave binaries, #3755.

PR #3807, opened 2023-12-15
Move sim certificates to build directory

nick-mobilecoin:
Previously the sim enclave signing certificates were in the source
directory and used a random key. This caused issues with incremental
builds and at timing issues where the enclaves were in the process of
creating a certificate while the untrusted code was too. This resulted
in attestation errors due to signature mismatches, which would force
someone to rebuild again. Now the sim signing certificates are in the
build directory and use a common default seed.

### DIFF
--- a/attest/verifier/build.rs
+++ b/attest/verifier/build.rs
@@ -21,11 +21,9 @@ use p256::{
 use rand::{RngCore, SeedableRng};
 use rand_hc::Hc128Rng;
 use std::{
-    env,
-    fs::write,
+    env, fs,
     path::{Path, PathBuf},
     sync::{Arc, Mutex},
-    time::SystemTime,
 };
 
 use p256::pkcs8::{
@@ -50,10 +48,8 @@ lazy_static::lazy_static! {
             },
             Err(_) => {
                 cargo_emit::warning!(
-                    "Using thread_rng() to generate mock attestation report signatories for simulation-mode enclaves"
+                    "Using default seed to generate mock attestation report signatories for simulation-mode enclaves"
                 );
-                let mut csprng = rand::thread_rng();
-                csprng.fill_bytes(&mut seed[..]);
             }
         }
         Hc128Rng::from_seed(seed)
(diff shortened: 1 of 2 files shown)


nick-mobilecoin:
* `main`
  * **PR #3807** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3807?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/mobilecoin/3807?utm_source=stack-comment).

attest/verifier/build.rs
```diff
@@ -50,10 +48,8 @@ lazy_static::lazy_static! {
             },
             Err(_) => {
                 cargo_emit::warning!(
-                    "Using thread_rng() to generate mock attestation report signatories for simulation-mode enclaves"
+                    "Using default seed to generate mock attestation report signatories for simulation-mode enclaves"
                 );
-                let mut csprng = rand::thread_rng();
-                csprng.fill_bytes(&mut seed[..]);
```
nick-mobilecoin, line 56: This was discussed offline, we don't think there's any benefit to using a random seed for sim builds since they aren't secure. The random seed per generation causes issues at times trying to attest, due to the timing of enclaves and untrusted executing this build script concurrently

attest/verifier/build.rs
```diff
@@ -193,50 +183,13 @@ fn generate_sim_files(data_path: impl AsRef<Path>) {
...
-            .to_std()
-            .expect("Failed to convert to std duration");
-
-    if generated_files.iter().any(|file| {
-        file.metadata()
-            .and_then(|f| f.modified().map(|t| t < regenerate_time))
-            .unwrap_or(true)
-    }) {
-        return true;
-    }
```
nick-mobilecoin, line 227: were running into enough issues with certificate contention causing a  developer to rebuild everything, it seems like IFF the developer has a 30 day old certificate on their machine having them rebuild every 30 days is still cheaper than daily or more for the previous behavior.

eranrund (approved):

PR #3815, opened 2023-12-28
Fix missing punctuation in common README

umbr4g3:
Minor fix for missing punctuation in a README file.


common/README.md
```diff
@@ -2,8 +2,8 @@
 
 This crate contains several things right now:
 
-- Common structs that are used at API boundaries of higher-level crates
-  Putting the structs in `common` instead of in a larger crate allows to break
+- Common structs that are used at API boundaries of higher-level crates.
+  Putting the structs in `common` instead of in a larger crate allows us to break
```
nick-mobilecoin, line 6: Since this sentence is being fixed, perhaps it should also get updated to be something like
```suggestion
  Putting the structs in `common` instead of in a larger crate allows us to break up
```

I don't think the intent was to "break" the dependencies so much as make them smaller

nick-mobilecoin:
Sorry this is taking a bit longer than ideal
I'm tracking down the CLA signing process per, https://github.com/mobilecoinfoundation/mobilecoin/blob/b063a81ee0bc4d6b34d18a0eac5eb323c759b339/CONTRIBUTING.md?plain=1#L43

nick-mobilecoin:
@umbr4g3 if you can e-mail me your full name and email address(which will probably be your from address), I can get a CLA sent out for you to sign

You can find my e-mail in the combined commit https://github.com/mobilecoinfoundation/mobilecoin/pull/3815/commits/e715155c5bd2b774b64b5f7a2aca3f16ea214a23 or in a less scrapey format as `nick at mobilecoin dot com`

nick-mobilecoin (approved):

PR #3817, opened 2024-01-03
mobilecoind: fix optimization tx's when using nonzero token ids

cbeck88:
without this, we hit an error in "select_utxos_for_optimization"

### DIFF
--- a/mobilecoind/src/payments.rs
+++ b/mobilecoind/src/payments.rs
@@ -1778,12 +1778,12 @@ mod test {
             utxos[3].value = 300 * MILLIMOB_TO_PICOMOB;
             utxos[4].value = 2000 * MILLIMOB_TO_PICOMOB;
             utxos[5].value = 1000 * MILLIMOB_TO_PICOMOB;
-            utxo[0].token_id = eusd;
-            utxo[1].token_id = eusd;
-            utxo[2].token_id = eusd;
-            utxo[3].token_id = eusd;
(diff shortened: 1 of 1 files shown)


mobilecoind/src/payments.rs
```diff
@@ -1762,6 +1765,38 @@ mod test {
         }
     }
 
+    #[test]
+    fn test_select_utxos_for_optimization_works_with_eusd() {
```
nick-mobilecoin, line 1769: Not understanding how this test pertains to the code change?

The code change "normalized" the inputs being passed to `select_utxos_for_optimization()` to be from the same token id. This test seems to be testing `select_utxos_for_optimization()`, but not the result of the input normalization.

cbeck88: I'm just trying to follow the patterns that exist and add test coverage where it doesn't exist.

If it were easy to construct a TransactionManager instance in a test, we could do that, but it's rather complicated, what with connection manager and all that, and I don't see any example to go off of.

If you prefer not to have this test I'm happy to remove it.

nick-mobilecoin: I guess I kind of glanced over the test name it does say "works with eusd"
From that perspective I can see how this improves case coverage.
I've approved it will need some changes to actually compile.

nick-mobilecoin (approved):

PR #3858, opened 2024-01-12
remove deprecated and unused build/ops files

jgreat:
### Motivation

Remove unused build and deploy tools.  These files haven't been in active use since v2 or v3 and we transitioned to GitHub Actions.


docker/rust-toolchain
```diff
@@ -1,2 +0,0 @@
-[toolchain]
```
nick-mobilecoin, line 1: I think this needs to get moved to the root
THere is a symlink from the root rust-toolchain to this

nick-mobilecoin (changes requested):
I think the rust-toolchain needs to be moved to the root, it's now a broken symlink

docker/rust-toolchain
jgreat: That explains the build failure :)

nick-mobilecoin (approved):

joekottke (approved):

PR #3859, opened 2024-01-12
use mobilecoinofficial/gh-actions where possible

jgreat:
### Motivation

Simplify and centralize maintenance of common upstream actions.  We can maintain action versions in one place instead of individually across all of our repos.

Use some simplified combined actions to replace repetitive multi step jobs (like mobilecoinofficial/gh-actions/docker)

- Replace actions/checkout with mobilecoinofficial/gh-actions/checkout
- Replace docker* actions with combined mobilecoinofficial/gh-actions/docker
- Clean up unused "skip" if statements. Skipping parts of the workflow will just end up with failed later steps.


.github/workflows/android-bindings-dispatch.yml
```diff
@@ -10,7 +10,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: mobilecoinofficial/gh-actions/checkout@v0
```
nick-mobilecoin, line 13: are we sure v0 is the one we want?
I see that it's released up to version v0.0.18

But looking at the history I see
```
*   6a49472 (HEAD -> main, tag: v0.0.18, tag: v0.0, tag: v0, origin/main, origin/HEAD) Merge pull request #9 from mobilecoinofficial/add-zstd-install
```

Are we deleting and recreating tags v0.0 and v0?
That may be an issue if a collaborator doesn't blow away the repo and reclone when working on it.
i.e. when v0.0.19 comes down my local repo will continue to keep v0 on the v0.0.18 tag 

my understanding is that most GH actions repos keep a `v0` branch around and that's how the actions move forward, by grabbing the tip of that branch

nick-mobilecoin (approved):
Should these get updated too?
```
.github/workflows/dependent-repos.yml
21:    - uses: actions/checkout@v4
41:    - uses: actions/checkout@v4
```

.github/workflows/android-bindings-dispatch.yml
jgreat: First question the checkouts on dependent-repos.yml.

Those jobs are doing something sort of unexpected and not checking out to the standard path. I see why, but I didn't think about it when I made the MCO checkout action.  I can add a pass through for that option, but have not yet. Let me know if you want me to do that before merging this.

Second question - tags. 

When you start with no tags, the auto versioning tool starts at v0.0.1.  v0 and v0.0 are over written by each update until we need to bump the major/minor versions for some reason. Overwriting the major/minor seems to be [common practice](https://docs.github.com/en/actions/creating-actions/about-custom-actions#using-tags-for-release-management)

jgreat: To address the tag concern.  Each action run pulls down the current v0 tag at runtime, so the version of the local repo won't determine what git sha you're actually using.  One run could be v0.0.18 and the next could be v0.0.19. If you hit the timing right different jobs in a single run could use different versions.  A re-run will use what ever is the current version at run time.  

This also means we can revert out of a bad change by resetting the v0 tag to an older release, but usually better practice to move forward, even if you're undoing a change.

joekottke (approved):

.github/workflows/android-bindings-dispatch.yml
nick-mobilecoin: > First question the checkouts on dependent-repos.yml.

TBH, i didn't think full service had been updated to the new dcap so surprised it was building before...???
I think we do it after as the lack of failure for dcap makes me lean toward it's not as much of a sanity check as it could be.

PR #3873, opened 2024-01-22
Set block_version to 4 and exceed by to 0 in charts.

jgreat:
### Motivation

Setting good defaults for the release. 

- Block version set to `4`
- Set the scale up `exceedBlockHeightBy` to `0` to work around ledger store reporting 0 when no blocks in the latest store.

### DIFF
--- a/.github/workflows/mobilecoin-dev-cd.yaml
+++ b/.github/workflows/mobilecoin-dev-cd.yaml
@@ -342,14 +342,14 @@ jobs:
 ###############################################
 # Deploy current version to namespace block v3
 ###############################################
-  deploy-current-bv3-release:
+  deploy-current-bv4-release:
     uses: ./.github/workflows/mobilecoin-workflow-dev-deploy.yaml
     needs:
     - bootstrap-v5-bv3
     - charts
     - generate-metadata
     with:
-      block_version: 3
+      block_version: 4
       chart_repo: ${{ needs.generate-metadata.outputs.chart_repo }}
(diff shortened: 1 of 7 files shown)


.github/workflows/mobilecoin-dev-cd.yaml
```diff
@@ -342,14 +342,14 @@ jobs:
 ###############################################
 # Deploy current version to namespace block v3
```
nick-mobilecoin, line 343: nit
```suggestion
# Deploy current version to namespace block v4
```

nick-mobilecoin (approved):
double checking

From the name bootstrap, i'm interpreting that to mean it's loading up the older instance and that's why it's still block version 3
from `mobilecoin-dispatch-dev-deploy.yaml`
```
  bootstrap:
    needs:
    - list-values
    if: inputs.bootstrap_version != 'none'
    uses: ./.github/workflows/mobilecoin-workflow-dev-bootstrap.yaml
    with:
      block_version: '3'
      chart_repo: ${{ inputs.chart_repo }}
      namespace: ${{ inputs.namespace }}
      version: ${{ inputs.bootstrap_version }}
    secrets: inherit
```

jgreat:
> double checking
> 
> From the name bootstrap, i'm interpreting that to mean it's loading up the older instance and that's why it's still block version 3 from `mobilecoin-dispatch-dev-deploy.yaml`

I think this one should be 3 until the 6.0 release is complete.

PR #3879, opened 2024-01-23
Add support for setting tx private key in tx builder, expose in mobilecoind api

cbeck88:
This can be used to create a relatively easy to use idempotence strategy for mobilecoind payments.

From a vendor or an exchange's point of view, it's really nice if you can have an "idempotence key" in an API that sends money somewhere. Here idempotence means, if I make this request, and retry in case of connection failures, there's no way that it can lead to a payment being sent twice.

For example, in mobilecoind, there is a `SendPayment` API which is meant to be the easiest way to send money to a public address. However, there is no idempotence key on this API, and so if I make the request, then due to connectivity issues, the response is not delivered, but the request was handled successfully, a retry will cause a new payment to be sent, with new UTXOs selected. So in realistic deployment environments, this API cannot actually be used safely by an exchange or a vendor. Some fraction of the time, random networking issues will cause payments to be delivered twice, and loss of funds.

It's best for the idempotence mechanism to be grounded in blockchain rules, because local wallet state is comparatively ephemeral.

The mobilecoin blockchain does not have a built-in notion of idempotence keys, but it does have two points of exclusion for transactions -- a UTXO cannot be spent twice, due to the key image mechanism, and a TxOut public key is unique across the whole chain.

One way that an idempotence mechanism can be created is by fixing the tx private key of (one or more) TxOut's in a transaction that you create. The tx private key is the only random element used when the TxOut public key, target key, and amount commitment are generated for a given recipient, token id, and value, so fixing the recipient, amount, and tx private key, fixes these fields of the TxOut, in particular the public key.

From an exchange or vendor's point of view, fixing the tx private key has additional benefits -- if you know the tx private key, amount, and recipient, you can prove the amount and recipient of a TxOut that you sent, even if the recipient themselves does not cooperate. This can be helpful to resolve disputes. Alternative existing mechanisms like shared secret and confirmation number don't accomplish this, because they can only be checked by the recipient, whereas this tx private key mechanism can be checked by anyone.

Existing wallets do not expose the tx private key, it is generally discarded instead. This is perhaps very good from a privacy point of view, but it basically gives up on robust dispute resolution.

This PR gives an exchange or vendor a relatively easy way to get both things. Such a user could decide to derive the tx_private_key of a TxOut, for instance, by a hash of the payment id or withdrawal id, together with a 32 byte secret.

As an example, the private spend key could be the secret used in this hash, however, that has some unexpected consequences like, rotating your private key could cause a loss of idempotence. As an alternative, a user could even choose to use an entirely new secret here which is not derived from the mnemonic, so that they can still change their mnemonic without risking loss of idempotence (and loss of funds).

Instead of mandating any such mechanism, this PR allows the application developer to decide exactly how this should be done. The new `tx_private_key` field of an `Outlay` is optional, but when specified, is expected to be a 32-byte hash. When used this way, a call to the `SendPayment` API can become idempotent.

---

Note that, this is an alternative to an approach developed some time ago in the mobile SDKs, when mobile wallets needed to establish idempotence for payments. The alternative approach is to seed the RNG used with the transaction builder, in order to control the tx private keys. This seed was often a hash using the payment id.

The main drawback of this is that it is fairly brittle.

* Invoking transaction builder functions in a different order can break idempotence, so changing code that uses it may become hazardous.
* Upgrading the seeded CSPRNG algortihm can break idempotence, and so becomes hazardous.

Additionally, from the user point of view, it only accomplishes one of my goals. It is a strategy for controlling the tx private key, but it doesn't help me learn the tx private key, and none of the existing APIs return that value.

I believe that this approach is the simplest change that accomplishes both goals, and is backwards compatible from an API perspective. Application developers are free to use either approach to establish and maintain idempotence.

### DIFF
--- a/mobilecoind/api/proto/mobilecoind_api.proto
+++ b/mobilecoind/api/proto/mobilecoind_api.proto
@@ -118,6 +118,9 @@ enum TxStatus {
 message Outlay {
     uint64 value = 1;
     external.PublicAddress receiver = 2;
+    // Optional tx private key to use for this tx out. This can be chosen explicitly as a mechanism for idempotence.
+    // Must be 32 pseudorandom bytes or none. Will be reduced to curve25519 scalar mod order.
+    bytes tx_private_key = 3;
 }
 
 // Structure used in specifying the list of outputs in a transaction.
@@ -126,6 +129,9 @@ message OutlayV2 {
     uint64 value = 1;
     external.PublicAddress receiver = 2;
     uint64 token_id = 3;
+    // Optional tx private key to use for this tx out. This can be chosen explicitly as a mechanism for idempotence.
+    // Must be 32 pseudorandom bytes or none. Will be reduced to curve25519 scalar mod order.
+    bytes tx_private_key = 4;
 }
 
 // Structure used to refer to a TxOut in the ledger that is presumed to be spendable.
--- a/mobilecoind/src/conversions.rs
+++ b/mobilecoind/src/conversions.rs
@@ -65,11 +65,8 @@ fn bytes_to_tx_private_key(bytes: &[u8]) -> Result<Option<RistrettoPrivate>, Con
         return Ok(None);
     }
 
-    if let Ok(bytes) = <&[u8; 32] as TryFrom<&[u8]>>::try_from(bytes) {
-        Ok(Some(RistrettoPrivate::from_bytes_mod_order(bytes)))
-    } else {
-        Err(ConversionError::ArrayCastError)
-    }
+    let bytes = <&[u8; 32] as TryFrom<&[u8]>>::try_from(bytes)?;
+    Ok(Some(RistrettoPrivate::from_bytes_mod_order(bytes)))
 }
 
 impl From<&Outlay> for api::Outlay {
--- a/mobilecoind/src/conversions.rs
+++ b/mobilecoind/src/conversions.rs
@@ -333,8 +333,13 @@ mod test {
             PublicAddress::try_from(proto.get_receiver()).unwrap(),
             public_addr
         );
+    }
 
-        // Proto -> Rust
+    #[test]
+    fn test_outlay_conversion_with_tx_private_key() {
+        let mut rng: StdRng = SeedableRng::from_seed([1u8; 32]);
+        let public_addr = AccountKey::random(&mut rng).default_subaddress();
+
         assert_eq!(rust, Outlay::try_from(&proto).unwrap());
 
         // Rust -> Proto, with tx private key
--- a/mobilecoind/src/conversions.rs
+++ b/mobilecoind/src/conversions.rs
@@ -333,14 +333,15 @@ mod test {
             PublicAddress::try_from(proto.get_receiver()).unwrap(),
             public_addr
         );
+
+        // Proto -> Rust
+        assert_eq!(rust, Outlay::try_from(&proto).unwrap());
     }
 
     #[test]
     fn test_outlay_conversion_with_tx_private_key() {
         let mut rng: StdRng = SeedableRng::from_seed([1u8; 32]);
         let public_addr = AccountKey::random(&mut rng).default_subaddress();
-
-        assert_eq!(rust, Outlay::try_from(&proto).unwrap());
 
         // Rust -> Proto, with tx private key
         let rust = Outlay {
--- a/mobilecoind/src/conversions.rs
+++ b/mobilecoind/src/conversions.rs
@@ -1,4 +1,4 @@
-// Copyright (c) 2018-2023 The MobileCoin Foundation
+// Copyright (c) 2018-2024 The MobileCoin Foundation
 
 //! Utilities for converting between `mobilecoind` and `mobilecoind_api` data
 //! types.
@@ -10,6 +10,7 @@ use crate::{
 use mc_account_keys::PublicAddress;
 use mc_api::ConversionError;
 use mc_common::HashMap;
+use mc_crypto_keys::RistrettoPrivate;
 use mc_mobilecoind_api as api;
 use mc_transaction_core::{
     ring_signature::KeyImage,
@@ -59,12 +60,27 @@ impl TryFrom<&api::UnspentTxOut> for UnspentTxOut {
     }
 }
 
+fn bytes_to_tx_private_key(bytes: &[u8]) -> Result<Option<RistrettoPrivate>, ConversionError> {
+    if bytes.is_empty() {
+        return Ok(None);
+    }
+
+    if let Ok(bytes) = <&[u8; 32] as TryFrom<&[u8]>>::try_from(bytes) {
+        Ok(Some(RistrettoPrivate::from_bytes_mod_order(bytes)))
+    } else {
+        Err(ConversionError::ArrayCastError)
+    }
+}
+
 impl From<&Outlay> for api::Outlay {
     fn from(src: &Outlay) -> Self {
         let mut dst = Self::new();
 
         dst.set_value(src.value);
         dst.set_receiver((&src.receiver).into());
+        if let Some(key) = src.tx_private_key {
+            dst.set_tx_private_key(key.to_bytes().to_vec());
+        }
 
         dst
     }
@@ -76,8 +92,13 @@ impl TryFrom<&api::Outlay> for Outlay {
     fn try_from(src: &api::Outlay) -> Result<Self, Self::Error> {
         let value = src.value;
         let receiver = PublicAddress::try_from(src.get_receiver())?;
+        let tx_private_key = bytes_to_tx_private_key(src.get_tx_private_key())?;
 
-        Ok(Self { value, receiver })
+        Ok(Self {
+            value,
+            receiver,
+            tx_private_key,
+        })
     }
 }
 
@@ -88,6 +109,9 @@ impl From<&OutlayV2> for api::OutlayV2 {
         dst.set_value(src.amount.value);
         dst.set_token_id(*src.amount.token_id);
         dst.set_receiver((&src.receiver).into());
+        if let Some(key) = src.tx_private_key {
+            dst.set_tx_private_key(key.to_bytes().to_vec());
+        }
 
         dst
     }
@@ -99,8 +123,13 @@ impl TryFrom<&api::OutlayV2> for OutlayV2 {
     fn try_from(src: &api::OutlayV2) -> Result<Self, Self::Error> {
         let amount = Amount::new(src.value, TokenId::from(src.token_id));
         let receiver = PublicAddress::try_from(src.get_receiver())?;
+        let tx_private_key = bytes_to_tx_private_key(src.get_tx_private_key())?;
 
-        Ok(Self { amount, receiver })
+        Ok(Self {
+            amount,
+            receiver,
+            tx_private_key,
+        })
     }
 }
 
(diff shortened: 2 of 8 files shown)


mobilecoind/api/proto/mobilecoind_api.proto
```diff
@@ -118,6 +118,9 @@ enum TxStatus {
 message Outlay {
     uint64 value = 1;
     external.PublicAddress receiver = 2;
+    // Optional tx private key to use for this tx out. This can be chosen explicitly as a mechanism for idempotence.
+    // Must be 32 pseudorandom bytes or none. Will be reduced to curve25519 scalar mod order.
+    bytes tx_private_key = 3;
```
cbeck88, line 123: this, and the similar change to `OutlayV2`, are the API change for mobilecoind

transaction/builder/src/transaction_builder.rs
```diff
@@ -447,6 +447,56 @@ impl<FPR: FogPubkeyResolver> TransactionBuilder<FPR> {
...
+    /// you to easily determine and record the tx_private_key, which you may
+    /// want for other reasons as described.
+    ///
+    /// # Arguments
+    /// * `amount` - The amount of this output
+    /// * `recipient` - The recipient's public address
+    /// * `tx_private_key` - Optionally, a specific tx_private_key to use for
+    ///   this output.
+    /// * `rng` - RNG used to generate blinding for commitment
+    pub fn add_output_with_tx_private_key<RNG: CryptoRng + RngCore>(
```
cbeck88, line 494: this is a new transaction builder API that allows adding an output with a specified tx private key

mobilecoind/src/conversions.rs
```diff
@@ -59,12 +60,27 @@ impl TryFrom<&api::UnspentTxOut> for UnspentTxOut {
...
+fn bytes_to_tx_private_key(bytes: &[u8]) -> Result<Option<RistrettoPrivate>, ConversionError> {
+    if bytes.is_empty() {
+        return Ok(None);
+    }
+
+    if let Ok(bytes) = <&[u8; 32] as TryFrom<&[u8]>>::try_from(bytes) {
+        Ok(Some(RistrettoPrivate::from_bytes_mod_order(bytes)))
+    } else {
+        Err(ConversionError::ArrayCastError)
+    }
```
nick-mobilecoin, line 72:
```suggestion
    let bytes = <&[u8; 32] as TryFrom<&[u8]>>::try_from(bytes)?;
    Ok(Some(RistrettoPrivate::from_bytes_mod_order(bytes)))
```

mobilecoind/src/conversions.rs
```diff
@@ -298,6 +327,24 @@ mod test {
...
+        };
+        let proto = api::Outlay::from(&rust);
+
+        assert_eq!(proto.value, rust.value);
+        assert_eq!(
+            PublicAddress::try_from(proto.get_receiver()).unwrap(),
+            public_addr
+        );
+
+        // Proto -> Rust
```
nick-mobilecoin, line 340: suggest using separate test case, easier to isolate failures, easier to parallelize execution, prevents the test turning into a large workflow/recipe style test
```suggestion
    }

    #[test]
    fn test_outlay_conversion_with_tx_private_key() {
        let mut rng: StdRng = SeedableRng::from_seed([1u8; 32]);
        let public_addr = AccountKey::random(&mut rng).default_subaddress();

```

transaction/builder/src/transaction_builder.rs
```diff
@@ -447,6 +447,56 @@ impl<FPR: FogPubkeyResolver> TransactionBuilder<FPR> {
         amount: Amount,
         recipient: &PublicAddress,
         rng: &mut RNG,
+    ) -> Result<TxOutContext, TxBuilderError> {
+        self.add_output_with_tx_private_key(amount, recipient, None, rng)
+    }
+
+    /// Add a non-change output to the transaction, optionally with a specified
+    /// tx_private_key.
```
nick-mobilecoin, line 455: This "optionally" part seems odd to me. I may be failing to recall but don't think many APIs have "do_x_with_y" and the y is optional. One usually calls that method/function to provide y.

Suggest making a common backend that this and add_output() call into. And this common backend allows an optional key. Suggest this private backend be private for now.

```mermaid
flowchart LR
A[add_output_with_tx_private_key] --> B[common_backend]
C[add_output] --> B
```

nick-mobilecoin (commented):
Did a look over the functionality and what not, going to see about tracking someone down for the higher level assurance that moving the private key up is an ok thing to do.

transaction/builder/src/transaction_builder.rs
cbeck88: I suggest, take a look at the mobilecoind code that calls this, because in that case, we have an optional RistrettoPrivate parameter. if I have a function call that can just take my optional parameter, that's very convenient. Otherwise I have to write an if else loop and there is more boilerplate and more room for error.

> but don't think many APIs have "do_x_with_y" and the y is optional

I think there are many examples of this in our API when it would be more convenient for the callers.

From my point of view, the only reason to retain `add_output` at all is backwards compatibility, and making it call `add_output_with_tx_private_key` seems fine. We can shorten the name of the new function if you prefer it.

cbeck88:
> Did a look over the functionality and what not, going to see about tracking someone down for the higher level assurance that moving the private key up is an ok thing to do.

Yeah, the `tx_private_key` is unfortunately named, it is not really a private key. It is more like a nonce.

cbeck88:
> Sorry forgot to mention, 5.2.3 is released so `release/v6.0` is probably what this PR should target

I developed this against 5.2.3 because I am likely going to deploy it a day or two from now. I'm happy to retarget this PR against release 6 if you prefer though.

nick-mobilecoin:
> > Sorry forgot to mention, 5.2.3 is released so `release/v6.0` is probably what this PR should target
> 
> I developed this against 5.2.3 because I am likely going to deploy it a day or two from now. I'm happy to retarget this PR against release 6 if you prefer though.

5 should more or less be done, we have one fix post 5.2.3 so that full service for 5.x can build on M1 macs. But everything else should be going into 6.x or main.

eranrund (commented):
I think adding support for this is fine - I'm going to let @nick-mobilecoin make the final decision on when this is ready to get merged, but I don't see any issues with it.

One thing I'd like you to consider is an alternative approach to idempotency: in places where I needed to ensure I don't double-send a transaction, my approach has been a multi-step process where I first use mobilecoind (or full-service) to generate a transaction, then I store the transaction in my app's database, then once it has been committed to the db, I try and submit it (and it either gets accepted by consensus, or rejected due to double spending or duplicate tx out pub keys).

It does require the generated tx response to include the tx out private keys for you to achieve the secondary goal of being able to prove what the outputs you created are and who they are destined to.

Not saying you should necessarily go with this approach, but I think since often you already are storing state in your application (and it is from this state that you will be deriving your deterministic tx private key for each tx you want to send), doing this two phase generation-and-then-submission might also make sense. A potential advantage (or disadvantage, depending on the use-case) of storing the entire generated tx and then trying to submit as many times as necessary (e.g. until you get a valid GRPC response from consensus that either indicates acceptance or refusal) is that you are also using the same inputs every time. With just the SendPayment API you are still not guaranteed that the same inputs are selected every time you call it, even if the outputs are now deterministic.

transaction/builder/src/transaction_builder.rs
cbeck88: Lmk what you prefer nick, happy to go either way

cbeck88:
https://github.com/mobilecoinfoundation/mobilecoin/pull/3879#discussion_r1463500573

> This and Outlay2 haven't diverged yet, and I'm not seeing a reason to have them diverge.
>
> It may fall into YAGNI, but I think keeping them symmetrical is cheap insurance for future changes

I am not sure, I thought about this briefly.

Right now, OutlayV2 is basically "outlay with specific token id", while Outlay is "outlay with token id inferred from context". So even if they are compatible as protobufs, they are not compatible semantically. If a user passed an outlay as outlay v2, it may cause them to select MOB when that is not what they intended, or if outlay v2 is passed as outlay, it may cause their specification of a token id to be ignored.

So making them diverge is perhaps safer. If field 3 of Outlay is now "bytes", then if you try to pass an OutlayV2 as an Outlay by accident, you will get a decoding error, rather than silently getting a different meaning. So that might be preferable. That's my 2 cents.

cbeck88:
> One thing I'd like you to consider is an alternative approach to idempotency: in places where I needed to ensure I don't double-send a transaction, my approach has been a multi-step process where I first use mobilecoind (or full-service) to generate a transaction, then I store the transaction in my app's database, then once it has been committed to the db, I try and submit it (and it either gets accepted by consensus, or rejected due to double spending or duplicate tx out pub keys).

Thanks, I think this is also a good way to achieve idempotence. I'm going to think about this some more for my case.

I wonder why Moby team did not approach idempotence that way before...

For my purposes, I still think it is useful to have the option of fixing the tx private key, if only because it is a simple way that i can also know the tx private key. I have a hunch it may be useful for SCIs also.

dolanbernard:
> I wonder why Moby team did not approach idempotence that way before...

I think the reason for this is that idempotence was needed for invite with payment. The invited user may not join immediately. If the whole transaction was stored, that would require the inputs to not be spent before that happens. So inviting multiple users could make all but the first to accept fail, for example.

I think the only real concern with this private key method is that it makes transaction builder easier to use incorrectly. We wouldn't want somebody thinking they should use sequential keys, for instance. But the name "key" does strongly imply that insecurely generated values should not be used.

The RNG strategy that we use for idempotence now is more fragile in theory, but unit testing for accidental breaking is very easy. Testing for planned changes (say between versions of transaction builder) is more difficult, though I'm not sure how likely that is.

transaction/builder/src/transaction_builder.rs
nick-mobilecoin: I'm fine keeping it as currently proposed, the argument of of not iffing based on function name, just args sold me on it

cbeck88:
> I have a hunch it may be useful for SCIs also.

I probably should make this work on the mcd sci builder also... gonna circle back on that

cbeck88:
this is now rebased on release v6

nick-mobilecoin (approved):

nick-mobilecoin:
@cbeck88  let me know if you can't see this output, https://github.com/mobilecoinfoundation/mobilecoin/actions/runs/7818115531/job/21327573452
I'm not sure what the deal is with CI and the fork here, but I pushed up another branch to force CI and that link is the job with the failures.

cbeck88:
i can see your link but I have green checks on the PR -- I'm going to try to investigate and fix

cbeck88:
i think this should be fixed now @nick-mobilecoin , lmk and feel free to merge

nick-mobilecoin:
@cbeck88 it showed up now for run, looks like some linter errors. 
The CD steps may fail due to permissions so we may merge without those.

cbeck88:
Will fix, thanks

cbeck88:
this time i ran ./tools/lint.sh in the container, and it passed, so i think we should be good

cbeck88:
```
        Running `/__w/mobilecoin/mobilecoin/target/enclave/release/build/mc-consensus-enclave-impl-763795e5a21c20b4/build-script-build`
  [mc-consensus-enclave-impl 6.0.0] thread 'main' panicked at /__w/mobilecoin/mobilecoin/consensus/enclave/impl/build.rs:79:31:
  [mc-consensus-enclave-impl 6.0.0] Failed parsing minting trust root public key PEM file: MalformedFraming
  [mc-consensus-enclave-impl 6.0.0] stack backtrace:
```

this looks like an error in the intfrastructure as opposed to a code problem

nick-mobilecoin:
> ```
>         Running `/__w/mobilecoin/mobilecoin/target/enclave/release/build/mc-consensus-enclave-impl-763795e5a21c20b4/build-script-build`
>   [mc-consensus-enclave-impl 6.0.0] thread 'main' panicked at /__w/mobilecoin/mobilecoin/consensus/enclave/impl/build.rs:79:31:
>   [mc-consensus-enclave-impl 6.0.0] Failed parsing minting trust root public key PEM file: MalformedFraming
>   [mc-consensus-enclave-impl 6.0.0] stack backtrace:
> ```
> 
> this looks like an error in the intfrastructure as opposed to a code problem

I agree, the CD steps require org permissions, but all the CI steps are green, https://github.com/mobilecoinfoundation/mobilecoin/actions/runs/7832347572/job/21372689035?pr=3879

So I'm merging

cbeck88:
thank you

PR #3910, opened 2024-02-14
Verify all of CheckKeyImagesResponse

nick-mobilecoin:
Previously the router_integration test only verified the `results` field
of the `CheckKeyImagesResponse`. Now the entire structure is verified.

### DIFF
--- a/fog/block_provider/src/local.rs
+++ b/fog/block_provider/src/local.rs
@@ -14,8 +14,8 @@ use std::time::Duration;
 
 #[derive(Clone)]
 pub struct LocalBlockProvider<L: Ledger + Clone + Sync> {
-    ledger: L,
-    watcher: Option<WatcherDB>,
+    pub ledger: L,
+    pub watcher: Option<WatcherDB>,
 }
 
 impl<L: Ledger + Clone + Sync> LocalBlockProvider<L> {
--- a/fog/ledger/server/tests/router_integration.rs
+++ b/fog/ledger/server/tests/router_integration.rs
@@ -4,7 +4,6 @@ use mc_account_keys::{AccountKey, PublicAddress};
 use mc_api::watcher::TimestampResultCode;
 use mc_blockchain_types::BlockVersion;
 use mc_common::{
-    logger,
     logger::{log, Logger},
     time::SystemTimeProvider,
 };
@@ -15,7 +14,6 @@ use mc_fog_ledger_server::{
     sharding_strategy::EpochShardingStrategy, KeyImageStoreServer, LedgerRouterConfig,
     LedgerRouterServer, LedgerStoreConfig, ShardingStrategy,
 };
-use mc_fog_ledger_test_infra::ShardProxyServer;
 use mc_fog_test_infra::get_enclave_path;
 use mc_fog_types::common::BlockRange;
 use mc_fog_uri::{FogLedgerUri, KeyImageStoreUri};
@@ -30,7 +28,7 @@ use serde::{Deserialize, Serialize};
 use std::{
     collections::HashMap,
     net::{IpAddr, Ipv4Addr, SocketAddr},
-    path::{Path, PathBuf},
+    path::PathBuf,
     str::FromStr,
     sync::Arc,
     time::Duration,
@@ -38,6 +36,7 @@ use std::{
 use tempfile::TempDir;
 use url::Url;
 
+const POLL_INTERVAL: Duration = Duration::from_millis(10);
 const TEST_URL: &str = "http://www.my_url1.com";
 const CHAIN_ID: &str = "local";
 
@@ -68,18 +67,18 @@ fn create_store_config(
         client_auth_token_max_lifetime: Default::default(),
         omap_capacity,
         sharding_strategy: ShardingStrategy::Epoch(EpochShardingStrategy::new(block_range)),
-        poll_interval: Duration::from_millis(250),
+        poll_interval: POLL_INTERVAL,
     }
 }
 
 fn add_block_to_ledger(
-    ledger_db: &mut LedgerDB,
+    block_provider: &mut LocalBlockProvider<LedgerDB>,
     recipients: &[PublicAddress],
     key_images: &[KeyImage],
     rng: &mut (impl CryptoRng + RngCore),
-    watcher: &WatcherDB,
 ) -> u64 {
     let amount = Amount::new(10, Mob::ID);
+    let ledger_db = &mut block_provider.ledger;
     let block_data = mc_ledger_db::test_utils::add_block_to_ledger(
         ledger_db,
         BlockVersion::MAX,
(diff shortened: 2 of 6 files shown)


fog/ledger/test_infra/src/lib.rs
```diff
@@ -316,70 +307,3 @@ impl Ledger for MockLedger {
         unimplemented!()
     }
 }
-
```
nick-mobilecoin, line 319: all of this was used to mimic a load balancer.
However we saw in real world deployment that the router did not work nicely with a load balancer.
So instead of leaving this extra complexity in the test, decided to remove it and focus the tests a bit more on the store behavior
```mermaid
flowchart  LR
A(Router) --> B(Load Balancer 1)
A --> C(Load Balancer 2)
A --> D(Load Balancer 3)
B --> F[Store A]
B --> G[Store B]
B --> H[Store C]
C --> I[Store D]
C --> J[Store E]
C --> K[Store F]
D --> L[Store G]
D --> M[Store H]
D --> N[Store I]
```

fog/ledger/server/tests/router_integration.rs
```diff
@@ -68,18 +67,18 @@ fn create_store_config(
         client_auth_token_max_lifetime: Default::default(),
         omap_capacity,
         sharding_strategy: ShardingStrategy::Epoch(EpochShardingStrategy::new(block_range)),
-        poll_interval: Duration::from_millis(250),
+        poll_interval: POLL_INTERVAL,
```
nick-mobilecoin, line 70: when the poll interval was made configurable it got changed from 10ms to 250ms everywhere, even in tests.
Moving this back to 10ms to make the tests run faster

fog/block_provider/src/local.rs
```diff
@@ -14,8 +14,8 @@ use std::time::Duration;
 
 #[derive(Clone)]
 pub struct LocalBlockProvider<L: Ledger + Clone + Sync> {
-    ledger: L,
-    watcher: Option<WatcherDB>,
+    pub ledger: L,
+    pub watcher: Option<WatcherDB>,
```
nick-mobilecoin, line 18: The LocalBlockProvider is pretty much only used in tests now.
I needed mutable access to `ledger` so instead of creating a mutable accessor method chose to make the field public.

fog/ledger/server/tests/router_integration.rs
```diff
@@ -166,22 +165,9 @@ fn create_store(
     store
 }
 
-fn create_shard(config: &ShardConfig, _logger: Logger) -> ShardProxyServer {
```
nick-mobilecoin, line 169: A bit of overloaded terminology the "shards" in this test were trying to mimic  "load balancers" with multiple store copies. Unfortunately they mimicry failed to represent real world usage as the routers won't balance, they latch on to which ever store copy they first talk to.

fog/ledger/server/tests/router_integration.rs
```diff
@@ -433,6 +364,13 @@ async fn smoke_test() {
             response.results[0].timestamp_result_code,
             TimestampResultCode::TimestampFound as u32
         );
+        // The genesis block will add 1 to the total number of blocks, it has
+        // one transaction which will also add 1 to the total number of
+        // transactions
+        assert_eq!(response.num_blocks, blocks_to_add + 1);
+        assert_eq!(response.global_txo_count, new_transactions + 1);
+        assert_eq!(response.latest_block_version, *BlockVersion::MAX);
+        assert_eq!(response.max_block_version, *BlockVersion::MAX);
```
nick-mobilecoin, line 373: This verification was missing and is what was exposed when the shards didn't scale up.
The clients rely on these values to know the status, https://github.com/mobilecoinofficial/MobileCoin-Swift/blob/70e92f44dd88d7b7cd4ac6175e188b2acb895217/Sources/Common/Fog/FogKeyImageChecker.swift#L79

The intent is that a follow on PR will update the tests to add one block at a time and ensure that this response has info for the newest block after the addition of each block.

fog/ledger/server/tests/router_integration.rs
```diff
@@ -254,74 +230,44 @@ fn create_router_client(
...
-            ));
-            tempdirs.push(watcher_db_dir);
-            tempdirs.push(ledger_db_dir);
-        }
+    let watcher_db_dir = TempDir::new().expect("Couldn't create temporary path for watcher DB");
+    let ledger_db_dir = TempDir::new().expect("Couldn't create temporary path for ledger DB");
+    let ledger = recreate_ledger_db(ledger_db_dir.path());
+    let watcher = setup_watcher_db(watcher_db_dir.path().to_path_buf(), logger.clone());
+    let mut block_provider = LocalBlockProvider::new(ledger, watcher);
+    seed_block_provider(&mut block_provider);
```
nick-mobilecoin, line 241: Consolidated the block provider and only seed it with the genesis block here so that a follow on PR can add one block at a time and verify the state of the system after each block

fog/ledger/server/tests/router_integration.rs
```diff
@@ -106,40 +106,44 @@ fn add_block_to_ledger(
...
         let recipients: Vec<_> = block.keys().cloned().collect();
         let key_images: Vec<_> = block.values().flat_map(|x| x.clone()).collect();
 
-        add_block_to_ledger(ledger, &recipients, &key_images, &mut rng, watcher);
+        add_block_to_ledger(block_provider, &recipients, &key_images, &mut rng);
     }
+
+    // The stores are running on separate threads. We wait twice as long as
+    // their POLL_INTERVAL to ensure they've had time to process the new blocks
+    std::thread::sleep(POLL_INTERVAL * 2);
```
nick-mobilecoin, line 133: I'm not a fan of waits in tests, but I think I was able to correctly associate the wait time with why.
However I'm up for alternatives if one can think of better ways, I think it can only reliably be done with an API on the stores themselves to kick them directly instead of from the run loop.

fog/ledger/test_infra/src/lib.rs
sugargoat: Roger - IIUC, this is essentially dead code in our real deployments (no load balancer), so no need to keep maintaining this test code.

fog/ledger/server/tests/router_integration.rs
sugargoat: For my understanding, the reason for the poll interval being at 250ms is resource responsibility? vs, in the test, it's fine to run faster.

fog/ledger/server/tests/router_integration.rs
sugargoat: Seems like the API is CBB - maybe worth flagging as `// CBB: if the sleeps here are causing issues at some point in the future, we could instead develop an API to trigger XYZ`

sugargoat (approved):
LGTM ✨

PR #3911, opened 2024-02-15
Fix fog ledger router returning incorrect block number

nick-mobilecoin:
Previously when a fog ledger store was spun up prior to the block chain
having blocks in its block range it would return back 0 for the block
number. This resulted in the collated response having 0 for the total
number of blocks. Now stores will always respond back with the number of
blocks their block provider has.


fog/ledger/server/tests/router_integration.rs
```diff
@@ -36,10 +36,32 @@ use std::{
 use tempfile::TempDir;
 use url::Url;
 
-const POLL_INTERVAL: Duration = Duration::from_millis(10);
+// The duration is a balancing act of being fast enough to not slow down the
```
nick-mobilecoin, line 39: I had to bump this up, I was getting intermittent local failures when at 10ms

fog/ledger/server/src/db_fetcher.rs
```diff
@@ -234,59 +237,60 @@ impl<
                     e
                 );
                 std::thread::sleep(Self::ERROR_RETRY_FREQUENCY);
-                *next_block_index
+                // We errored so we don't know how many blocks are in the ledger
+                return 0;
```
nick-mobilecoin, line 241: returning zero here could still cause us to be in the same situation where 0 is returned for all of the stores, but this is an error case that would only happen if we have some other issue so it seems tolerable to do here, we could also fall back to `shared_state.highest_processed_block_count`

fog/ledger/server/src/db_fetcher.rs
```diff
@@ -234,59 +237,60 @@ impl<
...
+            Ok(blocks) => blocks,
+        };
 
-                let mut span = block_span_builder(&tracer, "poll_block", *next_block_index)
-                    .with_start_time(start_time)
-                    .start(&tracer);
+        let latest_block = blocks.latest_block;
 
-                span.set_attribute(TELEMETRY_BLOCK_INDEX_KEY.i64(*next_block_index as i64));
+        if let Some(next_block) = blocks.results.get(0).and_then(|r| r.as_ref()) {
```
nick-mobilecoin, line 248: This is a little hard to see in the GH ui
BeyondCompare seems to handle it nicer, 
image

fog/ledger/server/tests/router_integration.rs
```diff
@@ -115,22 +134,23 @@ fn seed_block_provider(block_provider: &mut LocalBlockProvider<LedgerDB>) {
...
         add_block_to_ledger(block_provider, &recipients, &key_images, &mut rng);
     }
 
-    // The stores are running on separate threads. We wait twice as long as
-    // their POLL_INTERVAL to ensure they've had time to process the new blocks
-    std::thread::sleep(POLL_INTERVAL * 2);
+    // The stores are running on separate threads. We wait 3(stores) + 1
+    // their POLL_INTERVAL to ensure they've all had time to process the new 
+    // blocks
+    std::thread::sleep(POLL_INTERVAL * 4);
```
nick-mobilecoin, line 153: not fond of bumping more, but I was running into issues where not all of the stores were being updated in time.

PR #3923, opened 2024-02-22
Add merging to BlockRange

nick-mobilecoin:


fog/types/src/common.rs
```diff
@@ -61,6 +61,51 @@ impl BlockRange {
     pub fn is_empty(&self) -> bool {
         self.len() == 0
     }
+
+    /// Merge the first contiguous ranges into a range
+    ///
+    /// The ranges should be sorted to produce consistent results.
+    /// They are not sorted here in order to properly indicate the index of the
+    /// last range merged.
```
nick-mobilecoin, line 69: I'm up for suggestions on a better description here.

fog/types/src/common.rs
```diff
@@ -61,6 +61,51 @@ impl BlockRange {
...
+    /// let (index, merged_range) = BlockRange::merge_from_ranges(ranges.iter()).unwrap();
+    /// assert_eq!(index, 0);
+    /// assert_eq!(merged_range, BlockRange::new(3, 8));
+    /// ```
+    ///
+    /// Returns the index in `ranges` of the last block range merged, and the
+    /// merged range
+    pub fn merge_from_ranges<'a>(
+        ranges: impl IntoIterator<Item = &'a BlockRange>,
+    ) -> Option<(usize, BlockRange)> {
```
nick-mobilecoin, line 89: My thought is that the caller is going to want to know which was the last index used for the range and that will be the one whose values of max_tx_count and what not are used

nick-mobilecoin:
from internal convo
> I’m curious:
> 1. how you imagine using index
> 2. what utility there is in creating a merged_range with `merged_range.start_block != 0`

1. The index was added to support determining how to populate the `CheckedKeyImageResponse`, https://github.com/mobilecoinfoundation/mobilecoin/blob/617786eff8016635411a673f7b8460fcc8db33ca/fog/ledger/enclave/impl/src/lib.rs#L262. My hope is that with the index the caller will then be able to copy that response directly from the shard that provided the "final" range that got merged.  I'm up for ideas on this one. I think we need to ensure that the caller doesn't accidentally provide a larger global txo count than the number of blocks. Lets assume there are 3 txos per block and the ranges are (0,10), (10, 12), (20, 22). The intent is that our max block returned is 12, I'm thinking we also need to ensure the `gobal_txo_count` is 36 and not 66.

2. I don't think there is any use in providing a merged range that doesn't start at 0, for the system. However for this function I think it's already complicated enough with the index piece. My thought was that the caller would do something like (non compiling code)
```rust
let zero_range = BlockRange::new(0,1);
let received_ranges = ranges.sort(); // some logic somewhere that keeps the indices when sorted.
let (index, merged_range) = BlockRange.merge_from_ranges([&zero_range].chain(received_ranges));
```
Up for suggestions as I recognize this may be placing too much room for error by the callers

holtzman:
> from internal convo
> 
> > I’m curious:
> > 
> > 1. how you imagine using index
> > 2. what utility there is in creating a merged_range with `merged_range.start_block != 0`
> 
> 1. The index was added to support determining how to populate the `CheckedKeyImageResponse`, https://github.com/mobilecoinfoundation/mobilecoin/blob/617786eff8016635411a673f7b8460fcc8db33ca/fog/ledger/enclave/impl/src/lib.rs#L262
>    . My hope is that with the index the caller will then be able to copy that response directly from the shard that provided the "final" range that got merged.  I'm up for ideas on this one. I think we need to ensure that the caller doesn't accidentally provide a larger global txo count than the number of blocks. Lets assume there are 3 txos per block and the ranges are (0,10), (10, 12), (20, 22). The intent is that our max block returned is 12, I'm thinking we also need to ensure the `gobal_txo_count` is 36 and not 66.
> 2. I don't think there is any use in providing a merged range that doesn't start at 0, for the system. However for this function I think it's already complicated enough with the index piece. My thought was that the caller would do something like (non compiling code)
> 
> ```rust
> let zero_range = BlockRange::new(0,1);
> let received_ranges = ranges.sort(); // some logic somewhere that keeps the indices when sorted.
> let (index, merged_range) = BlockRange.merge_from_ranges([&zero_range].chain(received_ranges));
> ```
> 
> Up for suggestions as I recognize this may be placing too much room for error by the callers

Ah. Fessing up to initiating the internal conversation, and this makes sense to me.

fog/types/src/common.rs
```diff
@@ -61,6 +61,51 @@ impl BlockRange {
...
+        let Some(mut merged_range) = first else {
+            return None;
+        };
+        let mut index = 0;
+
+        for range in ranges {
+            if range.start_block <= merged_range.end_block
+                && merged_range.end_block <= range.end_block
+            {
+                merged_range.end_block = merged_range.end_block.max(range.end_block);
```
holtzman, line 101: small nit -- since this line is only reached when `merged_range.end_block` is less than or equal to `range.end_block` there is no need for a max operator.  I suppose the max could be left in case the enclosing logic changes in the future.
```suggestion
                merged_range.end_block = range.end_block;
```

fog/types/src/common.rs
holtzman:
```suggestion
    /// The "ranges" input must be pre-sorted by increasing start_block to produce correct results.
    /// Sorting is left to the caller so as to return a meaningful "index" which indicates an
    /// input range that contains an "end_block" at the top of the first contiguous range of blocks.
```

holtzman:
> Up for suggestions as I recognize this may be placing too much room for error by the callers

What if each `BlockRange` included the `query_response` from the corresponding shard? Then instead of doing a `merge_from_range`,  do a`best_query_response` where all of the shard `query_responses` are sorted by `start_block` and looked at in order until a discontinuity is found (or the end of the sorted BlockRange list is reached) and the `query_response` from before the discontinuity (or final one in list if there is no discontinuity) is returned?

Then `num_blocks`, `global_txo_count`, and `latest_block_version` could be taken from that `query_response`.

Some logic would also be needed in case none of the shards reported results that included block 0.

holtzman (approved):

nick-mobilecoin:
> > Up for suggestions as I recognize this may be placing too much room for error by the callers
> 
> What if each `BlockRange` included the `query_response` from the corresponding shard? Then instead of doing a `merge_from_range`, do a`best_query_response` where all of the shard `query_responses` are sorted by `start_block` and looked at in order until a discontinuity is found (or the end of the sorted BlockRange list is reached) and the `query_response` from before the discontinuity (or final one in list if there is no discontinuity) is returned?
> 
> Then `num_blocks`, `global_txo_count`, and `latest_block_version` could be taken from that `query_response`.
> 
> Some logic would also be needed in case none of the shards reported results that included block 0.

This may be a better alternative. The shards are going to need to return  a different type then the `CheckKeyImagesResponse` in order to accomodate the range. So it may make sense to create a dedicated `NewResponseType::merge_from(responses)` and contain the logic there.

I'll keep this PR open/unmerged for now and build the other logic on top of this to see how it plays out.

nick-mobilecoin:
Follow on PR is https://github.com/mobilecoinfoundation/mobilecoin/pull/3933

PR #3933, opened 2024-02-27
Fix fog ledger router returning incorrect block number

nick-mobilecoin:
Previously when a fog ledger store was spun up prior to the block chain
having blocks in its block range it would return back 0 for the block
number. This resulted in the collated response having 0 for the total
number of blocks. Now stores will respond back with the range of
blocks that they have processed and the collation logic will combine the
overlapping ranges to come up with the highest available block count.

### DIFF
--- a/fog/ledger/enclave/impl/src/lib.rs
+++ b/fog/ledger/enclave/impl/src/lib.rs
@@ -172,6 +172,8 @@ where
         })?;
 
         let mut resp = CheckKeyImagesResponse {
+            // `num_blocks` is a count, `end_block` is an exclusive index.
+            // A block range of [0, 5) would have a count of 5 blocks.
             num_blocks: untrusted_key_image_query_response
                 .processed_block_range
                 .end_block,
@@ -274,6 +276,9 @@ where
 
         let untrusted_response =
             merge_untrusted_responses(shard_query_responses.iter().map(|r| &r.untrusted_response));
+
+        // `num_blocks` is a count, `end_block` is an exclusive index.
+        // A block range of [0, 5) would have a count of 5 blocks.
         let num_blocks = untrusted_response.processed_block_range.end_block;
         let global_txo_count = untrusted_response.last_known_block_cumulative_txo_count;
         let latest_block_version = untrusted_response.latest_block_version;
@@ -504,12 +509,17 @@ mod tests {
     type UntrustedResponseTuple = ((u64, u64), u64, u32, u32);
     #[parameterized(
         all_consecutive = { vec![((0, 5), 15, 0, 0), ((5, 8), 24, 1, 2)], ((0, 8), 24, 1, 2) },
+
+        // `merge_untrusted_responses()` will force a range from (0, 0] so this
+        // response with (1, 8] will be a gap and won't be merged resulting in
+        // a (0, 0] response
         gap_at_start_block = { vec![((1, 8), 24, 1, 2)], ((0, 0), 0, 0, 0) },
         gap_in_middle = { vec![((0, 5), 15, 0, 0), ((8, 10), 24, 1, 2)], ((0, 5), 15, 0, 0) },
         contained_overlap_before = { vec![((5, 8), 24, 1, 2), ((0, 10), 30, 3, 3)], ((0, 10), 30, 3, 3) },
         contained_overlap_after = { vec![((0, 10), 30, 3, 3), ((5, 8), 24, 1, 2)], ((0, 10), 30, 3, 3) },
         contained_overlap_intermediate = { vec![((0, 10), 30, 3, 3), ((5, 8), 24, 1, 2), ((7, 12), 36, 4, 4)], ((0, 12), 36, 4, 4) },
         unsorted = { vec![((5, 8), 24, 1, 2), ((8, 10), 30, 2, 3), ((0, 5), 15, 0, 0)], ((0, 10), 30, 2, 3) },
+        none = { vec![], ((0, 0), 0, 0, 0) },
     )]
     fn merge_the_untrusted_responses(
         untrusted_responses: Vec<UntrustedResponseTuple>,
--- a/fog/ledger/enclave/impl/src/lib.rs
+++ b/fog/ledger/enclave/impl/src/lib.rs
@@ -32,14 +32,25 @@ use mc_fog_ledger_enclave_api::{
     Error, KeyImageData, KeyImageResult, LedgerEnclave, OutputContext, Result,
     UntrustedKeyImageQueryResponse,
 };
-use mc_fog_types::ledger::{
-    CheckKeyImagesRequest, CheckKeyImagesResponse, GetOutputsRequest, GetOutputsResponse,
+use mc_fog_types::{
+    common::BlockRange,
+    ledger::{
+        CheckKeyImagesRequest, CheckKeyImagesResponse, GetOutputsRequest, GetOutputsResponse,
+    },
 };
 use mc_oblivious_traits::ORAMStorageCreator;
 use mc_sgx_compat::sync::Mutex;
 use mc_sgx_report_cache_api::{ReportableEnclave, Result as ReportableEnclaveResult};
+use serde::{Deserialize, Serialize};
 
 mod oblivious_utils;
+
+#[derive(Debug, Serialize, Deserialize)]
+/// Response from a shard enclave to a router enclave for the key image query
+struct ShardKeyImageResponse {
+    untrusted_response: UntrustedKeyImageQueryResponse,
+    results: Vec<KeyImageResult>,
+}
 
 /// In-enclave state associated to the ledger enclaves
 pub struct SgxLedgerEnclave<OSC>
@@ -161,7 +172,9 @@ where
         })?;
 
         let mut resp = CheckKeyImagesResponse {
-            num_blocks: untrusted_key_image_query_response.highest_processed_block_count,
+            num_blocks: untrusted_key_image_query_response
+                .processed_block_range
+                .end_block,
             results: Default::default(),
             global_txo_count: untrusted_key_image_query_response
                 .last_known_block_cumulative_txo_count,
@@ -252,40 +265,29 @@ where
             .into_iter()
             .map(|(responder_id, query_response)| {
                 let plaintext_bytes = self.ake.backend_decrypt(&responder_id, &query_response)?;
-                let query_response: CheckKeyImagesResponse =
-                    mc_util_serial::decode(&plaintext_bytes)?;
+                let query_response: ShardKeyImageResponse =
+                    mc_util_serial::deserialize(&plaintext_bytes)?;
 
                 Ok(query_response)
             })
-            .collect::<Result<Vec<CheckKeyImagesResponse>>>()?;
+            .collect::<Result<Vec<_>>>()?;
 
-        let num_blocks = shard_query_responses
-            .iter()
-            .map(|query_response| query_response.num_blocks)
-            .min()
-            .expect("this is only None when the iterator is empty but we early-exit in that case");
-        let global_txo_count = shard_query_responses
-            .iter()
-            .map(|query_response| query_response.global_txo_count)
-            .min()
-            .expect("this is only None when the iterator is empty but we early-exit in that case");
-        let latest_block_version = shard_query_responses
-            .iter()
-            .map(|query_response| query_response.latest_block_version)
-            .max()
-            .expect("this is only None when the iterator is empty but we early-exit in that case");
+        let untrusted_response =
+            merge_untrusted_responses(shard_query_responses.iter().map(|r| &r.untrusted_response));
+        let num_blocks = untrusted_response.processed_block_range.end_block;
+        let global_txo_count = untrusted_response.last_known_block_cumulative_txo_count;
+        let latest_block_version = untrusted_response.latest_block_version;
+        let max_block_version = max(latest_block_version, *MAX_BLOCK_VERSION);
 
-        let plaintext_results: Vec<KeyImageResult> = shard_query_responses
+        let plaintext_results = shard_query_responses
             .into_iter()
             .flat_map(|query_response| query_response.results)
-            .collect();
+            .collect::<Vec<_>>();
 
         let oblivious_results = oblivious_utils::collate_shard_key_image_search_results(
             client_query_request.queries,
             &plaintext_results,
         );
-
-        let max_block_version = max(latest_block_version, *MAX_BLOCK_VERSION);
 
         let client_query_response = CheckKeyImagesResponse {
             num_blocks,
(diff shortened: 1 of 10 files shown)


fog/ledger/server/tests/router_integration.rs
```diff
@@ -518,23 +576,4 @@ async fn overlapping_stores() {
     assert_eq!(response.global_txo_count, new_transactions + 1);
     assert_eq!(response.latest_block_version, *BlockVersion::MAX);
     assert_eq!(response.max_block_version, *BlockVersion::MAX);
-
```
nick-mobilecoin, line 521: This is now tested in the pre verify before calling populate_block_provider()

fog/ledger/enclave/impl/src/lib.rs
```diff
@@ -252,41 +265,30 @@ where
...
                 let plaintext_bytes = self.ake.backend_decrypt(&responder_id, &query_response)?;
-                let query_response: CheckKeyImagesResponse =
-                    mc_util_serial::decode(&plaintext_bytes)?;
+                let query_response: ShardKeyImageResponse =
+                    mc_util_serial::deserialize(&plaintext_bytes)?;
 
                 Ok(query_response)
             })
-            .collect::<Result<Vec<CheckKeyImagesResponse>>>()?;
+            .collect::<Result<Vec<_>>>()?;
```
sugargoat, line 273: I'm wasn't familiar with this syntax - after a quick search, it looks like the compiler will infer the vector type, which iiuc will be `Vec<EnclaveMessage<NonceSession>>`

fog/ledger/enclave/impl/src/lib.rs
```diff
@@ -352,14 +350,59 @@ where
...
+    //   5-----------------20      Range 1
+    //       12--------15          Range 2
+    // The second range is fully contained in the first, so it is removed. This
+    // could happen in instances when the shards overlap but the newer shard is
+    // still catching up
+
+    // Note that
+    // [`dedup_by`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.dedup_by)
+    // states:
+    // > The elements are passed in opposite order from their order in the slice
```
sugargoat, line 380: Thank you for this explanation!

fog/ledger/server/tests/router_integration.rs
```diff
@@ -40,6 +40,25 @@ const POLL_INTERVAL: Duration = Duration::from_millis(10);
 const TEST_URL: &str = "http://www.my_url1.com";
 const CHAIN_ID: &str = "local";
 
+fn assert_key_image_unspent(key: &KeyImage, result: &KeyImageResult) {
+    assert_eq!(result.key_image, *key);
+    // None in the status means not spent.
+    assert_eq!(result.status(), Ok(None));
+    assert_eq!(
+        result.timestamp_result_code,
+        TimestampResultCode::TimestampFound as u32
```
sugargoat, line 49: Why is the TimestampFound for an unspent key_image?

fog/ledger/server/tests/router_integration.rs
```diff
@@ -115,22 +134,24 @@ fn seed_block_provider(block_provider: &mut LocalBlockProvider<LedgerDB>) {
     add_block_to_ledger(block_provider, &recipients, &[], &mut rng);
 }
 
-fn populate_block_provider(
+fn populate_block_provider<'a>(
     block_provider: &mut LocalBlockProvider<LedgerDB>,
-    blocks_config: &BlockConfig,
+    blocks_config: impl IntoIterator<Item = &'a HashMap<PublicAddress, Vec<KeyImage>>>,
```
sugargoat, line 139: Curious about this change - will come back after understanding more of the PR as a whole

fog/ledger/server/src/db_fetcher.rs
```diff
@@ -234,69 +225,70 @@ impl<
                     e
                 );
                 std::thread::sleep(Self::ERROR_RETRY_FREQUENCY);
-                *next_block_index
+                // We errored so we don't know how many blocks are in the ledger
+                return 0;
```
sugargoat, line 229: What's the expected client behavior in this case?

fog/ledger/server/src/db_fetcher.rs
```diff
@@ -234,69 +225,70 @@ impl<
...
+            let _active = mark_span_as_active(span);
+
+            // Get the timestamp for the block.
+            let timestamp =
+                if next_block.block_timestamp_result_code == TimestampResultCode::TimestampFound {
+                    next_block.block_timestamp
+                } else {
+                    tracer.in_span("poll_block_timestamp", |_cx| {
+                        self.block_provider
+                            .poll_block_timestamp(*next_block_index, watcher_timeout)
```
sugargoat, line 254: If the watcher is stuck (e.g. it's not watching the whole network so it can't get the timestamp), what happens here?

fog/ledger/server/src/db_fetcher.rs
```diff
@@ -234,69 +225,70 @@ impl<
...
+            tracer.in_span("add_records_to_enclave", |_cx| {
+                self.add_records_to_enclave(*next_block_index, records);
+            });
+
+            *next_block_index += 1;
+            let mut processed_block_range = self.sharding_strategy.get_block_range();
+            processed_block_range.end_block = *next_block_index;
+            self.update_db_poll_shared_state(&latest_block, processed_block_range);
         }
+        // Adding 1 as indices are 0 based, but "number of blocks" is 1 based.
```
sugargoat, line 280: Thank you for explaining!

fog/ledger/enclave/impl/src/lib.rs
```diff
@@ -161,7 +172,9 @@ where
         })?;
 
         let mut resp = CheckKeyImagesResponse {
-            num_blocks: untrusted_key_image_query_response.highest_processed_block_count,
+            num_blocks: untrusted_key_image_query_response
+                .processed_block_range
+                .end_block,
```
sugargoat, line 177: Just confirming, `end_block` is zero-indexed, so if `end_block` is 5 on a range of [0,5], there's actually 6 blocks in the response?

fog/ledger/enclave/impl/src/lib.rs
```diff
@@ -252,41 +265,30 @@ where
...
-            .min()
-            .expect("this is only None when the iterator is empty but we early-exit in that case");
-        let latest_block_version = shard_query_responses
-            .iter()
-            .map(|query_response| query_response.latest_block_version)
-            .max()
-            .expect("this is only None when the iterator is empty but we early-exit in that case");
+        let untrusted_response =
+            merge_untrusted_responses(shard_query_responses.iter().map(|r| &r.untrusted_response));
+        let num_blocks = untrusted_response.processed_block_range.end_block;
```
sugargoat, line 277: same question regarding `num_blocks` vs `end_block` for the range (and if the range doesn't start with 0, should it be `end_block - start_block`, or no because this is referring to the total number of blocks that have been processed?)

fog/ledger/enclave/impl/src/lib.rs
```diff
@@ -457,4 +500,42 @@ mod tests {
...
+    #[parameterized(
+        all_consecutive = { vec![((0, 5), 15, 0, 0), ((5, 8), 24, 1, 2)], ((0, 8), 24, 1, 2) },
+        gap_at_start_block = { vec![((1, 8), 24, 1, 2)], ((0, 0), 0, 0, 0) },
+        gap_in_middle = { vec![((0, 5), 15, 0, 0), ((8, 10), 24, 1, 2)], ((0, 5), 15, 0, 0) },
+        contained_overlap_before = { vec![((5, 8), 24, 1, 2), ((0, 10), 30, 3, 3)], ((0, 10), 30, 3, 3) },
+        contained_overlap_after = { vec![((0, 10), 30, 3, 3), ((5, 8), 24, 1, 2)], ((0, 10), 30, 3, 3) },
+        contained_overlap_intermediate = { vec![((0, 10), 30, 3, 3), ((5, 8), 24, 1, 2), ((7, 12), 36, 4, 4)], ((0, 12), 36, 4, 4) },
+        unsorted = { vec![((5, 8), 24, 1, 2), ((8, 10), 30, 2, 3), ((0, 5), 15, 0, 0)], ((0, 10), 30, 2, 3) },
+    )]
+    fn merge_the_untrusted_responses(
```
sugargoat, line 514: Great set of tests, ty

fog/ledger/server/tests/router_integration.rs
```diff
@@ -464,37 +495,64 @@ async fn overlapping_stores() {
...
+        for keys in block.values() {
+            for key in keys {
+                let response = test_environment
+                    .router_client
+                    .check_key_images(&[*key])
+                    .await
+                    .expect("check_key_images failed");
+                // We should always get a result to prevent side channel
+                // attacks from determining whether a key image was spent.
+                assert_eq!(response.results.len(), 1);
```
sugargoat, line 515: 🙏

sugargoat (approved):
This looks great! Left several comments mostly for my own edification

fog/ledger/enclave/impl/src/lib.rs
nick-mobilecoin: End block is zero indexed, however it is exclusive, https://github.com/mobilecoinfoundation/mobilecoin/blob/21d3af4d7122bb4bfeae7ff4cff3d78b9b1a323b/fog/types/src/common.rs#L12
(Also I have no idea why that's in fog/types...)
so [0, 5) would be block indices 0, 1, 2, 3, 4

fog/ledger/enclave/impl/src/lib.rs
nick-mobilecoin: it will be a `Vec<ShardKeyImageResponse>`, image from rust rover, 
image

I chose to use type inference here as I changed the type of `query_response` and it had to be changed in multiple places so decided to use inference to minimize future maintenance.
I will admit type inference is something newer to me coming from C, so it may be a hammer I swing a bit too much.

fog/ledger/enclave/impl/src/lib.rs
nick-mobilecoin: two strikes on me possibly not communicating in the code well enough.

I think it's only in two places thus it's not much burden to put a comment on them.
Generally always comment a behavior/assumption feels like something else needs to change
let me know if you think a comment in these 2 locations would suffice or if perhaps a simple helper method of `range_to_num_blocks()` would be better

fog/ledger/enclave/impl/src/lib.rs
sugargoat: Ah perfect ty!

fog/ledger/enclave/impl/src/lib.rs
sugargoat: Comment should be sufficient!

fog/ledger/enclave/impl/src/lib.rs
```diff
@@ -252,41 +265,30 @@ where
             .into_iter()
             .map(|(responder_id, query_response)| {
                 let plaintext_bytes = self.ake.backend_decrypt(&responder_id, &query_response)?;
```
sugargoat, line 267: High level question - which aspects need to be constant time?

fog/ledger/server/src/db_fetcher.rs
sugargoat: From live discussion - we'll want to double check with infrastructure that there's alerting in place for when the watcher is stuck

fog/ledger/enclave/impl/src/lib.rs
nick-mobilecoin: I believe this was discussed in the live walkthrough, at this time `oblivious_utils::collate_shard_key_image_search_results` is the constant time behavior. It's behavior should not have changed in this PR.

fog/ledger/server/src/db_fetcher.rs
nick-mobilecoin: created https://sentz.atlassian.net/browse/SENTZ-2354

fog/ledger/server/tests/router_integration.rs
nick-mobilecoin: I rediscovered why!
Previously `BlockConfig` was a vector, this forced one to always create a `vec![blah]` when calling this, by making this take a generic iterator, one can pass a vector, or a `[block1, block2]` directly
basically makes it a bit more ergonomic for the callers

fog/ledger/enclave/impl/src/lib.rs
nick-mobilecoin: added code comment

fog/ledger/enclave/impl/src/lib.rs
nick-mobilecoin: added code comment

fog/ledger/enclave/impl/src/lib.rs
```diff
@@ -457,4 +500,42 @@ mod tests {
             mc_fog_types::ledger::KeyImageResultCode::NotSpent as u32
         );
     }
+
+    type UntrustedResponseTuple = ((u64, u64), u64, u32, u32);
+    #[parameterized(
+        all_consecutive = { vec![((0, 5), 15, 0, 0), ((5, 8), 24, 1, 2)], ((0, 8), 24, 1, 2) },
+        gap_at_start_block = { vec![((1, 8), 24, 1, 2)], ((0, 0), 0, 0, 0) },
```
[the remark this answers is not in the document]
nick-mobilecoin: added comment to help clarify

fog/ledger/server/tests/router_integration.rs
```diff
@@ -40,6 +40,21 @@ const POLL_INTERVAL: Duration = Duration::from_millis(10);
 const TEST_URL: &str = "http://www.my_url1.com";
 const CHAIN_ID: &str = "local";
 
+fn assert_key_image_unspent(key: &KeyImage, result: &KeyImageResult) {
+    assert_eq!(result.key_image, *key);
+    // None in the status means not spent.
+    assert_eq!(result.status(), Ok(None));
```
nick-mobilecoin, line 46: This used to have 
```rust
    assert_eq!(
        result.timestamp_result_code,
        TimestampResultCode::TimestampFound as u32
    );
```
like below, it was copied from the original test logic, however after looking over more the timestamp code applies to the spent_at value in the proto file. The spent at should only be populated if the key image is found, this is what `status()` does behind the scenes. As such verifying the timestamp result doesn't apply

fog/ledger/server/tests/router_integration.rs
nick-mobilecoin: this verification was copied from the original test logic, however after looking over more the timestamp code applies to the spent_at value in the proto file. The spent at should only be populated if the key image is found, this is what `status()` does behind the scenes. As such verifying the timestamp result doesn't apply

fog/ledger/server/src/db_fetcher.rs
nick-mobilecoin: I refactored this now, and leveraged the `Option<>` type to have the caller contain all the behavior of what to do in these situations.

This previous implementation was spreading that handling between teh caller and here.

fog/ledger/server/src/db_fetcher.rs
```diff
@@ -170,42 +171,32 @@ impl<
...
         let mut next_block_index = block_range.start_block;
         loop {
-            if block_range.contains(next_block_index) {
-                loop {
-                    let num_blocks = self.load_block_data(&mut next_block_index);
+            loop {
+                let Some(num_blocks) = self.load_block_data(&mut next_block_index) else {
+                    std::thread::sleep(Self::ERROR_RETRY_FREQUENCY);
+                    continue;
+                };
```
nick-mobilecoin, line 178: note this logic changed in addressing other review comments
previously the sleep was in `load_block_data()` and this would return 0
now `load_block_data()` returns an `Option<>` and leaves the logic of what to do here instead of being split between the two and `load_block_data()` relying on this function handling `0` nicely

PR #3943, opened 2024-03-11
Use `mc-crypto-memo-mac` for computing hmac

nick-mobilecoin:
Previously there were two implementations of `compute_category1_hmac()`
one in `mc-transaction-extra` and one in `mc-crypto-memo-mac`. Now there
is only one implementation in `mc-crypto-memo-mac`


transaction/extra/src/memo/mod.rs
```diff
@@ -48,7 +48,7 @@
 //! | 0x0204          | Destination With Payment Intent Id Memo           |
 
 pub use self::{
-    authenticated_common::{compute_authenticated_sender_memo, compute_category1_hmac},
+    authenticated_common::compute_authenticated_sender_memo,
     authenticated_sender::AuthenticatedSenderMemo,
```
nick-mobilecoin, line 52: This was never fully exposed publicly, newer versions of rust will throw a warning for unused code here.
With there being an implementation in `mc-crypto-memo-mac` it seemed best to consolidate to one implementation.

transaction/extra/src/memo/authenticated_common.rs
```diff
@@ -64,13 +30,13 @@ pub fn validate_authenticated_sender(
     let shared_secret =
         receiving_subaddress_view_private_key.key_exchange(sender_address.spend_public_key());
 
-    let expected_hmac = compute_category1_hmac(
+    let expected_hmac = mc_crypto_memo_mac::compute_category1_hmac(
         shared_secret.as_ref(),
         tx_out_public_key,
         memo_type_bytes,
-        memo_data,
+        &memo_data[..48].try_into().expect("length mismatch"),
```
sugargoat, line 37: Can you say a little more about the slicing here?

sugargoat (approved):
LGTM - nice cleanup

transaction/extra/src/memo/authenticated_common.rs
sugargoat: Thank you! Yes, I was curious about slicing `memo_data[..48]`, and the confidence level in `memo_data` being constructed correctly for that slice. It looks like the 16 bytes at the end will b the hmac, and we're fairly confident (plus catching the panic and raising `length mismatch` if there's an issue)

PR #3947, opened 2024-03-26
impl TryFrom<String> trait for AccountId

holtzman:
### Motivation

The external transaction signer data types includes a 32 byte `AccountId` which is typically conveyed in JSON and other external code as a 64 digit hex string.  It is useful to be able to create an AccountId from such a representation.

### Changes

* this PR adds a `From<String>` implementation for `AccountId` allowing an object to be instantiated from a 64 character hex string.

### DIFF
--- a/transaction/signer/src/types.rs
+++ b/transaction/signer/src/types.rs
@@ -182,6 +182,16 @@ impl AsRef<[u8; 32]> for AccountId {
     }
 }
 
+/// Create [AccountId] object from hex-encoded string
+impl From<String> for AccountId {
+    fn from(value: String) -> Self {
+        let mut byte_array = [0u8; 32];
+        hex::decode_to_slice(value, &mut byte_array)
+            .expect("failed to decode account_id hex string");
+        Self(byte_array)
+    }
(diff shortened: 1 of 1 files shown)


transaction/signer/src/types.rs
```diff
@@ -182,6 +182,16 @@ impl AsRef<[u8; 32]> for AccountId {
     }
 }
 
+/// Create [AccountId] object from hex-encoded string
+impl From<String> for AccountId {
+    fn from(value: String) -> Self {
+        let mut byte_array = [0u8; 32];
+        hex::decode_to_slice(value, &mut byte_array)
+            .expect("failed to decode account_id hex string");
```
starkriedesel, line 190: Do we really want to `expect` and panic on failure here? Maybe implementing `TryFrom` and returning a `Result<AccountId, hex::FromHexError>` would be better?

holtzman: definitely better. updated PR to impl TryFrom trait.

sugargoat (approved):
LGTM!

PR #3951, opened 2024-04-02
6.0 consensus chart refactor

jgreat:
### Motivation

Refactor consensus-node chart to be more flux and helm diff friendly.  Remove some of the configuration abstractions.  This is a bit of a trade off in configuration complexity having the user provide secrets/configmaps instead of generating objects from a "configuration" chart.

- Run `actionlint` with `shellcheck` and clean up lint error and adopt better practices. This is where a lot of the quotes and missing `EOF` additions come from.
- Remove `consensus-node-config`, `mc-core-common-config`, `mc-core-dev-env-setup` charts. These functions have been moved into `mobilecoin-workflow-dev-setup-envrionment.yaml` GHA workflow.
- Update `consensus-node` chart to use external configmaps/secrets and remove any functionality that required a `lookup` call.  Some values will need to be provided at deployment time.
- Update `consensus-node` chart README with new required k8s objects and example values.
- Remove unneeded dev setup scripts.
- Remove v1 token config helper and examples.


.internal-ci/helm/consensus-node/templates/node-deployment.yaml
```diff
@@ -26,9 +26,6 @@ spec:
       labels:
         app: consensus-node
         {{- include "consensusNode.labels" . | nindent 8 }}
-        {{- if .Values.loadBalancedNode }}
```
joekottke, line 29: Nice.  Removal of that vestige that never worked.

joekottke (approved):

PR #3965, opened 2024-04-16
CI - Update CI jobs to remove deprecated actions

jgreat:
### Motivation

Move CI jobs to just run `cargo` commands in a simple run. 

- remove unneeded rust-setup
- remove unused mobilecoin-cache-seeds


.github/actions/run-mc-tests/action.yml
```diff
@@ -34,6 +34,7 @@ runs:
       # Run tests, with JUnit XML output (configured via .config/nextest.toml)
       # at `target/nextest/ci/junit.xml`
       run: |
+        . /opt/intel/sgxsdk/environment
```
nick-mobilecoin, line 37: Do we really need this?
I thought it would be taken care of by the following entries in the docker file
```
ENV SGX_SDK=/opt/intel/sgxsdk
ENV PATH=/opt/cargo/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/opt/intel/sgxsdk/bin:/opt/intel/sgxsdk/bin/x64
ENV PKG_CONFIG_PATH=/opt/intel/sgxsdk/pkgconfig
ENV LD_LIBRARY_PATH=/opt/intel/sgxsdk/sdk_libs
```

.github/workflows/ci.yml
```diff
@@ -348,7 +327,9 @@ jobs:
         env:
           PGHOST: postgres
           PGUSER: postgres
+        shell: bash
         run: |
+          . /opt/intel/sgxsdk/environment
```
nick-mobilecoin, line 332: similar to the test action is sourcing this file actually needed?

nick-mobilecoin (approved):
looks good might be able to trim out a couple of liens

.github/actions/run-mc-tests/action.yml
jgreat: These should, but I think PATH and maybe PKG_CONFIG_PATH and LD_LIBRARY_PATH might get overridden by the GHA runtime :(  

The setup_rust custom action sourced and pushed the values to $GITHUB_ENV, but it also did a lot of redundant things that we just shouldn't need to run and maintain.

PR #3971, opened 2024-04-22
Use the current job id in the uploaded core dump name

nick-mobilecoin:
The newer upload artifacts requires a unique name for artifacts. In
order to ensure that multiple failing test steps can upload core dumps
the artifact name has been prefixed with the
[job id](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context)

### Future Work

[Soundtrack of this PR]()

### DIFF
--- a/.github/actions/upload-core-dumps/action.yml
+++ b/.github/actions/upload-core-dumps/action.yml
@@ -7,6 +7,6 @@ runs:
     - name: Upload core dumps
       uses: mobilecoinofficial/gh-actions/upload-artifact@v0
       with:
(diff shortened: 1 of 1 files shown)


nick-mobilecoin:
meant to fix failures such as https://github.com/mobilecoinfoundation/mobilecoin/actions/runs/8787562561/job/24113045515
```
Error: Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run
```

holtzman (approved):

PR #3976, opened 2024-05-08
Allow ledger from archive to update a ledger database

nick-mobilecoin:
Previously mc-ledger-from-archive would always create a new ledger from
the origin block. Now mc-ledger-from-archive will update a pre-existing
ledger database.

### DIFF
--- a/ledger/from-archive/src/main.rs
+++ b/ledger/from-archive/src/main.rs
@@ -5,12 +5,12 @@
 
 mod config;
 
-use std::path::Path;
 use clap::Parser;
 use config::LedgerFromArchiveConfig;
-use mc_common::logger::{create_app_logger, log, Logger, o};
+use mc_common::logger::{create_app_logger, log, o, Logger};
 use mc_ledger_db::{create_ledger_in, Ledger, LedgerDB};
 use mc_ledger_sync::ReqwestTransactionsFetcher;
+use std::path::Path;
 
 fn main() {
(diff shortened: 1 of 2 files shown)


ledger/from-archive/src/main.rs
```diff
@@ -73,3 +60,33 @@ fn main() {
...
 }
+
+fn ledger_db(logger: &Logger, ledger_path: impl AsRef<Path>, transactions_fetcher: &ReqwestTransactionsFetcher) -> LedgerDB {
+    match ledger_path.as_ref().exists() {
+        true => {
+            log::info!(logger, "Opening existing ledger at {}", ledger_path.as_ref().display());
+            LedgerDB::open(ledger_path.as_ref()).expect("Could not open existing ledger")
+        }
+        false => create_ledger_db(logger, ledger_path, transactions_fetcher)
+    }
```
nick-mobilecoin, line 71: I couldn't think of a better way to handle this, there is a timing issue here in that if 2 processes try to do this logic or similar at the same time then one will overwrite the other.

I don't think it's any worse than what we had before where this would just always stomp over the ledger db

jgreat: I think this is fine.  We run ledger-from-archive as a startup process.  There should not be any overlap between this and consensus or ledger-distribution.

jgreat (approved):

joekottke (approved):

sugargoat (approved):
Beautiful

PR #3980, opened 2024-05-10
Change ledger distribution start block logic

nick-mobilecoin:
Previously ledger distribution required an argument and potentially a
state file in order to know what block to start uploading from. Now
ledger distribution will look at the current local ledger database and
walk back from the highest block finding which block has not been
uploaded yet.

### DIFF
--- a/ledger/distribution/src/main.rs
+++ b/ledger/distribution/src/main.rs
@@ -7,7 +7,7 @@
 pub mod uri;
 
 use crate::uri::{Destination, Uri};
-use clap::{Parser, ValueEnum};
+use clap::Parser;
 use mc_api::{block_num_to_s3block_path, blockchain, merged_block_num_to_s3block_path};
 use mc_blockchain_types::{BlockData, BlockIndex};
 use mc_common::logger::{create_app_logger, log, o, Logger};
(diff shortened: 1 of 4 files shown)


ledger/distribution/src/main.rs
```diff
@@ -203,6 +183,45 @@ impl BlockHandler for S3BlockWriter {
...
+                let result = runtime.block_on(self.s3_client.head_object(req));
+                match result {
+                    Ok(_) => OperationResult::Ok(true),
+                    Err(RusotoError::Service(HeadObjectError::NoSuchKey(_))) => {
+                        OperationResult::Ok(false)
+                    }
+                    Err(e) => {
+                        log::warn!(self.logger, "Failed to talk to S3 {e:?}, retrying...");
+                        OperationResult::Retry(e)
+                    }
```
nick-mobilecoin, line 218: Unfortunately rusoto won't make this call without AWS credentials
So I have not been able to test this out locally and I'm not sure the CI/CD will fully exercise.
This may need to be tested out in a dev environment somehow

ledger/distribution/src/main.rs
```diff
@@ -420,15 +409,83 @@ fn main() {
...
+        let temp_dir = TempDir::new().unwrap();
+        let distribution_path = temp_dir.path();
+        let mut block_handler = LocalBlockWriter::new(distribution_path.into(), logger);
+
+        for expected_block_count in [0, 1, 3, 5] {
+            // Note: the `0` case won't write any blocks.
+            // This will always re-write the blocks, but that's fine for this test.
+            for block_index in 0..expected_block_count {
+                block_handler.write_single_block(&ledger.get_block_data(block_index).unwrap());
+            }
```
nick-mobilecoin, line 480: I wanted to flush out the logic and ensure there wasn't an off by one for computing the start block.
While I wasn't able to test the `S3::block_exists()` implementation, I think this proves out the `first_block_to_handle()` logic.

jgreat (approved):

PR #3988, opened 2024-05-13
Put service metrics behind a feature

nick-mobilecoin:
Previously service metrics were always available from `mc-util-metrics.
This resulted in crates like mc-sgx-report-cache-untrusted depending on
grpcio. Now service metrics are behind the feature `service_metrics` and
crates need to opt in to use it.


holtzman (approved):
lgtm

PR #3996, opened 2024-06-05
serialize/deserialize u64 subaddress [index] as quoted decimal string

holtzman:
When serializing the `TxoUnsynced` struct, the `subaddress` member, which is a `u64`, should be represented as a `string` containing the decimal numeric representation of the `u64`, rather than as an unquoted decimal number directly, and when deserializing it, parse the `subaddress` as a string containing a decimal number.

### Motivation

Currently, the `subaddress` member is being serialized as a number. This is unsafe in the face of parsers that do not allocate enough bits to parsing numbers, for example many JavaScript JSON parsers, resulting in loss of precision, and thereby alteration of the value, as rounding occurs. 

This is an issue that we have encountered in general with serialized `u64`s, and specifically with the serializing of the `TxoUnsynced` struct. Our common practice has become to serialize `u64`s as strings containing the decimal numeric value of the `u64` so as to be human readable while surviving intermediate parsing and processing intact, prior to be deserialized again.

### DIFF
--- a/transaction/signer/src/types.rs
+++ b/transaction/signer/src/types.rs
@@ -53,6 +53,7 @@ pub struct TxoSyncReq {
 #[derive(Clone, PartialEq, Debug, Serialize, Deserialize)]
 pub struct TxoUnsynced {
     /// Subaddress for unsynced TxOut
+    #[serde(with = "string")]
     pub subaddress: u64,
 
     /// tx_out_public_key for unsynced TxOut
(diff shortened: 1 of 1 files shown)


nick-mobilecoin (approved):
looks good. 
Would have been nice to have a sanity test or two showing the behavior, but it's on par with the rest of the file

Is the only other implementation of this logic in full service?  I thought there would have been another implementation available in this repo

holtzman:
> looks good. Would have been nice to have a sanity test or two showing the behavior, but it's on par with the rest of the file
> 
> Is the only other implementation of this logic in full service? I thought there would have been another implementation available in this repo

Thanks!

It is a bit anti-pattern for the serialization to be implemented in mobilecoin.git rather than full-service for the full-service use-case. However, this repo was was already implementing serialization for the other struct member, `tx_out_public_key` so it made sense to implement it here for `subaddress` as well.

The struct is used by other repos that submodule mobilecoin, as well.

PR #4005, opened 2024-09-04
fog-view poll timing

jgreat:
### Motivation

Make fog-view database polling interval configurable. 

Before this change 1 view-store generated about 300 postgresql transactions/s in development.
After this change 1 view-store generated about 150 postgresql transactions/s in development.

Moving from default of 100ms to 250ms reduced number of transactions/s by just over half.  

- add `MC_DB_POLLING_INTERVAL_MS` env var for fog-view-store
- add pg_stat_statements lib to the dev postgres install.
- fix a couple of minor comment typos.

### Future work
Investigate view queries.  Look in to only doing update queries on stores with incomplete block ranges.

### DIFF
--- a/fog/view/server/src/server.rs
+++ b/fog/view/server/src/server.rs
@@ -251,9 +251,6 @@ where
     logger: Logger,
 }
 
-/// How long to wait between polling db
-const DB_POLL_INTERNAL: Duration = Duration::from_millis(100);
-
 impl<E, DB, SS> DbPollThread<E, DB, SS>
 where
     E: ViewEnclaveProxy,
@@ -350,7 +347,7 @@ where
         log::debug!(logger, "Db poll thread started");
 
         let mut worker = DbPollThreadWorker::new(
-            config,
+            config.clone(),
             stop_requested,
             enclave,
             db,
@@ -369,7 +366,8 @@ where
                 WorkerTickResult::HasMoreWork => {}
 
                 WorkerTickResult::Sleep => {
-                    sleep(DB_POLL_INTERNAL);
+                    // This waits between polling. Is this doubled up with sleeps in DbFetcher?
+                    sleep(config.db_polling_interval_ms);
                 }
             }
         }
@@ -434,7 +432,7 @@ pub enum WorkerTickResult {
     Sleep,
 }
 
-/// Telemetry: block indes currently being worked on.
+/// Telemetry: block index currently being worked on.
 const TELEMETRY_BLOCK_INDEX_KEY: Key = telemetry_static_key!("block-index");
 
 impl<E, DB, SS> DbPollThreadWorker<E, DB, SS>
@@ -457,6 +455,7 @@ where
 
         let db_fetcher = DbFetcher::new(
             db.clone(),
+            config.db_polling_interval_ms,
             db_fetcher_readiness_indicator.clone(),
             sharding_strategy.clone(),
             config.block_query_batch_size,
--- a/fog/view/server/src/server.rs
+++ b/fog/view/server/src/server.rs
@@ -345,9 +345,9 @@ where
         logger: Logger,
     ) {
         log::debug!(logger, "Db poll thread started");
-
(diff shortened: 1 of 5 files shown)


joekottke (approved):
While I'm not a Rust dev, this generally does make sense.  But I'd rather have someone else confirm it.

fog/view/server/src/server.rs
```diff
@@ -369,7 +366,8 @@ where
                 WorkerTickResult::HasMoreWork => {}
 
                 WorkerTickResult::Sleep => {
-                    sleep(DB_POLL_INTERNAL);
+                    // This waits between polling. Is this doubled up with sleeps in DbFetcher?
```
nick-mobilecoin, line 369: It's not doubled up, the db fetcher looks to be launched on a separate thread.

I think this thread could be running at a faster rate with no extra load on the db as it is pulling form the db fetcher and storing in the enclave
Polling faster here won't move faster than the 250ms from the db_fetcher, but it would reduce the overall latency. Right now worst case latency is 250ms(db fetcher) + 250ms (this) to get the data into the enclave.

The faster this moves the closer the worst case latency get's to 250ms

fog/view/server/test-utils/src/lib.rs
```diff
@@ -227,6 +227,7 @@ impl RouterTestEnvironment {
                     sharding_strategy,
                     postgres_config: Default::default(),
                     block_query_batch_size: 2,
+                    db_polling_interval_ms: Default::default(),
```
nick-mobilecoin, line 230: I wonder if this is causing the fog test failures or if those are just the flakey ones
```suggestion
                    db_polling_interval_ms: Duration::from_millis(250),
```

nick-mobilecoin (approved):
looks good, some nits on naming and ownership logic

fog/view/server/test-utils/src/lib.rs
jgreat: Is that `db_polling_interval_ms: Default::default()` not picking up 250ms from the clap default config values?  

I could set this back to 100ms for the tests.

fog/view/server/src/server.rs
jgreat: Should we set the worker tick sleep to something less than the db polling interval?  I just assumed that it should be the same because they shared the same const. 

I think the down side to reducing the worker sleep, is that we could end up in a situation where the view server reports ready prematurely. 

IIUC It looks like worker loop just sets the service to ready when the fetcher stops returning records.  Even if the interval timings are the same this could be problematic if the time it takes to fetch records from the db exceeds the interval time + processing time the worker takes to get its tasks done.  Could explain why view gets pretty unhappy in high network latency scenarios. 

This also seems like it could be a place where we could put logic in to control database polling based on current processed block height.  Can we shutdown the dbFetcher thread from here?

fog/view/server/test-utils/src/lib.rs
nick-mobilecoin: No the default here would be `Duration::default()` which is probably 0

fog/view/server/src/server.rs
nick-mobilecoin: I think for now we can keep the tick the same and see what happens.
From the comment here, [MAX_QUEUED_RECORDS](https://github.com/mobilecoinfoundation/mobilecoin/blob/8acb950f86af0ca2cfc8424ab18252ec3c4a0b66/fog/view/server/src/db_fetcher.rs#L30) I think the worker will be the slower one

> Can we shutdown the dbFetcher thread from here?
I didn't think so, but it looks like view is using `max` when merging the shard data, [LastKnownData](https://github.com/mobilecoinfoundation/mobilecoin/blob/8acb950f86af0ca2cfc8424ab18252ec3c4a0b66/fog/view/enclave/impl/src/types.rs#L167)
Ledger was using `min()` so it broke things. We maybe able to turn off fetcher once we're done.

nick-mobilecoin (approved):

PR #4009, opened 2024-09-26
fog-view - blue/green and zone based deployments.

jgreat:
### Motivation

Set up deployments to allow us to specify blue/green and specific availability zones. 

Instead of a monolithic install for fog-view components we break the install out into shared network components and the "stacks" that container the shard sets. 

- `fog-view-services` - which contains the `ingress` and supporting routing objects. The ingress routing can be updated by modifying the `color: ` selector label on the `fog-view-service-router` service object.
- `fog-view-fsg` - the fogShardGenerator object.  This object will now be deployed as a blue/green set and statically tied to an Availability Zone so we can ensure balancing.

A common deployment will use these helm deployments.

```
fog-view-service  # Ingress and network objects

fog-view-fsg-bl-z1  # fogShardGenerator blue on zone 1
fog-view-fsg-bl-z2  # fogShardGenerator blue on zone 2
fog-view-fsg-bl-z3  # fogShardGenerator blue on zone 3
```

On an upgrade we will add a green set.  Once the green set is loaded we can switch the selector labels on the fog-view-service-router service to start routing traffic.
```
fog-view-fsg-gr-z1  # fogShardGenerator green on zone 1
fog-view-fsg-gr-z2  # fogShardGenerator green on zone 2
fog-view-fsg-gr-z3  # fogShardGenerator green on zone 3
```

### Future Work

- rework fog-ledger to use this deployment pattern


joekottke (approved):
Not a blocking comment, but I remember that we ran into issues with the store names getting too long in the past, does this change still fit within the character limit?

jgreat:
I believe so. I tested it up to 1m block.  If we do run into a limitation we can use fullnameOverride to tweak the runtime names.

PR #4013, opened 2024-10-23
Fix ledger distribution resync

nick-mobilecoin:


ledger/distribution/src/main.rs
```diff
@@ -212,6 +212,13 @@ impl BlockHandler for S3BlockWriter {
                     Err(RusotoError::Service(HeadObjectError::NoSuchKey(_))) => {
```
jgreat, line 212: nit: I don't think this breaks anything but I don't believe that we will ever hit this since this error is valid for a HEAD request.  IIUC rusoto-s3 is parsing the body of the return of the error, but a HEAD request will never have a body.  This seem like a bug in rusoto_s3: it shouldn't expose a NoSuchKey error for head_object.

jgreat (approved):

ledger/distribution/src/main.rs
nick-mobilecoin: After testing with a temp s3 bucket and looking at https://docs.aws.amazon.com/cli/latest/reference/s3api/head-object.html I concur this has been removed in [`55b319c` (#4013)](https://github.com/mobilecoinfoundation/mobilecoin/pull/4013/commits/55b319c6ca0cd7b47224b110676030c47cbf5b78)

PR #4022, opened 2024-12-13
Upgrade rust to 1.77.0 nightly - 2024-02-01

jgreat:
### Highlights

- Update `nextest.toml` to limit number of concurrent tests to match available container cores in CI. 
- Update CI cargo builds to limit concurrent jobs to match available container cores.
- Update workspace name logic since cargo-metadata "stabilized" (changed) the format of package ids.

**`clippy::unconditional-recursion`** - This may be a bug in clippy. Set `allow` for these warnings and check later rust versions. 
- util/u64-ratio/src/lib.rs
- util/repr-bytes/src/lib.rs
- connection/src/sync.rs

**`clippy::question-mark`** use ? notation instead of return None
- fog/types/src/common.rs

**`clippy::useless-vec`**
- fog/view/enclave/impl/src/types.rs
- fog/test_infra/src/db_tests.rs
- mobilecoind-json/src/data_types.rs

**`clippy::thread-local-initializer-can-be-made-const`**
- fog/distribution/src/main.rs


.config/nextest.toml
```diff
@@ -7,6 +7,8 @@ failure-output = "immediate-final"
 status-level = "slow"
 # Add retires for flaky tests
 retries = { backoff = "exponential", count = 3, delay = "5s" }
+# Limit threads to a "large" gha container max
+test-threads = 22
```
nick-mobilecoin, line 11: ❓ are these (and ci.yml) thread call outs needed?
> Default: number of logical CPUs

Main concern is if we move to different machines we may try to spin up 22 threads when there's only 12 or similar.

nick-mobilecoin (approved):

.config/nextest.toml
jgreat: The problem I'm trying to solve here is that the containers are limited to 22 cores, but cargo/nexttest has no way of knowing that since the system still reports the 48 cores the host has.

jgreat: 2 large containers can be scheduled onto the same 48 core node.  If two builds/tests are running at the same time, both runs will be competing for resources thinking they have full access to all 48 cores.

What I'm hoping for is that builds and tests are actually faster and more reliable if we don't over load the nodes and force them into extra context switching.

jgreat: Ok doesn't look like it made the difference I hoped,  I will revert when I submit the 1.78.0-nightly upgrade.

PR #4024, opened 2025-01-30
upgrade rust 1.79.0-nightly

jgreat:
### Clean up Cargo.toml formatting and remove unnecessary feature.

- warning: `dev_dependencies` is deprecated in favor of `dev-dependencies` and will not work in the 2024 edition
- warning: `default_features` is deprecated in favor of `default-features` and will not work in the 2024 edition
- warning: `proc_macro` is deprecated in favor of `proc-macro` and will not work in the 2024 edition
- warning: the cargo feature `named-profiles` has been stabilized in the 1.57 release and is no longer necessary to be listed in the manifest

### warning: non-local `impl` definition, they should be avoided as they go against expectation 
`#[derive(Zeroize)]` and `#[derive(Display)]`

- upgrade `displaydoc` to 0.2.5
- upgrade `zeroize` to 1.8.1
- upgrade `zeroize_derive` (lock files) to `1.4.21

### warning: Compiler version doesn't include clang or GCC: "cc" "--version"

- update cc (cc-rs) version to 1.2.10

### error: usage of a legacy numeric method
help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#legacy_numeric_constants

- crypto/message-cipher/src/aes_impl.rs
- crypto/noise/src/cipher_state.rs
- common/src/logger/loggers/mod.rs
- crypto/digestible/derive/test/tests/structs.rs
- crypto/digestible/tests/basic.rs
- consensus/scp/src/msg.rs
- transaction/builder/src/signed_contingent_input_builder.rs
- transaction/builder/src/transaction_builder.rs
- t3/connection/src/lib.rs
- peers/src/connection.rs
- mobilecoind-json/src/bin/main.rs
- fog/sample-paykit/src/client.rs
- mobilecoind-dev-faucet/src/lib.rs
- fog/block_provider/src/mobilecoind.rs
- mobilecoind/src/service.rs
- fog/ingest/server/src/connection.rs

### error: importing a legacy numeric constant
help: remove this import and use the associated constant from the primitive type instead

- attest/core/src/ias/verify.rs

### error: usage of a legacy numeric constant
help: use the associated constant instead

- attest/core/src/ias/verify.rs

### error: clamp-like pattern without using clamp function

- attest/core/src/quote.rs

### error: function cannot return without recursing
- fog/ingest/enclave/api/src/error.rs

### error: transmute used without annotations
help: consider adding missing annotations: `transmute::<u32, mc_blockchain_types::BlockVersion>`

- ledger/db/src/ledger_db.rs

### DIFF
--- a/consensus/service/src/background_work_queue.rs
+++ b/consensus/service/src/background_work_queue.rs
@@ -13,6 +13,8 @@ enum QueueMsg<T> {
     StopRequested,
 }
 
+// ThreadSpawnFailed is never read.
+#[allow(dead_code)]
 #[derive(Debug)]
 pub enum BackgroundWorkQueueError {
     ThreadSpawnFailed(io::Error),
--- a/consensus/service/src/background_work_queue.rs
+++ b/consensus/service/src/background_work_queue.rs
@@ -13,14 +13,17 @@ enum QueueMsg<T> {
     StopRequested,
 }
 
-// ThreadSpawnFailed is never read.
-#[allow(dead_code)]
-#[derive(Debug)]
+#[derive(Debug, displaydoc::Display)]
 pub enum BackgroundWorkQueueError {
+    /// Failed to spawn a thread {0}
     ThreadSpawnFailed(io::Error),
+    /// Failed to send a message to the queue
     SendFailed,
+    /// Failed to receive a message from the queue
     RecvFailed,
+    /// Trying to start a queue that is already started
     AlreadyStarted,
+    /// Error joining the threads {0}
     JoinFailed(String),
 }
 
--- a/fog/sql_recovery_db/src/models.rs
+++ b/fog/sql_recovery_db/src/models.rs
@@ -1,5 +1,9 @@
 // Copyright (c) 2018-2022 The MobileCoin Foundation
 
+// Some of these fields are detected as "dead_code" because they are not used in
+// the codebase.
+#![allow(dead_code)]
+
 use super::schema::*;
 use crate::sql_types::{SqlCompressedRistrettoPublic, UserEventType};
 use mc_fog_types::common::BlockRange;
--- a/fog/sql_recovery_db/src/models.rs
+++ b/fog/sql_recovery_db/src/models.rs
@@ -9,7 +9,7 @@ use mc_fog_types::common::BlockRange;
 
 #[derive(Debug, Queryable)]
 pub struct IngressKey {
-    pub _ingress_public_key: SqlCompressedRistrettoPublic,
+    pub ingress_public_key: SqlCompressedRistrettoPublic,
     pub start_block: i64,
     pub pubkey_expiry: i64,
     pub retired: bool,
--- a/fog/sql_recovery_db/src/models.rs
+++ b/fog/sql_recovery_db/src/models.rs
@@ -1,7 +1,6 @@
 // Copyright (c) 2018-2022 The MobileCoin Foundation
 
-// Some of these fields are detected as "dead_code" because they are not used in
-// the codebase.
+// clippy/rust-analyzer is unable to detect struct fields used in tests?
 #![allow(dead_code)]
 
 use super::schema::*;
@@ -10,7 +9,7 @@ use mc_fog_types::common::BlockRange;
 
 #[derive(Debug, Queryable)]
 pub struct IngressKey {
-    pub ingress_public_key: SqlCompressedRistrettoPublic,
+    pub _ingress_public_key: SqlCompressedRistrettoPublic,
     pub start_block: i64,
     pub pubkey_expiry: i64,
     pub retired: bool,
--- a/fog/view/protocol/src/user_rng_set.rs
+++ b/fog/view/protocol/src/user_rng_set.rs
@@ -171,23 +171,3 @@ impl From<DecodeError> for TxOutRecoveryError {
         Self::ProstDeserializationFailed
     }
 }
-
-// Is this error type ever used?
-#[allow(dead_code)]
-#[derive(Debug)]
-pub enum RngSetError {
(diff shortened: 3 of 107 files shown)


consensus/service/src/background_work_queue.rs
```diff
@@ -13,6 +13,8 @@ enum QueueMsg<T> {
     StopRequested,
 }
 
+// ThreadSpawnFailed is never read.
```
jgreat, line 16: Clippy detected that ThreadSpawnFailed(io::Error) in this enum is never read.

fog/sql_recovery_db/src/models.rs
```diff
@@ -1,5 +1,9 @@
 // Copyright (c) 2018-2022 The MobileCoin Foundation
 
+// Some of these fields are detected as "dead_code" because they are not used in
```
jgreat, line 3: `dead_code`: This is sql schema, I would think all these fields should be used someplace in the code? At least for writing to the database.

Detected: 
```

pub struct IngressKey {
    pub ingress_public_key: SqlCompressedRistrettoPublic,
}


pub struct IngestInvocation {
    pub ingress_public_key: SqlCompressedRistrettoPublic,
    pub egress_public_key: Vec<u8>,
    pub last_active_at: chrono::NaiveDateTime,
    pub start_block: i64,
    pub decommissioned: bool,
    pub rng_version: i32,
}

pub struct IngestedBlock {
    pub id: i64,
    pub ingest_invocation_id: i64,
    pub ingress_public_key: SqlCompressedRistrettoPublic,
    pub block_number: i64,
    pub cumulative_txo_count: i64,
    pub block_signature_timestamp: i64,
    pub proto_ingested_block_data: Vec<u8>,
}

File: fog/view/protocol/src/user_rng_set.rs [hunk 3]
```diff
@@ -172,6 +172,8 @@ impl From<DecodeError> for TxOutRecoveryError {
     }
 }
 
+// Is this error type ever used?
+#[allow(dead_code)]
```
jgreat, line 176: `dead_code`: is `RngSetError` used?

mobilecoind-dev-faucet/src/worker.rs
```diff
@@ -56,6 +56,8 @@ pub struct UtxoRecord {
 }
 
 /// A tracker the worker keeps for UTXO records it hands to faucet threads.
+// utxo field never used?
+#[allow(dead_code)]
 pub struct UtxoTracker {
     pub utxo: UnspentTxOut,
```
jgreat, line 62: `dead_code`:  `utxo` not used?

fog/sql_recovery_db/src/models.rs
nick-mobilecoin: This one is interesting, these are all `Queryable` meaning that they're structs being retrieved from the database. None of them implement `Insertable` so they're never used for inserting into the database.
Since they're `Queryable` the structs have to have all the fields that the database rows have so as written these fields need to be present.

Suggest removing the blanket `#![allow(dead_code)]` and instead use leading underscore

The leading underscore is a case by case basis to say I know this field isn't used (yet) and it's intenional.
```rust
#[derive(Debug, Queryable)]
pub struct IngressKey {
    pub _ingress_public_key: SqlCompressedRistrettoPublic,
    pub start_block: i64,
    pub pubkey_expiry: i64,
    pub retired: bool,
    pub lost: bool,
}

#[derive(Debug, Queryable)]
pub struct IngestInvocation {
    pub id: i64,
    pub _ingress_public_key: SqlCompressedRistrettoPublic,
    pub _egress_public_key: Vec<u8>,
    pub _last_active_at: chrono::NaiveDateTime,
    pub _start_block: i64,
    pub _decommissioned: bool,
    pub _rng_version: i32,
}

#[derive(Debug, Queryable)]
pub struct IngestedBlock {
    pub _id: i64,
    pub _ingest_invocation_id: i64,
    pub _ingress_public_key: SqlCompressedRistrettoPublic,
    pub _block_number: i64,
    pub _cumulative_txo_count: i64,
    pub _block_signature_timestamp: i64,
    pub _proto_ingested_block_data: Vec<u8>,
}
```

fog/view/protocol/src/user_rng_set.rs
nick-mobilecoin: it doesn't look like this one is used. It should be removed

mobilecoind-dev-faucet/src/worker.rs
nick-mobilecoin: yah it's never used.
```rust
pub struct UtxoTracker {
    receiver: oneshot::Receiver<SubmitTxResponse>,
    received: Option<SubmitTxResponse>,
}

impl UtxoTracker {
    /// Make a new tracker and associated record for a given utxo
    pub fn new(utxo: UnspentTxOut) -> (Self, UtxoRecord) {
        let (sender, receiver) = oneshot::channel();

        let record = UtxoRecord {
            utxo,
            sender,
        };

        let tracker = Self {
            receiver,
            received: None,
        };

        (tracker, record)
    }
```

sgx/build/Cargo.toml
```diff
@@ -9,7 +9,7 @@ readme = "README.md"
 rust-version = "1.76.0"
 
 [dependencies]
-cc = "1.0"
+cc = "1.2.10"
```
nick-mobilecoin, line 12: similar to displaydoc and zeroize consider the following if possible
```suggestion
cc = "1"
```

nick-mobilecoin (approved):
looks good, if possible it would be good to only use the major version or the "0.x" version for 0 version crates, to minimize file changes when dependencies are updated in the future

PR #4025, opened 2025-02-01
Update to rust 1.81.0 nightly

jgreat:
- Lots of inline doc cleanup
- Add `#![allow(renamed_and_removed_lints)]` to pass generated code. TODO: update and refactor for prost or protobuf:3

### warning: unexpected `cfg` condition value `feature =`
Fix: add missing features to Cargo.toml

- ledger/sync/Cargo.toml
- mobilecoind/Cargo.toml
- sgx/panic/Cargo.toml
- sgx/slog/Cargo.toml
- sgx/sync/Cargo.toml
- sgx/urts/Cargo.toml
- transaction/signer/Cargo.toml

### error: the borrowed expression implements the required traits

- mobilecoind/src/utxo_store.rs

### error: using `map_err` over `inspect_err`

- fog/view/load-test/src/main.rs
- fog/view/server/src/router_request_handler.rs
- util/metered-channel/src/lib.rs

### error: using `map` over `inspect`

- util/metered-channel/src/lib.rs


fog/types/src/view.rs
```diff
@@ -539,7 +539,9 @@ impl FogTxOut {
     ///
     /// Returns:
     /// * TxOut,
+    ///
     /// Or
+    ///
     /// * An error if recovery failed
```
nick-mobilecoin, line 545: ⛏️ This was poorly formatted previously with an `Or` between 2 bulleted list items
I think this portion of the comment is useless, it doesn't provide any more information than the return type `Result<TxOut, FogTxOutError>`
```suggestion
```
Other options are a normal list
```rust
    /// Returns:
    /// * TxOut
    /// * An error if recovery failed
```
spelling it out
```rust
    /// Returns:
    /// TxOut on success or an error if recovery failed
```

nick-mobilecoin: Might consider it for the other functions that have
But I'm also fine if these stay as they are, they're no worse then they were previously
```rust
/// * blah
///
/// Or
///
/// * baz
```

nick-mobilecoin (approved):

fog/types/src/view.rs
jgreat: I didn't really like adding in the extra lines, but it was finding something that the doc formater liked that kept the original intent.

PR #4028, opened 2025-03-20
Update Build/Run OS images to 24.04

jgreat:
### Primary Changes

- Update workflow container to 24.04 base
- Update runtime containers to 24.04 base
- Update python dependencies for new python3 version (3.12)
- Update python integration test scripts to use virtualenv
- Update `mob prompt` to use new 24.04 base image.
- Update Intel SGX SDK to 2.25
- Bump SECURITY_VERSION for SGX SDK update
- Bump cargo package versions for SGX SDK update 

### Additional changes.

- Enable mobilecoind GRPC integration tests. 
- Add .devcontainer.json for IDE devcontainer integration.

### Motivation

Update aging software and dependencies.

### Future Work

- Update cargo/rust dependencies.


.internal-ci/test/mobilecoind-integration-test.sh
```diff
@@ -25,6 +25,8 @@ pushd "${strategies_dir}" >/dev/null || exit 1
 
 echo "-- Install requirements"
 echo ""
+python3 -m venv venv
+source venv/bin/activate
 pip3 install -r requirements.txt
```
nick-mobilecoin, line 30: Absolutely not advocating this be done now, just planting the seed

We probably want to consider moving over to https://github.com/astral-sh/uv at least for any new python stuff
The main benefits are:
- speed (for local setup) 
- pinning transitive dependencies with a lock file
- One can call out the python version and it will install it
- One can run the tests via one command `uv run pytest` versus this multiline that might leave one in the venv

nick-mobilecoin (approved):

.internal-ci/test/mobilecoind-integration-test.sh
jgreat: I think if we're going to use something I'd probably go with poetry since we already use it in other projects.

PR #4044, opened 2025-07-10
release/7.0 - revert chart rename, use /charts oci/package repo namespace

jgreat:
Since we're starting to publish charts as OCI packages to Github we need a differentiator between docker images and helm chart images names. 

We can technically publish a chart and docker image as the same name/tag, but its confusing in the gh packages ui.

Changing the name of the package ends up making the new chart incompatible with helm upgrade process (immutable selector fields with the chart name as the value),

This lead to extended downtime/messy workarounds in the TestNet upgrade process since we needed to delete/reapply charts instead of a clean upgrade process.

We're going to revert the chart name changes and instead namespace them under `/charts` 

New chart repo URL pattern: `oci://ghcr.io/mobilecoinfoundation/charts/<name>`


joekottke (approved):
Did you make any changes in the `_helpers.tpl` that need to be undone?

PR #4045, opened 2025-07-10
release/7.0 - cache CI results

jgreat:
- CI - Add caching to ci tasks to reduce runtimes when Rust/Go code doesn't change.
- CI - Ignore changes to markdown files

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,12 +2,22 @@ name: CI
 
 on:
   pull_request:
-  push:
-    branches:
-      - main
-      - 'release/**'
     paths-ignore:
     - '**.md'
+    - '.github/workflows/mobilecoin-*'
+    - '.github/workflows/dependent-repos.yml'
+    - '.github/workflows/refresh-ledger-bootstrap.yaml'
+    - '.github/workflows/android-bindings-dispatch.yml'
+  push:
+    branches:
+    - main
+    - 'release/**'
+    paths-ignore:
+    - '**.md'
+    - '.github/workflows/mobilecoin-*'
+    - '.github/workflows/dependent-repos.yml'
+    - '.github/workflows/refresh-ledger-bootstrap.yaml'
+    - '.github/workflows/android-bindings-dispatch.yml'
   workflow_dispatch:
 
 env:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -24,112 +26,186 @@ jobs:
...
-        uses: mobilecoinofficial/gh-actions/checkout@v0
-      - name: Cargo build (SW)
-        shell: bash
-        run: cargo build --locked
-      - name: Check dirty git
-        uses: ./.github/actions/check-dirty-git
+    - name: Check out code
+      uses: mobilecoinofficial/gh-actions/checkout@v0
+
+    - name: Cache Rust Binaries
```
nick-mobilecoin, line 45: ❓ This isn't actually caching the binaries is it?
My understanding is this is only caching the `success` sentinal file and using that to determin if this job needs to run

.github/workflows/ci.yml
```diff
@@ -24,112 +26,186 @@ jobs:
...
     container: mobilecoin/fat-sgx-builder:v24.4.2
-
     steps:
-      - name: Check out code
-        uses: mobilecoinofficial/gh-actions/checkout@v0
-      - name: Cargo build (SW)
-        shell: bash
-        run: cargo build --locked
-      - name: Check dirty git
-        uses: ./.github/actions/check-dirty-git
```
nick-mobilecoin, line 35: was it intentional that these went away?
I'm a bit torn on their value to begin with
I like the idea of ensuring we haven't modified any code, but when we have generated code files that are used in part of the build it doesn't seem as beneficial

nick-mobilecoin (approved):
thanks for the speed up, might consider renaming the cache step as I don't think it's actually caching the binaries

.github/workflows/ci.yml
jgreat: There's no real point in caching/storing the binaries.  We are just marking the jobs as successful for this revision of the code.  If any of the rust files change, that will invalidate the cache and new tests will run.

.github/workflows/ci.yml
jgreat: I did intentionally pull the the check dirty git jobs.  I can put them back if we want, or just put them on the general build steps (build-dev, build-prod).  I don't see them being needed on each step that does a build.

PR #4064, opened 2025-09-18
Generate tonic protos for api crates

nick-mobilecoin:


Cargo.toml
```diff
@@ -342,6 +342,8 @@ textwrap = "0.16"
 tiny-bip39 = "1"
 tokio = "1"
 toml = "0.8"
+tonic = { version = "0.9", default-features = false }
+tonic-build = "0.9"
```
nick-mobilecoin, line 346: Version 0.9 is the one that works with prost 0.11

api/src/lib.rs
```diff
@@ -6,6 +6,15 @@ mod autogenerated_code {
     // Include the auto-generated code.
     include!(concat!(env!("OUT_DIR"), "/protos-auto-gen/mod.rs"));
 }
+
+pub mod tonic {
```
nick-mobilecoin, line 10: I figure push these into a `tonic` namespace to avoid conflicting with the `pub use crate::{autogenerated_code::*` below

t3/api/src/lib.rs
```diff
@@ -15,6 +15,19 @@ mod autogenerated_code {
     }
 }
 
+pub mod tonic {
+    pub mod external {
+        pub mod v1 {
+            tonic::include_proto!("external.v1");
```
nick-mobilecoin, line 21: not sure on this nesting here, may need to wait until the actual tonic implementations of gRPC is done

eranrund (approved):

PR #4070, opened 2025-09-26
Update tracing to use otlp

nick-mobilecoin:


util/telemetry/src/lib.rs
```diff
@@ -48,7 +48,15 @@ pub fn versioned_tracer(
     version: Option<&'static str>,
     schema_url: Option<&'static str>,
 ) -> BoxedTracer {
-    tracer_provider().versioned_tracer(name, version, schema_url, None)
```
nick-mobilecoin, line 51: versioned_tracer is still present, but it's deprecated

util/telemetry/src/otlp.rs
```diff
@@ -35,21 +35,27 @@ pub fn setup_default_tracer_with_tags(
...
-    opentelemetry_jaeger::new_agent_pipeline()
-        .with_service_name(service_name)
+    let provider = opentelemetry_otlp::new_pipeline()
+        .tracing()
+        .with_exporter(opentelemetry_otlp::new_exporter().tonic())
         .with_trace_config(trace::Config::default().with_resource(Resource::new(tags)))
-        .install_simple()
-        .map_err(Error::Trace)
-        .map(Some)
+        .install_batch(opentelemetry_sdk::runtime::Tokio)
```
nick-mobilecoin, line 56: This used to use `install_simple()`.
The `install_simple()` will do a full publish of a span/trace when the span closes
The batch processor will queue up finished spans locally and send them out.

util/telemetry/src/otlp.rs
```diff
@@ -35,21 +35,27 @@ pub fn setup_default_tracer_with_tags(
...
+        .map_err(Error::Trace)?;
+    global::set_tracer_provider(provider);
```
nick-mobilecoin, line 58: It's suggested that libraries don't call `global::set_trace_provder()` and that applications should be the only ones, but these functions seem to be app specific functions

util/telemetry/README.md
```diff
@@ -1,21 +1,22 @@
...
+We are using opentelemtry as the backend for collecting traces and displaying them.
 
-Note that you need to include **both** environment variables if you want to change the defaults.
-
-Once Jaeger is running, you should be able to start seeing traces on its frontend.
+There are two environment variables that affect the exporting of traces:
+- `MC_TELEMETRY` - This needs to be set to `1` or `true` in order for traces to
+  be exported
+- `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` - this is where your OpenTelemetry
+  collector is running. The default value is `127.0.0.1:4317`.
```
nick-mobilecoin, line 15: @jgreat you'll want to look here for setting it up
I'm not sure where fog test client's config files are, I saw that for the fog services they have [jaegerTracing](https://github.com/mobilecoinfoundation/mobilecoin/blob/067f635f51a039eef02d9c48658fa0aa3af5c8a8/.internal-ci/helm/fog-ledger-fsg/values.yaml#L196C1-L196C14) set to `false`

jgreat (approved):

PR #4076, opened 2025-10-02
Make TransactionBuilder derive Clone and DRY some unit tests (SENTZ-4373)

eranrund:
### Motivation

By making `TransactionBuilder` derive `Clone` we can shorten quite a few unit tests by cutting down on boilerplate involved in constructing one for each test scenario inside unit tests that contain multiple scenarios.

### Future Work

It's possible there are more unit tests that can benefit from this sort of change, possibly in `full-service` as well.

### DIFF
--- a/transaction/builder/src/transaction_builder.rs
+++ b/transaction/builder/src/transaction_builder.rs
@@ -73,7 +73,7 @@ pub struct TxOutContext {
 ///
 /// This is generic over MemoBuilder to allow injecting a policy for how to
 /// use the memos in the TxOuts.
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 pub struct TransactionBuilder<FPR: FogPubkeyResolver> {
     /// The block version that we are targeting for this transaction
     block_version: BlockVersion,
@@ -925,7 +925,7 @@ pub(crate) fn create_fog_hint<RNG: RngCore + CryptoRng, FPR: FogPubkeyResolver>(
 pub mod transaction_builder_tests {
     use super::*;
     use crate::{
-        test_utils::{create_output, get_input_credentials, get_ring, get_transaction},
+        test_utils::{create_output, get_input_credentials, get_transaction},
         BurnRedemptionMemoBuilder, DefragmentationMemoBuilder, EmptyMemoBuilder,
         GiftCodeCancellationMemoBuilder, GiftCodeFundingMemoBuilder, GiftCodeSenderMemoBuilder,
         RTHMemoBuilder,
@@ -937,7 +937,7 @@ pub mod transaction_builder_tests {
         burn_address, burn_address_view_private, AccountKey, ShortAddressHash,
         CHANGE_SUBADDRESS_INDEX, DEFAULT_SUBADDRESS_INDEX, GIFT_CODE_SUBADDRESS_INDEX,
     };
-    use mc_crypto_ring_signature_signer::{InputSecret, NoKeysRingSigner, OneTimeKeyDeriveData};
+    use mc_crypto_ring_signature_signer::{InputSecret, NoKeysRingSigner};
     use mc_fog_report_validation_test_utils::{FullyValidatedFogPubkey, MockFogResolver};
     use mc_transaction_core::{
         constants::{MAX_INPUTS, MAX_OUTPUTS, MILLIMOB_TO_PICOMOB},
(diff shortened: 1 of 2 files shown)


transaction/builder/src/transaction_builder.rs
```diff
@@ -2824,46 +2732,11 @@ pub mod transaction_builder_tests {
...
-                .collect();
-
-            let input_credentials = InputCredentials::new(
-                ring,
-                membership_proofs,
-                real_index,
-                OneTimeKeyDeriveData::OneTimeKey(onetime_private_key),
-                *alice.view_private_key(),
-            )
-            .unwrap();
```
eranrund, line 2858: The rest of the test didn't care about how the input was constructed and `get_input_credentials` used by `single_input_transaction_builder` does a similar thing and produces a valid input.

transaction/builder/src/transaction_builder.rs
```diff
@@ -3117,25 +2982,15 @@ pub mod transaction_builder_tests {
...
-        let input_credentials = get_input_credentials(
-            block_version,
-            Amount::new(100, token_id),
-            &sender,
-            &fog_resolver,
-            &mut rng,
-        );
+
+        let transaction_builder =
+            single_input_transaction_builder(block_version, token_id, 500, &fog_resolver, &mut rng);
```
eranrund, line 2990: I changed this to a higher amount since there are tests below that try to spend more than `100`, and I figured we don't want anything to artificially fail due to insufficient funds vs due to the scenario the test is actually testing for.

transaction/builder/src/transaction_builder.rs
eranrund, line 3024: This tries to spend `200` (+ fees).

eranrund, line 3028: This test tries to spend `200 + fees`, which is why I bumped the amount above.

transaction/builder/src/transaction_builder.rs
```diff
@@ -3216,20 +3058,14 @@ pub mod transaction_builder_tests {
...
-                Amount::new(10, Mob::ID),
-                fog_resolver.clone(),
-            )
-            .unwrap();
-            transaction_builder.add_input(input_credentials.clone());
+            let mut transaction_builder = transaction_builder.clone();
 
             transaction_builder
-                .add_output(Amount::new(100, token_id), &burn_address(), &mut rng)
+                .add_output(Amount::new(490, Mob::ID), &burn_address(), &mut rng)
```
eranrund, line 3067: The original test used Mob as the fee token and tried to create change in a different token.
Here it got changed to use `token_id` as the fee token and `Mob` as the change.
The amount also got updated so that the `sum of inputs = sum of outputs - change`, even if inputs and outputs are in different token ids.

transaction/builder/src/transaction_builder.rs
```diff
@@ -3247,29 +3083,14 @@ pub mod transaction_builder_tests {
...
-                &mut rng,
-            );
-            transaction_builder.add_input(input_credentials);
-
             let TxOutContext {
                 tx_out_public_key: burn_output_public_key,
                 ..
             } = transaction_builder
-                .add_output(Amount::new(110, token_id), &burn_address(), &mut rng)
+                .add_output(Amount::new(497, token_id), &burn_address(), &mut rng)
```
eranrund, line 3096: Here (and in other places below), the amount got updated so that the transaction math is valid.

transaction/builder/src/transaction_builder.rs
```diff
@@ -3247,29 +3086,14 @@ pub mod transaction_builder_tests {
...
 
-            let mut transaction_builder = TransactionBuilder::new(
-                block_version,
-                Amount::new(10, token_id),
-                fog_resolver.clone(),
-            )
-            .unwrap();
-
+            let mut transaction_builder = transaction_builder.clone();
             transaction_builder.set_fee(3).unwrap();
```
nick-mobilecoin, line 3257: this change threw me for a loop
I couldn't understand how the test used to pass
I saw the 10 fee being set up in the `new()` and overlooked the fee being reset to 3, not sure why  it was done that way instead of initialized to 3, but test seems a bit easier to follow now

nick-mobilecoin (approved):

transaction/builder/src/transaction_builder.rs
eranrund: I think some tests used different amounts/fees/tokens just to exercise that functionality and not have everything use the exact same inputs/configuration.

PR #4078, opened 2025-10-03
DRY some code in TransactionBuilder tests

eranrund:
The tests are very long. This PR DRYs up some repeating patterns in the test.

### DIFF
--- a/transaction/builder/src/transaction_builder.rs
+++ b/transaction/builder/src/transaction_builder.rs
@@ -991,6 +991,64 @@ pub mod transaction_builder_tests {
         transaction_builder
     }
 
+    #[track_caller]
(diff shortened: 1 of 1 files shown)


transaction/builder/src/transaction_builder.rs
```diff
@@ -991,6 +991,64 @@ pub mod transaction_builder_tests {
         transaction_builder
     }
 
+    #[track_caller]
```
nick-mobilecoin, line 994: 👍 I didn't realize this was a built in macro, I thought one had to make verification/test helpers as macros in order to get the assert to happen in the correct location

nick-mobilecoin (approved):

PR #4079, opened 2025-10-03
TxBlueprint::to_unsigned_tx can process change outputs after regular outputs (SENTZ-4374)

eranrund:
Building transactions using the `TransactionBuilder` no longer forces the user to have the change output be the last output added.
This is no longer a limitation because the code that converts a `TxBlueprint` to an `UnsignedTx` can process the change output(s) after processing other outputs.

### DIFF
--- a/transaction/builder/src/memo_builder/burn_redemption_memo_builder.rs
+++ b/transaction/builder/src/memo_builder/burn_redemption_memo_builder.rs
@@ -143,3 +143,49 @@ impl MemoBuilder for BurnRedemptionMemoBuilder {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use assert_matches::assert_matches;
+    use mc_account_keys::AccountKey;
+    use mc_crypto_keys::RistrettoPublic;
+    use mc_util_from_random::FromRandom;
+    use rand::{rngs::StdRng, SeedableRng};
+
+    #[test]
+    fn make_memo_for_output_fails_after_change_output() {
+        let mut rng: StdRng = SeedableRng::from_seed([0u8; 32]);
+        let mut builder = BurnRedemptionMemoBuilder::new([3u8; BurnRedemptionMemo::MEMO_DATA_LEN]);
+        builder.enable_destination_memo();
+        builder
+            .make_memo_for_output(
+                Amount::new(100, Mob::ID),
+                &burn_address(),
+                MemoContext {
+                    tx_public_key: &RistrettoPublic::from_random(&mut rng),
+                },
+            )
+            .unwrap();
+        builder
+            .make_memo_for_change_output(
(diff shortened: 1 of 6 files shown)


transaction/builder/src/memo_builder/burn_redemption_memo_builder.rs
```diff
@@ -143,3 +143,49 @@ impl MemoBuilder for BurnRedemptionMemoBuilder {
...
+
+        assert_matches!(
+            builder.make_memo_for_output(
+                Amount::new(100, Mob::ID),
+                &burn_address(),
+                MemoContext {
+                    tx_public_key: &RistrettoPublic::from_random(&mut rng),
+                }
+            ),
+            Err(NewMemoError::MultipleOutputs)
```
eranrund, line 188: The way the burn memo builder is written, there's no actual way to get it to return `OutputsAfterChange`

transaction/builder/src/memo_builder/defragmentation_memo_builder.rs
```diff
@@ -126,3 +126,48 @@ impl MemoBuilder for DefragmentationMemoBuilder {
...
+
+        assert_matches!(
+            builder.make_memo_for_output(
+                Amount::new(100, Mob::ID),
+                &PublicAddress::default(),
+                MemoContext {
+                    tx_public_key: &RistrettoPublic::from_random(&mut rng),
+                }
+            ),
+            Err(NewMemoError::MultipleOutputs)
```
eranrund, line 170: The way the defrag memo builder is written, there's no actual way to get it to return `OutputsAfterChange`

nick-mobilecoin: 🤦 `defrag` my mind was back on the FS logic we want to have to fragment which would have multiple outputs.

transaction/builder/src/tx_blueprint.rs
```diff
@@ -103,8 +103,16 @@ impl TxBlueprint {
...
+        outputs.sort_by_key(|x| match x {
+            TxBlueprintOutput::Change { .. } => 1,
+            _ => 0,
+        });
+
         let mut outputs_and_secrets = Vec::new();
-        for output in self.outputs.clone() {
+        for output in outputs.clone() {
             outputs_and_secrets.push(build_output(&mut memo_builder, self, output)?);
         }
```
nick-mobilecoin: suggest getting rid of the extra clone, also removes the Vec::new().
I know these should generally be small quantities, often 3, but seems a bit more idomatic
```rust
        let mut outputs_and_secrets = outputs
            .into_iter()
            .map(|output| build_output(&mut memo_builder, self, output))
            .collect::<Result<Vec<_>, _>>()?;
```

nick-mobilecoin (approved):

PR #4088, opened 2025-11-05
Use postgres 18.1.1 in CD runs

nick-mobilecoin:


.github/workflows/mobilecoin-workflow-dev-setup-environment.yaml
```diff
@@ -331,7 +331,7 @@ jobs:
         action: helm-deploy
         chart_repo: https://charts.bitnami.com/bitnami
         chart_name: postgresql
-        chart_version: 15.2.2
+        chart_version: 18.1.1
```
nick-mobilecoin, line 334: we keep getting stuck trying to set up postgress in CD runs, per https://sentzglobal.slack.com/archives/CH11X49TQ/p1761246438210629
this is updating postgres to a newer version.

jgreat (approved):

.github/workflows/mobilecoin-workflow-dev-setup-environment.yaml
jgreat: Longer term we need to modify the toolbox so we can deploy without a "version" to get latest.

PR #4089, opened 2025-11-06
Add tracing context extractor and injector

nick-mobilecoin:
Add a tracing context extractor and injector to be able to support
distributed tracing with grpcio.

### DIFF
--- a/consensus/service/Cargo.toml
+++ b/consensus/service/Cargo.toml
@@ -42,7 +42,7 @@ mc-util-grpc = { workspace = true }
 mc-util-metered-channel = { workspace = true }
 mc-util-metrics = { workspace = true }
 mc-util-serial = { workspace = true }
-mc-util-telemetry = { workspace = true, features = ["otlp"] }
+mc-util-telemetry = { workspace = true }
 mc-util-uri = { workspace = true }
 
 base64 = { workspace = true }
--- a/util/telemetry/Cargo.toml
+++ b/util/telemetry/Cargo.toml
@@ -7,16 +7,11 @@ license = "GPL-3.0"
 readme = "README.md"
 rust-version = { workspace = true }
 
-[lib]
-path = "src/lib.rs"
-
-[features]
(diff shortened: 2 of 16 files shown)


consensus/service/Cargo.toml
```diff
@@ -42,7 +42,7 @@ mc-util-grpc = { workspace = true }
 mc-util-metered-channel = { workspace = true }
 mc-util-metrics = { workspace = true }
 mc-util-serial = { workspace = true }
-mc-util-telemetry = { workspace = true, features = ["otlp"] }
```
nick-mobilecoin, line 45: I removed the otlp feature forom mc_util_telemtry, that's more or less all the crate does is support otlp so seems unecessary

util/telemetry/src/extractor.rs
```diff
@@ -0,0 +1,31 @@
...
+        })
+    }
+
+    fn keys(&self) -> Vec<&str> {
+        self.0.iter().map(|(k, _)| k).collect()
+    }
+}
+
+/// Extract trace context from gRPC Context
+pub fn extract_context(rpc_ctx: &RpcContext) -> Context {
```
nick-mobilecoin, line 27: not fond of the double `context` usage here (grpc vs trace), but I think it's going to be par for the course

util/telemetry/Cargo.toml
```diff
@@ -7,16 +7,11 @@ license = "GPL-3.0"
...
-[lib]
-path = "src/lib.rs"
-
-[features]
-otlp = ["dep:opentelemetry-otlp"]
-
 [dependencies]
 cfg-if = { workspace = true }
 displaydoc = { workspace = true }
+grpcio = { workspace = true }
```
nick-mobilecoin, line 13: not fond of the grpcio dependency, but in the future we'll _hopefully_ be using tonic and down lower we depend on tonic for exporting the trace info so longer term it shouldn't be adding an extra dependency

eranrund (approved):

PR #4090, opened 2025-11-06
Add distribitued tracing through fog ledger router

nick-mobilecoin:
Previously fog ledger router didn't support tracing. Now it supports
propagating the tracing context from the caller all the way to the
shards.

### DIFF
--- a/fog/ledger/server/src/bin/key_image_store.rs
+++ b/fog/ledger/server/src/bin/key_image_store.rs
@@ -18,6 +18,28 @@ fn main() {
     mc_common::setup_panic_handler();
     let config = LedgerStoreConfig::parse();
 
+    let _runtime = if mc_util_telemetry::telemetry_enabled() {
+        // Create a Tokio runtime for telemetry OTLP exporter
+        let runtime = tokio::runtime::Builder::new_multi_thread()
+            .enable_all()
+            .build()
+            .expect("tokio runtime");
+        Some(runtime)
+    } else {
+        None
+    };
+
+    let _runtime_guard = _runtime.as_ref().map(|runtime| runtime.enter());
+
+    let _tracer = mc_util_telemetry::setup_default_tracer_with_tags(
+        env!("CARGO_PKG_NAME"),
+        &[(
+            "client_responser_id",
+            config.client_responder_id.to_string(),
+        )],
+    )
+    .expect("Failed setting telemetry tracer");
+
     let enclave_path = env::current_exe()
         .expect("Could not get the path of our executable")
         .with_file_name(ENCLAVE_FILE);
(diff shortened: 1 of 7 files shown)


fog/ledger/server/src/bin/key_image_store.rs
```diff
@@ -18,6 +18,28 @@ fn main() {
...
+        None
+    };
+
+    let _runtime_guard = _runtime.as_ref().map(|runtime| runtime.enter());
+
+    let _tracer = mc_util_telemetry::setup_default_tracer_with_tags(
+        env!("CARGO_PKG_NAME"),
+        &[(
+            "client_responser_id",
+            config.client_responder_id.to_string(),
```
nick-mobilecoin, line 38: like view is doing I used the responsder id to have a way to differentiate the shards

fog/ledger/server/src/bin/router.rs
```diff
@@ -16,6 +16,28 @@ fn main() {
     mc_common::setup_panic_handler();
     let config = LedgerRouterConfig::parse();
 
+    let _runtime = if mc_util_telemetry::telemetry_enabled() {
```
nick-mobilecoin, line 19: As the file name says this is the router, it didn't support tracing before, but now it should

fog/ledger/server/src/key_image_service.rs
```diff
@@ -107,6 +108,9 @@ impl<E: LedgerEnclaveProxy> KeyImageService<E> {
         &mut self,
         request: attest::NonceMessage,
     ) -> Result<attest::NonceMessage, EnclaveError> {
+        let tracer = mc_util_telemetry::tracer!();
+        let _guard =
+            Context::current_with_span(tracer.start("check_key_image_store_auth")).attach();
```
nick-mobilecoin, line 113: Other places in mobilecoin used the `in_span()` method, but wanted to avoid unnecessary nesting so just use the guard version

FYI the implementation of `in_span()`
```rust
        let span = self.start(name);
        let cx = Context::current_with_span(span);
        let _guard = cx.clone().attach();
        f(cx)
```

eranrund (approved):

fog/ledger/server/src/untrusted_tx_out_service.rs
```diff
@@ -48,12 +50,13 @@ impl UntrustedTxOutService {
         let TxOutInfoByPublicKeyResponse {
             results,
             latest_block,
-        } = self
-            .block_provider
-            .get_tx_out_info_by_public_key(tx_out_pub_keys.as_slice())
-            .map_err(|err| {
-                rpc_internal_error("get_tX_out_info_by_public_key", err, &self.logger)
-            })?;
+        } = tracer.in_span("get_tx_out_info_by_public_key", |_| {
```
nick-mobilecoin, line 53: used an in-span here to focus on the potential network call

PR #4091, opened 2025-11-08
Fix fog ledger distributed tracing

nick-mobilecoin:
Previously the traces in fog ledger were not properly propagating the
trace context through. Now they are.

### DIFF
--- a/fog/ledger/server/src/bin/key_image_store.rs
+++ b/fog/ledger/server/src/bin/key_image_store.rs
@@ -32,7 +32,7 @@ fn main() {
     let _runtime_guard = _runtime.as_ref().map(|runtime| runtime.enter());
 
     let _tracer = mc_util_telemetry::setup_default_tracer_with_tags(
-        env!("CARGO_PKG_NAME"),
+        "fog_ledger_store",
         &[(
             "client_responser_id",
             config.client_responder_id.to_string(),
--- a/fog/ledger/server/src/key_image_service.rs
+++ b/fog/ledger/server/src/key_image_service.rs
@@ -185,8 +185,9 @@ impl<E: LedgerEnclaveProxy> KeyImageStoreApi for KeyImageService<E> {
     ) {
         let _timer = SVC_COUNTERS.req(&ctx);
         let tracer = mc_util_telemetry::tracer!();
-        let _guard = mc_util_telemetry::extract_context(&ctx)
-            .with_span(tracer.start("auth"))
+        let tracing_ctx = mc_util_telemetry::extract_context(&ctx);
+        let _guard = tracing_ctx
+            .with_span(tracer.start_with_context("auth", &tracing_ctx))
             .attach();
         mc_common::logger::scoped_global_logger(&rpc_logger(&ctx, &self.logger), |logger| {
             if let Err(err) = self.authenticator.authenticate_rpc(&ctx) {
@@ -220,8 +221,9 @@ impl<E: LedgerEnclaveProxy> KeyImageStoreApi for KeyImageService<E> {
     ) {
         let _timer = SVC_COUNTERS.req(&ctx);
         let tracer = mc_util_telemetry::tracer!();
-        let _guard = mc_util_telemetry::extract_context(&ctx)
-            .with_span(tracer.start("multi_key_image_store_query"))
+        let tracing_ctx = mc_util_telemetry::extract_context(&ctx);
+        let _guard = tracing_ctx
+            .with_span(tracer.start_with_context("multi_key_image_store_query", &tracing_ctx))
             .attach();
 
         mc_common::logger::scoped_global_logger(&rpc_logger(&ctx, &self.logger), |logger| {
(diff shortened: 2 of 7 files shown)


fog/ledger/server/src/bin/key_image_store.rs
```diff
@@ -32,7 +32,7 @@ fn main() {
     let _runtime_guard = _runtime.as_ref().map(|runtime| runtime.enter());
 
     let _tracer = mc_util_telemetry::setup_default_tracer_with_tags(
-        env!("CARGO_PKG_NAME"),
+        "fog_ledger_store",
```
nick-mobilecoin, line 35: when debugging this it was a bit difficult when both of the services were called mc-fog-ledger

I'm up for suggestions here, but I think having separate names is better

Image

fog/ledger/server/src/key_image_service.rs
```diff
@@ -185,8 +185,9 @@ impl<E: LedgerEnclaveProxy> KeyImageStoreApi for KeyImageService<E> {
     ) {
         let _timer = SVC_COUNTERS.req(&ctx);
         let tracer = mc_util_telemetry::tracer!();
-        let _guard = mc_util_telemetry::extract_context(&ctx)
-            .with_span(tracer.start("auth"))
+        let tracing_ctx = mc_util_telemetry::extract_context(&ctx);
+        let _guard = tracing_ctx
+            .with_span(tracer.start_with_context("auth", &tracing_ctx))
```
nick-mobilecoin: One of the problems with distributed tracing not working:

`tracer.start()` will use the currently attached context for creating the span

This means the previous code would:
1. create a span in the root context of the fog-ledger-store
2. Extract the context from the caller
3. add the span to the extracted context, however the span is associated with the root context
4. attach the context with span, thus I think attaching the root span

IDK, tbh this circular dependencies between the tracer/context/spans makes me go 🤯 Maybe there's a reason but this raw api is difficult to understand the usage.


The fix
Create the span with the extracted context ala `start_with_context`, then when it's called with attach() it works correctly
one could also do
```rust
 let _root_guard = mc_util_telemetry::extract_context(&ctx).attach();
 let _guard = create_context(&tracer, "name").attach();
```
but I felt the double guards felt awkward

fog/ledger/server/src/router_handlers.rs
```diff
@@ -354,9 +354,10 @@ async fn route_query(
     request: &MultiKeyImageStoreRequest,
     shard_clients: Vec<Arc<KeyImageStoreApiClient>>,
 ) -> Result<Vec<(Arc<KeyImageStoreApiClient>, MultiKeyImageStoreResponse)>, RouterServerError> {
-    let responses = shard_clients
-        .into_iter()
-        .map(|shard_client| query_shard(request, shard_client));
+    let tracer = mc_util_telemetry::tracer!();
+    let responses = shard_clients.into_iter().map(|shard_client| {
+        query_shard(request, shard_client).with_context(create_context(&tracer, "query_shard"))
```
nick-mobilecoin, line 359: I felt it was nicer to have a span per shard being queried in the caller

fog/ledger/server/src/router_service.rs
```diff
@@ -155,6 +155,7 @@ impl<E: LedgerEnclaveProxy> FogKeyImageApi for LedgerRouterService<E> {
                 shards.values().cloned().collect(),
                 logger.clone(),
             )
+            .with_context(tracing_ctx)
```
nick-mobilecoin, line 158: antoher problem
Since this is an async function we need the trace context to be added to the future otherwise the span won't work correctly for when the future is polled

util/telemetry/src/otlp.rs
```diff
@@ -36,6 +36,8 @@ pub fn setup_default_tracer_with_tags(
         return Ok(None);
     }
 
+    opentelemetry::global::set_text_map_propagator(TraceContextPropagator::new());
```
nick-mobilecoin, line 39: another problem, if one doesn't add the trace propagator, to the actual global propagator it doesn't know to keep track of the trace context for injecting

fog/ledger/server/src/key_image_service.rs
```diff
@@ -185,8 +185,9 @@ impl<E: LedgerEnclaveProxy> KeyImageStoreApi for KeyImageService<E> {
     ) {
         let _timer = SVC_COUNTERS.req(&ctx);
         let tracer = mc_util_telemetry::tracer!();
-        let _guard = mc_util_telemetry::extract_context(&ctx)
-            .with_span(tracer.start("auth"))
+        let tracing_ctx = mc_util_telemetry::extract_context(&ctx);
```
eranrund: In some code below you've used the name `parent_ctx`, I found it a little easier to read when that name was used instead of `tracing_ctx`.

eranrund (approved):
Agree this API is very much not intuitive. The resulting trace is awesome though!

fog/ledger/server/src/key_image_service.rs
nick-mobilecoin: updated to `parent_ctx` here and in other locations in [7848847](https://github.com/mobilecoinfoundation/mobilecoin/pull/4091/commits/7848847ecfe00ce9949ab59e1644218860d08f09)

PR #4092, opened 2025-11-11
Propagate trace context through mobilecoind block_provider

nick-mobilecoin:
Previously the trace context was lost once a request was made to the
mobilecoind block provider. Now the trace context propagates through to
the Mobilecoind backend.

### DIFF
--- a/fog/block_provider/src/mobilecoind.rs
+++ b/fog/block_provider/src/mobilecoind.rs
@@ -3,7 +3,7 @@
 use crate::{
     BlockDataWithTimestamp, BlockProvider, BlocksDataResponse, Error, TxOutInfoByPublicKeyResponse,
 };
-use grpcio::{ChannelBuilder, EnvBuilder};
+use grpcio::{CallOption, ChannelBuilder, EnvBuilder, MetadataBuilder};
 use mc_api::watcher::TimestampResultCode;
 use mc_blockchain_types::{Block, BlockData, BlockIndex};
 use mc_common::logger::{log, Logger};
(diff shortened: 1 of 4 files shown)


mobilecoind/src/service.rs
```diff
@@ -2699,6 +2700,11 @@ macro_rules! build_api {
                     request: $service_request_type,
                     sink: UnarySink<$service_response_type>,
                 ) {
+                    let tracer = mc_util_telemetry::tracer!();
```
nick-mobilecoin, line 2703: This ends up creating a span for every mobilecoind function, but it will only be for those that are run with MC_TELEMETRY=1 so figure it's ok and shouldn't be too spammy

fog/block_provider/src/mobilecoind.rs
```diff
@@ -193,10 +194,17 @@ impl BlockProvider for MobilecoindBlockProvider {
...
+        // shouldn't prevent the request from being processed.
+        let _ = meta_builder.inject_context(&Context::current());
+        let call_opt = CallOption::default().headers(meta_builder.build());
         let request = GetTxOutResultsByPubKeyRequest {
             tx_out_public_keys: tx_out_pub_keys.iter().map(|pk| pk.into()).collect(),
         };
-        let response = self.client.get_tx_out_results_by_pub_key(&request)?;
+        let response = self
+            .client
+            .get_tx_out_results_by_pub_key_opt(&request, call_opt)?;
```
nick-mobilecoin, line 207: I realized with the newer block provider trait we might benefit by propagating the context down to the networked block provider.

mobilecoind/src/service.rs
nick-mobilecoin: See #4094 i've reworked this, as I was wrong, it's a bit spammy

eranrund (approved):

PR #4093, opened 2025-11-11
Add distributed tracing to fog view

nick-mobilecoin:


fog/view/server/src/bin/main.rs
```diff
@@ -39,7 +39,7 @@ fn main() {
     let _runtime_guard = _runtime.as_ref().map(|runtime| runtime.enter());
 
     let _tracer = mc_util_telemetry::setup_default_tracer_with_tags(
-        env!("CARGO_PKG_NAME"),
+        "fog-view-store",
```
nick-mobilecoin, line 42: similar to ledger I wanted unique names per binary

Image

fog/view/server/src/fog_view_router_service.rs
```diff
@@ -61,6 +61,10 @@ where
         requests: RequestStream<FogViewRouterRequest>,
         responses: DuplexSink<FogViewRouterResponse>,
     ) {
+        let tracer = mc_util_telemetry::tracer!();
+        let parent_ctx = mc_util_telemetry::extract_context(&ctx);
+        let tracing_ctx = parent_ctx.with_span(tracer.start_with_context("request", &parent_ctx));
```
nick-mobilecoin, line 66: I know many of htese are repeated
I could potentially move to a macro, but I wanted to do some digging around to see if grpcio has some kind of middleware layer the context could be extracted in and we attach the `parent_ctx`  as the current one.

fog/view/server/src/bin/main.rs
nick-mobilecoin: a bit more tracing eye candy, not sure the different shades of brown was a good choice for jaeger, but you can see test client -> view router -> shard
image

I wasn't able to get the test client actually seeing the balances locally to start generating blocks, so punted on local run for now

eranrund (approved):

PR #4094, opened 2025-11-12
Remove some stray traces/spans

nick-mobilecoin:
Previously all of the Mobilecoind API was creating spans. This was a bit
noisy when the caller wasn't propagating a trace context. Now
Moibilecoind will only create spans if it was provided a propagated
trace context.

### DIFF
--- a/fog/sample-paykit/src/client.rs
+++ b/fog/sample-paykit/src/client.rs
@@ -39,7 +39,6 @@ use mc_transaction_core::{
     Amount, FeeMap, TokenId,
 };
 use mc_transaction_extra::{MemoType, SenderMemoCredential, SignedContingentInput};
-use mc_util_telemetry::{block_span_builder, telemetry_static_key, tracer, Key, Span};
 use mc_util_uri::{ConnectionUri, FogUri};
 use rand::Rng;
 use std::collections::{HashMap, HashSet};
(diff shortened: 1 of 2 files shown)


mobilecoind/src/service.rs
```diff
@@ -2700,11 +2700,18 @@ macro_rules! build_api {
...
                 ) {
-                    let tracer = mc_util_telemetry::tracer!();
                     let parent_ctx = mc_util_telemetry::extract_context(&ctx);
-                    let _guard = parent_ctx
-                        .with_span(tracer.start_with_context(stringify!($service_function_name), &parent_ctx))
-                        .attach();
+                    // Mobilecoind is often called by polling threads that don't
+                    // have a trace context (e.g. db polling thread). In those
+                    // cases, we don't want to record spans as they would be orphaned.
+                    let _guard = if parent_ctx.span().span_context().is_valid() {
```
nick-mobilecoin, line 2707: I was wrong in https://github.com/mobilecoinfoundation/mobilecoin/pull/4092#discussion_r2515791362 about these spans not being a big deal, got a bit of noise when trying to run a service locally ~4sec for the db pollers

Image

fog/sample-paykit/src/client.rs
```diff
@@ -218,14 +212,6 @@ impl Client {
                 err
             })?;
 
-        let tracer = tracer!();
```
nick-mobilecoin, line 221: This was causing a duplicate span that the caller was already tracking, 

Image

eranrund (approved):

PR #4095, opened 2025-11-21
Make PrintableWrapper::b58_encode infallible

nick-mobilecoin:
previously the PrintableWrapper::b58_encode would return a result. This
was a hold over from when the protobuf crate was used to encode the
PrintableWrapper to a byte vector internally, which could fail. Now
PrintableWrapper::b58_encode is infallible and returns a String.

### DIFF
--- a/api/src/display.rs
+++ b/api/src/display.rs
@@ -42,12 +42,12 @@ fn calculate_checksum(data: &[u8]) -> [u8; 4] {
 /// encoded string, with a checksum prepended to it.
 impl printable::PrintableWrapper {
     /// Converts the proto to bytes and then encodes as b58
-    pub fn b58_encode(&self) -> Result<String, Error> {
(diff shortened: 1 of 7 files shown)


api/src/display.rs
```diff
@@ -42,12 +42,12 @@ fn calculate_checksum(data: &[u8]) -> [u8; 4] {
 /// encoded string, with a checksum prepended to it.
 impl printable::PrintableWrapper {
     /// Converts the proto to bytes and then encodes as b58
-    pub fn b58_encode(&self) -> Result<String, Error> {
+    pub fn b58_encode(&self) -> String {
```
nick-mobilecoin, line 45: note: I first noticed this walking down some full-service calls that didn't seem like they should be fallible, so this will likely result in full-service needing some minor changes after this goes in.

eranrund (approved):

PR #4097, opened 2026-06-11
Adds SECURITY.md

sugargoat:
### Motivation

We would like to specify the format and the destination for security vulnerability reports.

[Soundtrack of this PR: Technicolor Beat](https://open.spotify.com/track/5pLpkaIRobcvPnUmclNv6o?si=7f9f36b63bc34654)

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,325 @@
+# Security Policy
+
+The security of Sentz is fundamental to our mission. We greatly appreciate the efforts of security researchers, auditors, and members of the security community who help identify vulnerabilities and responsibly disclose them.
+
+This document describes how to report security vulnerabilities, the information we need to investigate a report, and the expectations we have for responsible security research.
+
+If you believe you have discovered a security vulnerability, please report it to:
+
+**[security@sentz.com](mailto:security@sentz.com)**
+
+Please do not publicly disclose a vulnerability until we have had a reasonable opportunity to investigate and remediate the issue.
+
+---
+
+# Security Bounty Program
+
+Sentz operates a security bounty program and may provide rewards for eligible vulnerability reports.
+
+Eligibility, scope, severity, and reward determinations are made at Sentz's sole discretion and may vary based on factors including:
+
+* Security impact
+* Exploitability
+* Affected users or systems
+* Report quality
+* Originality of the finding
+* Existing mitigations
+
+Not all reports will qualify for bounty consideration.
+
+---
+
+# Scope
+
+Unless otherwise specified, we welcome reports relating to security vulnerabilities affecting Sentz-controlled products, services, infrastructure, and codebases.
+
+Examples may include:
+
+* Sentz mobile applications
+* Sentz wallet infrastructure
+* Sentz blockchain network and protocol implementations
+* Consensus and transaction validation logic
+* Cryptographic systems and key management
+* Public APIs and backend services
+* Web applications and web properties
+* Developer SDKs and client libraries
+* Payment and transaction infrastructure
+* Privacy-preserving systems and protocols
+
+---
+
+# Privacy Vulnerabilities
+
+Privacy is a core security property of the Sentz ecosystem.
+
+We encourage reports involving:
+
+* Metadata leakage
+* User deanonymization
+* Transaction graph analysis weaknesses
+* Privacy bypasses
+* Correlation attacks
+* Information disclosure through protocol behavior
+* Side-channel attacks
+* Cryptographic privacy failures
+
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -0,0 +1,325 @@
...
+
+If you believe you have discovered a security vulnerability, please report it to:
+
+**[security@sentz.com](mailto:security@sentz.com)**
+
+Please do not publicly disclose a vulnerability until we have had a reasonable opportunity to investigate and remediate the issue.
+
+---
+
+# Security Bounty Program
```
nick-mobilecoin: ⛏️ it may be desired for appearance, but it seems that the a header is usually sufficient to indicate a new section/topic

```suggestion
# Security Bounty Program
```

if the section separator is removed it should be done for the other areas as appropriate

SECURITY.md
```diff
@@ -0,0 +1,325 @@
...
+
+# Scope
+
+Unless otherwise specified, we welcome reports relating to security vulnerabilities affecting Sentz-controlled products, services, infrastructure, and codebases.
+
+Examples may include:
+
+* Sentz mobile applications
+* Sentz wallet infrastructure
+* Sentz blockchain network and protocol implementations
```
nick-mobilecoin, line 38: I'm not sure where the separations are between the company and the foundation
My understanding is at this time the `Sentz mobile applications` is more or less all the company and *not* part of the foundation. While this repo is part of the foundation.

Are we ok calling out company owned here?
I could definitly see it in that users of the Sentz app won't have a dedicated repo to see the security policy or similar, just want to be sure we're ok [crossing the streams](https://youtu.be/wyKQe_i9yyo?t=49)

SECURITY.md
```diff
@@ -0,0 +1,325 @@
...
+
+Identify any AI systems used during the research process.
+
+Please distinguish between:
+
+### Findings directly observed and validated by the researcher
+
+and
+
+### Hypotheses, attack paths, or analyses suggested by AI systems
```
nick-mobilecoin: Should these be emphasized with bold or italics instead of a level 3 header?

```suggestion
**Findings directly observed and validated by the researcher**

and

**Hypotheses, attack paths, or analyses suggested by AI systems**
```

SECURITY.md
```diff
@@ -0,0 +1,325 @@
...
+* Required prerequisites.
+* Whether the attack has been demonstrated or remains theoretical.
+* Evidence supporting claimed impact.
+* Economic analysis where relevant.
+
+Where practical, include a proof-of-concept demonstrating exploitation.
+
+---
+
+## AI-Assisted Security Research
```
nick-mobilecoin: Should this be a level 1 header to indicate a start of a new section?

I don't think the `6. ` will be renumbered by a markdown renderer as it's not a list. So it seems a bit odd to have
```
(h2) 6. title
(h2) AI...
(h2) 7. title

File: SECURITY.md [hunk 5]
```diff
@@ -0,0 +1,325 @@
...
+
+As AI-assisted security research becomes increasingly common, we welcome submissions that leverage AI tools.
+
+However, to enable efficient review and accurate assessment, we request that AI-assisted submissions follow the format above, and also include an AI Assistance Disclosure.
+
+## 7. AI Assistance Disclosure
```
nick-mobilecoin: see comment above, is 7. the right number here? looking at the render, https://github.com/mobilecoinfoundation/mobilecoin/blob/security-md/SECURITY.md it doesn't get re-numbered

nick-mobilecoin (approved):

SECURITY.md
sugargoat: (Lol at the ghostbuster's reference.)

Yeah, I'm ok with crossing the streams here, because I think we would want to cast a wide net, so any security issue within the ecosystem is to be sent to the dev team at security@sentz.com, and then we can triage from there.

SECURITY.md
sugargoat: Good sug, ty!

SECURITY.md
sugargoat: Ah yeah, I wasn't quite sure how to do this. It's an "optional addition to the form", so maybe from 7. on it should be h3 - let me fix that up

SECURITY.md
sugargoat: Updated formatting to remove the numbers and make these sub-headings of an "Additional Requirements for AI" - I think it's clearer now, ty!

SECURITY.md
sugargoat: Ok updated!

